PR #5, opened 2020-06-05
Add Atom theme

kognise:
I added my Obsidian Atom theme!

Here's the screenshot I included, not sure if it would be cleaner to include only the dark/light one though. I couldn't get a GIF to look nice.

❤️


ericaxu:
⚛️ :atom:

PR #26, opened 2020-11-04
Add PDF Exporter plugin

Nigecat:


lishid:
I took a quick look at your code and it seems the plugin just copies the HTML over to a new BrowserWindow? This will most likely break things given you lose all the CSS from the app, especially because LaTeX depends on CSS to render.

Also this functionality is already built into Obsidian (in the next insider version) and will come with full CSS support. Do you think it's enough to supersede this plugin?

Nigecat:
Ah I didn't know about that. That would definitely work better and I'll probably just wait for that to be released.

lishid:
Thanks again for your contribution! I hope this doesn't discourage you from making more plugins 👍

PR #41, opened 2020-11-11
Adds Extract Highlights

akaalias:
Hi Obsidian!

I made a small plugin based on this forum conversation: https://forum.obsidian.md/t/extract-highlights-from-note/7867

I decided to make a minimal version of the functionality for members to test out and then expand based on their feedback.

Have a look and if you think it's worth including, please merge it in. 

Thanks so much, 

Alexis

### DIFF
--- a/community-plugin-stats.json
+++ b/community-plugin-stats.json
@@ -1,45 +1,46 @@
 {
   "nldates-obsidian": {
-    "0.3.0": 122,
+    "0.3.1": 60,
+    "0.3.0": 129,
     "0.2.6": 233,
     "0.2.5": 101,
     "0.2.4": 121,
     "0.2.3": 90,
-    "downloads": 667
+    "downloads": 734
   },
   "hotkeysplus-obsidian": {
-    "0.2.1": 230,
-    "downloads": 230
+    "0.2.1": 247,
+    "downloads": 247
   },
   "obsidian-git": {
-    "1.4.0": 732,
+    "1.4.0": 797,
     "1.3.0": 23,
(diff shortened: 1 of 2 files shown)


lishid:
Few suggestions:
- Use `new Notice('text')` instead of a status bar (https://github.com/akaalias/extract-highlights-plugin/blob/master/main.ts#L32)
- Don't use clipboard from electron, use the browser version `navigator.clipboard.writeText();`. This makes your plugin portable to web & mobile in the future.
- Auto-format your code to have consistent spaces or tabs.

akaalias:
Hi @lishid thank you so much for your feedback and your helpful. I'll get to that as soon as this week is over and resubmit then. 

LOVE Obsidian and Publish by the way. Thank you for making something I didn't know I needed :)

akaalias:
Hi @lishid, 

I've created a new release with your suggested edits and resolved the merge-conflict. Let me know if this works. 

By the way: I really like the Notice() idea, feels better. Thanks!

lishid:
Seems like your merge screwed up the whole file... Can you rebase off master and fix that?
The end result should look like this:
https://github.com/obsidianmd/obsidian-releases/pull/45/files

akaalias:
Hi @lishid I'm sorry this is taking so long. I've resolved the conflicts against current master status but somehow all lines got replaced instead of adding mine at the bottom. If this is too messy, I think you can delete this pull request and I'll start over with it. Thanks!

lishid:
Yeah... it would be much better if you can try to start a new branch from scratch

PR #53, opened 2020-11-15
add move-to-note plugin

jsbernier:
added move-to-note plugin


lishid:
Any particular reason why it seems like you're writing generated code by hand?
Also, would it make sense to add this functionality to https://github.com/lynchjames/note-refactor-obsidian instead of having a separate plugin for it?

jsbernier:
Im dev from .net but noob to typescript/javascript dev. Did not take the time needed to setup or learn new tools.

You are right, il would make sense.

________________________________

lishid:
Oh got it. I've been doing quite some work in C# previously and I think TypeScript is a great alternative to JavaScript that resembles more of the type safe world 😄 

James is also pretty cool to work with, and I think he can help you if you wish to make a PR to his repo :)

PR #55, opened 2020-11-16
Add Nucleus theme

Lukalot:
Adds Nucleus dark theme to the community themes list.

### DIFF
--- a/community-themes.json
+++ b/community-themes.json
@@ -343,5 +343,13 @@
         "repo": "obsidian-molecule",
         "screenshot": "images/Screenshot1.png",
         "compatibleModes": ["light"]
+    },
+    {
+        "name": "Nucleus",
+        "author": "Lukalot",
+        "github": "Lukalot",
+        "repo": "nucleus-obsidian",
+        "screenshot": "images/preview.png",
+        "compatibleModes": ["dark"]
     }
 ]


ericaxu:
Thanks for the PR!

This theme looks very close to the default theme, other than some personalized colors. Did I miss anything?

Lukalot:
Well, the tagging is semi-heavily augmented, the tags have a more classical rounded background. Most notably there's a special "Metatag" system when a tag is prefixed with `#-` as opposed to only a hashtag appears differently, which I use heavily in my personal notes. Spacing in the sidebar is more dense and headings have floating indicators next to them and don't effect the main body when in preview mode, which I think is another feature non of the other themes have currently (to my knowledge). The blacks have more blue, which is admittedly only a color change, and the small mid-tone border on the view window wraps across the top entirely (also subtle, not a big change I suppose). There are quite apparent style changes for both the Calendar and Code Copy Button plugins which make a big difference. Calendar with the default theme, for example, has a black rectangle where the dates end (or it did last time I checked). The copy code button is padded, recolored, and reshaped to fit with the rest of the theme.

Overall, I loved the original theme (hence the strong similarities in most places) but this theme brings several special features, color customizations, indicators and plugin styling improvements.

If you look specifically at a preview mode note you should see some of the more noticeable changes.  If you don't think they're significant enough to warrant a community theme entry, that's totally understandable. I can make some more significant changes later and pr if you think that would be better.

ericaxu:
Thanks so much for understanding! @Lukalot 

We have an upcoming change to separate custom CSS into themes and snippets, to make remixing easier, and I think your CSS would be a good snippet candidate once we have a community section for that. I hope that's understandable! Thanks so much for your contribution to the community.

PR #57, opened 2020-11-17
Add wikilinks-to-mdlinks

agathauy:


lishid:
Please use `workspace.getActiveLeafOfViewType` instead of accessing `activeLeaf` directly.

You should also check for `cmEditor.hasFocus` because getCursor always returns the last cursor even if there's no focus.

agathauy:
Hi @lishid! I put in the `cmEditor.hasFocus` just now, however upon checking the Obsidian API github repo I can't quite figure out what `workspace.getActiveLeafOfViewType` does, or how it would change my usage of activeLeaf. 

So far I have just tried a few different variations of using it instead of activeLeaf, but I seem to be using it incorrectly 😅

agathauy:
@lishid ok nevermind got it! Will be pushing an updated version :D

agathauy:
Pushed an updated version 👍

lishid:
Make sure you test the result from `getActiveLeafOfViewType` (in your case `currentView`) for nulls.

Also recommend using a check callback instead. This way your command doesn't show up if the user isn't focused on a markdown pane.
https://github.com/obsidianmd/obsidian-sample-plugin/blob/318cb2b055e06748f56c25c4d4c624eabcababb2/main.ts#L19

lishid:
Here's the docs on check callback for your reference
https://github.com/obsidianmd/obsidian-api/blob/d0b0562a5f8251153a72cc397f08876dbb28222b/obsidian.d.ts#L369

agathauy:
Ok thank you made the changes you recommended. And thank you for providing the links! 😀

lishid:
Great work! Actually my bad on `hasFocus` because when you trigger the command palette it loses focus.

What you'll probably want to do instead is check if the view type is source mode.
https://github.com/obsidianmd/obsidian-api/blob/d0b0562a5f8251153a72cc397f08876dbb28222b/obsidian.d.ts#L1129

agathauy:
@lishid Hmm I changed the checkCallback to this upon your prior recs: 
```
				const currentView = this.app.workspace.getActiveLeafOfViewType(MarkdownView)
				if ((currentView == null) || (currentView.getMode() !== 'source'))  {
					return
				}
```

However I noticed that the setting doesn't appear anymore in the command palette. Yeah haha forgot to check with the hasFocus before too.

agathauy:
It seems like using checkCallback doesn't allow the plugin setting to appear, even if it passed that check.
I just do a `return this.toggleLink()` at the end though.

However, if I don't use checkCallback, the setting appears in the command palette.

lishid:
You need to `return true` so it shows up. Check the `returns` documentation at:
https://github.com/obsidianmd/obsidian-api/blob/d0b0562a5f8251153a72cc397f08876dbb28222b/obsidian.d.ts#L372

lishid:
The purpose of checkCallback is to "check" whether your command can be executed at the present moment based on pane configuration, etc. Your function should only return true when the configuration is correct.

You need to read `checking` to determine whether to actually run your `toggleLink` function. Otherwise, your command will run the moment when the command palette shows up, regardless of whether the user runs your command or not.

agathauy:
@lishid Ok got it! Updated it again. Things work now with checkCallback. Hope I got it right this time haha. Thanks again for the help and explaining the api 😄

lishid:
Looking great!

PR #59, opened 2020-11-21
Update community-plugins.json

phibr0:
# Add Cycle through Panes Plugin

This Plugin allows the User to switch between active/open Tabs in Obsidian using `ctrl + Tab`, just like in a Browser!


phibr0:
In Combination with Mr. Jackphils Jump to Link plugin and Vim Keybindings, this plugin allows you to control Obsidian entirely without a Mouse!

PR #62, opened 2020-11-22
Add System Dark Mode plugin

kepano:


lishid:
I think your .gitignore is ignoring the .ts file instead of the .js file 😂

kepano:
@lishid that should be fixed, thanks

lishid:
Cool. A couple of things suggest adding/tweaking:
- I don't see a way to disable/unload the plugin, probably a good idea to add that? You can use `this.register(() => ...unload callback...);` to register for an unload inline (inside the onload function) to help you unhook the media watch listener.
- A quick auto-format of your code would help fix the tabs vs spaces mix.
- Would suggest switching the function syntax from `functionname = () => {` to just `functionname() {`

kepano:
@lishid This is done. Thanks for your help!

PR #66, opened 2020-11-24
Add obsidian-dangling-links plugin

graydon:
My first obsidian plugin, hopefully I did it near enough to correctly to work.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -329,5 +329,13 @@
         "description": "Use YouTube to improve your pronunciation. YouGlish plugin gives you fast, unbiased answers about how words is spoken by real people and in context.",
         "author": "Noureddine Haouari",
         "repo": "nhaouari/obsidian-youglish-plugin"
+    },
+    {
+        "id": "obsidian-dangling-links",
+        "name": "Dangling links panel",
+        "author": "Graydon Hoare",
+        "description": "Adds a panel showing any dangling links in a vault.",
+        "repo": "graydon/obsidian-dangling-links",
+        "branch": "main"
     }
 ]


lishid:
Are you just looking for links that don't resolve to files? The link resolution algorithm is a lot more complicated than what you seem to be testing for `!allFiles.has(link.link)`. It splits the link into path & subpath and then resolves relative/absolute/unique file names case insensitive. Right now we already compute those in metadataCache.unresolvedLinks (not publicly exposed) but it does not store actual file location of where those links are.

graydon:
Yes, that sort of link is the only sort I've made across hundreds of notes.

I don't want to spend more days on this; I already spent a full day of my workweek (and note: I am a paying customer and this is an obvious missing feature in Obsidian). If you want to make the necessary API public, or heck, even implement the feature, go for it. As it is I'm happy with this level of funcionality in the plugin. It finds all my dangling links.

[I am not the only one](https://forum.obsidian.md/t/how-to-find-links-without-notes/2578/13) who needs something like this. It's surely more ergonomic than running a python script over a vault. I figured I'd share it here. I'm sure it could be improved further, but I have an actual other job.

graydon:
(I'm sorry that comes across as a bit prickly -- if you have actionable suggestions I'm willing to spend a little more energy, and I do appreciate the attention / review. It's just not super useful to tell me there's an API I can't call, implementing an algorithm I can't see, that I should be .. reimplementing .. by guesswork? To cover cases I don't need?)

lishid:
You're right, I forgot to include the solution details. You can use these functions to properly resolve links:

`getLinkpath`
https://github.com/obsidianmd/obsidian-api/blob/3efab21fc0cfef3ae75951beb66aadd479ae610d/obsidian.d.ts#L836
This will split the path & subpath and only get the path part. This will remove things like `#heading` from the link.

`MetadataCache.getFirstLinkpathDest`
https://github.com/obsidianmd/obsidian-api/blob/3efab21fc0cfef3ae75951beb66aadd479ae610d/obsidian.d.ts#L1275
This will resolve a link path (you'll have to pass the sourcePath which is "from which file does this link originate") and give you the target file, or null if it's not found.

You're probably looking for links that gets a null destination.

I don't mind merging it right now, or I can wait until you find time to work out the details. Let me know what's best.

graydon:
Ok, thanks. Updated with changes, lightly tested, seems to work: 

https://github.com/graydon/obsidian-dangling-links/commit/71afd750dbaf2722a7626b2633d5c0042e10a3d5

Added new release 0.0.2

lishid:
Great work!

autonomygaps:
Fantastic work.  This is a total game-changer!  It facilitates a workflow in which I can Wiki-wrap a key term when I'm writing or reviewing text in Obsidian, without stopping to flesh out a new note but with the confidence that there is a way to periodically review the "open loops" of "non-existent" files that look at later.  The graph view lets you do this, in a way, but this is so much faster and clearer.

Two feature requests, related to the fact that I currently have hundreds of dangling links: 

1. Folding by folders, like a bulleted outline, so that you can focus on the dangling links in a particular folder.
2. Sorting by Date Modified, Date Created, and Note Title. (Current it seems to be alpha order, but first by folder)

Thanks!

graydon:
@autonomygaps in theory all this can be done at the HTML-and-CSS level I think but it's not my strong suit -- it'd go a lot faster if @lishid could provide the markup that's used for the current backlinks frame, which I think implements exactly what you're after.

Any chance we could see that @lishid ? I imagine it's some mix of `<summary>` and `<details>` and snippet-text and selectors for sorting?

lishid:
I can look into exposing the API for that although it's quite a complicated component (it's also got some crazy optimizations to unload the html elements when out-of-view).

The interface shouldn't be too bad but we want to make sure it's ready so that there won't be subsequent major breaking changes to the API in the future.

PR #67, opened 2020-11-24
Add obsidian-juliandate

THeK3nger:
First plugin. I converted one of my most used text expansion for astronomical observations into a native plugin.


lishid:
https://github.com/THeK3nger/obsidian-juliandate/blob/d0775c96d0ae93aa4716ced572671e625054d0b4/src/main.ts#L10

Don't use a class for your plugin settings - `loadData` returns an regular object. Use an interface instead. (You might also want to change this on your plugin template since I've seen at least 2 others do the same, probably following your template)

https://github.com/THeK3nger/obsidian-juliandate/blob/d0775c96d0ae93aa4716ced572671e625054d0b4/src/main.ts#L31

Don't use activeLeaf directly (You'll error out when activeLeaf is null), use `Workspace.getActiveViewOfType(MarkdownView)`.

THeK3nger:
> Don't use a class for your plugin settings - loadData returns an regular object. Use an interface instead. (You might also want to change this on your plugin template since I've seen at least 2 others do the same, probably following your template)

Done. I've also updated the template.

> Don't use activeLeaf directly (You'll error out when activeLeaf is null), use Workspace.getActiveViewOfType(MarkdownView).

And done. Is this a new API, right? Because some older plugin I checked out still use `activeLeaf` directly.

lishid:
Looks great!

https://github.com/THeK3nger/obsidian-juliandate/blob/7ab99291367434a4f66644f28dd2f9e292a5dc4e/src/main.ts#L43
You'll want a null check on the view here - this will fail if the user isn't focused on a MarkdownView.

> Is this a new API, right?

Yeah this was added specifically because plugin authors had to null check `activeLeaf` and then `activeLeaf.view instanceof MarkdownView`. Now it's rolled into one and returns null if either doesn't match.

community-plugins.json
```diff
@@ -322,5 +322,13 @@
...
         "repo": "deathau/cm-typewriter-scroll-obsidian",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-juliandate",
+        "name": "Julian Date Quick Insert",
+        "author": "thek3nger",
+        "description": "Add a shortuct to insert the current Julian Date for astronomical observations.",
+        "repo": "THeK3nger/obsidian-juliandate",
+        "branch": "master"
```
lishid, line 332: Also you don't need to specify this if your branch is master.

THeK3nger:
Fixed both. 👍

lishid:
Seems like you've rewrote the whole file? It's probably the line endings that got changed.

Any chance you can fix that?

THeK3nger:
Weird. I used the GitHub's online editor exactly to avoid line endings issues in the first place. Oh, well... It should be fixed now.

lishid:
Yay!

PR #68, opened 2020-11-24
Add Youglish Plugin

nhaouari:


lishid:
Couple of small things you may want to change:

https://github.com/nhaouari/obsidian-youglish-plugin/blob/db4ff838ac68d18b2c412718a1e00be90a2a36c1/main.ts#L49
Use `workspace.getActiveViewOfType` and test for null

https://github.com/nhaouari/obsidian-youglish-plugin/blob/db4ff838ac68d18b2c412718a1e00be90a2a36c1/main.ts#L60
Same change as above.

https://github.com/nhaouari/obsidian-youglish-plugin/blob/db4ff838ac68d18b2c412718a1e00be90a2a36c1/main.ts#L50
Use `sourceMode` instead of `currentMode`.

https://github.com/nhaouari/obsidian-youglish-plugin/blob/db4ff838ac68d18b2c412718a1e00be90a2a36c1/main.ts#L128
Your settings don't seem to be saved across restart. I'd recommend saving it using `plugin.saveData`/`plugin.loadData`.

PR #70, opened 2020-11-27
Adds Dangerzone Plugin

akaalias:
Hi @lishid,

I'd like to submit my Dangerzone Writing plugin (See https://forum.obsidian.md/t/dangerzone-flowstate-like-plugin-prototype/8776) for details with this PR.

Let me know if this is okay!

Thank you, 

Alexis

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -337,5 +337,12 @@
         "description": "Adds a panel showing any dangling links in a vault.",
         "repo": "graydon/obsidian-dangling-links",
         "branch": "main"
+    },
+    {
+        "id": "dangerzone-writing-plugin",
+        "name": "Dangerzone Writing",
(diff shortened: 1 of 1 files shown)


lishid:
Couple of suggestions:

- Auto format your code so your spacing is consistent and semicolon is all added.
- Clear the `setInterval` when your plugin unloads. Probably want to cleanup the whole `countdown` object.
- I would suggest not storing all codemirror editors as a reference, because you're inducing a memory leak when panes are closed but you're holding on a reference to it.
https://github.com/akaalias/dangerzone-writing-plugin/blob/ad53097930e79a551bd4067023fae72d142048c1/main.ts#L10
Instead, your `handleKeyDown` function could unhook itself if the plugin has been disabled.

PR #78, opened 2020-12-07
Add flashcards plugin

reuseman:


lishid:
https://github.com/reuseman/flashcards-obsidian/blob/98b16c0de42ac4ccf4c54751808cbf2c608c8d20/main.ts#L30
Not sure if this was tested or not, but loadData() returns an object (through `JSON.parse()`) which will not have the prototype of your `Settings` class. You should use an interface for settings instead.

Otherwise this looks great!

reuseman:
I've followed your suggestion. Now it should be fixed, let me know

lishid:
Great work!

PR #79, opened 2020-12-09
Add completed-area plugin.

DahaWong:


lishid:
https://github.com/DahaWong/obsidian-completed-area/blob/bdb8b38ed5e354416e48631c28a4630f8fd42a09/src/main.ts#L55
You probably want some kind of debounce here instead of a setTimeout?

https://github.com/DahaWong/obsidian-completed-area/blob/bdb8b38ed5e354416e48631c28a4630f8fd42a09/src/main.ts#L28
Can you provide a way to disable your plugin? Right now if you enable then disable your plugin, it won't actually disable.
I recommend setting a flag `disabled` in the `onunload` function and returning on those events if `disabled` has been set. This is because keeping track of the CodeMirror instances to unhook the events will prevent garbage collection, which will cause a memory leak.

https://github.com/DahaWong/obsidian-completed-area/blob/bdb8b38ed5e354416e48631c28a4630f8fd42a09/src/main.ts#L67
Missing a null check for `activeLeaf` here :)

DahaWong:
> https://github.com/DahaWong/obsidian-completed-area/blob/bdb8b38ed5e354416e48631c28a4630f8fd42a09/src/main.ts#L55
> You probably want some kind of debounce here instead of a setTimeout?
> 
> https://github.com/DahaWong/obsidian-completed-area/blob/bdb8b38ed5e354416e48631c28a4630f8fd42a09/src/main.ts#L28
> Can you provide a way to disable your plugin? Right now if you enable then disable your plugin, it won't actually disable.
> I recommend setting a flag `disabled` in the `onunload` function and returning on those events if `disabled` has been set. This is because keeping track of the CodeMirror instances to unhook the events will prevent garbage collection, which will cause a memory leak.
> 
> https://github.com/DahaWong/obsidian-completed-area/blob/bdb8b38ed5e354416e48631c28a4630f8fd42a09/src/main.ts#L67
> Missing a null check for `activeLeaf` here :)

Thanks for your advice. I have to admit that it's not a good idea to listen to `keydown/up` at CodeMirror all the time, I'll fix those codes related.

PR #80, opened 2020-12-09
Add Citations plugin

hans:
Hi, I've just created a very basic integration with Zotero. Added here. Thanks devs for all the hard work!


lishid:
Great work! Would suggest making auto-formatting your code a habit :) since there's a lot of spacing inconsistencies around the files

PR #82, opened 2020-12-11
Add tab anywhere plugin

nadavspi:
https://github.com/nadavspi/obsidian-tab-anywhere-plugin

Not sure if there's already a way to configure Code Mirror via the UI, but I didn't find it. 
Let me know if you have a suggestion for how to set the option right away since it seems like the codemirror event only executes when CodeMirror is first initialized.


nadavspi:
Oops, found a bug.

PR #87, opened 2020-12-17
add theme 2042

StefanoCecere:


ericaxu:
Hi Stefano, thanks for the pull request!

If I'm not mistaken, it seems like you only changed some colors and widths and it's mostly still the same as the default dark theme. I'm afraid it's not significant enough to be included in the community theme list, sorry!

StefanoCecere:
ok, no problem. you know: the default dark isn't the best... but community ones are customized too much.
since Obsidian core doesn't allow to change the base colors (H, em, strong, links, font and width) and, more important, if you have a dozen vaults it is a pain to keep them syncronized with a custom theme, the only solution was to have a "community" theme!

ericaxu:
Ah, I see.

I'm not sure how you're synchronizing your vaults, is it possible for your sync service to ignore frequently changes files in `.obsidian` folder such as `workspace` but still sync the rest? 🤔

PR #89, opened 2020-12-18
Add Extended MathJax plugin

xldenis:
Adds an Extended MathJax plugin which has a fixed global preamble as well as support for additional mathjax packages (for now `mdchem` and `bussproofs`). I'm reserving the name `obsidian-latex` because I'm working on an additional feature which would allow actually calling latex for math rendering instead of mathjax in a future release.


Vinzent03:
I know this is the wrong repo, but @xldenis has issues deactivated. If you try to take a look at the readme in obsidian store, you get a `Plugin ID mismatch`, because the id in this repo is different to the one in his `manifest.json`.

xldenis:
Sorry about that!

PR #90, opened 2020-12-19
Add Apple Reminders Plugin

urishiraval:
Hey,

Please approve my [Plugin](https://github.com/urishiraval/obsidian-apple-reminders-plugin) for the Obsidian Folio.

Warm Regards,
A Huge Fan of your Team's Work.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -464,5 +464,13 @@
         "author": "Xavier Denis",
         "description": "Enables additional MathJax packages and adds a global preamble for MathJax",
         "repo": "xldenis/obsidian-latex"
+    },
+    {
+        "id": "obsidian-apple-reminders-plugin",
+        "name": "Apple Reminders",
+        "author": "Rishi Raval",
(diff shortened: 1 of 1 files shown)


lishid:
Hooking recursive mutation observers on the whole workspace might be a performance issue. Might want to find a better way to do things.

urishiraval:
@lishid I see...

I'm basically copying the method that the todoist plugin uses.

But is there any alternative that you can think of that would work better?

lishid:
Hmm just looking at the way it's working, I'm guessing it might not work well in a document that's long enough such that you can scroll the item out of view, and when you do, the renderer will unload itself.

We internally have optimizations in preview mode that removes any elements outside of view range (with a bit of buffer space).

lishid:
To be honest, I can't think of a great solution right now to manage lifecycles of rendered parts yet. I'll think of a better way to help manage lifecycle in the future.

urishiraval:
I see....I'm sure it will be a worthy addition to the Dev environment. Looking forward to it!

Thanks @lishid

PR #92, opened 2020-12-20
Add Neo4j Graph View plugin

HEmile:
Repo: https://github.com/HEmile/semantic-markdown-converter
EDIT: Tested on macOS Catalina 10.15.7 and Windows 10


lishid:
https://github.com/HEmile/semantic-markdown-converter/blob/4d6678bd2f8f5601e8b5f46407c6db7a6ad8154c/obsidian%20plugin/main.ts#L33

Recommend you to use an interface instead of a class for your settings here - loadData returns the result of a JSON.parse() which is a plain object.

PR #97, opened 2020-12-25
Adding PDF to Markdown plugin

akaalias:
Hi @lishid! 

I've put this PDF to Markdown generator together and think the community would enjoy using it.

Please have a look and if you find anything during QA, please let me and I'll go and fix it. 

Thank you so much, 

Alexis

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -505,5 +505,13 @@
         "description": "A plugin for templating in Obsidian, powered by Nunjucks.",
         "repo": "garyng/obsidian-temple",
         "branch": "master"
+    },
+    {
+        "id": "pdf-to-markdown-plugin",
+        "name": "PDF to Markdown",
+        "author": "Alexis Rondeau",
+        "description": "Save a PDF's text (headings, paragraphs, lists, etc.) to a Markdown file.",
+        "repo": "akaalias/obsidian-extract-pdf",
+        "branch": "main"
     }
 ]


lishid:
https://github.com/akaalias/obsidian-extract-pdf/blob/c11f3c746ddcf76f6e9da372f98e464bf448b414/src/main.ts#L45
Just use `Workspace.getActiveFile()`, and do a null check on it 👍 

https://github.com/akaalias/obsidian-extract-pdf/blob/c11f3c746ddcf76f6e9da372f98e464bf448b414/src/main.ts#L47
Use `file.extension === 'pdf'` instead, which normalizes the extension case for you.

https://github.com/akaalias/obsidian-extract-pdf/blob/c11f3c746ddcf76f6e9da372f98e464bf448b414/src/main.ts#L59

Use this instead of passing the full file path:
```
let arrayBuffer = await app.vault.readBinary(file);
let result = await pdfjsLib.getDocument(arrayBuffer).promise;
```

https://github.com/akaalias/obsidian-extract-pdf/blob/c11f3c746ddcf76f6e9da372f98e464bf448b414/src/main.ts#L63
Any idea why you're using the `.then()` instead of awaiting the promise?

https://github.com/akaalias/obsidian-extract-pdf/blob/c11f3c746ddcf76f6e9da372f98e464bf448b414/src/main.ts#L85
Set the second parameter to empty string here. `sourcePath` means from which document this linktext should be resolved from. It's typically used to resolve relative paths but there no guarantee it won't be used for other purposes.

https://github.com/akaalias/obsidian-extract-pdf/blob/c11f3c746ddcf76f6e9da372f98e464bf448b414/src/main.ts#L103
Forgot to implement this 😄 

https://github.com/akaalias/obsidian-extract-pdf/blob/c11f3c746ddcf76f6e9da372f98e464bf448b414/src/main.ts#L110
Probably not needed?

lishid:
We also include a full copy of pdf.js, but it's lazy-loaded. Perhaps we should expose that to reduce the size and redundancy of your plugin's dependency?

akaalias:
Hi @lishid, thank you for your QA suggestions. I've implemented them today. To your question about pdf.js – I think that's a good idea!

- [x]  https://github.com/akaalias/obsidian-extract-pdf/blob/c11f3c746ddcf76f6e9da372f98e464bf448b414/src/main.ts#L45 Just use Workspace.getActiveFile(), and do a null check on it 👍
- [x] https://github.com/akaalias/obsidian-extract-pdf/blob/c11f3c746ddcf76f6e9da372f98e464bf448b414/src/main.ts#L47 Use file.extension === 'pdf' instead, which normalizes the extension case for you.
- [x] https://github.com/akaalias/obsidian-extract-pdf/blob/c11f3c746ddcf76f6e9da372f98e464bf448b414/src/main.ts#L59 Use this instead of passing the full file path:
```
let arrayBuffer = await app.vault.readBinary(file);
let result = await pdfjsLib.getDocument(arrayBuffer).promise;
```
- [x] https://github.com/akaalias/obsidian-extract-pdf/blob/c11f3c746ddcf76f6e9da372f98e464bf448b414/src/main.ts#L63 Any idea why you're using the .then() instead of awaiting the promise?
- [x] https://github.com/akaalias/obsidian-extract-pdf/blob/c11f3c746ddcf76f6e9da372f98e464bf448b414/src/main.ts#L85 Set the second parameter to empty string here. sourcePath means from which document this linktext should be resolved from. It's typically used to resolve relative paths but there no guarantee it won't be used for other purposes.
- [x] https://github.com/akaalias/obsidian-extract-pdf/blob/c11f3c746ddcf76f6e9da372f98e464bf448b414/src/main.ts#L103
Forgot to implement this 😄
- [x] https://github.com/akaalias/obsidian-extract-pdf/blob/c11f3c746ddcf76f6e9da372f98e464bf448b414/src/main.ts#L110
Probably not needed?

PR #98, opened 2020-12-26
Add move text plugin

DomasB:
A simple plugin to move line or selection using keyboard shortcuts.
https://github.com/DomasB/obsidian-move-text

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -505,5 +505,13 @@
         "description": "A plugin for templating in Obsidian, powered by Nunjucks.",
         "repo": "garyng/obsidian-temple",
         "branch": "master"
+    },
+    {
+        "id": "obsidian-move-text",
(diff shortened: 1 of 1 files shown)


lishid:
Just curious how this is different from the "Swap line up"/"Swap line down" hotkeys that are already present. Sounds like they're similar? They're just not bound to any keys by default.

Let me know 😄

DomasB:
@lishid, not different at all :D
Would say that naming of the built in command is not very discoverable. When searching for this functionality the first word that comes to mind is "move" and "swap" up/down sounds a bit unnatural as a sentence.

Closing this PR.

lishid:
Ah ok! That sounds fair. Sorry you had to go through all the trouble!

DomasB:
No problem :) was good introduction to the API. Might write more plugins later.

PR #106, opened 2021-01-03
Add autocomplete plugin

Yeboster:


lishid:
https://github.com/Yeboster/autocomplete-obsidian/blob/33de5d0b02d31963c86b87dced7cdce6b1216911/src/main.ts#L34

Use the new CodeMirror hook: https://github.com/obsidianmd/obsidian-sample-plugin/blob/cfe4d17ce7dd3346af83e5578d87556f97285951/main.ts#L45

Also your plugin does not unhook the CodeMirror keyup event. Should probably do that using this utility:
https://github.com/obsidianmd/obsidian-api/blob/3b187dd5322c16e5970b17b60bc59daac838723f/obsidian.d.ts#L2725

https://github.com/Yeboster/autocomplete-obsidian/blob/33de5d0b02d31963c86b87dced7cdce6b1216911/src/main.ts#L83
You should probably use this API, which properly null-check `activeLeaf`
https://github.com/obsidianmd/obsidian-api/blob/3b187dd5322c16e5970b17b60bc59daac838723f/obsidian.d.ts#L2690

Hmm actually inspecting the code, I think this would fail if you trigger the autocomplete, then click on another pane - `getCurrentEditor` isn't guaranteed to return the previously "attached" CodeMirror instance which means you will have a dangling CodeMirror that has your keyboard overrides.

Yeboster:
Hello @lishid,
Thanks for the feedback! I've changed the code with your suggestions and fixed the hanging hooks.
I've also tried to manage the multiple pane layout by checking if the editor is the same as the one that triggered the autocomplete, if not I remove the view (and also the keybindings).

I want to refactor the code properly to remove some workarounds and improve the general structure and flow (by taking the general workflow from the [Codemirror hint addon](https://github.com/codemirror/CodeMirror/tree/master/addon/hint), but right now I'm in the exam session so I cannot dedicate so much time to this. Still, I wanted to publish it to have some feedback from the users.

I'd like to hear from you if you have some advice.

lishid:
Nice, those are some great fixes. Only minor things left:

https://github.com/Yeboster/autocomplete-obsidian/blob/a45cfa23ff0757c3ed9813fb292fb970b6a22eb0/src/main.ts#L84
The `.bind` isn't doing any good here. `keyupListener` is already a bound function (by the use of `() => {}` syntax).

https://github.com/Yeboster/autocomplete-obsidian/blob/a45cfa23ff0757c3ed9813fb292fb970b6a22eb0/src/main.ts#L106
Small typescript nit: Function signature says you'll return null but you return undefined instead...

https://github.com/Yeboster/autocomplete-obsidian/blob/a45cfa23ff0757c3ed9813fb292fb970b6a22eb0/src/autocomplete.ts#L65
`Node.firstChild` is probably what you're looking for here

https://github.com/Yeboster/autocomplete-obsidian/blob/a45cfa23ff0757c3ed9813fb292fb970b6a22eb0/src/autocomplete.ts#L75
We have a helper function called `HTMLElement.toggleClass(cls: string, value: boolean)` that does exactly what you're trying to do here. Just do `child.toggleClass('is-selected', index === selectedIndex);`.

https://github.com/Yeboster/autocomplete-obsidian/blob/a45cfa23ff0757c3ed9813fb292fb970b6a22eb0/src/autocomplete.ts#L91
You might be interested in https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

Yeboster:
Thanks for the suggestions, highly appreciated!
> https://github.com/Yeboster/autocomplete-obsidian/blob/a45cfa23ff0757c3ed9813fb292fb970b6a22eb0/src/main.ts#L84
> The `.bind` isn't doing any good here. `keyupListener` is already a bound function (by the use of `() => {}` syntax).

Good catch! I was experimenting with `.bind` to use class functions but that was ugly (because I'd have stored the bind ref into a global variable). So kept the lambda as global ref and missed that bind.

> https://github.com/Yeboster/autocomplete-obsidian/blob/a45cfa23ff0757c3ed9813fb292fb970b6a22eb0/src/main.ts#L106
> Small typescript nit: Function signature says you'll return null but you return undefined instead...

Fixed! That hurt my eyes too 😆 

> https://github.com/Yeboster/autocomplete-obsidian/blob/a45cfa23ff0757c3ed9813fb292fb970b6a22eb0/src/autocomplete.ts#L65
> `Node.firstChild` is probably what you're looking for here

That was exactly what I was looking for. Instead of `firstChild`, I've used [firstElementChild](https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/firstElementChild) which returns the first HTMLElement.

> https://github.com/Yeboster/autocomplete-obsidian/blob/a45cfa23ff0757c3ed9813fb292fb970b6a22eb0/src/autocomplete.ts#L75
> We have a helper function called `HTMLElement.toggleClass(cls: string, value: boolean)` that does exactly what you're trying to do here. Just do `child.toggleClass('is-selected', index === selectedIndex);`.

Didn't know of that helper; I'm going to use it from now on.
 
> https://github.com/Yeboster/autocomplete-obsidian/blob/a45cfa23ff0757c3ed9813fb292fb970b6a22eb0/src/autocomplete.ts#L91
> You might be interested in https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

Initially, I was using `scrollIntoView`, but unfortunately, it didn't work well with nested scrolls because in Obsidian there's the pane scroll and then the autocomplete inside it. So when the DOM tries to scroll to it, it scrolls both the parent and the child. So I've looked a bit into the [CM Hint addon](https://github.com/codemirror/CodeMirror/tree/master/addon/hint) and they've used the [getBoundingClientRect](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) as a solution. I'll look forward to using that logic, but if you have a different solution, please let me know.

lishid:
> Initially, I was using scrollIntoView, but unfortunately, it didn't work well with nested scrolls because in Obsidian there's the pane scroll and then the autocomplete inside it. So when the DOM tries to scroll to it, it scrolls both the parent and the child. So I've looked a bit into the CM Hint addon and they've used the getBoundingClientRect as a solution. I'll look forward to using that logic, but if you have a different solution, please let me know.

Ah ok, that makes sense... well... too bad 😆

PR #107, opened 2021-01-04
Adds PDF Highlights plugin

akaalias:
Hi @lishid!

I'm super excited about this plugin and would love to share it with the community. 

Since we spoke about using Obsidian's PDFJS – In this case here, that won't be possible because I had to create a really customized version of PDFJS in order to retrieve the annotation text. I'm using my fork from `https://github.com/akaalias/pdfjs-dist` to get it done. 

Question here: How can I import it so it doesn't overload Obsidian's PDFJS? 

Currently I'm doing: 

``` 
import pdfjsCustom from "node_modules/pdfjs-dist/build/pdf";
import worker from "node_modules/pdfjs-dist/build/pdf.worker.entry";
... 

pdfjsCustom.GlobalWorkerOptions.workerSrc = worker;
...
var loadingTask = pdfjsCustom.getDocument(arrayBuffer);
...
```

I've also re-used your suggestion on file-handling from before:

```
let file = this.app.workspace.getActiveFile();
if (file === null) return;
if (file.extension !== 'pdf') return;
let arrayBuffer = await this.app.vault.readBinary(file);
```

Much nicer. 

Anyways, when you do have the time, please QA the plugin over at https://github.com/akaalias/obsidian-extract-pdf-highlights and let me know what I can do to launch it!

Thank you so much, 

Alexis

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -458,7 +458,7 @@
         "author": "Zach Raines",
         "description": "Export vault files in a format amenable to pasting into a TeX document",
         "repo": "raineszm/obsidian-export-to-tex"
-    }, 
+    },
     {
         "id": "obsidian-latex",
         "name": "Extended MathJax",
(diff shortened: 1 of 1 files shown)


lishid:
I see that you're using `main` on your repo - might need to add it here, or push a `master` branch instead :)

lishid:
As long the PDF viewer works with your plugin loaded, I don't think there'd be any other issues 👍

akaalias:
Thank you for your review!

Yes, the whole 'main' branch thing is new to me. I've used the default instructions from github where they changed it to main but have changed the default branch now to 'master' -> https://github.com/akaalias/obsidian-extract-pdf-highlights.

My pdfjs doesn't interfere with the PDF viewer and rendering. Just wonder if there's a way to keep them completely separated.

lishid:
I don't know if it's separated or not, but if it does not interfere I think there's not much to worry about.

Also it seems like your new commit rewrote the whole file. Must be a line ending issue - could you push a new one to fix that?

akaalias:
Ah, I used the web-ui to resolve the conflict. Even though I only edited the last 10 lines or so, it re-wrote the whole file. Sorry about that. 

Do you know how I can revert that in my branch? I'm such a git-klutz :\ Or we close this PR and I create a fresh one. Let me know what's better for you!

akaalias:
I reset to c251aa44217079e44b66077d445e3fca31c9c430, edited and force-pushed to my fork.

PR #108, opened 2021-01-04
Adding completed-task-display

heliostatic:


lishid:
Great stuff!

https://github.com/heliostatic/completed-task-display/blob/8fd9c897caacb337232cff108ac2ce753abce86a/main.ts#L2
This doesn't seem used, any idea what that's for? I see it's not bundled in to the release file, so that's good.

https://github.com/heliostatic/completed-task-display/blob/8fd9c897caacb337232cff108ac2ce753abce86a/main.ts#L22
It doesn't seem like the setting is being applied on load. In either case I think your plugin should do fine without a setting.
I would suggest just wiping everything about setting and just all the functionality within the ribbon icon code :)

heliostatic:
> Great stuff!
> 
> https://github.com/heliostatic/completed-task-display/blob/8fd9c897caacb337232cff108ac2ce753abce86a/main.ts#L2
> This doesn't seem used, any idea what that's for? I see it's not bundled in to the release file, so that's good.
> 
> https://github.com/heliostatic/completed-task-display/blob/8fd9c897caacb337232cff108ac2ce753abce86a/main.ts#L22
> It doesn't seem like the setting is being applied on load. In either case I think your plugin should do fine without a setting.
> I would suggest just wiping everything about setting and just all the functionality within the ribbon icon code :)

I've removed the unused `zlib` and settings in https://github.com/heliostatic/completed-task-display/releases/tag/1.0.2

PR #109, opened 2021-01-05
add youhavebeenstaring plugin

fxal:
Hi there,
I've created a minimalistic reminder in the status bar for how long the vault is open.
All the best,
Felix

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -553,11 +553,18 @@
         "description": "Fine-grained control of Discordian Theme",
         "repo": "radekkozak/discordian-plugin"
     },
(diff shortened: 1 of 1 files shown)


lishid:
That's a fun plugin!

Recommend adding semicolons 😆 because it does make a difference in some cases, read the answer here:
https://stackoverflow.com/a/1169596

Also sorry about this, but you'll have to rebase the commit to insert your plugin to the json again because of the merge conflict.

fxal:
Thanks for the feedback!
There is now a next version with semicolons :)

PR #111, opened 2021-01-10
Add obsidian-query2table

avirut:
https://github.com/avirut/obsidian-query2table

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -591,6 +591,13 @@
         "repo": "fxal/obsidian-youhavebeenstaring-plugin"
     },
     {
+        "id": "obsidian-metatemplates",
+        "name": "metatemplates",
+        "author": "avirut",
+        "description": "Generate notes from templates using YAML front-matter",
+        "repo": "avirut/obsidian-metatemplates"
+    },  
+    {
         "id": "obsidian-query2table",
         "name": "query2table",
         "author": "avirut",
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -589,5 +589,12 @@
         "author": "Felix Almer",
         "description": "Tells you in the status bar for how long you've been staring at your obsidian vault. Well - at least how long your vault is open.",
         "repo": "fxal/obsidian-youhavebeenstaring-plugin"
+    },
+    {
+        "id": "obsidian-query2table",
+        "name": "query2table",
+        "author": "avirut",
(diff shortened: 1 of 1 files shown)


lishid:
Couple of suggestions:

- Auto-format your code so you don't have a mix of spaces & tabs (see https://github.com/avirut/obsidian-query2table/blob/master/main.ts for how bad they look)
- [Code](https://github.com/avirut/obsidian-query2table/blob/790451b00f038016cba2ff5d4dd6eb151acd9044/main.ts#L32) I think I already suggested this in discord, but don't do that - instead, create your postProcessor in the `onload` and use `Plugin.registerMarkdownPostProcessor` instead of `MarkdownPreviewRenderer.registerPostProcessor`. This takes care of unloading for you, so you can then remove the call to `MarkdownPreviewRenderer.unregisterPostProcessor`.
  
  Creating the function inside `onload()` would bind `this` to the plugin, so you won't have to use [this hack](https://github.com/avirut/obsidian-query2table/blob/790451b00f038016cba2ff5d4dd6eb151acd9044/main.ts#L47) to find a reference to your plugin.
- [Missing null check on `fm`](https://github.com/avirut/obsidian-query2table/blob/790451b00f038016cba2ff5d4dd6eb151acd9044/main.ts#L74)
- Might want to add a disclaimer that this plugin takes over the search pane when rendering, which resets the user's search. This also means that having two `query2table` blocks open (whether in the same file, or in two panes) won't work at all.

Overall, pretty cool project! I think this could be made into a `View` instead of being part of the markdown spec, similar to a search pane, but a more advanced one. It'd certainly require a lot more re-work and pretty much an entire rewrite though.

avirut:
I've updated source and created a new release addressing (I think all of) the issues you mentioned here. I also just wanted to note that while you're correct that I can't run two `query2table` blocks in the same file, it does seem to work with two different files in different panes even when they are both simultaneously open. I imagine this might be because it doesn't seem to re-render on every update, but only when changes are made to the codeblock itself, so once a query has been compiled to a list of files, those files must be remaining available for the lifetime of that codeblock. Not sure though, as I don't really have a good understanding of what's going on behind the scenes.

lishid:
Cool. Now you'll just have to resolve the merge conflict with your other plugin, and I can merge it :)

(beware that doing this on github's interface might lead to a switch in line endings and cause the entire file to be changed)

PR #112, opened 2021-01-10
Add obsidian-metatemplates

avirut:
https://github.com/avirut/obsidian-metatemplates

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -589,5 +589,12 @@
         "author": "Felix Almer",
         "description": "Tells you in the status bar for how long you've been staring at your obsidian vault. Well - at least how long your vault is open.",
         "repo": "fxal/obsidian-youhavebeenstaring-plugin"
-    }
+    },
+    {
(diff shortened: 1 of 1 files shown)


lishid:
Quick review:

- Same suggestion as the other plugin, auto-format the code would help a lot
- [Not sure if this is a good idea... `---` can be used for headings or as `<hr>`](https://github.com/avirut/obsidian-metatemplates/blob/9e2636bdba1531f581c7dd0df2e4d6e0d6537ffc/main.ts#L157)
- [I recommend reading the template files when you construct the modal instead of having to maintain an up-to-date list of templates](https://github.com/avirut/obsidian-metatemplates/blob/9e2636bdba1531f581c7dd0df2e4d6e0d6537ffc/main.ts#L78)
- [Recommend always using `===` everywhere](https://github.com/avirut/obsidian-metatemplates/blob/9e2636bdba1531f581c7dd0df2e4d6e0d6537ffc/main.ts#L55)

PR #113, opened 2021-01-11
Add tgrosinger/recent-files-obsidian

tgrosinger:


lishid:
https://github.com/tgrosinger/recent-files-obsidian/blob/50620674f2e7b1540b573fcb7862a3b59d47e8fe/main.ts#L14
Might want to call your interface name a bit different name, since File is something that confuses with the browser API.

https://github.com/tgrosinger/recent-files-obsidian/blob/50620674f2e7b1540b573fcb7862a3b59d47e8fe/main.ts#L63
You can do this in `load`, which will unregister the event on unload automatically:
```
	load() {
		super.load();
		this.registerEvent(this.app.workspace.on('file-open', this.update));
	}
```

https://github.com/tgrosinger/recent-files-obsidian/blob/50620674f2e7b1540b573fcb7862a3b59d47e8fe/main.ts#L73
Try our utilities `createDiv('nav-folder mod-root')`. You can substitute `createXX({cls: 'classes'})` with just the `createXX('classes')` if you aren't setting anything else like `text`.

https://github.com/tgrosinger/recent-files-obsidian/blob/50620674f2e7b1540b573fcb7862a3b59d47e8fe/main.ts#L103
I think you're looking for `.unshift()`.

https://github.com/tgrosinger/recent-files-obsidian/blob/50620674f2e7b1540b573fcb7862a3b59d47e8fe/main.ts#L120
`openedFile` could be null. Make sure you null checks here and also in `updateData`.

https://github.com/tgrosinger/recent-files-obsidian/blob/50620674f2e7b1540b573fcb7862a3b59d47e8fe/main.ts#L91
I'd recommend just use `navFile.onClickEvent((evt) => this.handleFileClick(evt, currentFile));` so that you don't have to try and find the file from the DOM element.

https://github.com/tgrosinger/recent-files-obsidian/blob/50620674f2e7b1540b573fcb7862a3b59d47e8fe/main.ts#L242
Consider making it a base64 url so it always works, and doesn't send requests when a suspicious user is doing a network capture :)

tgrosinger:
Wow, thank you for the code review @lishid! I'll let you know when I've made those changes.

tgrosinger:
Done. Those were really good suggestions, thank you.

The only one I did not change was about using `createDiv` when making `rootEl`. I don't want to use `createDiv` there because it would attach it to the child immediately, but I would prefer to build it up first, then swap it out with the existing contents. Kind of like a double-buffer. Let me know if you disagree though.

lishid:
You can call `createDiv` bare (there's a global one) that doesn't attach it to anything.

tgrosinger:
Ah perfect. Switched to using that.
Thanks

lishid:
Great work as always!

PR #115, opened 2021-01-13
Add Imgur Plugin

gavvvr:


lishid:
https://github.com/gavvvr/obsidian-imgur-plugin/blob/c3cf8116758bf39fb6a84557dd61883bbe37c3cc/main.ts#L13
Suggest using Object.assign [like we do it in the sample plugin](https://github.com/obsidianmd/obsidian-sample-plugin/blob/master/main.ts#L61).

https://github.com/gavvvr/obsidian-imgur-plugin/blob/c3cf8116758bf39fb6a84557dd61883bbe37c3cc/main.ts#L48
Doesn't seem like you're proxying the paste event anymore - does this mean the original paste event won't work anymore?

https://github.com/gavvvr/obsidian-imgur-plugin/blob/c3cf8116758bf39fb6a84557dd61883bbe37c3cc/main.ts#L45
This function is called each time you change the config, but it's never unhooked properly.
The underlying implementation is just:
```ts
this.app.workspace.iterateCodeMirrors(callback);
this.registerEvent(this.app.workspace.on('codemirror', callback));
```
You should probably register the event once onload instead of registering more events every time the setting changes.

Actually - You should probably just always register the event, and in the handler, only do something if `clientId` is set. Instead of binding and unbinding the original handler.

gavvvr:
Hi @lishid. Thanks for the review!

- `Object.assign` - 👌🏻
- If the image gets uploaded remotely, I do not want it to also be pasted to filesystem. So yes, I get rid of default behaviour and preserve it in `cmAndHandlersMap` to restore when plugin gets unloaded.
- I see the problem with calling `registerCodeMirror` each time config changes
> You should probably just always register the event, and in the handler, only do something if clientId is set
Agree. Please see this patch: https://github.com/gavvvr/obsidian-imgur-plugin/commit/590244f7ec8f4a819af680f37add114a59056bf1
Now seems to be correct and safe. Let me know if it is fine and I will push it to main branch.

lishid:
> If the image gets uploaded remotely, I do not want it to also be pasted to filesystem. So yes, I get rid of default behaviour and preserve it in cmAndHandlersMap to restore when plugin gets unloaded.

I'm referring to when the user is not pasting an image - the other behavior of pasting HTML (converted to markdown) gets broken. I think what you'll want to do here is to check for image uploads - if there aren't any, then use the default behavior. If there are then skip calling the default handler like you are doing right now.

gavvvr:
Did not know Obsidian converts HTML to MarkDown. Cool!
Then here is another change: https://github.com/gavvvr/obsidian-imgur-plugin/commit/932062901b9328a5b182b6d646f7741a002b589b

lishid:
Looks good!

PR #116, opened 2021-01-14
Add Checklist plugin

delashum:


lishid:
Great plugin!

https://github.com/delashum/obsidian-checklist-plugin/blob/edd154ab05d0b66c2a96bf776b263810a3f232a4/src/svelte/App.svelte#L41
Might want to turn this into a debounce instead of a setTimeout, so when there are many file changes (for example, git, or sync), your plugin doesn't set many timers to re-compute multiple times.

https://github.com/delashum/obsidian-checklist-plugin/blob/edd154ab05d0b66c2a96bf776b263810a3f232a4/src/view.ts#L29
I don't think there's an `onDestroy` hook. You should probably cleanup in `onClose` instead.

delashum:
Yep you're right on `onDestroy`, mixed that up with svelte. I'll get that fixed.

With the timeout, It wasn't intended to replace a debounce. In the `on('modify')` callback it seems like it's called before the fileCache is updated, so the timeout was to give the cache time to update before i reparse everything.

Also, i may be incorrect here as there is no official documentation on this, but it seems like the `on('modify')` is already debounced. This is purely based on my observations of it's behavior. Do you know anything about that? If it is the case that modifications on many files don't debounce, then i definitely see value in also debouncing this.

lishid:
> With the timeout, It wasn't intended to replace a debounce. In the on('modify') callback it seems like it's called before the fileCache is updated, so the timeout was to give the cache time to update before i reparse everything.

Ah you are correct, MetadataCache resolves after the modify event. There's an unexposed event called `resolve` on MetadataCache that you can listen for instead of the `modify` event on Vault, which fires after the cache has been recomputed.

> Also, i may be incorrect here as there is no official documentation on this, but it seems like the on('modify') is already debounced. This is purely based on my observations of it's behavior. Do you know anything about that? If it is the case that modifications on many files don't debounce, then i definitely see value in also debouncing this.
I don't believe it is. You might be seeing the debounced saving behavior from the editor, but if you use an external editor that can modify many files at once (like mass search & replace), or you use git, or dropbox to sync multiple files at once, it could certainly happen.

delashum:
Alright, I released `1.0.1` which includes the bug fix for `onClose`

lishid:
Alright. If you can fix the merge conflict, I'll pull it in!

delashum:
fixed! thanks

lishid:
Seems like your plugin's ID here (obsidian-checklist-plugin) is different that in your repo (obsidian-checklist), which is causing download failure.

delashum:
Yep, totally missed that. Just created a new release with the correct ID

PR #117, opened 2021-01-15
Add page heading from links plugin

beet:
Adds https://github.com/beet/page-headings-obsidian-plugin

I like to create new pages by first inserting a link into the parent page like `[[My child page]]`, and then clicking on the link to create it. I then have to manually insert a heading matching the filename like `# My child page`.

With this plugin enabled, a `file-open` callback constructs a heading any time a blank file is opened, either from having just created it from a link, or it already existing, and inserts it into the page so it's one less thing to do.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -653,5 +653,6 @@
         "author": "Mark Beattie",
         "description": "Inserts a heading into blank pages from the filename",
         "repo": "beet/page-headings-obsidian-plugin",
+        "branch": "main"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -646,5 +646,12 @@
         "author": "Michal Bureš",
         "description": "Convert file path to uri for easier use of links to local files outside of Obsidian",
         "repo": "MichalBures/obsidian-file-path-to-uri"
+    },
+    {
+        "id": "page-heading-from-links",
+        "name": "Page Heading From Links",
+        "author": "Mark Beattie",
+        "description": "Inserts a heading into blank pages from the filename",
+        "repo": "beet/page-headings-obsidian-plugin",
     }
 ]


lishid:
You seem to have committed the `.js` file instead of the `.ts` file 😅 

Also your repo is on `main` which you'll need to add the flag in our JSON file, or you can push your branch to `master` instead.

https://github.com/beet/page-headings-obsidian-plugin/blob/7705620278ce1ae70b342e22f150bd7dc050613b/main.js#L36
You're missing a lot of checks here: activeLeaf could be null, activeLeaf could be a non-markdown type, etc.

Instead, use `Workspace.getActiveViewOfType(MarkdownView)` and perform a null check on that.

https://github.com/beet/page-headings-obsidian-plugin/blob/7705620278ce1ae70b342e22f150bd7dc050613b/main.js#L42
You should probably use `view.file` (that you got from getActiveViewOfType) instead of the one passed in.

https://github.com/beet/page-headings-obsidian-plugin/blob/7705620278ce1ae70b342e22f150bd7dc050613b/main.js#L33
Your comment says "When opening any blank file", but you're really just checking if the first line is empty. Might want to check if `view.data === ''` instead of searching for the content inside the editor.

lishid:
Overall pretty cool plugin! I'm sure many people are looking for exactly this.

beet:
@lishid awesome, thanks for the feedback!

beet:
@lishid I've addressed your suggested changes with https://github.com/beet/page-headings-obsidian-plugin/commit/a43e46205e6a9de624657d753223037b3776cbfd and set the branch to `main`.

lishid:
Looks good. Did you want to also commit the typescript file instead?

If you can please rebase to fix the merge conflict I'll have it merged asap!

beet:
@lishid I've rebased and resolved the merge conflicts, sorry about the two commits; couldn't work out how to squash them.

I'm not primarily a JavaScript developer, so I just followed https://github.com/obsidianmd/obsidian-sample-plugin and the other community plugins in my `.obsidian` dir which all seem to have `main.js`, so I didn't make a TypeScript file I'm afraid.

Is that info outdated, or have I missed a fundamental step in the plugin creation process?

lishid:
Ah so you just wrote the whole thing in the generated `js` file? Interesting...
Well technically that's totally allowed, I have no problems with that.
We just recommend using TypeScript because it provides extra type-checking and helps other people read your source code better. The only downside is that you'll have to install a bunch of node-js libraries to compile the TypeScript code down to JavaScript for it to run. It's up to you though.

PR #118, opened 2021-01-15
Add Text Expander

konodyuk:


lishid:
Recommend doing an auto-format to keep the code clean 😄 

https://github.com/konodyuk/obsidian-text-expander/blob/b73e50a4de3a13ee685538a304110af8f81fd94f/main.ts#L268
You may be interested in our helper functions for generating DOM elements:
`createEl(...)` (create globally) and `HTMLElement.createEl(...)` (create and attach to parent). It can create elements and set its css classes, innertext, etc.

For example, `descEl.appendChild(document.createElement('br'));` becomes `descEl.createEl('br')`

PR #124, opened 2021-01-21
Update community-plugins.json

xpgo:
Add the Folder Note plugin

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -661,5 +661,13 @@
         "description": "Add icons to your notes",
         "repo": "visini/obsidian-icons-plugin",
         "branch": "main"
+    },
+    {
+        "id": "folder-note-plugin",
+        "name": "Folder Note",
+        "author": "xpgo",
+        "description": "Add description note to a folder",
+        "repo": "xpgo/obsidian-folder-note-plugin",
+        "branch": "main"
     }
 ]


lishid:
https://github.com/xpgo/obsidian-folder-note-plugin/blob/422eea49555aaefbc3052273c5835226b0c976b8/main.ts#L70

The correct usage here would be `openLinkText(folderNotePath, '', false, { active: true })`
Same for the other usage below that one.


https://github.com/xpgo/obsidian-folder-note-plugin/blob/422eea49555aaefbc3052273c5835226b0c976b8/main.ts#L64
Might want to make the note name configurable, as well as the content that gets put into it.

Actually, I suggest combining these two functions into a single one with a flag for whether to create the file.

xpgo:
@lishid Thanks for the good suggestions. I am new to TypeScript and Obsidian. I have updated the plugin based on your suggestions and created another pull request.

lishid:
Good work! One last thing I'm confused about:

https://github.com/xpgo/obsidian-folder-note-plugin/blob/23215e0e2b71e73b09465a2c0d45f258f7cf6711/main.ts#L85

I'm not sure what this is for? It doesn't seem to stick if you close and re-open the vault anyway.

xpgo:
I used the code there  to:
1. add a class name 'has-folder-note' to the folder-element which has note, so the folder text can be rendered with different color indicating it has description.
2. add a class name 'is-folder-note' to the folder-note-file-element, so the element will be hidden by css rules.

Since the folder note name can be configurable, I have to select the folder-note-file-element dynamically. It doesn't matter whether the vault is closed and reopen, once the user clicked a folder, these class names will be inserted immediately.

I am not good at CSS and Typescript, did I use the bad way to implement this function?

lishid:
> once the user clicked a folder

Yeah that's what I meant - if you restart the app, the class name won't be added, and therefore it will show the files until you click on the folder again

xpgo:
Yes, I also noticed this issue. Add class names to all the related elements at the loading stage may solve the problem, but it may delay the response if there are many such notes. I think the good way it to implement similar feature inside Obsidian. Or maybe there are other methods. 

However, I don't think it is a very annoying thing right now, because:
1. Actually,  when reopen a vault, all or most of folders are in collapse state, the note file will not be shown. Even the user saved the workspace state,  only one or few folder note file is shown. And it can be hidden again by click the folder.
2. Now the default note name is set to the same as folder name in the plugin, it will looks better at the start.

There maybe some perfect strategy, I will think about it. Thanks.

lishid:
Sounds good, merged!

PR #125, opened 2021-01-21
Add Min3ditorHotkeys plugin

d-sauer:
Additional editor hotkeys inspired by coding editors


lishid:
https://github.com/d-sauer/Obsidian-Min3ditorHotkeys-plugin/blob/aab1be23eaa76f10c567c43360c0f56a4c37f12b/main.ts#L100

You should use `Workspace.getActiveViewOfType(MarkdownView)` and do a null check on it instead. Currently you have no error checking at all here.

https://github.com/d-sauer/Obsidian-Min3ditorHotkeys-plugin/blob/aab1be23eaa76f10c567c43360c0f56a4c37f12b/main.ts#L102

Doesn't seem like it would add a newline if you're selecting some text? Maybe I missed something?

d-sauer:
Thank you for the feedback.
Let me resolve usage of `Workspace.getActiveViewOfType(MarkdownView)`, null check and I'll add an additional check.

I do add new (before/after) if there is a selected text

lishid:
Sounds good.

> I do add new (before/after) if there is a selected text

If you trace through `duplicateLines()`, then `this.getSelectedText(editor)`, then `if (editor.somethingSelected())`, then I don't think there was a newline through this route.

d-sauer:
Hi, just added additional `null` check and refactored a code a bit for better readability of the flow.
Let me know if you have any other concerns?

lishid:
Ok looks good. Your branch is master, so you should probably remove the `"branch": "main"` part from the PR.

d-sauer:
thx, also removed `"branch": "main"` as I'm using `master` name

Vinzent03:
All these hotkeys are already present in hotkeys++ plugin.

PR #126, opened 2021-01-22
Vantage v1.0.0.

ryanjamurphy:
Hope this works! 🤞

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -669,5 +669,13 @@
         "description": "Add description note to a folder",
         "repo": "xpgo/obsidian-folder-note-plugin",
         "branch": "main"
+    },
+    {
+        "id": "vantage-obsidian",
+        "name": "Vantage - Advanced search builder",
+        "author": "ryanjamurphy",
+        "description": "Build advanced search queries in Obsidian",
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/ryanjamurphy/vantage-obsidian/blob/bc1b9319fc5017ae08e2a5037b80300696dec9de/main.ts#L4

Probably shouldn't bundle moment.js in, use the one from `window.moment`.

https://github.com/ryanjamurphy/vantage-obsidian/blob/bc1b9319fc5017ae08e2a5037b80300696dec9de/main.ts#L86

Use `Workspace.getActiveViewOfType(MarkdownView)` and null check the result.

https://github.com/ryanjamurphy/vantage-obsidian/blob/bc1b9319fc5017ae08e2a5037b80300696dec9de/main.ts#L121

Oh wow damn...

Few things that will make your life a lot easier:
- `createDiv(...)` instead of `createEl('div', ...)`
- `createEl/Div({cls: 'class-names'})` instead of `createEl({}); element.addClass('class-names');`
- `parent.createEl/Div()` automatically inserts the child into the parent.

PR #129, opened 2021-01-24
add Sort & Permute lines plugin

Vinzent03:


lishid:
https://github.com/Vinzent03/obsidian-sort-and-permute-lines/blob/4a2b89c93fd8b4b326bda0e6d8a51e97f7818b46/main.ts#L11

This won't work if the user has more than one pane. You should rather get the current active CodeMirror instance when your command is triggered:
```ts
let view = this.app.getActiveViewOfType(MarkdownView);
if (!view) return;

let cm = view.sourceMode.cmEditor;
// Do something with cm
```

https://github.com/Vinzent03/obsidian-sort-and-permute-lines/blob/4a2b89c93fd8b4b326bda0e6d8a51e97f7818b46/main.ts#L84

This seems rather inefficient - remember that your sort function is going to be calling this function O(n log n) times, and you're constructing a new RegExp, and blowing up each line to an array of each individual characters, multiple times per line. This will probably cause a lot of strain on the GC every time you trigger the command.

This also means your sort functions just doesn't work with non-english languages - the lengths will be wrong for languages that uses accented characters like `é`, and they're going to be 0 for east-asian languages like Chinese or Japanese.

I suggest doing this instead:

```ts
// Create this globally
const {compare} = new Intl.Collator(undefined, {
	usage: 'sort',
	sensitivity: 'base',
	numeric: true,
});

let sortfunc = (a: string, b: string) => compare(a.trim(), b.trim());

// Use it as the sort function
sortAlphabetically() {
	...
	lines.sort(sortfunc);
	...
}
```

Vinzent03:
Thank you for your advice.
I had the same thought with codemirror today.

My aim with regex was to ignore syntax like `- my line` and `[[link]]`. But your sort method is definitely better. Thank you.
Pushed a hopefully better version.

lishid:
Looks like you found the ignorePunctuation flag which I was going to suggest. Good work!

One last small suggestion I would make is to use `getEditor()` in the command handlers themselves and pass along the reference to CodeMirror to `getLines` and `setLines`, instead of having those functions get the editor themselves.

Also, merged!

PR #131, opened 2021-01-24
Add Things logbook plugin

liamcain:
https://github.com/liamcain/obsidian-things-logbook

This plugin will periodically fetch tasks from the [Things 3](https://culturedcode.com/things/) SQLite DB and insert them into their own section within your daily notes.

Note: The plugin is macOS-only (and requires Things to be installed) so I do an initial upfront check in `onload` for the platform.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -692,10 +692,18 @@
         "repo": "d-sauer/Obsidian-Min3ditorHotkeys-plugin"
     },
     {
-       "id":"obsidian-jsonifier",
-       "name":"JSONifier",
-       "author":"Kjell Connelly",
-       "description":"JSON.stringify() or JSON.parse() highlighted text and copy to clipboard.",
-       "repo":"KjellConnelly/obsidian-jsonifier"
+        "id":"obsidian-jsonifier",
+        "name":"JSONifier",
+        "author":"Kjell Connelly",
+        "description":"JSON.stringify() or JSON.parse() highlighted text and copy to clipboard.",
+        "repo":"KjellConnelly/obsidian-jsonifier"
+    },
+    {
+        "id": "things-logbook",
+        "name": "Things Logbook",
+        "author": "liamcain",
+        "description": "Sync your Things logbook with your daily notes",
+        "repo": "liamcain/obsidian-things-logbook",
+        "branch": "main"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -692,11 +692,11 @@
         "repo": "d-sauer/Obsidian-Min3ditorHotkeys-plugin"
     },
     {
-        "id":"obsidian-jsonifier",
-        "name":"JSONifier",
-        "author":"Kjell Connelly",
-        "description":"JSON.stringify() or JSON.parse() highlighted text and copy to clipboard.",
-        "repo":"KjellConnelly/obsidian-jsonifier"
+        "id": "obsidian-jsonifier",
+        "name": "JSONifier",
+        "author": "Kjell Connelly",
+        "description": "JSON.stringify() or JSON.parse() highlighted text and copy to clipboard.",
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/liamcain/obsidian-things-logbook/blob/ebb33cac248fa9da72c22509d5b7dc6b68eaab21/src/index.ts#L119

Probably best to do this serially instead of all in parallel?

https://github.com/liamcain/obsidian-things-logbook/blob/ebb33cac248fa9da72c22509d5b7dc6b68eaab21/src/index.ts#L206

Look into the new way to load options with Object.assign() in the sample plugin - it helps setting the defaults better.

https://github.com/liamcain/obsidian-things-logbook/blob/ebb33cac248fa9da72c22509d5b7dc6b68eaab21/src/index.ts#L202

An easy way to make sure you don't schedule multiple times is to cancelTimeout right before setting the new one :)

lishid:
[Very interesting string haha](https://github.com/liamcain/obsidian-things-logbook/blob/ebb33cac248fa9da72c22509d5b7dc6b68eaab21/src/constants.ts#L2). Great work nonetheless!

liamcain:
Thanks for the Code Review! Will have some fixes soon.
> Probably best to do this serially instead of all in parallel?

Agreed serial makes more sense than parallel here 👍 

> Look into the new way to load options with Object.assign() in the sample plugin - it helps setting the defaults better.

Two things here

1. I think you might want to update the `sample-plugin` to use the following instead
```
this.settings = Object.assign({}, DEFAULT_SETTINGS, await this.loadData())
```  

As it is now, loading data will mutate the DEFAULT_SETTINGS object which might lead to unexpected results.

2. I didn't do it that way on purpose. I didn't want to save the defaults as user settings unless they specifically overrode them. But actually for this plugin, that might not make the most sense since I don't want to change the default on someone and break their sync... Okay, I'll fix that.

lishid:
> As it is now, loading data will mutate the DEFAULT_SETTINGS object which might lead to unexpected results.

Makes sense, good catch!

lishid:
I've updated the sample plugin. Thanks again.

Let me know when you're ready to have the PR merged.

liamcain:
Okay, just cut a [new release](https://github.com/liamcain/obsidian-things-logbook/commit/f7c4a3b4911d5e4bae5daa63ed331b36d4d35320) with the fixes you suggested (plus some other improvements to the settings UX).

Thanks again for the code review, should be good to go now!

community-plugins.json
```diff
@@ -692,10 +692,18 @@
         "repo": "d-sauer/Obsidian-Min3ditorHotkeys-plugin"
     },
     {
-       "id":"obsidian-jsonifier",
-       "name":"JSONifier",
-       "author":"Kjell Connelly",
-       "description":"JSON.stringify() or JSON.parse() highlighted text and copy to clipboard.",
-       "repo":"KjellConnelly/obsidian-jsonifier"
+        "id": "obsidian-jsonifier",
```
liamcain, line 695: This entry was indented with 3 spaces, just added an extra space before the keys. And another space between the keys and values.

lishid:
Almost there 😆 

https://github.com/liamcain/obsidian-things-logbook/blob/7bc57712a5166202d2ee50073b574e2cdad7f8a1/src/index.ts#L138

This is running it in parallel, so you'll actually be seeing `Things Logbook sync complete` before the sync is actually done.

Maybe just iterate the object with `for...in...hasOwnProperty` instead?

liamcain:
aah, great catch! I was assuming it was out of order because of a delay in the FSWatcher but that doesn't make sense since I'm reading the files via the Vault 🤦

Just updated it with `for...of` [here](https://github.com/liamcain/obsidian-things-logbook/commit/38d3fb76627655b562617bcb0405f61301450b1d).

PR #133, opened 2021-01-25
add "Shortcuts for starred files" plugin.

Vinzent03:
I don't know it it is the best way to acces the starred files, but it works.
(https://github.com/Vinzent03/obsidian-shortcuts-for-starred-files)


lishid:
That'll work, but don't forget starred could be a query instead of a file.

Vinzent03:
Thanks for the tip. Added a catch for that.

PR #136, opened 2021-01-27
Adding Obsidian Query Language plugin

jplattel:
@lishid this is the new pull-request which is properly formatted


lishid:
Thanks!

PR #137, opened 2021-01-28
Added my Markdown WYSIWYG Editor Plugin

Reocin:


lishid:
Pretty cool!

If I understand this correctly, it adds a GUI and a command way of inserting markups for markdown, is that right?
I was just a little confused that it's named "WYSIWYG Editor" because normally WYSIWYG stands for "what you see is what you get", which typically refers to the rendering engine to not show any markup of the language.
Maybe consider renaming the plugin to avoid the confusion?

You also forgot your github user name in the repo string in the PR.

https://github.com/Reocin/obsidian-markdown-wysiwyg-editor-plugin/blob/80d179b787426a2ca16fe086728483f60f434b1c/src/main.ts#L62
I suggest adding a call here to `Workspace.iterateCodeMirrors` to unhook the `cm.on('keyup')` you added earlier, to properly cleanup when unloading. Make sure you're passing the reference to the original function you registered with keyup!

Reocin:
Hello Lishid
Thanks for you replay. Yes you are right. It has a command way an so i should really rename it. At the beginning it wasn't the plan to add this function, but as always plans are changing XD.
I changed the name of the plugin to Markdown-Editor. Its general so it has also potential to grow in future. Is that good for you ?
I already add the unhook function thanks for your input.
regards
Daniel

lishid:
I'll close this and we'll move forward on your new PR #138

PR #138, opened 2021-01-28
Added my new Plugin Markdown Editor (renamed version)

Reocin:
As you suggested i renamed my plugin to Markdown Editor. This time my username (Reocin) should be in the repo.

regards
Reocin

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -734,5 +734,13 @@
         "description": "This plugin allows you to query notes and represent data within Obsidian",
         "repo": "jplattel/obsidian-query-language",
         "branch": "main"
+    },
+    {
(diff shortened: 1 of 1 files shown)


lishid:
Your PR is fine and I see you addressed the suggestions, great work!

I'm still concerned about the name right now, since "markdown editor" still sounds way too vague for an average user.

Think of it from this perspective: When a user is looking at the list of plugins, and they see "markdown editor", they would be super confused since Obsidian is a markdown editor, why is there a markdown editor plugin?

Ideally you should name it something that makes it obvious what the plugin does - something like markdown-formatting-shortcuts would be much better, but feel free to spend some of your creativity with the name 😄

PR #139, opened 2021-01-29
Added my new plugin, renamed version V2

Reocin:
As you wished I renamed it. I took your suggestion but changed shortcut to assistant. What do you think?


Reocin:
Thanks :-)

PR #140, opened 2021-01-31
Removes YAML front matter template plugin

m-orfanos:
The plugin isn't currently not working in the latest build (I can't even get it to run...). I don't have time right now to fix the issue, it will have to wait at least another 2-3 weeks. I'd rather remove it from the list until then so it doesn't affect anyone. Sorry for the trouble.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -555,14 +555,6 @@
         "branch": "master"
     },
     {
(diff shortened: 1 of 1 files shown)


lishid:
Sounds good.

PR #142, opened 2021-01-31
added Better PDF plugin

MSzturc:


lishid:
https://github.com/MSzturc/obsidian-better-pdf-plugin/blob/ea50255f5994f08a2f50ecb85e2075b7c3d53af0/main.ts#L126
Given you aren't unloading anything, you won't need to make a child to manage the lifecycle of this.

Just inline the entire PDF rendering code in your post processor should be fine.

One thing to make sure though:
[Here you're calling a callback which splits the async function](https://github.com/MSzturc/obsidian-better-pdf-plugin/blob/ea50255f5994f08a2f50ecb85e2075b7c3d53af0/main.ts#L143). You should instead use `for (let node of nodes)` to ensure your async function is continuous. [Same here, make this an `await`](https://github.com/MSzturc/obsidian-better-pdf-plugin/blob/ea50255f5994f08a2f50ecb85e2075b7c3d53af0/main.ts#L78).

MSzturc:
Thank you for your constructive feedback. I tried to include all the comments in the code. Please look over it again, you may find further points that need to be improved.

lishid:
That's a good step yeah.

I was probably unclear in my last comment:
I think you should just remove the class `PDFRenderNode` and move everything from its `onload()` inside your post-processor, replacing [this line](https://github.com/MSzturc/obsidian-better-pdf-plugin/blob/b2e20828a16398476f9fd67f8f761fb114cf488c/main.ts#L149).

MSzturc:
I've removed PDFRenderNode class and refactored parameterReading into separate method. Please look over it again.

lishid:
Looks great! Let's get this merged. Seems like there's a conflict so you might have to rebase your PR or send it fresh again.

MSzturc:
All conflicts are now resolved. Feel free to merge my PR

PR #143, opened 2021-02-01
Add obsidian-dataview

blacksmithgu:
Adds the obsidian dataview plugin.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -757,5 +757,12 @@
         "author": "MSzturc",
         "description": "Insert, scale, rotate and cut-out pdf pages into your notes",
         "repo": "MSzturc/obsidian-better-pdf-plugin"
+    },
+    {
+        "id": "dataview",
+        "name": "Dataview",
+        "author": "Michael Brenan",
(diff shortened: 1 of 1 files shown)


lishid:
Impressive work!

Just some small suggestions:

https://github.com/blacksmithgu/obsidian-dataview/blob/4e0e2246ca850ec60d7aff0239535cbacdea7e38/src/main.ts#L27
You'll want to test `Workspace.layoutReady` first - if you enable your plugin after the workspace was ready, your callback will never be fired.

Given that [you could be waiting indefinitely here](https://github.com/blacksmithgu/obsidian-dataview/blob/4e0e2246ca850ec60d7aff0239535cbacdea7e38/src/main.ts#L56) if something goes wrong, you should probably group the whole code of your post processor inside a `MarkdownRenderChild` so that in `onunload` you can set some variable to stop your infinite timeouts.

blacksmithgu:
Thanks for the feedback - I went ahead and fixed both issues and published a new release with them. Also fixed the merge conflict.

lishid:
Great, merged!

PR #145, opened 2021-02-01
Add tag-wrangler

pjeby:


lishid:
It's finally here!
Couple of things I'm curious about:

[What does this do?](https://github.com/pjeby/tag-wrangler/blob/78fe013c33cf39aee00a7b62552bf80c4f1c17cd/src/plugin.js#L15) I can't find another reference for `lastFocus`

Any particular reason you're registering events in capture mode? I've rarely seen people use them except for specific situations.

https://github.com/pjeby/tag-wrangler/blob/78fe013c33cf39aee00a7b62552bf80c4f1c17cd/src/plugin.js#L72
Can probably use addEventListener instead? Though I don't think it matters that much

https://github.com/pjeby/tag-wrangler/blob/78fe013c33cf39aee00a7b62552bf80c4f1c17cd/src/renaming.js#L59
FYI We also do this case insensitive, not that anyone would use `tAgS` but some might use `Tag/Tags`

pjeby:
>What does this do? I can't find another reference for lastFocus

At one point I was trying to restore editor focus upon canceling the menu; it turned out to be futile to even attempt and this was a leftover remnant, so I'll get rid of it.

> Any particular reason you're registering events in capture mode?

My understanding is that capture mode lets you catch an event *before* it goes to child elements.  And I needed it in my (still-WIP) WYSIWYG editor plugin in order to grab click events on todo checkboxes away before CodeMirror could process them, so I got in the habit when overriding things that might already have event handlers.

(Context: my WYSIWYG editor plugin does things like style the `[ ]` and `[x]` in CodeMirror so they look like actual checkboxes, and when you click on them it toggles them.  It also has overrides for link hovering and clicking so that you get hover previews without needing Ctrl, and so you can ctrl-click links to open them.  Over time I expect it to grow more, especially once I get folds working for things like images.)

> Can probably use addEventListener instead? Though I don't think it matters that much

Yeah, probably.  That was pure laziness.  ;-)

> FYI We also do this case insensitive, not that anyone would use tAgS but some might use Tag/Tags

As it happens, I just released 0.2.2 a few minutes ago with proper case insensitivity and improved warning messages for clashes (including not warning of a clash if all you're doing is renaming to make your tags case-consistent).

lishid:
Ok sounds good, let me know when you're ready to have the PR merged.

> My understanding is that capture mode lets you catch an event before it goes to child elements. And I needed it in my (still-WIP) WYSIWYG editor plugin in order to grab click events on todo checkboxes away before CodeMirror could process them, so I got in the habit when overriding things that might already have event handlers.

Yeah that's fair. CodeMirror can pass through the event if you want to handle it before CodeMirror I believe using `cmEditor.on('mousedown', callback)`, you can preventDefault (or return a magic value) to stop CodeMirror from handling it.

pjeby:
Merge any time, the lastfocus thing isn't bothering anything AFAIK.  ;-)  I did check it in on the trunk in case anybody cares, but the next release will be likely be a big refactoring of the innards, and unlikely to land before the weekend.  In the meantime, if there are any bugs, might as well know now.  ;-)

PR #147, opened 2021-02-03
Add Note Folder Autorename plugin

pjeby:
This was an interesting challenge: I had to hook a lot of Obsidian to safely ensure that all links are updated correctly.  The filemanager's renameFile fires off various async processes and it depends on a clean metadata state when it starts.  So doing another rename in response to the operation safely requires waiting for all that to finish.

I *think* I managed it in a reasonably clean way, but if you have any suggestions I'm open to improvements.


pjeby:
Rebased to remove conflicts.

PR #150, opened 2021-02-04
Add Periodic Notes plugin

liamcain:
https://github.com/liamcain/obsidian-periodic-notes

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -764,5 +764,13 @@
         "author": "Michael Brenan",
         "description": "Advanced queries over your vault for the data-obsessed.",
         "repo": "blacksmithgu/obsidian-dataview"
+    },
+    {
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/liamcain/obsidian-periodic-notes/blob/dc4e88b11b9a03e279401b0722e7da59c4fbb234/src/index.ts#L67
Does it make sense to use Object.assign here?

liamcain:
> https://github.com/liamcain/obsidian-periodic-notes/blob/dc4e88b11b9a03e279401b0722e7da59c4fbb234/src/index.ts#L67
Does it make sense to use Object.assign here?

Good point, It's safe without the Object.assign (for initial installs), but I'll want to update this logic if I add new setting keys in an update.

PR #151, opened 2021-02-05
Update community-plugins.json

Darakah:
Submitting a plugin for review


lishid:
https://github.com/Darakah/obsidian-paper-cut/blob/f28d01c445af19effd5d425f165e261aa628ea3d/main.ts#L16
I think you can just use `node.setAttribute()` directly?

https://github.com/Darakah/obsidian-paper-cut/blob/f28d01c445af19effd5d425f165e261aa628ea3d/main.ts#L22
Recommend you to rename `MyPlugin`, `MyPluginSettings`, `SampleSettingTab` to better names 😂 

https://github.com/Darakah/obsidian-paper-cut/blob/f28d01c445af19effd5d425f165e261aa628ea3d/main.ts#L39
Here you should probably just use `status.addClass('papercut')`

https://github.com/Darakah/obsidian-paper-cut/blob/f28d01c445af19effd5d425f165e261aa628ea3d/main.ts#L49
You're testing the "truthy-ness" of this, won't need the `!= undefined`. Just `if (this.app.workspace.getActiveFile())` will suffice. Same for the next check after.

https://github.com/Darakah/obsidian-paper-cut/blob/f28d01c445af19effd5d425f165e261aa628ea3d/main.ts#L52
Don't use cachedData directly. Use `vault.read(file)` if you plan to write back to the file, or `vault.readCached(file)` if you don't.
Also you should probably only grab the file data if your frontmatter filter passes.

https://github.com/Darakah/obsidian-paper-cut/blob/f28d01c445af19effd5d425f165e261aa628ea3d/main.ts#L66
This is bad, don't do this. Instead, add a new field like "statusBarEl" to your class and assign the value you get from `addStatusBarItem` to `this.statusBarEl` which can be then used to get the element.

https://github.com/Darakah/obsidian-paper-cut/blob/f28d01c445af19effd5d425f165e261aa628ea3d/main.ts#L140
Probably want to add saveSettings here and after.

Darakah:
Hey @lishid 

Much thanks for you detailed remarks and explanations.
I have addressed all the mentioned problems and auto-formatted it. Please let me know if anything else still needs to be changed. 

Thanks in advance.

lishid:
https://github.com/Darakah/obsidian-paper-cut/blob/15bac16ff7f90a82bf1e1c3659dfa71a324d069e/main.ts#L17
I think most of these can be inline variables, they aren't being shared for anything else other than in `cut()`

https://github.com/Darakah/obsidian-paper-cut/blob/15bac16ff7f90a82bf1e1c3659dfa71a324d069e/main.ts#L53
You should probably query this once and save it into a variable, instead of copy-pasting this function call 4 times.

https://github.com/Darakah/obsidian-paper-cut/blob/15bac16ff7f90a82bf1e1c3659dfa71a324d069e/main.ts#L59
You should use `vault.read(file)` instead of `vault.adapter.read(file.path)`.
Convert your function into an `async` function and call `await vault.read()`, instead of using .then().


---

General algorithm question: You seem to be testing for the note size `this.page_len > this.settings.PAPER_CUT_LIMIT`, but you trim it according to a different algorithm `tmp_content.replace(/[^a-zA-Z0-9]/g, '').length < this.settings.PAPER_CUT_LIMIT`.

Wouldn't this lead to cases where the file is longer, but the filtered content is shorter, so you keep trimming and overwriting the file with no changes?

Also, this obviously would not work for any non-alphabetical latin languages. Think east asian languages like Chinese or Japanese.
Why not just use the file size anyway?

Darakah:
Thanks for the fast reply.

For `this.page_len > this.settings.PAPER_CUT_LIMIT` i agree this should be modified to avoid that case. 

Not using file size because that would limit the usage of empty lines / spaces / tabs which many use to better format their note. The aim was to consider only information providing characters. And yes, didn't think about those languages. I think the best option would be to simply remove only the \s characters this way even if it is those different languages it would still work. (i went with this because i did not account for all those markdown syntax bold / italic / highlights and so on which if used could greatly increase the apparent page size while the actual information is not increasing ... maybe will add the most comment ones along with the whitespaces) 

Will address the remarks later and send an update. Thanks a lot.

Darakah:
Was just rechecking the code actually `this.page_len` is also under the same trimmed rules https://github.com/Darakah/obsidian-paper-cut/blob/15bac16ff7f90a82bf1e1c3659dfa71a324d069e/main.ts#L60

Darakah:
Thanks for the remarks. I have addressed all the highlighted points and limited the removed characters to whitespaces and the formatting characters such as `-` and `*` which should make it applicable to any language. Adjusted the design of the function checks to a clearer layout.

lishid:
Great. I think it's ready to merge, though you'll have to rebase your PR off master since there are conflicts right now.

Please also auto-format your code! It still has a mix of tabs and spaces.

Darakah:
Done, modified the PR to current master. Many thanks.

PR #152, opened 2021-02-05
Obsidian Comments Plugin

Darakah:


lishid:
- Suggest running main.ts through an auto-formatter.
- Don't use `var`, always use `let` or `const`.
- Don't cast to `any` unless you have a good reason to. If you're casting anything from obsidian's API to `any` you're likely doing something wrong.

https://github.com/Darakah/obsidian-comments-plugin/blob/03a6d57b78a4afe2b0f687eb18348eff41727ba4/main.ts#L100
Same as other plugin, don't use this.

https://github.com/Darakah/obsidian-comments-plugin/blob/03a6d57b78a4afe2b0f687eb18348eff41727ba4/main.ts#L96
Consider naming your variables properly. Also consider creating them inline where they're needed so they are scoped to the correct `{}` block, e.g. `for(let i = 0 ...)`.

https://github.com/Darakah/obsidian-comments-plugin/blob/03a6d57b78a4afe2b0f687eb18348eff41727ba4/main.ts#L16
I don't see the need for these 3 helpers, especially given we have better enhancements for the DOM creation API:
- `createDiv/createSpan/createEl()` to create elements.
- `parentElement.createDiv/Span/El` to create elements and attach them to the `parentElement`.
- `createEl({ ... })` has tons of options to set the inner text, attributes, css-classes, etc.

lishid:
Recommend remove or justify (in a `//` comment) if you use `as any`.

https://github.com/Darakah/obsidian-comments-plugin/blob/03a6d57b78a4afe2b0f687eb18348eff41727ba4/main.ts#L254
What's `(this as any).view` used for?

https://github.com/Darakah/obsidian-comments-plugin/blob/03a6d57b78a4afe2b0f687eb18348eff41727ba4/main.ts#L290
`getEditor` once and reuse?

Darakah:
Dear Licat,

I have addressed all of the problems highlighted here and from the other plugin review and updated the plugin code accordingly. Please let me know if anything else still needs to be modified. 

Thanks in advance

lishid:
Looking great! Final things:

https://github.com/Darakah/obsidian-comments-plugin/blob/4d15934833cd41e0d95953477bc64f9a8ab9df8b/main.ts#L132
I feel like this is a bit dangerous - you're setting raw HTML that you got from a markdown file directly into the HTML of an element on page. If someone had wrote `<script> ... </script>` in a way that would be picked up by your selector, they would be able to execute arbitrary code, which is a security problem.

https://github.com/Darakah/obsidian-comments-plugin/blob/4d15934833cd41e0d95953477bc64f9a8ab9df8b/main.ts#L243
You forgot to save settings for these two.

Darakah:
Thanks for the fast response Licat!

Point 2: I am not saving as these 2 settings are just telling the user which variable to change in the css file and not really updating any value (ik it is best to remove them all together but it would be good pointer for those that dono what to do)

Point 1: The thing is, i did this so that if the internal sections are actually also html elements (which is the case most of the time for me as i use it to create better notes), if i do `.setText` it of course will look ugly and not render. If you believe it is a security issue i will just convert it to `.setText` that way it will not be executed. 

QUESTION: i thought scripts are not executed by the obsidian compiler (as i tried to use a couple and it did not work) isn't it the same for what is rendering the elements in the side panel? wouldn't it just ignore it?

Many thanks!

lishid:
> QUESTION: i thought scripts are not executed by the obsidian compiler (as i tried to use a couple and it did not work) isn't it the same for what is rendering the elements in the side panel? wouldn't it just ignore it?

Your plugin can execute script, which may have security holes depending on how it's written. Obsidian's markdown renderer sanitizes any HTML it outputs, but your plugin is reading the file content directly, and then putting un-sanitized user data directly in `innerHTML`, which isn't something we can prevent.

Darakah:
Modified to `.setText` taking only the innerHTML element.

Many thinks for all the help.

PR #154, opened 2021-02-06
Add Block Render Plugin for Obsidian

xpgo:
Block Render Plugin: Render different code blocks in a note file to specific preview.


lishid:
https://github.com/xpgo/obsidian-block-render-plugin/blob/72f0b0b0fd154e983ea99f6e76e571d97ef3b253/src/block-markmap.ts#L81
Any chance this can cause id collisions?

https://github.com/xpgo/obsidian-block-render-plugin/blob/72f0b0b0fd154e983ea99f6e76e571d97ef3b253/src/block-xcards.ts#L116
Consider using `readCached` if you aren't writing the file back to disk.

https://github.com/xpgo/obsidian-block-render-plugin/blob/72f0b0b0fd154e983ea99f6e76e571d97ef3b253/src/block-xcards.ts#L270
Don't recommend writing raw HTML especially when it can include user-generated content with no sanitization. Use proper DOM APIs to create the elements, set their attributes and text content, and append them to parent elements.

https://github.com/xpgo/obsidian-block-render-plugin/blob/72f0b0b0fd154e983ea99f6e76e571d97ef3b253/src/main.ts#L33
You may want to consider using the [new API to register for code blocks](https://github.com/obsidianmd/obsidian-api/blob/9b69506127863102e1102cdeb4d14f523d08583e/obsidian.d.ts#L1718) instead of manually looking for the `code.language-xxx` blocks.

xpgo:
Thanks for the suggestions, I have made some changes accordingly:

> https://github.com/xpgo/obsidian-block-render-plugin/blob/72f0b0b0fd154e983ea99f6e76e571d97ef3b253/src/block-markmap.ts#L81
> Any chance this can cause id collisions?

Now, the id is removed.

> https://github.com/xpgo/obsidian-block-render-plugin/blob/72f0b0b0fd154e983ea99f6e76e571d97ef3b253/src/block-xcards.ts#L116
> Consider using `readCached` if you aren't writing the file back to disk.

I have tried to use `this.app.vault.cachedRead()`, but the function needs a parameter of TFile. The API does not provide an entrance to create a TFile object from path string, and I have only the path string. There is only a function `getAbstractFileByPath()`, but I need a function like `getTFileByPath()`, do you have any suggestion?

> https://github.com/xpgo/obsidian-block-render-plugin/blob/72f0b0b0fd154e983ea99f6e76e571d97ef3b253/src/block-xcards.ts#L270
> Don't recommend writing raw HTML especially when it can include user-generated content with no sanitization. Use proper DOM APIs to create the elements, set their attributes and text content, and append them to parent elements.

Now I have changed them to DOM Elements.

> https://github.com/xpgo/obsidian-block-render-plugin/blob/72f0b0b0fd154e983ea99f6e76e571d97ef3b253/src/main.ts#L33
> You may want to consider using the [new API to register for code blocks](https://github.com/obsidianmd/obsidian-api/blob/9b69506127863102e1102cdeb4d14f523d08583e/obsidian.d.ts#L1718) instead of manually looking for the `code.language-xxx` blocks.

It seems that the new API is not supported by some old verions of Obsidian, moreover, how can I unregister specific code language if users disabled some types of code blocks though the plugin settings?

lishid:
> I have tried to use this.app.vault.cachedRead(), but the function needs a parameter of TFile. The API does not provide an entrance to create a TFile object from path string, and I have only the path string. There is only a function getAbstractFileByPath(), but I need a function like getTFileByPath(), do you have any suggestion?

```ts
let file = vault.getAbstractFileByPath(filepath);
if (file && file instanceof TFile) {
  let contents = vault.cachedRead(file);
}
```

> It seems that the new API is not supported by some old verions of Obsidian, moreover, how can I unregister specific code language if users disabled some types of code blocks though the plugin settings?

That is correct - You can use the minVersion in your manifest.json to avoid old Obsidian from accidentally installing your plugin, but it's very rare to see older versions of Obsidian in the wild because of our auto-updater. Generally you'll want to target the latest version of Obsidian whenever possible just because we may start deprecating older APIs.

As for unregistering, I do agree that there isn't an API for doing that, but I don't think it makes sense to let the user to configure whether to enable them - if they don't use say abcjs, they can just avoid writing code blocks with abcjs. If you do want users to configure these, the best way is to split your plugin into each type of their own. I think that makes the user experience better as well since right now it's unclear what your plugin does just from reading the name and description "Render different code blocks in a note file to specific preview"

xpgo:
> ```ts
> let file = vault.getAbstractFileByPath(filepath);
> if (file && file instanceof TFile) {
>   let contents = vault.cachedRead(file);
> }
> ```

Thanks for the suggestion, I have changed the codes based on this. 

> That is correct - You can use the minVersion in your manifest.json to avoid old Obsidian from accidentally installing your plugin, but it's very rare to see older versions of Obsidian in the wild because of our auto-updater. Generally you'll want to target the latest version of Obsidian whenever possible just because we may start deprecating older APIs.

I can only update Obsidian to the version 0.10.11. The new API is added after 0.11.12 (Inside builde). I will use the new API in the future. 

> As for unregistering, I do agree that there isn't an API for doing that, but I don't think it makes sense to let the user to configure whether to enable them - if they don't use say abcjs, they can just avoid writing code blocks with abcjs. If you do want users to configure these, the best way is to split your plugin into each type of their own. I think that makes the user experience better as well since right now it's unclear what your plugin does just from reading the name and description "Render different code blocks in a note file to specific preview"

I have changed the description to be more specific. I think you are right about this. We should use only one plugin for one kind of code block, especially for users. And it also could be offensive to other developers. Personally, I just want to use a single plugin to render different code blocks. Thanks all the same.

lishid:
Sounds good. Let me know how you'd like to proceed - we can also wait for 0.10.12 to go public in the next few days. Splitting the plugins would be a great idea as well, I'm happy to pull them in together too.

xpgo:
I think it'd be better to split it to multiple plugins.
I will wait for the new release of 0.10.12,  and test the plugins for a while.
Thank you.

ericaxu:
0.10.13 is now released. Any chance we can move forward with this?

PR #156, opened 2021-02-07
Add obsidian-show-file-path plugin

ravimashru:


lishid:
Might want to rename "MyPlugin" 😂

ravimashru:
Ooops... thanks for pointing that out 😂

PR #157, opened 2021-02-07
Obsidian Timelines Plugin

Darakah:
Many thanks for any remarks!


liamcain:
Hi, I found a couple issues and made some suggestions for how you could improve the code here: https://github.com/Darakah/obsidian-timelines/issues/1

Darakah:
Hello liam, 

Many thanks for the detailed remarks, greatly appreciated especially the common practices. 

- Regarding the `if(!Number.isInteger(note_id)){ ` the line before it is supposed to parse out the `-` used to separate the different fields and thus `if the user inputed date` is actually valid, it will be a proper integer, otherwise it is not a valid date and that note will be discarded. 

https://github.com/Darakah/obsidian-timelines/blob/685dddb902b990085af34c94fe6170112daf1c8a/main.ts#L88

- ` if(!timeline_dates[i]){ ` regarding this, I was actually in the middle of trying to figure it out. Planning on dropping it once i find it. But what did you mean by the `alternation logic`? 

- Regarding the user text escaping will use the escape function to cast the different fields to avoid this case. Many thanks for pointing this out as i had overlooked it.

Darakah:
Dear Liam, 

Just pushed version 0.1.1 to address all your remarks.

**_-> Style-nit:_** Adjusted all variable names to a single convention of  camelCase. Thank you for the remark.

**_-> string interpolation:_** As suggested, I replaced all string concatenations with the string interpolation to avoid any unwanted casting. 

**_-> Unused:_** deleted https://github.com/Darakah/obsidian-timelines/blob/685dddb902b990085af34c94fe6170112daf1c8a/main.ts#L25

**_-> code formatted:_** reformatted using VS Code and replaced all indentations to tabs.

**_-> Escaping possible breaking characters in user strings:_** Used regular expression to replace possible breaking characters.

lishid:
https://github.com/Darakah/obsidian-timelines/blob/4aea2f4b2df933e601de8572a2d3f01854eeb530/main.ts#L96
Here you are in fact storing user data to be put in your built HTML. A variety of attacks can be done here, for example, creating a `data-img=")'&quot;&lt;script&gt;alert(0)&lt;/script&gt;"`

Just don't write HTML like that - it's error prone, and one of the easiest ways of making your code vulnerable due to the large vector of DOM injection attacks.

Build up the DOM with `createEl()` and properly abstracting parts of your code into functions will help a lot.

https://github.com/Darakah/obsidian-timelines/blob/4aea2f4b2df933e601de8572a2d3f01854eeb530/main.ts#L13
You aren't properly typing the "MultiList" here. Not sure if Typescript is giving you an error on build, but you should specify what each of those list/levels are. This kind of function really obscures what's going on.

Ideally, you should create a type for everything you're storing - for example
```ts
interface MyDataContainer {
  date: string;
  title: string;
  //...
}

type NoteData = MyDataContainer[];
type AllNotes = NoteData[];
```

Darakah:
Hey Licat, Many thanks for your remarks, Great help as always!

**_Point 2:_** It is being inferred from this line (and the one in the else statement below it) https://github.com/Darakah/obsidian-timelines/blob/4aea2f4b2df933e601de8572a2d3f01854eeb530/main.ts#L96
Your method is much clearer, replaced it accordingly.

**_Point 1:_** Replaced the HTML building part with Create Elements as suggested. 

Please let me know if anything else still needs to be modified. 

----

**_Questions:_**

So in the previous version the following was happening:

> Page Text Is read using Obsidian API vault.read (1) -> Page content was converted into a div with the `innerHTML` set to that content (2) -> querySelector was applied to the created element to obtain the Timeline span containing the note info (3) -> the different attributed are saved in a Multi-array (4) -> a string concatenation of the elements (only string concat operation) (5) -> the created HTML string element is written to the note using Obsidian API setLines (6) -> the added text to the main note will be parsed by the default Obsidian Parser as if I had written it directly in the note (no?) (7) -> when this text is read Obsidian Parser SANATIZATION takes place (8)

- At what point in the above mentioned set does the `script` / possible `DOM injection` takes place? The only time it is not treated as normal text is during step `(2)` when it is set to the innerHTML of the created element to query it. Simply setting the innerHTML causes a script included inside to run?? 

Sorry for the long question, Just wanna properly understand how this takes place.

Many thanks for all your helpful remarks as always Licat!!!

lishid:
Hmm my bad, I was under the impression that you display the built HTML somewhere on the page. Still, you should probably use a proper HTML parser rather than just creating a `div` and setting its innerHTML. This is because there's still a chance some event handlers on the DOM may fire, which could execute malicious scripts.

Look into using https://developer.mozilla.org/en-US/docs/Web/API/DOMParser which generates an element you can query almost identically to how it would work with a new element. This API is specifically designed to be safe to use for this purpose.

https://github.com/Darakah/obsidian-timelines/blob/0dc5c576157ea739289a15338567a32413995959/main.ts#L153
Seems like you are using the raw DOM api to create your elements here rather than our recommended shorthands. That's fine with me, but I just want to recommend our helper functions one last time because it could make your life much easier and your code shorter and more concise. Here's an example:

```ts
let noteBackgImg = document.createElement('div')
noteBackgImg.setAttribute('class', 'thumb')
noteBackgImg.setAttribute('style', `background-image: url(${getElement(timelineNotes, timelineDates[i], j, 2)});`)
noteCard.appendChild(noteBackgImg);
```
This can all be replaced by a single line, which sets the attributes and append to the parent element all at once:
```ts
noteCard.createDiv({cls: 'thumb', attr: {style: `background-image: url(${getElement(timelineNotes, timelineDates[i], j, 2)});`}});
```

> (That's one of my favorite plugins as i learned most of what i used from there)
Yeah.... I still feel weird about how they do the HTML. Maybe we can create an issue there later.

Darakah:
Dear Licat,

As suggested:
1. Using DOM parser to read user date from note to avoid any possible DOM attack. (thanks for the pointer will use this method from now on)
2. Using Obsidian helper functions to create the different elements to make the code cleaner and more concise.

Please let me if there are any other problems or if anything else can be further improved.

Many thanks! 
With each review you greatly help :)

liamcain:
I took another pass and found 2 small things (added them to https://github.com/Darakah/obsidian-timelines/issues/1). Otherwise the rest looks fine to me.

Darakah:
Hey Liam,

Thanks for your feedback, posted a reply on the mentioned issue.

Darakah:
Hey Licat, 

Anything else?

lishid:
Should be ready to merge. Your repo is `main`, so you'll probably want to make a master branch, or indicate that in the PR.

Darakah:
Dear Licat,

Indicated the branch in the PR.

Thanks in advance

lishid:
It's in, great work!

PR #155, opened 2021-02-07
Add Various Complements Plugin

tadashi-aikawa:
URL: https://github.com/tadashi-aikawa/obsidian-various-complements-plugin

Best regards.

PR #158, opened 2021-02-08
Add Various Complements

tadashi-aikawa:
This PR is a continuation of https://github.com/obsidianmd/obsidian-releases/pull/155.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -801,5 +801,13 @@
         "description": "Edit ini files in Obsidian",
         "repo": "deathau/ini-obsidian",
         "branch": "main"
+    },
+    {
+        "id": "various-complements",
+        "name": "Various Complements",
+        "author": "tadashi-aikawa",
+        "description": "This plugin enables you to complement input in markdown files",
+        "repo": "tadashi-aikawa/obsidian-various-complements-plugin",
+        "branch": "main"
     }
 ]


liamcain:
Hi @tadashi-aikawa, one small issue:

https://github.com/tadashi-aikawa/obsidian-various-complements-plugin/blob/main/main.ts#L84
```ts
      checkCallback: (checking: boolean) => {
        if (checking && !this.app.workspace.getActiveViewOfType(MarkdownView)) {
          return false;
        }

        this.execAutoComplete();
        return true;
      },
```

If checking is true, you don't want to call `this.execAutoComplete();` at all. Checking means that command palette is filtering through all the commands to see if it _can_ run. Instead you probably want something like this:

```ts
      checkCallback: (checking: boolean) => {
        if (checking) {
          // command should only be available if current view is a MarkdownView
          return this.app.workspace.getActiveViewOfType(MarkdownView);
        }
        this.execAutoComplete();
      },
```

tadashi-aikawa:
Thank you, @liamcain!

I didn't understand the argument `checking` correctly. I have fixed it by https://github.com/tadashi-aikawa/obsidian-various-complements-plugin/commit/691e4ea2c8f2e3ad1ab6956bc049ef7957054cf5 👍

tadashi-aikawa:
> Create new GitHub release using your new version number as the "Tag version". Use the exact version number, don't include a prefix v. See here for an example: https://github.com/obsidianmd/obsidian-sample-plugin/releases

I missed it.., I'll fix it soon.

tadashi-aikawa:
OK. I have fixed 😄

PR #159, opened 2021-02-08
Add "Creases" plugin

liamcain:
## Repo location
https://github.com/liamcain/obsidian-creases

## Release Checklist
- [x] I have tested this on Windows, macOS, and Linux (if applicable)
- [x] The Github release contains all required files
  - [x] `main.js`
  - [x] `manifest.json`
  - ~~`styles.css` _(optional)_~~
- [x] Github Release name matches the exact version number specified in your manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -816,6 +816,13 @@
         "author": "darakah",
         "description": "Create a timeline view of all notes with the specified combination of tags",
         "repo": "Darakah/obsidian-timelines",
+    },
+    {
+        "id": "creases",
+        "name": "Creases",
+        "author": "liamcain",
+        "description": "Mark sections of markdown to be collapsed by default",
+        "repo": "liamcain/obsidian-creases",
         "branch": "main"
     }
 ]


ericaxu:
Awesome template! Might steal it 😉

liamcain:
@ericaxu yeah was making it as a draft to suggest you could add as a github template. Was trying to think of some other things that might be helpful too.

liamcain:
Okay, so this plugin is definitely less useful in 0.11.0 but I still think it's nice for collapsing autogenerated content like templates.

I just pushed a new version that's compatible with the foldManager and adds a "refold" command that refolds the content only to the creases that you set.

Let me know what you think @lishid

liamcain:
Going to play aound with 0.11.0 a bit more to see if 1.) I still want/need this behavior 2.) there's a way to make this less jank if I just use the foldManager

lishid:
Yup, that makes sense. Let me know what you think.

PR #163, opened 2021-02-13
Adds Daily Activity plugin

trydalch:


lishid:
Couple of things:

- Don't import `moment`, use `window.moment` so you aren't bundling a full copy of moment.js to your plugin
- Clean up the plugin's sample settings, modal and setting tab.
- Make sure you follow the release process (by uploading the manifest and js to your github release) so users can download your release from inside Obsidian.

liamcain:
Also make sure that your release name matches the version number (`0.0.1`)

@trydalch I did a pass, let me know if you have any questions!

## Code Review
- https://github.com/trydalch/obsidian-daily-activity/blob/ed2af8617507b46585bc46a97a187dce42b636f3/src/ActivityLogger.ts#L22 - store `window.moment()` in a temp variable so you aren't creating a new moment object for each file in the vault. (perf improvement)
- https://github.com/trydalch/obsidian-daily-activity/blob/ed2af8617507b46585bc46a97a187dce42b636f3/src/ActivityLogger.ts#L51 make to sure do a check if activeView is null. I know you have a check in the command's callback, but it doesn't hurt to be extra safe here, especially because the commands are async (potential bug)
- https://github.com/trydalch/obsidian-daily-activity/blob/ed2af8617507b46585bc46a97a187dce42b636f3/src/ActivityLogger.ts#L35 you probably want to wrap the basename in `getLinkpath()` from the obsidian API here to handle notes inside folders. (potential bug)

trydalch:
Thanks for the reviews. I believe I've addressed all of your comments. Thank you for the feedback.

liamcain:
Nice job! Just a few small configuration issues:
- The version in `manifest.json` needs to match the release version (`0.0.2`) - (otherwise Obsidian will fail on installing from the gallery)
- authorUrl in the `manifest.json` should either be a personal website or `https://github.com/trydalch`
- Not too important, but the `version` in package.json should probably match your manifest.json version (`0.0.2`)

trydalch:
Ah, I missed that. Thank you. Should all be updated now.

PR #164, opened 2021-02-13
Add plugin: Active note to window title

jplattel:
This plugin sets the window title to include the current open note


lishid:
Maybe restore the original window title on unload?

jplattel:
Oh that's a good one, will fix that! :)

liamcain:
@jplattel you'll need to rebase

jplattel:
Ugh, doing this in the Github interfaces results in the whole file being rewritten.. I'll just open a new one..

PR #165, opened 2021-02-14
[RFC] Add github template

liamcain:
- I introduced 2 templates that will populate the pull request description when the user selects one.
  - new plugin
  - new theme
- the plugin template has a proposed checklist for plugin authors to fill out before submitting
- I added a Github action that lints the JSON files (since I've noticed some PRs mess up the formatting or add an extra comma)

## Open Questions
- Any concerns over the proposed Plugin Release Checklist?
- Is there a JSON linter with fewer deps than eslint that we could use? t's crazy how big the lock file is...

**Small note:** I added `ericaxu` as an auto-assignee for themes. LMK if there are any concerns there!


ericaxu:
Yep, I can be the auto-assignee. Should it be pull request template instead of issue template, by the way? Is it because issue templates give the option to choose from, where as pull request templates don't?

liamcain:
> Should it be pull request template instead of issue template, by the way? Is it because issue templates give the option to choose from, where as pull request templates don't?

🤦 no, it was just a typo, good catch!

liamcain:
Okay, so I tested it out and you made a good point that Github doesn't have a good UI selection screen for multiple pull request templates, it only does that for issue templates. I combined the templates into a single one.

Here's what it looks like:
image

Vinzent03:
I would add that it is necassary to use the same id in `community-plugins.json` and `manifest.json`.

Some used different ids, why the plugin couldn't be loaded.

Vinzent03 (approved):

lishid:
Perhaps we should ignore package-lock.json?

argenos:
@liamcain should the plugin template mention something about not setting hotkeys by default? I remember this was discussed as a best practice in Discord.

liamcain:
> Perhaps we should ignore package-lock.json?

@lishid, we could. I prefer keeping lock files committed to the repo though so if any issues crop up, they are reproduce-able. But it's probably less important for just linting json.

ericaxu:
When can we get this in? It's still a pain to find the repo by looking at the PR...

liamcain:
@ericaxu, for the linting, I think you'll need to enable Github Actions on this repo

image

The "only specific actions" is probably a good option for this.

@lishid, I removed the lock file since it's probably overkill for such a simple linting action.

---

I think this should be good to merge now; we can always iterate on the template in the future.

PR #166, opened 2021-02-14
Adding daily stats plugin

dhruvik7:
New plugin for tracking daily word count

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -834,5 +834,12 @@
         "description": "This outputs a list of the files changed/created today in the active file",
         "author": "trydalch",
         "repo": "trydalch/obsidian-daily-activity"
+    },
+    {
+        "id": "obsidian-daily-stats",
+        "name": "Daily Stats",
+        "description": "Track your daily word count",
+        "author": "Dhruvik Parikh",
+        "repo": "dhruvik7/obsidian-daily-stats"
     }
 ]


liamcain:
- https://github.com/dhruvik7/obsidian-daily-stats/blob/c81fb4a38cd6eb329d2ba1d0d8e18937f0145a6f/main.ts#L40 Don't pass `this` as context to the callback here. You're better off binding `this` to the callback handlers like so: `this.app.workspace.on("quick-preview", this.onQuickPreview.bind(this))`
- https://github.com/dhruvik7/obsidian-daily-stats/blob/c81fb4a38cd6eb329d2ba1d0d8e18937f0145a6f/main.ts#L64 You should use file.path instead here to properly handle folders
- https://github.com/dhruvik7/obsidian-daily-stats/blob/c81fb4a38cd6eb329d2ba1d0d8e18937f0145a6f/main.ts#L61 you should instead use `getActiveViewOfType(MarkdownView)` here
- https://github.com/dhruvik7/obsidian-daily-stats/blob/master/main.ts#L92 you should check the full path here. It's also concerned best practice to use `hasOwnProperty` instead of the `in` operator for checking if an object contains a key
- https://github.com/dhruvik7/obsidian-daily-stats/blob/c81fb4a38cd6eb329d2ba1d0d8e18937f0145a6f/main.ts#L64 this triggers every keystroke. It can be expensive for larger files. The built-in wordcount plugin does this work in a background thread using webworkers. Another option would be to use the [debounce function](https://github.com/obsidianmd/obsidian-api/blob/81ab85ade4552c9116c1e10d009127e62019c923/obsidian.d.ts#L591) provided by the API

dhruvik7:
Thanks for the review, @liamcain 
I've made all of the updates you recommended. On the last point, the original reason why I went with quick-preview instead of counting words on a timeout was that it allowed me to avoid having to do a file read, which I found added a ton of latency to the updates. It seems to me that the webworker approach would face the same issue?

In any case, I ended up adding debouncing, as per your recommendation, which should limit the computational overhead while maintaining fast updates.

liamcain:
Looks good! 

One last thing: you should change your minAppVersion in the manifest.json to 0.10.12 since that's when the debounce function was introduced.

dhruvik7:
Done!

liamcain:
Nice, LGTM

lishid:
Couple of more suggestions:

https://github.com/dhruvik7/obsidian-daily-stats/blob/7d1dc8461044149201125558349e312a9c75e402/main.ts#L36
You can write this in the field definition block so you don't have to type out the Debouncer definition:
```ts
settings: DailyStatsSettings;
statusBarEl: HTMLElement;
currentWordCount: number;
today: string;
debouncedUpdate = debounce(this.updateWordCount.bind(this), 400, false);
```

https://github.com/dhruvik7/obsidian-daily-stats/blob/7d1dc8461044149201125558349e312a9c75e402/main.ts#L108
You might want to do `(d.getMonth() + 1)`

Otherwise looks good to me!

PR #167, opened 2021-02-15
Add Render Block Markmap Plugin for Obsidian

xpgo:
Add a plugin for rendering code block of markmap to embedded mindmap in preview mode.


liamcain:
- https://github.com/xpgo/obsidian-render-block-markmap/blob/421c0faf2590d7de9eeb1127c3080e9d28d18759/src/main.ts#L54 this shouldn't be necessary
- https://github.com/xpgo/obsidian-render-block-markmap/blob/main/README.md#features small typo - `entile` should be `entire`
- https://github.com/xpgo/obsidian-render-block-markmap/blob/b68ee086ea7ccf14b567f0c77340533a16747c39/src/main.ts#L28 - small UX note - this toggle seems redundant with just disabling the plugin entirely.

xpgo:
>     * https://github.com/xpgo/obsidian-render-block-markmap/blob/421c0faf2590d7de9eeb1127c3080e9d28d18759/src/main.ts#L54 this shouldn't be necessary

Yeah, it looks unnecessary, but the trick did solve a problem. 
The svg element of markmap needs initial position information of its parent element to determine the size of its children, or there will be lots of uncatched errors and the initial svg will not be shown. I am not good at js and ts, there may be some better way to solve the problem, but it seems a magic trick after many tests.

>     * https://github.com/xpgo/obsidian-render-block-markmap/blob/main/README.md#features small typo - `entile` should be `entire`

It has been amended, thanks.

>     * https://github.com/xpgo/obsidian-render-block-markmap/blob/b68ee086ea7ccf14b567f0c77340533a16747c39/src/main.ts#L44 this should be `Object.assign({}, DEFAULT_SETTINGS, await this.loadData());` to avoid mutating the DEFAULT_SETTINGS object.

The settings codes have been removed.

>     * https://github.com/xpgo/obsidian-render-block-markmap/blob/b68ee086ea7ccf14b567f0c77340533a16747c39/src/main.ts#L28 - small UX note - this toggle seems redundant with just disabling the plugin entirely.


Yes, I left it there for future usage. Now I have removed the whole settings codes.

>     * https://github.com/xpgo/obsidian-render-block-markmap/blob/b68ee086ea7ccf14b567f0c77340533a16747c39/src/block-markmap.ts#L15 you should use `let` instead of `var` throughout this file

They have been changed, thanks.

liamcain:
> Yeah, it looks unnecessary, but the trick did solve a problem.
The svg element of markmap needs initial position information of its parent element to determine the size of its children, or there will be lots of uncatched errors and the initial svg will not be shown. I am not good at js and ts, there may be some better way to solve the problem, but it seems a magic trick after many tests.

Hmm, looking at the code, it shouldn't have any effect on the layout, especially because it's not it's not early exiting from the register callback, it's just querying an element and returning.

I'll checkout the code to see what you mean.

---

Is there any reason you added `viewHeight` as an instance variable to the MarkmapProcessor? As far as code structure, having those functions in a class doesn't provide much benefit so I'm curious what the intent is there.

Code looks fine otherwise. You'll need to rebase this PR though to fix the conflict.

liamcain:
FWIW, I was able to reproduce the issue you're running into with the refreshEl. Seems that it's getting undefined for the size until the dom is queried. Not sure why but you could change the refreshEl function to be:

image

xpgo:
> I'll checkout the code to see what you mean.

yeah, I also noticed that the qureySelector() is the key codes to avoid the issue.

> 
> Is there any reason you added `viewHeight` as an instance variable to the MarkmapProcessor? As far as code structure, having those functions in a class doesn't provide much benefit so I'm curious what the intent is there.

Because I want to parser more options in the future, and there may be multiple code blocks of markmap with different options in a note. 

> 
> Code looks fine otherwise. You'll need to rebase this PR though to fix the conflict.

I will close the PR and make another clean one when ready. Thanks.

PR #169, opened 2021-02-17
Add Vault Changelog plugin

MrZeroo00:
Add  a plugin to maintain a changelog of recently edited files in your vault.

This initial release contains the minimum number of features required:

- Manual and automatic changelog update
- Configurable changelog file path
- Configurable number of items to show in the changelog


phibr0:
Looks good to me! (I am no TypeScript Expert tho so maybe someone else should look over it too)

liamcain:
https://github.com/MrZeroo00/obsidian-vault-changelog/blob/30c1b457c00690004399adfdeacab13d7756dc91/main.ts#L74 Maybe filter this out before the slice so that the number of recently changed files is correct?

https://github.com/MrZeroo00/obsidian-vault-changelog/blob/30c1b457c00690004399adfdeacab13d7756dc91/main.ts#L2 might be easier to use window.moment instead of dayjs. Avoids adding a new dependency.

https://github.com/MrZeroo00/obsidian-vault-changelog/blob/30c1b457c00690004399adfdeacab13d7756dc91/main.ts#L165 toggling this off doesn't remove the event listeners

https://github.com/MrZeroo00/obsidian-vault-changelog/blob/30c1b457c00690004399adfdeacab13d7756dc91/main.ts#L64 In the demo, it looks like the CHANGELOG is embedded with existing note content but as far as I can tell, it will just overwrite whatever is in the changelog file. If that's the case it might make sense to warn the users somewhere since they could lose content by overwriting without realizing.

---

Really cool plugin, I think this would be killer for Obsidian Publish vaults!

MrZeroo00:
Thanks @liamcain for your feedback!

> https://github.com/MrZeroo00/obsidian-vault-changelog/blob/30c1b457c00690004399adfdeacab13d7756dc91/main.ts#L74 
> Maybe filter this out before the slice so that the number of recently changed files is correct?

Good suggestion, done [here](https://github.com/MrZeroo00/obsidian-vault-changelog/commit/0d4c4750c7f846b925f2e9b9558ee848ba2dc37c).

> https://github.com/MrZeroo00/obsidian-vault-changelog/blob/30c1b457c00690004399adfdeacab13d7756dc91/main.ts#L2 
> might be easier to use window.moment instead of dayjs. Avoids adding a new dependency.

Done [here](https://github.com/MrZeroo00/obsidian-vault-changelog/commit/4f61aee378c0dc2c1a22b150fa8efe266991ea0f) although I had to add a dirty [hack](https://github.com/MrZeroo00/obsidian-vault-changelog/commit/4f61aee378c0dc2c1a22b150fa8efe266991ea0f#diff-564c860307257bfac9176f554a35aa858c350a59d75c0dcd0a5d63a480805bd0R80) in order to compile without errors.
Do you have ideas for a proper way to do this? Most plugins I've seen use their own instance of Moment.

> https://github.com/MrZeroo00/obsidian-vault-changelog/blob/30c1b457c00690004399adfdeacab13d7756dc91/main.ts#L165
> toggling this off doesn't remove the event listeners

I couldn't find a way to remove the event listeners apart from forcing a plugin unload/load.
From the [API's doc](https://github.com/obsidianmd/obsidian-api/blob/81ab85ade4552c9116c1e10d009127e62019c923/obsidian.d.ts#L480), the event will be detached when unloading. Is it enough? What would be a proper way to fix this?

> https://github.com/MrZeroo00/obsidian-vault-changelog/blob/30c1b457c00690004399adfdeacab13d7756dc91/main.ts#L64
> In the demo, it looks like the CHANGELOG is embedded with existing note content but as far as I can tell, it will just overwrite whatever is in the changelog file. If that's the case it might make sense to warn the users somewhere since they could lose content by overwriting without realizing.

You're right, this is misleading. I've [updated the readme](https://github.com/MrZeroo00/obsidian-vault-changelog/commit/e903d26c332608534211aaa7a1ae75dc83edb3ef) to reflect it.

liamcain:
> Done here although I had to add a dirty hack in order to compile without errors.
Do you have ideas for a proper way to do this? Most plugins I've seen use their own instance of Moment.

You can `declare` it like so: https://github.com/liamcain/obsidian-calendar-ui/blob/76febf71efc19f80496de9e9d3341220954c83a7/src/index.ts#L10 That requires adding @types/moment your devDependencies but it doesn't add moment to your bundle at all.

> I couldn't find a way to remove the event listeners apart from forcing a plugin unload/load.
From the API's doc, the event will be detached when unloading. Is it enough? What would be a proper way to fix this?

I think you can keep the event registered to the plugin instance and just call `this.app.vault.off()`. Note: `off` expects the event name and the callback passed in so you might want to update your event handlers to look more like this:

```ts
      this.registerEvent(
        this.app.vault.on( "modify", this.watchVaultChange)
      );
```

The wrapper function with the async await is unnecessary anyway.

MrZeroo00:
> I think you can keep the event registered to the plugin instance and just call `this.app.vault.off()`. Note: `off` expects the event name and the callback passed in so you might want to update your event handlers to look more like this:
> 
> ```ts
>       this.registerEvent(
>         this.app.vault.on( "modify", this.watchVaultChange)
>       );
> ```
> 
> The wrapper function with the async await is unnecessary anyway.

`this.watchVaultChange` ultimately calls `this.app.vault.modify` which returns a [Promise](https://github.com/obsidianmd/obsidian-api/blob/81ab85ade4552c9116c1e10d009127e62019c923/obsidian.d.ts#L2630). Is the async/await really unnecessary? (I'm touching my limits on Typescript here)
Trying to remove it lead to errors.

liamcain:
> this.watchVaultChange ultimately calls this.app.vault.modify which returns a Promise. Is the async/await really unnecessary? (I'm touching my limits on Typescript here)

`async/await` in Javascript is really just syntactical sugar for writing promises. When you see an `await` you can just pretend that everything after that line is wrapped in a `then` clause if that makes it easier to think about.

Say you have:
```ts
await functionThatReturnsAPromise();
console.log('called after await');
```

That's the equivalent of:
```ts
functionThatReturnsAPromise().then(() => {
  console.log('called after await'); 
});
```

---

So, looking at your code, we have:
```ts
      this.registerEvent(
        this.app.vault.on(
          "modify",
          async (file) => await this.watchVaultChange(file)
        )
      );
```

`watchVaultChange` has the type definition of `(file: TFile) => Promise<void>`.

 And the wrapper function that you created has the same type definition. Meaning, the wrapper function is (_mostly_) functionally equivalent.  The async/await here would only make a difference if you had something after the `watchVaultChange` call.

For example:

```ts
async (file) => {
  await this.watchVaultChange(file);
  console.log('called after watchVaultChange has finished');
```

Okay, I said _mostly_. There's one caveat to this change: the _caller_ is different. Another JS/TS gotcha if you're not familiar with it. [`this`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) doesn't necessarily refer to the current object instance. It's meaning changes based on the context that it's called. So if you are passing a function (i.e. `this.watchVaultChange` as a callback to be called outside the current Plugin context, you'll need to use [`bind`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind).

TL;DR: Use the following snippet - 

```ts
this.registerEvent(this.app.vault.on("modify", this.watchVaultChange.bind(this)));

**Comment by MrZeroo00** [comment]
Wow! Thanks for this great explanation. I'm glad I contributed with this plugin, I've learned something!

I've implemented the changes you suggest in this [commit](https://github.com/MrZeroo00/obsidian-vault-changelog/commit/6acf94d19ad41cd0b07d7f4cafd26fed9aba996a).

**Comment by liamcain** [comment]
Okay, just a couple small nitpicks:

https://github.com/MrZeroo00/obsidian-vault-changelog/blob/6acf94d19ad41cd0b07d7f4cafd26fed9aba996a/main.ts#L69 You should use `!==` since Javascript is dumb and `!=` will cast values and other weird stuff and you could end up with the wrong thing here.

https://github.com/MrZeroo00/obsidian-vault-changelog/blob/6acf94d19ad41cd0b07d7f4cafd26fed9aba996a/main.ts#L51 Use `===` instead.

https://github.com/MrZeroo00/obsidian-vault-changelog/blob/6acf94d19ad41cd0b07d7f4cafd26fed9aba996a/main.ts#L39 My bad for not being more clear about this. The line where you reassign `this.watchVaultChange` should be inside the `onload` function so that it doesn't get called more than once.

Everything else is looking good!

**Comment by MrZeroo00** [comment]
Done!
I've amended the async/await commit and fixed the rest in a different commit.
Thanks for your review :)

**Review by liamcain — APPROVED** [review]

**Comment by MrZeroo00** [comment]
Awesome! Thanks again @liamcain for your time.

**Comment by liamcain** [comment]
@MrZeroo00  Did you mean to close this?

**Comment by MrZeroo00** [comment]
Yes, is there a different workflow?

**Comment by liamcain** [comment]
Most people want to merge it 🤣

**Comment by MrZeroo00** [comment]
🤦🏼‍♂️
I thought you merged it but you just approved the changes.

**Comment by ericaxu** [comment]
@lishid People want to merge this 🔥

**Comment by lishid** [comment]
Crap I wrote some suggestions but lost the tab on the Chrome update, so I forgot to follow up.

https://github.com/MrZeroo00/obsidian-vault-changelog/blob/17e38683bf453340925b56961b634d039a491d97/main.ts#L50
Might want to make this function debounced, since events are awaited for - you might end up with multiple parallel executions of `writeChangelog`.

https://github.com/MrZeroo00/obsidian-vault-changelog/blob/17e38683bf453340925b56961b634d039a491d97/main.ts#L85
Probably better to use `if( file instanceof TFile)` instead of casting it directly.

https://github.com/MrZeroo00/obsidian-vault-changelog/blob/17e38683bf453340925b56961b634d039a491d97/main.ts#L132
I think the `String()` isn't necessary? If you're running issues here then there might be a bug we need to fix.

Looks good otherwise!

**Comment by MrZeroo00** [comment]
> Crap I wrote some suggestions but lost the tab on the Chrome update, so I forgot to follow up.
> 
> https://github.com/MrZeroo00/obsidian-vault-changelog/blob/17e38683bf453340925b56961b634d039a491d97/main.ts#L50
> Might want to make this function debounced, since events are awaited for - you might end up with multiple parallel executions of `writeChangelog`.

What would you suggest to this? I'm tempted to use something like [async-mutex](https://www.npmjs.com/package/async-mutex) but it seems overkill.

> https://github.com/MrZeroo00/obsidian-vault-changelog/blob/17e38683bf453340925b56961b634d039a491d97/main.ts#L85
> Probably better to use `if( file instanceof TFile)` instead of casting it directly.

Fixed [in this commit](https://github.com/MrZeroo00/obsidian-vault-changelog/commit/c55fdb17b23b19aba662804e522265c4a839df9c).

> https://github.com/MrZeroo00/obsidian-vault-changelog/blob/17e38683bf453340925b56961b634d039a491d97/main.ts#L132
> I think the `String()` isn't necessary? If you're running issues here then there might be a bug we need to fix.

You're right, it's unnecessary. Fixed [here](https://github.com/MrZeroo00/obsidian-vault-changelog/commit/bdb3ad14c324756d6023976c4cf91cfa82d685b2).

**Comment by liamcain** [comment]
> What would you suggest to this? I'm tempted to use something like async-mutex but it seems overkill.

You can just use the [`debounce`](https://github.com/obsidianmd/obsidian-api/blob/81ab85ade4552c9116c1e10d009127e62019c923/obsidian.d.ts#L591) function provided by the Obsidian API. You could wrap the function in debounce in your onLoad function.

Something like this should work:
```ts
this.watchVaultChange = debounce(this.watchVaultChange.bind(this), 200, false);
```

You'll also need to rebase this PR to resolve the conflicts.

MrZeroo00:
Fixed [here](https://github.com/MrZeroo00/obsidian-vault-changelog/commit/43b9b89d7e27562716b6a670c114ac8f4857d614), updated the release and rebased the PR.

MrZeroo00:
Awesome! Thanks for the reviews :)

PR #170, opened 2021-02-17
Add TODO plugin

larslockefeer:
A plugin for text-based GTD in Obsidian.

Features in this initial release:
- Aggregates all outstanding TODOs in your vault and lists them in a single view
- Split out TODOs by type ("Today", "Scheduled", "Inbox" and "Someday/Maybe")
- Schedule a TODO for a specific date by adding a tag #YYYY-DD-MM
- Mark a TODO as Someday/Maybe by adding a tag #someday
- Complete TODOs from the list view
- Quickly jump to the file in which a TODO is found from the list view

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -864,5 +864,13 @@
         "description": "Status bar item with vault statistics such as number of notes, files, attachments, and links",
         "author": "Bryan Kyle",
         "repo": "bkyle/obsidian-vault-statistics-plugin"
+    },
+    {
+        "id": "obsidian-plugin-todo",
+        "name": "Obsidian TODO",
+        "description": "Text-based GTD in Obsidian. Collects all outstanding TODOs from your vault and presents them in lists Today, Scheduled, Inbox and Someday/Maybe.",
+        "author": "Lars Lockefeer",
+        "repo": "larslockefeer/obsidian-plugin-todo",
+        "branch": "main"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -867,7 +867,7 @@
     },
     {
         "id": "obsidian-plugin-todo",
-        "name": "Obsidian TODO",
+        "name": "Obsidian TODO | Text-based GTD",
         "description": "Text-based GTD in Obsidian. Collects all outstanding TODOs from your vault and presents them in lists Today, Scheduled, Inbox and Someday/Maybe.",
         "author": "Lars Lockefeer",
         "repo": "larslockefeer/obsidian-plugin-todo",


liamcain:
https://github.com/larslockefeer/obsidian-plugin-todo/blob/5066b137419295c1ea0abd9371ba1412b99d049c/ui/TodoItemView.ts#L107 Just a heads up that Javascript .sort isn't stable so the order of tasks in the same group could change between renders. You could use lodash's sortBy instead

https://github.com/larslockefeer/obsidian-plugin-todo/blob/5066b137419295c1ea0abd9371ba1412b99d049c/model/TodoIndex.ts#L44 since you're writing back to disk, you should use .read() here instead of cachedRead.

https://github.com/larslockefeer/obsidian-plugin-todo/blob/5066b137419295c1ea0abd9371ba1412b99d049c/model/TodoIndex.ts#L82 I don't think you can assume these are always TFiles. You should properly check the type with instanceof since these could be new folders getting created/renamed.

https://github.com/larslockefeer/obsidian-plugin-todo/blob/5066b137419295c1ea0abd9371ba1412b99d049c/model/TodoIndex.ts#L57 nit: maybe use console.debug or put this behind a debug flag so the plugin isn't too noisy in the console.

https://github.com/larslockefeer/obsidian-plugin-todo/blob/5066b137419295c1ea0abd9371ba1412b99d049c/main.ts#L11 this looks unused.

---

Looks great! Just an idea but it might help to put GTD in the name. There are a bunch of "TODO" plugins coming out so it might help differentiate it a bit 😄

larslockefeer:
Thanks for your review @liamcain! I addressed most of your comments and republished the release assets.

The only comment I haven't addressed is the one about `Array.sort` not being stable yet. To my understanding, since Node v12, [sorting is stable](https://twitter.com/mathias/status/1120700101637353473) and [Electron 11.0](https://www.electronjs.org/blog/electron-11-0) bundles this version. The current Obsidian release uses Electron 11.1.0 so I think I should be safe?

liamcain:
> The only comment I haven't addressed is the one about Array.sort not being stable yet. To my understanding, since Node v12, sorting is stable and Electron 11.0 bundles this version. The current Obsidian release uses Electron 11.1.0 so I think I should be safe?

Oh, nice! TIL

community-plugins.json
```diff
@@ -849,5 +849,13 @@
         "author": "Joost Plattel",
         "repo": "jplattel/open-note-to-window-title",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-plugin-todo",
+        "name": "Obsidian TODO",
```
liamcain, line 855: I think this is the `name` that gets searched/displayed rather than the `name` in the manifest, if you want to change it here too.

liamcain (approved):
LGTM! 🚀

community-plugins.json
larslockefeer: Check! Updated it, thanks!

lishid:
https://github.com/larslockefeer/obsidian-plugin-todo/blob/cb044cb057692e8817fdb9d2b58dd30cc035109c/main.ts#L33
You might be interested in the onLayoutReady callback which does exactly this for you, which is available from 0.11.0 and up.

lishid:
Otherwise, looks great! Might need you to fix the PR so I can merge though.

larslockefeer:
Thanks @lishid, rebased to resolve the conflicts.

PR #171, opened 2021-02-17
added obsidian-vault-statistics-plugin

bkyle:


liamcain:
https://github.com/bkyle/obsidian-vault-statistics-plugin/blob/6719f7f9a2b77ad14cce47fb34fb6723d259159b/main.ts#L87 Instead of doing setTimeout, it might be simpler to use the [debounce function provided but the Obisidian API](https://github.com/obsidianmd/obsidian-api/blob/81ab85ade4552c9116c1e10d009127e62019c923/obsidian.d.ts#L591) to accomplish the same thing 

https://github.com/bkyle/obsidian-vault-statistics-plugin/blob/6719f7f9a2b77ad14cce47fb34fb6723d259159b/main.ts#L33 Instead of using a Record here, you might want to make this a Typescript interface. It will be functionally equivalent but you'll have add type safety to help point out any potential bugs.

Since the plugin is pulling info from metadata, you might want to listen for the `resolved` event on the metadataCache instead of putting listeners on the vault. Info about it here: https://github.com/obsidianmd/obsidian-api/blob/81ab85ade4552c9116c1e10d009127e62019c923/obsidian.d.ts#L1577

bkyle:
@liamcain thanks for the feedback.  I've made the suggested changes.  Please review at your convenience.

liamcain:
Looks good!

One small issue: your Github Release tag and your manifest.json file have a `v` in it https://github.com/bkyle/obsidian-vault-statistics-plugin/blob/dc6e5823f02aa0074f45b77cd154962de7ab950a/manifest.json#L4 the plugin installer expects the [exact version number](https://github.com/obsidianmd/obsidian-sample-plugin#releasing-new-releases), otherwise it might not load correctly.

You'll also need to rebase this PR since some other plugins have gotten merged in.

bkyle:
Done :)

liamcain (approved):

PR #172, opened 2021-02-18
Add Meld Encrypt plugin

meld-cp:
This plugin allows you to encrypt portions of your notes allowing you to store sensitive details along with other information.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -849,5 +849,13 @@
         "author": "Joost Plattel",
         "repo": "jplattel/open-note-to-window-title",
         "branch": "main"
+    },
+    {
+        "id": "meld-encrypt",
+        "name": "Meld Encrypt",
+        "description": "Hide secrets in your notes",
+        "author": "meld-cp",
+        "repo": "meld-cp/obsidian-encrypt",
+        "branch": "main"
     }
 ]


lishid:
Very cool plugin overall! Some suggestions:

Recommend auto-formatting your code to normalize some of the spacing between parenthesis and brackets.

https://github.com/meld-cp/obsidian-encrypt/blob/dfe34f36ce17f0baacc4fea7aeb6580747856d00/src/main.ts#L2
https://github.com/meld-cp/obsidian-encrypt/blob/dfe34f36ce17f0baacc4fea7aeb6580747856d00/src/main.ts#L95
Perhaps consider using the browser's built-in AES (it's what we use).

Just for the sake of users who might want to decrypt this later without Obsidian, you should probably document what the underlying library is doing:
- How is the key derived from the password string (crypto-es's docs only say "If you use a passphrase, then it will generate a 256-bit key.", but doesn't say how).
- What mode is AES used in (CBC or GCM, or something else).

For example, we use scrypt with a salt to derive a 256-bit key, which is then used with AES in GCM mode.

https://github.com/meld-cp/obsidian-encrypt/blob/dfe34f36ce17f0baacc4fea7aeb6580747856d00/src/main.ts#L30
Here I'd suggest using `getCursor('from')` and `getCursor('to')` instead of that, just because listSelections could give you multiple discontinuous selections if you were using the alt-click method to create multiple cursors.

https://github.com/meld-cp/obsidian-encrypt/blob/dfe34f36ce17f0baacc4fea7aeb6580747856d00/src/main.ts#L34
You're probably looking for the `getRange(from, to)` function.

meld-cp:
Thanks for the prompt feedback lishid.

I've taken your advice and removed the dependency on crypto-es.  Implementing SubtleCrypto was out of my comfort zone tbo, but it's working, and main.js is a whole ~500k lighter because of it.

I've added some details on how to manually decrypt if needed.

Your advice on using getCursor and getRange worked a treat thanks.

PR #174, opened 2021-02-18
add `obsidian-hotkeys-for-specific-files` plugin

Vinzent03:
In addition, I update the name of my other plugin `obsidian-shortcuts-for-starred-files`
https://github.com/Vinzent03/obsidian-hotkeys-for-specific-files


lishid:
https://github.com/Vinzent03/obsidian-hotkeys-for-specific-files/blob/44503973ffae6df0e5d3fdd29cede5985f054ca7/main.ts#L70
`index` is just `i` here right?

Vinzent03:
Not exactly. `index` keeps track of the index of the text field.  I need a "global" variable, because I need to know at which index I am, when I add a new text field. The for loop is used to create text fields for the already added files.

lishid:
Oh you're right, it's being used outside as well.

PR #175, opened 2021-02-19
Submit obsidian-quick-alias plugin

uonr:


lishid:
https://github.com/uonr/obsidian-quick-alias/blob/df5c664c4d3fdbee571464ae4332d88eeec01f7d/main.ts#L9

This works, but you're better off using `Workspace.getActiveViewOfType(MarkdownView)` and testing for null on that one. This will safely cast the returned View to MarkdownView as well, so you can use `getMode()` instead of getViewState and testing for mode in there.

uonr:
Close because I think this plugin just for personal use, please check another plugin https://github.com/obsidianmd/obsidian-releases/pull/182

PR #176, opened 2021-02-20
Initial release of csv-obsidian!

deathau:
### Features
- Open (and edit) CSV files right from Obsidian!
- Auto-saving
- Per-file setting for including headers (persisted in local storage)
- Markdown editing an preview for each individual cell (internal links aren't working correctly, yet)
- Sort the data by clicking on a column name
- Filter by column values
- Freeze columns
- Insert new columns/rows

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -879,5 +879,13 @@
         "description": "Set hotkeys for specific files and open them just with your keyboard.",
         "author": "Vinzent",
         "repo": "Vinzent03/obsidian-hotkeys-for-specific-files"
+    },
+    {
+        "id": "csv-obsidian",
+        "name": "CSV Editor",
+        "description": "Edit CSV files in Obsidian",
+        "author": "death_au",
+        "repo": "deathau/csv-obsidian",
+        "branch": "main"
     }
 ]


lishid:
https://github.com/deathau/csv-obsidian/blob/1aaa18a29bc1da00e30e4d9df29b3d77d52a63ee/src/main.ts#L14

You're still using the old (no longer recommended) style of saving data. Look up the new version in the sample plugin (and consider migrating your other plugins as well).

https://github.com/deathau/csv-obsidian/blob/1aaa18a29bc1da00e30e4d9df29b3d77d52a63ee/src/main.ts#L7
`inherits` and `basename` doesn't seem to be used.

https://github.com/deathau/csv-obsidian/blob/1aaa18a29bc1da00e30e4d9df29b3d77d52a63ee/src/main.ts#L152
I think you can just create a single function that takes no parameters and use that for all your callbacks that just save.

https://github.com/deathau/csv-obsidian/blob/1aaa18a29bc1da00e30e4d9df29b3d77d52a63ee/src/main.ts#L246
Any chance this parser can fail or throw exceptions? Might be worth calling reject if there is one.

lishid:
I think I'm good to merge this. Will do it as soon as the conflict is fixed!

deathau:
Thanks for the review!
- Settings code from an old template when this plugin doesn't have any settings? Of course, will fix.
- bad imports from vscode auto-completing the wrong thing? Thanks for the catch. Gone.
- I don't know why I was under the impression it wouldn't let me use a no-arg function, but that's an easy fix also
- Good point. I've added in some try/catch, and also an extra callback function for parse errors that will reject.

All of these fixes will be put in the _next_ release, as functionally they don't change anything for this one. Thanks for taking the time to have a look. I'm very grateful for the second pair of eyes.

lishid:
Yup, all suggestions are cosmetic/nits so it doesn't impact function.

As for the no-arg function, I just know that there were situations where it worked for me. Maybe a typescript upgrade would help, not sure.

For the error handling, I think try catch isn't needed insider the Promise function (I think it'd be automatically caught and rejected) but your library does things asynchronously you'll want to hook on some `error` or `fail` callbacks.

deathau:
ah, right, no need for try-catch. This is why I shouldn't code first thing in the morning 😆 
Thanks again.

Vinzent03:
I think the branch is not `main`. In your repo you use `master`. Cannot load the plugin in the store.

PR #177, opened 2021-02-21
add "Prettier Format" plugin

hipstersmoothie:


lishid:
Some suggestions:

https://github.com/hipstersmoothie/obsidian-plugin-prettier/blob/01b5bb6c2789847fdae09a40eea0354eceba664e/src/main.ts#L88
This is a source of memory leaks, since the app could be creating and destroying CodeMirror editors.
I would suggest instead in your onunload function to call `Workspace.iterateCodeMirrors` to unhook your keydown event.

You should also move this to `onload`.

https://github.com/hipstersmoothie/obsidian-plugin-prettier/blob/01b5bb6c2789847fdae09a40eea0354eceba664e/src/main.ts#L99
This is the old format for loading settings which doesn't work well. See the new format in [the sample plugin](https://github.com/obsidianmd/obsidian-sample-plugin/blob/master/main.ts).

https://github.com/hipstersmoothie/obsidian-plugin-prettier/blob/01b5bb6c2789847fdae09a40eea0354eceba664e/src/main.ts#L112
Use the new API `Workspace.getActiveViewOfType(MarkdownView)` instead.

https://github.com/hipstersmoothie/obsidian-plugin-prettier/blob/01b5bb6c2789847fdae09a40eea0354eceba664e/src/main.ts#L171
Not sure if it's a good idea, because users of your plugin might think it works when auto-saving, and it's not clear that you have to press Cmd+S (Also note `metaKey` is the Win key on Windows and Linux, and typically on those platforms you hit Ctrl+S instead of Win+S).

hipstersmoothie:
@lishid I am on a mac and when I presse my `cmd` key I get. So I think I need to check for both metaKey and ctrlKey.


Screen Shot 2021-02-22 at 10 54 25 AM

hipstersmoothie:
Is there a better way to hook into the document being saved? I updated the setting description to call out that it only works when saving via the hotkey. I guess if they changed that hotkey though my plugin wouldn't work.

lishid:
Yeah I can't think of a good way to do that, other than perhaps hooking up the vault modified event, compare it to the currently opened document, and performing the format on a delay. Gotta make sure your format save doesn't cause you to queue up another attempt too.

hipstersmoothie:
I'll see if I can do that

hipstersmoothie:
Okay it looks like obsidian periodically save the files automatically. From my experience using auto formatters you don't really want to always be formatting when modifiy/working. It's better to just do it when the user explicitly saves via hotkey.

Do you know if there is a way to get all the configured commands/hotkeys?

hipstersmoothie:
ooo `hotkeyManager` seems like what I want

hipstersmoothie:
Okay I think it's a bit better now. Will find the configured save hotkey and only run format if that hotkey was used 

https://github.com/hipstersmoothie/obsidian-plugin-prettier/blob/main/src/main.ts#L195-L217

lishid:
Looks good!

Some minor things maybe:
https://github.com/hipstersmoothie/obsidian-plugin-prettier/blob/663d419e2f2eecbb2038d5b430b5084aa921a143/src/main.ts#L118
`activeLeaf` should become `activeView`, and you should do a truthy check here instead of an instanceof (since getActiveViewOfType does an instanceof check internally and returns the properly typed MarkdownView).

https://github.com/hipstersmoothie/obsidian-plugin-prettier/blob/663d419e2f2eecbb2038d5b430b5084aa921a143/src/main.ts#L142
Should probably grab tabSize and indentWithTabs when the handler is executed, because those might be changed at runtime when users change settings.

https://github.com/hipstersmoothie/obsidian-plugin-prettier/blob/663d419e2f2eecbb2038d5b430b5084aa921a143/src/main.ts#L211
`Mod` is mapped to `metaKey` on MacOS and `ctrlKey` on Windows/Linux. We internally have a function to handle this but unfortunately I have not yet exposed this, so you might have to roll your own.

The `key` test is also slightly different in that we don't exactly use the `event.key` but a mapped version from `event.keyCode`. Unfortunately that's also not exported so there isn't much you can do here...

hipstersmoothie:
Ok, have the changes for the first 2 fixes.

> We internally have a function to handle this but unfortunately I have not yet exposed this, so you might have to roll your own.

I'm not sure I can detect this on my own, any tips?

> The key test is also slightly different in that we don't exactly use the event.key but a mapped version from event.keyCode. Unfortunately that's also not exported so there isn't much you can do here...

Any suggestion here either? is the implication that my plugin doesn't work as easily cross platform? It worked so far in my editor.

lishid:
> I'm not sure I can detect this on my own, any tips?
You'll have to test the user agent for mac and detect the `mod` as `isMac ? event.metaKey : event.ctrlKey`.

> Any suggestion here either? is the implication that my plugin doesn't work as easily cross platform? It worked so far in my editor.

I think the biggest issue is going to be if the user assigned the hotkey to special characters on a different language layout. For saving this is probably not a big deal.

Actually, a better idea is probably to monkey patch the save command's callback function and perform your task on the active editor using `Workspace.getActiveViewOfType(MarkdownView)`, then once you're done, call the original callback.

hipstersmoothie:
Done! Makes the code a bit simpler too

lishid:
Cool. One last thing you'll want to do is to either switch your repo's branch from main to master, or add `"branch": "main"` to your PR.

hipstersmoothie:
Done! accidentally removed that resolving a merge conflict

whitenoise-dev:
@hipstersmoothie  there is minor problem with the manifest.json. In your repo the version is 0.0.15 but in the released version is 0.0.14. The plugins keeps on asking for an update.

hipstersmoothie:
I'll take a look into it

hipstersmoothie:
Okay, should be fixed 🚀  While developing these plugin I also made a plugin to [auto](https://github.com/intuit/auto) to automate the release for any obsidian plugin, you can read more [here](https://github.com/zephraph/obsidian-tools/tree/main/packages/auto-plugin-obsidian).

PR #178, opened 2021-02-21
add table-of-contents plugin

hipstersmoothie:


lishid:
Pretty cool! Couple of suggestions:

https://github.com/hipstersmoothie/obsidian-plugin-toc/blob/95b3f849df512cbec7d364ca0420212bb2839be1/src/main.ts#L106
I don't think this is needed here.

https://github.com/hipstersmoothie/obsidian-plugin-toc/blob/95b3f849df512cbec7d364ca0420212bb2839be1/src/main.ts#L112
Don't use this - instead use `Workspace.getActiveViewOfType(MarkdownView)`.

https://github.com/hipstersmoothie/obsidian-plugin-toc/blob/95b3f849df512cbec7d364ca0420212bb2839be1/src/create-toc.ts#L51
You may be interested in our indexed version of headings, which are actually parsed from the markdown, so you don't run into the cases where your regex recognizes headings that are in fenced code blocks. Look for `App.metadataCache.getFileCache(file)`

hipstersmoothie:
I've made all the changes you wanted! That heading cache really made the code a bit more straightforward. Thanks for the tips!

lishid:
Ok getting close 👍 

https://github.com/hipstersmoothie/obsidian-plugin-toc/blob/62f9b667da555e28e628c479d5d75712c9ce62c7/src/main.ts#L113
Same as your other plugin - do a truthy check here and rename `activeLeaf` to `activeView`.

https://github.com/hipstersmoothie/obsidian-plugin-toc/blob/62f9b667da555e28e628c479d5d75712c9ce62c7/src/main.ts#L121
Seems like you're doing `getActiveFile` twice here. Not sure what the intent here but would be worth checking the logic here about what you'd want to be not null. You can also use `activeView.file` to get the view's file too as a shortcut.

https://github.com/hipstersmoothie/obsidian-plugin-toc/blob/62f9b667da555e28e628c479d5d75712c9ce62c7/src/main.ts#L139
This function seems very similar to the previous one. Might be worth combining the two?

hipstersmoothie:
All changes pushed!

lishid:
Cool looks good. I've merged the PR!

Probably missed this comment by the way:
> You can also use `activeView.file` to get the view's file too as a shortcut.
https://github.com/hipstersmoothie/obsidian-plugin-toc/blob/0272c76969b128ed9c6a0a93aec61fd9e2fe9564/src/main.ts#L150

hipstersmoothie:
Pushed an update for that too. Thanks for the code reviews!

PR #180, opened 2021-02-23
mochi cards exporter plugin

kalbetredev:
A plugin to export Markdown notes to Mochi cards

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -887,5 +887,13 @@
         "author": "death_au",
         "repo": "deathau/csv-obsidian",
         "branch": "main"
+    },
+    {
+        "id": "mochi-cards-exporter",
+        "name": "Mochi Cards Exporter",
+        "description": "Export Markdown notes to Mochi cards from within obsidian",
+        "author": "kalbetre",
+        "repo": "kalbetredev/mochi-cards-exporter",
+        "branch": "main"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -903,5 +903,5 @@
         "author": "kalbetre",
         "repo": "kalbetredev/mochi-cards-exporter",
         "branch": "main"
-    },
+    }
 ]


lishid:
https://github.com/kalbetredev/mochi-cards-exporter/blob/b4c28077e77256c8dab5a919b87613d924a22076/main.ts#L63
Might want to check the at the file you're exporting is indeed a markdown file with `file.extension === 'md'`.

https://github.com/kalbetredev/mochi-cards-exporter/blob/b4c28077e77256c8dab5a919b87613d924a22076/src/util/MochiExporter.ts#L99
You should be able to just return cards here, since it's an async function.

kalbetredev:
Thanks for the quick reviews. You are right I had to check if the current file is a markdown. Also about the direct return of the cards. I have applied all the changes.

lishid:
Looks good. You have another `return Promise` slightly after that I forgot to mention.

I would suggest passing the activeFile throughout your functions instead of always getting it from Workspace. This way you are sure the target file is always the same one.

If you can rebase your PR to resolve the merge conflict, I'll have it merged asap.

kalbetredev:
Thanks. If you have noticed, I retrieve the active file from the workspace only once when the exporter is initialized. And then all the functions use this activeFile. 

And ok I will do rebase

kalbetredev:
I have fixed the merge conflict.

lishid:
Here are the places that are getting the active file:

https://github.com/kalbetredev/mochi-cards-exporter/blob/40f44b5975993fddeb9d4796c204e4b8a7d758a2/main.ts#L33
https://github.com/kalbetredev/mochi-cards-exporter/blob/40f44b5975993fddeb9d4796c204e4b8a7d758a2/main.ts#L64
https://github.com/kalbetredev/mochi-cards-exporter/blob/40f44b5975993fddeb9d4796c204e4b8a7d758a2/src/util/MochiExporter.ts#L33

I think you should have the file passed in from the constructor, and also have it passed through `exportMochiCards`.

lishid:
Oh and I just noticed that you're using master. Should probably remove the `"branch": "main"` from the PR to make sure users can download your plugin correctly.

kalbetredev:
Ok got it. I have made all the necessary changes with regard to passing the active file to the exporter on the time of calling.  I have also renamed the master branch to main and made the required update.

Thank you.

lishid:
Great work!

kalbetredev:
Thank you very much.

PR #182, opened 2021-02-27
Submit "Furigana" plugin

uonr:


lishid:
The code is pretty simple and I don't have anything to say, but you'll need to fix this PR:
- `repo` should be `username/repo` and should not include the github part

uonr:
@lishid thanks, done.

PR #184, opened 2021-02-28
Obsidian Activity History plugin

Darakah:
## Repo URL

https://github.com/Darakah/obsidian-activity-history

## Release Checklist

- [x] The Github release contains all required files
  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css`
- [x] Github Release name matches the exact version number specified in your manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)


lishid:
https://github.com/Darakah/obsidian-activity-history/blob/1c8d276390716e0151654de1ff789b2d311449f4/src/main.ts#L25

You can just use an ES6 arrow function here to automatically have access to the outside `this` here:
```ts
setInterval(() => {
  // Can use this here
}, 0000);
```
https://github.com/Darakah/obsidian-activity-history/blob/1c8d276390716e0151654de1ff789b2d311449f4/src/main.ts#L38

Don't save settings on boot up. It's unnecessary and costs an extra save cycle the user will have to wait through when they boot up the app.

https://github.com/Darakah/obsidian-activity-history/blob/1c8d276390716e0151654de1ff789b2d311449f4/src/settings.ts#L165

Probably a better idea to `parseInt` here instead, then check for `isNaN`? 

https://github.com/Darakah/obsidian-activity-history/blob/1c8d276390716e0151654de1ff789b2d311449f4/src/utils.ts#L236

There's a debounce function exposed from our API now as well.

https://github.com/Darakah/obsidian-activity-history/blob/1c8d276390716e0151654de1ff789b2d311449f4/src/types.ts#L23

Your arrays are weird looking. `[Type]` means an array with a single element. I believe you mean to use `Type[]` instead?

Darakah:
Dear Licat, 

Thank you for the helpful remarks.

1. Replaced. Thanks a lot, much cleaner, didn't know about this.
2. Removed. Thanks for the suggestion.
3. Modified as suggested
4. Removed as It was not being used.
5. Modified as suggested.

lishid:
https://github.com/Darakah/obsidian-activity-history/blob/afd5267726e2b01fa435b6e1488b1dbc7409bd21/src/settings.ts#L169
 
Still using the `+` trick here - consider parsing the int into a variable at the beginning of your function and reusing that value.

Darakah:
Done, https://github.com/Darakah/obsidian-activity-history/blob/86c44ff49ad7ad85caf2d22c09d045cdcc655c81/src/settings.ts#L164

PR #185, opened 2021-02-28
Adds Footnote Shortcut plugin

akaalias:
Hi @lishid, 

This is a simple plugin based on this [forum discussion](https://forum.obsidian.md/t/footnote-shortcut/8872). 

3 members of the community have asked me to add it to the community plugins. 

I thought the idea was really good because I've also felt the barrier to footnotes was too high. I've since stopped using them personally.

Let me know what you think :) 

Thank you and have a great weekend. 

Alexis

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -948,5 +948,13 @@
         "description": "Render chess positions diagrams in note preview.",
         "author": "Davide Aversa",
         "repo": "THeK3nger/obsidian-chessboard"
+    },
+    {	
+        "id": "obsidian-footnotes",
+        "name": "Footnote Shortcut",
+        "description": "Makes creating footnotes in Obsidian more fun!",
+        "author": "Alexis Rondeau",
+        "repo": "akaalias/obsidian-footnotes",
+        "branch": "master"
     }
 ]


lishid:
https://github.com/akaalias/obsidian-footnotes/blob/75b3e1e13fd7697190ebdf4346150abddb75326e/main.ts#L23

Please use `Workspace.getActiveViewOfType(MarkdownView)` and do a truthy check on the result here.

akaalias:
Got it, this change is now part of release [0.0.3](https://github.com/akaalias/obsidian-footnotes/releases/tag/0.0.3)

=> https://github.com/akaalias/obsidian-footnotes/commit/184f164dafcb0f4049a266f0f1dbfbb422f1aa1c

liamcain:
https://github.com/akaalias/obsidian-footnotes/blob/2ad067659782be8240143dbffdb756a5678b32c4/main.ts#L92 If I understand this plugin correctly, hardcoding this to 6 will break when there are more than 9 footnotes.

https://github.com/akaalias/obsidian-footnotes/blob/2ad067659782be8240143dbffdb756a5678b32c4/main.ts#L12 while there's no hard rule for it, it's recommended that you don't provide a default hotkey to avoid conflicting with user settings.

https://github.com/akaalias/obsidian-footnotes/blob/2ad067659782be8240143dbffdb756a5678b32c4/main.ts#L46 Small nit: avoid using `var` and instead use `let` for variable declarations.

https://github.com/akaalias/obsidian-footnotes/blob/2ad067659782be8240143dbffdb756a5678b32c4/main.ts#L10 it looks like you changed this from checkCallback to callback, you might want to use checkCallback still and if `checking` is true, make sure you're in a MarkdownView. That will mean the command only shows in the command palette if the condition is met.

akaalias:
Hey @liamcain – Nice to me you. 

My name is Alexis and I'm building this plugin on behalf of a handful of members in the community. 

To your points: 

> https://github.com/akaalias/obsidian-footnotes/blob/2ad067659782be8240143dbffdb756a5678b32c4/main.ts#L92 If I understand this plugin correctly, hardcoding this to 6 will break when there are more than 9 footnotes.

The 6 was actually to place the cursor at the end of the line BUT, you're right – This would have degraded when a footnote index is double-digit or more. I've addressed this in the next release. Good catch.

> https://github.com/akaalias/obsidian-footnotes/blob/2ad067659782be8240143dbffdb756a5678b32c4/main.ts#L12 while there's no hard rule for it, it's recommended that you don't provide a default hotkey to avoid conflicting with user settings.

If you look at [the genesis of this plugin](https://forum.obsidian.md/t/footnote-shortcut/8872) and the [plugin README](https://github.com/akaalias/obsidian-footnotes), you'll see that this is designed as a "hotkey-first" feature that only makes sense as a hotkey for speed and usability. In fact, as this started out as a hotkey request for Obsidian core hotkey I hope the plugin goes away at some point. 

> https://github.com/akaalias/obsidian-footnotes/blob/2ad067659782be8240143dbffdb756a5678b32c4/main.ts#L46 Small nit: avoid using var and instead use let for variable declarations.

Addressed in the next issue.

> https://github.com/akaalias/obsidian-footnotes/blob/2ad067659782be8240143dbffdb756a5678b32c4/main.ts#L10 it looks like you changed this from checkCallback to callback, you might want to use checkCallback still and if checking is true, make sure you're in a MarkdownView. That will mean the command only shows in the command palette if the condition is met.

Maybe you can shed some light on this because I'm now using `callback: () =>` because `checkCallback(checking: boolean) =>` presented me with a very odd behaviour before:

- Given I have the footnotes plugin installed
- When  I press `Command-P`
- And I press `ESC` (or other keys)

**Expected behavior**
Noting, really. I just ESC'ed out of Command Palette. That should just do nothing, right?

**Observed behavior**
The footnotes plugin is falsely triggered and inserts a footnote.

Here's a demo of what I mean. 

command-p

liamcain:
Hi @akaalias, nice meeting you too! Your Journey plugin is really amazing.

---

> If you look at the genesis of this plugin and the plugin README, you'll see that this is designed as a "hotkey-first" feature that only makes sense as a hotkey for speed and usability. In fact, as this started out as a hotkey request for Obsidian core hotkey I hope the plugin goes away at some point.

I think it's great to encourage its usage via a hotkey! In this case it might make more sense to keep the default hotkey blank and encourage users to select their own. Your default of `mod shift 6` on Windows becomes `Windows Shift 6` which I think conflicts with a system-wide hotkey.

> Maybe you can shed some light on this because I'm now using callback: () => because checkCallback(checking: boolean) => presented me with a very odd behaviour before:

Yea, `checkCallback` is a little wonky and could be better documented in the API Spec. I wrote about how it should be used here: https://liamca.in/Obsidian/API+FAQ/commands/what+does+checking+mean+in+checkCallback

Also, it looks like you reformatted the entire `community-plugins.json` file in your last commit. Could you revert it so it maintains the same formatting?

akaalias:
Hey there! Thank you for the infos. 

> I think it's great to encourage its usage via a hotkey! In this case it might make more sense to keep the default hotkey blank and encourage users to select their own. Your default of mod shift 6 on Windows becomes Windows Shift 6 which I think conflicts with a system-wide hotkey.

I personally think it makes complete sense to have it work "out-of-the-box" BUT I'd rather base the decision on what the community expects. I've [posted the question here](https://forum.obsidian.md/t/footnote-shortcut/8872/43?u=akaalias), so let's see what comes back :) 

> Yea, checkCallback is a little wonky and could be better documented in the API Spec. I wrote about how it should be used here: https://liamca.in/Obsidian/API+FAQ/commands/what+does+checking+mean+in+checkCallback

Oh, okay, I would never have figured this out myself. The odd behavior I showed you is now gone with the next release.

> Also, it looks like you reformatted the entire community-plugins.json file in your last commit. Could you revert it so it maintains the same formatting?

I just used the "Resolve Conflict" button right in the browser. Do you know how to revert this commit?

DutchPete:
I am in favour of the `Cmd+Shift+6` (MacOS) precisely because it makes sense to have the "link" with the `^` that is used for footnotes. I don't understand why Liam states it is `mod+shift+6` on Windows, I would have thought the Windows equivalent is `CTRL+Shit+6`.

Anyway, if you leave the hotkey blanc I will assign `Cmd+Shift+6`.

akaalias:
@DutchPete Thank you for checking in! After installing the plugin, would you rather...

- **have** a default hotkey already set up for you
OR
- **not** have a default hotkey set up and require you configure it after installation

DutchPete:
I don't mind, do what you think is best. Perhaps a default hotkey is possible that behaves properly across platforms. If you do set one, I think it is use ful to have the `shift+6` ion it because of the caret-footnote link.

I-Pch:
Hi there 😊!

To be honest, I'm used to the plugins with no hot keys set up and therefore defining them later (even though I use the control palette more than any hot keys I've setup myself 😇).

I get the worries about some conflicts which may occur with other hot keys so I probably would be choose the security and go with no hot key setup 😊... But that's just me 😊.

Maybe you could just suggest the hot key in the description of the plugin, leaving the users free to use it or choose something that might suit them better.

liamcain:
> I just used the "Resolve Conflict" button right in the browser. Do you know how to revert this commit?

I don't think it's possible within the browser, but you can rebase your branch locally and `--force` push it.

akaalias:
> > I just used the "Resolve Conflict" button right in the browser. Do you know how to revert this commit?
> 
> I don't think it's possible within the browser, but you can rebase your branch locally and `--force` push it.

Sweet, I _believe_ I did that successfully [following these steps](https://medium.com/@topspinj/how-to-git-rebase-into-a-forked-repo-c9f05e821c8a)

Regarding hotkey or not: To move forward and get this in the hands of the community, I've taken the hotkey out in today's release [0.0.6](https://github.com/akaalias/obsidian-footnotes/releases/tag/0.0.6)

I'll see to explaining this clearly in the README. Let's do this: IF I get a lot of questions/issues such as "Why is this not working?" let's discuss again for a future release :)

DutchPete:
Good steps forward, Alexis. Couple of remarks.

1. You say "I've taken the hotkey out" but in 0.06 `Cmd+Shift+6` still works, which is great. After all, isn't a hotkey the idea behind this *Footnote shortcut* project?
2. I tested it in a note with pre-existing inline footnotes. The "Alexis" (= normal 😄 ) footnote , number 2 in this case, between 2 pre-existing footnotes has its text on a line below the number - see screenshot below.

image

3. The reindexing of inline footnotes and normal footnotes works fine, kudos for that. I can still see confusion on the horizon between numbers in Edit mode and in Preview mode. Referring to the example in point 2, the normal footnote, which has number 2 in Preview, has number 1 in Edit mode.

So, while we need not worry about selecting the right number for a new, normal footnote thanks to reindexing, there is still the mismatch between Edit and Preview numbering. My feeling is that this is an issue that cannot be eliminated.

akaalias:
Hey @DutchPete Thanks for that! 

> You say "I've taken the hotkey out" but in 0.06 Cmd+Shift+6 still works, which is great. After all, isn't a hotkey the idea behind this Footnote shortcut project? 

I had the same thing – It's because you have the hotkey from previous releases. 

> After all, isn't a hotkey the idea behind this Footnote shortcut project?

I agree that having a hotkey is the whole idea of the project but it seemed like having it set up by default was a show-stopper to get this into the Obsidian directory. 

@liamcain and @lishid please advise if a default hotkey – the whole raison d'etre of this plugin imho – is a show-stopper or not.

> I tested it in a note with pre-existing inline footnotes. The "Alexis" (= normal 😄 ) footnote , number 2 in this case, between 2 pre-existing footnotes has its text on a line below the number - see screenshot below.

Please file this as an issue under https://github.com/akaalias/obsidian-footnotes/issues because I couldn't reproduce the behaviour. 

> The reindexing of inline footnotes and normal footnotes works fine, kudos for that. 

That's actually default Obsidian Preview behavior. The plugin is not yet syncing between the indexes you see in edit and in Preview modes. That'll be part of [the reindexing feature](https://github.com/akaalias/obsidian-footnotes#automatically-re-index-footnotes)

Thanks!

liamcain (approved):

DutchPete:
> Please file this as an issue under https://github.com/akaalias/obsidian-footnotes/issues because I couldn't reproduce the behaviour.

Done.

akaalias:
Hi @lishid and @liamcain! Let me know if there's anything else missing to adding this plugin. Thanks so much :)

Calorion:
I strongly approve of having a default hotkey. Given that it doesn't do anything without one, users may get confused at its lack. If they don't like it they can change it.

PR #189, opened 2021-03-05
Update ITS Theme Image

SlRvb:
Just to display light and dark version of the theme in 1 image :).

### DIFF
--- a/community-css-themes.json
+++ b/community-css-themes.json
@@ -363,7 +363,7 @@
         "name": "ITS Theme",
         "author": "SlRvb",
         "repo": "SlRvb/Obsidian--ITS-Theme",
-        "screenshot": "Darkmode.png",
+        "screenshot": "ITS.png",
         "compatibleModes": ["dark", "light"],
         "branch": "main"
     }
--- a/community-css-themes.json
+++ b/community-css-themes.json
@@ -362,7 +362,7 @@
     {
         "name": "ITS Theme",
         "author": "SlRvb",
-        "repo": "Obsidian--ITS-Theme",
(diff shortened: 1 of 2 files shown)


ericaxu:
I don't think a PR is necessary -- you can just replace the old image but keep the name. Was it because the name would be inaccurate?

SlRvb:
Yeah it was because the name was inaccurate totally my own fault for not future proofing the name _sorry._  😅

ericaxu:
> Yeah it was because the name was inaccurate totally my own fault for not future proofing the name _sorry._ 😅

Thanks for the explanation. I think that's fine, but could please remove the previous listing so we don't have both showing up?

SlRvb:
Like remove the filename `Darkmode.png` from my repo to avoid it popping up on the themes page? I can do that.

ericaxu:
> Like remove the filename `Darkmode.png` from my repo to avoid it popping up on the themes page? I can do that. Unless that's not what you mean? Do you mean the previous commits?

I meant there's already an entry for your theme here: https://github.com/obsidianmd/obsidian-releases/blob/master/community-css-themes.json#L363

After this PR, I think it would show up twice 🤔

SlRvb:
So mine wouldn't replace just the line that I changed? 🤔

ericaxu:
I see where the issue is... your fork is too outdated, so in your fork your theme is still the last theme, but if you check the file in our master, there has been a few after yours. I'll update it and close this, that would be easier for both of us!

ericaxu:
@SlRvb Should be added in https://github.com/obsidianmd/obsidian-releases/commit/69f1209f51554c90cfa39b1c3958baca271eded1. Thanks!

PR #190, opened 2021-03-05
Add Obsidian Leaflet Plugin

valentine195:
Added the ability to create interactive maps using Leaflet.js.

Includes the ability to have persistent map markers (that can link to a note), drag-n-drop a note to create a marker, and link markers to notes to open.

Thanks!
Jeremy


lishid:
https://github.com/valentine195/obsidian-leaflet-plugin/blob/ce79e32c68fc64a7a54a8d292e4db9aa9f6eb7a4/src/main.ts#L204
Please use `let view = Workspace.getActiveViewOfType(MarkdownView)` and perform a truthy check on the result.

https://github.com/valentine195/obsidian-leaflet-plugin/blob/ce79e32c68fc64a7a54a8d292e4db9aa9f6eb7a4/src/main.ts#L187
Don't recommend doing this when you have complex data-types like `LeafletMap` in your settings object. This simply saves it using `JSON.stringify()`, which could generate a lot of useless things when you include things like DOM elements, etc. Keep your settings data type simple (primitives, arrays and plain objects only).

https://github.com/valentine195/obsidian-leaflet-plugin/blob/ce79e32c68fc64a7a54a8d292e4db9aa9f6eb7a4/src/main.ts#L163
Wouldn't this cause memory leaks? Doesn't seem to be cleaned up anywhere (e.g. when you close this file, etc).

https://github.com/valentine195/obsidian-leaflet-plugin/blob/ce79e32c68fc64a7a54a8d292e4db9aa9f6eb7a4/src/main.ts#L18
Maybe define/name what these `key` means? (Hint: You can write a name here instead of `key`)

General git stuff: Consider adding a gitignore for `.DS_Store` and `main.js`? Could just clone the sample plugin repo for those.

valentine195:
> https://github.com/valentine195/obsidian-leaflet-plugin/blob/ce79e32c68fc64a7a54a8d292e4db9aa9f6eb7a4/src/main.ts#L204
> Please use `let view = Workspace.getActiveViewOfType(MarkdownView)` and perform a truthy check on the result.

Done.

> https://github.com/valentine195/obsidian-leaflet-plugin/blob/ce79e32c68fc64a7a54a8d292e4db9aa9f6eb7a4/src/main.ts#L187
> Don't recommend doing this when you have complex data-types like `LeafletMap` in your settings object. This simply saves it using `JSON.stringify()`, which could generate a lot of useless things when you include things like DOM elements, etc. Keep your settings data type simple (primitives, arrays and plain objects only).

Done.

> https://github.com/valentine195/obsidian-leaflet-plugin/blob/ce79e32c68fc64a7a54a8d292e4db9aa9f6eb7a4/src/main.ts#L163
> Wouldn't this cause memory leaks? Doesn't seem to be cleaned up anywhere (e.g. when you close this file, etc).

I added an event listener to file deletion to remove the map marker data from the saved data. This data needs to persist through file closing, as it uses the data to rebuild the markers when the file is reopened. 

I'm still trying to think of a better way to store a reference to each markdown code block, to be able to store/retrieve the data in a better way, but this will prevent the data object from just continuously getting bigger. Not sure if you any ideas for this or not.

> https://github.com/valentine195/obsidian-leaflet-plugin/blob/ce79e32c68fc64a7a54a8d292e4db9aa9f6eb7a4/src/main.ts#L18
> Maybe define/name what these `key` means? (Hint: You can write a name here instead of `key`)

Done, thanks. First typescript project, still getting used to it.

> General git stuff: Consider adding a gitignore for `.DS_Store` and `main.js`? Could just clone the sample plugin repo for those.

Also done.

Thanks for the feedback.

lishid:
Apologies for the late reply.

> I added an event listener to file deletion to remove the map marker data from the saved data. This data needs to persist through file closing, as it uses the data to rebuild the markers when the file is reopened.
> 
> I'm still trying to think of a better way to store a reference to each markdown code block, to be able to store/retrieve the data in a better way, but this will prevent the data object from just continuously getting bigger. Not sure if you any ideas for this or not.

So in your case, it looks like you actually you'll need some lifecycle control over the maps. To do this, you should register a child when you render your block. [The docs on that are here](https://github.com/obsidianmd/obsidian-api/blob/81ab85ade4552c9116c1e10d009127e62019c923/obsidian.d.ts#L1197)

By implementing this API, you'll get notified when the block is removed, such as when the user edit out your block, or when the page is closed. Without doing this, you'll keep accumulating more map objects when the user navigates, and edits the document containing the map. This will probably lead to a memory leak.

Actually about that, perhaps you should also implement something [in the `onunload()` handler](https://github.com/valentine195/obsidian-leaflet-plugin/blob/1c31f3f94dd2a7c5ab7dd807226301897bd5ec38/src/main.ts#L112) to cleanup any in memory data structures when your plugin is disabled by the user?

valentine195:
Thanks, I'll look into this. The maps structure *shouldn't* accumulate when the user navigates or edits the document - I'm currently storing it with an id based on the source path of the file and the URL that is provided that is preventing maps from getting pushed every time the postprocessor is triggered: https://github.com/valentine195/obsidian-leaflet-plugin/blob/1c31f3f94dd2a7c5ab7dd807226301897bd5ec38/src/main.ts#L153

What I've been struggling with is capturing when the user changes the image url provided in the code block, as that wouldn't be captured by that and *would* push a new map, but I think the lifecycle management you've linked to will solve that issue.

> Actually about that, perhaps you should also implement something in the onunload() handler to cleanup any in memory data structures when your plugin is disabled by the user?

Okay, I'll clear everything out on unload.

valentine195:
Ok, both of these changes should be good to go

lishid:
Perfect. Can you fix the PR conflict, then we can get it merged.

valentine195:
Done!

PR #191, opened 2021-03-07
Add format-hotkeys-obsidian

anstosa:


anstosa:
Wow thanks for the code review @liamcain! Really appreciate it. Fixed everything according to your tips

> https://github.com/anstosa/format-hotkeys-obsidian/blob/fc0a6e5668659a319e86c60ee426ab1fddb597a1/src/main.ts#L313-L325 I'm still on my first pass, but I'm not quite sure I understand what's going on in this section. One thing that might help I think if you add types to your function arguments. Is looks like force is a mixed type here? Or at least optional. Is there any way to avoid the force flag entirely?

That function is typed at the top but it's SUPER complicated and I definitely need to do another pass there too. I think it probably needs to be two functions (`togglePrefix` and `setPrefix`) that share some utilities.

```
interface TogglePrefixBase {
  searches: string[];
  replace?: string[];
  force?: boolean;
  replaceMatches?: string[];
  remove?: (content: string, searches: TogglePrefixBase["searches"]) => string;
}

interface TogglePrefixLiteral extends TogglePrefixBase {
  prefix: string;
  add?: undefined;
}

interface TogglePrefixFunction extends TogglePrefixBase {
  prefix?: undefined;
  add: (content: string, searches: TogglePrefixBase["searches"]) => string;
}

type TogglePrefix = TogglePrefixLiteral | TogglePrefixFunction;
```

> https://github.com/anstosa/format-hotkeys-obsidian/blob/fc0a6e5668659a319e86c60ee426ab1fddb597a1/src/main.ts#L405-L411 It might make more sense to let CodeMirror do the heavy lifting here and use cm.execCommand('indentMore') and cm.execCommand('indentLess') - though I'm not sure how those commands handle multiple selections...

I had gone around those because I thought they were buggy but your comment made me look again and realize that there is a <kbd>Ctrl</kbd> + <kbd>[</kbd> keyboard conflict with vim mode. [I'll try to repair that silently or at least document how to fix it for vim users.](https://github.com/anstosa/format-hotkeys-obsidian/issues/1)

liamcain:
> Fixed everything according to your tips

Fixes look good to me!

> . I think it probably needs to be two functions (togglePrefix and setPrefix) that share some utilities

Yeah, I think pulling the `add` and `remove` methods out onto the Plugin class would help with readability too. Then you don't need `force` and can just call `add` and `remove` directly.

> I had gone around those because I thought they were buggy but your comment made me look again and realize that there is a Ctrl + [ keyboard conflict with vim mode. I'll try to repair that silently or at least document how to fix it for vim users.

At least for me, <kbd>Mod</kbd><kbd>[</kbd> anbd <kbd>Mod</kbd><kbd>]</kbd> already changes the indent, vim mode or not (and it doesn't affect the cursor position). Maybe these can just be removed from the plugin?

---

Let me know how you want to proceed and I can take another pass!

anstosa:
> At least for me, Mod[ anbd Mod] already changes the indent, vim mode or not (and it doesn't affect the cursor position). Maybe these can just be removed from the plugin?

🤦🏽‍♂️ removed.

> Yeah, I think pulling the add and remove methods out onto the Plugin class would help with readability too. Then you don't need force and can just call add and remove directly.

Done! Definitely feels much more reasonable after refactor.

I just pushed a new tag pending release by GitHub actions. I think we're ready to go?

liamcain:
Awesome, looks great! A couple small nits - 

https://github.com/anstosa/format-hotkeys-obsidian/blob/e5e0521f07581dab7bf11b3a27a4cda1b88d4c92/src/main.ts#L247
I'm pretty sure this REGEX_ANY is redundant since `prefixLines` already appends it.

https://github.com/anstosa/format-hotkeys-obsidian/blob/e5e0521f07581dab7bf11b3a27a4cda1b88d4c92/src/main.ts#L289 ES6 already provides an `every` function so you don't need to use lodash for this.

 https://github.com/anstosa/format-hotkeys-obsidian/blob/e5e0521f07581dab7bf11b3a27a4cda1b88d4c92/src/main.ts#L359 can use your `log()` function here instead

https://github.com/anstosa/format-hotkeys-obsidian/blob/e5e0521f07581dab7bf11b3a27a4cda1b88d4c92/src/main.ts#L324 I think this comment is stale?

---

Anyway, nothing blocking. :shipit:

liamcain (approved):

anstosa:
> https://github.com/anstosa/format-hotkeys-obsidian/blob/e5e0521f07581dab7bf11b3a27a4cda1b88d4c92/src/main.ts#L247
I'm pretty sure this REGEX_ANY is redundant since prefixLines already appends it.

Yup, good catch

> https://github.com/anstosa/format-hotkeys-obsidian/blob/e5e0521f07581dab7bf11b3a27a4cda1b88d4c92/src/main.ts#L289 ES6 already provides an every function so you don't need to use lodash for this.

Yeah I'll reduce my reliance on lodash so more of it can fall out in tree-shaking

> https://github.com/anstosa/format-hotkeys-obsidian/blob/e5e0521f07581dab7bf11b3a27a4cda1b88d4c92/src/main.ts#L359 can use your log() function here instead

Removed

> https://github.com/anstosa/format-hotkeys-obsidian/blob/e5e0521f07581dab7bf11b3a27a4cda1b88d4c92/src/main.ts#L324 I think this comment is stale?

Note stale but in the wrong place. Moved it.

Thank you so much for the detailed review!

lishid:
Code looks good. Your release seems a bit big for me at 1.64MB (Obsidian's app javascript file is 1.37MB right now). Perhaps you're including CodeMirror and/or lodash in your build?

Seems like for the things you're using lodash, you might as well just write a small util function. For CodeMirror you can probably use `import type {...} from 'codemirror';` instead to avoid pulling a whole copy of the library in.

liamcain:
Taking a peek at the output, it looks like the bulk of the `main.js` is from lodash and the inline sourcemaps. 

I think you need to use a different package to actually get treeshaking with lodash + rollup.

And it might make sense to just include sourcemaps in the dev build (so put it behind an env var)

lishid:
I'm only seeing `isString` and `times` from lodash. I'd recommend just use a helper function for those, and avoid porting in lodash:

isString: `if(typeof myvar === 'string')`
times: `new Array(n).fill('#').join('')`

anstosa:
Great points all. I did a bunch of production bundle optimization (now down to **6.25 KB**) including swapping lodash for first party utils, doing type imports from CodeMirror, removing the sourcemaps in production, and using terser in production.

lishid:
Uh oh, looks like the merge rewrote the whole file. I think github's UI does that and swaps the line endings :(

Any chance you can get it fixed, and force push a clean commit?

anstosa:
Oh god, of course it does. That's what I get for using the website instead of the command line 🤦🏽‍♂️. Yeah once sec

anstosa:
There we go, clean commit!

anstosa:
Thanks all! 🙏🏽

Obsidian is amazing. Excited to be part of it 😊

lishid:
Great work!

PR #192, opened 2021-03-10
Code block from selection plugin

derwish-pro:


liamcain:
https://github.com/derwish-pro/obsidian-code-block-from-selection/blob/a86ca5d79527c219756fb97703246d6257648dbd/main.ts#L43 you probably want a leading `\n` here right?

https://github.com/derwish-pro/obsidian-code-block-from-selection/blob/a86ca5d79527c219756fb97703246d6257648dbd/main.ts#L88 You might want to rename the class here from `SampleSettingTab`.

https://github.com/derwish-pro/obsidian-code-block-from-selection/blob/a86ca5d79527c219756fb97703246d6257648dbd/main.ts#L9 Empty string might be a better default value for this.

https://github.com/derwish-pro/obsidian-code-block-from-selection/blob/a86ca5d79527c219756fb97703246d6257648dbd/main.ts#L27 While there's not a hard rule on this, it's recommended you don't provide a default hotkey and instead leave it to the users to define one in the Hotkey menu for themselves. This avoids keybindings conflicting with existing user settings or OS defaults. For example, <kbd>Mod</kbd> <kbd>`</kbd> on macOS universally cycles through open application windows.

https://github.com/derwish-pro/obsidian-code-block-from-selection/blob/a86ca5d79527c219756fb97703246d6257648dbd/main.ts#L47 this will break if the active leaf is not a MarkdownView. [Here's a snippet that could help](https://liamca.in/Obsidian/API+FAQ/editor/get+the+active+CodeMirror+instance).

derwish-pro:
Fixed

liamcain (approved):

PR #193, opened 2021-03-11
Remember cursor position plugin

derwish-pro:


lishid:
https://github.com/derwish-pro/obsidian-remember-cursor-position/blob/e273b9bfcba85bf18b1062822fff174f525fa938/main.ts#L82
This trim is probably not necessary.

https://github.com/derwish-pro/obsidian-remember-cursor-position/blob/e273b9bfcba85bf18b1062822fff174f525fa938/main.ts#L81
Probably worth moving this code a few lines down just before it's needed to save work.

https://github.com/derwish-pro/obsidian-remember-cursor-position/blob/e273b9bfcba85bf18b1062822fff174f525fa938/main.ts#L218
Missing a null check here.

https://github.com/derwish-pro/obsidian-remember-cursor-position/blob/e273b9bfcba85bf18b1062822fff174f525fa938/main.ts#L37
Considering adding an error handling here, because if this read fails your entire plugin (events, settings, etc) won't work. Maybe move this a few lines down after you register everything too.

derwish-pro:
Fixed

lishid:
LGTM. Can you rebase one more time? I've fixed the file so github merges won't change newlines for the whole file every time.

PR #196, opened 2021-03-13
Consistent attachments and links plugin

derwish-pro:

## Repo URL

https://github.com/derwish-pro/obsidian-consistent-attachments-and-links/

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


lishid:
https://github.com/derwish-pro/obsidian-consistent-attachments-and-links/blob/92fc775e36c33c5ebc3a387ab7839fbeeb599ec8/src/main.ts#L36

Instead of dicing the path yourself, you can do `if(file instanceof TFile && file.extension === 'md')`

https://github.com/derwish-pro/obsidian-consistent-attachments-and-links/blob/92fc775e36c33c5ebc3a387ab7839fbeeb599ec8/src/main.ts#L7
You're using NodeJS's `path` package, which means your plugin is desktop only - it will crash on other platforms where node is not available. Look into writing your own helpers for the path manipulations, or change your plugin manifest to say isDesktopOnly: true.

https://github.com/derwish-pro/obsidian-consistent-attachments-and-links/blob/92fc775e36c33c5ebc3a387ab7839fbeeb599ec8/src/links-handler.ts#L18
Might want to consider using `app.metadataCache.getFirstLinkpathDest` instead of rolling your own here. It seems that your link resolution doesn't handle `#` fragments for subpaths. I think it also doesn't like the omission of the extension `.md`, and doesn't do case-insensitive matching.

https://github.com/derwish-pro/obsidian-consistent-attachments-and-links/blob/92fc775e36c33c5ebc3a387ab7839fbeeb599ec8/src/links-handler.ts#L42
Just a note that this is likely very unperformant for large vaults, but it doesn't seem to be used anywhere.

---

After reading through the codebase, I got an understanding that this requires the user to use a very opinionated way of managing their files (and obsidian file/path settings). I think in this case, you should probably add a disclaimer section at the beginning of your README to **require** the user to configure & use obsidian with your settings, such as disabling Obsidian's link updating algorithm, using relative paths, and disabling wikilinks.

Performance: There's a couple things I'd recommend keeping in mind - renaming a folder will issue a rename event for each of its child file/folders. I feel like the plugin would be exponentially doing work in that case, and a lot of it might not be cheap especially for large vaults (such as iterating through every file's links & embeds, and performing link resolution on those, and doing that for the whole vault for each rename event).
Nothing specific here I can recommend, but just something to keep in mind.

derwish-pro:
> Instead of dicing the path yourself, you can do if(file instanceof TFile && file.extension === 'md')

ok, it's convenient, thanks. however in that case the type is TAbstractFile which does not have this field.

> You're using NodeJS's path package, which means your plugin is desktop only - it will crash on other platforms where node is not available. Look into writing your own helpers for the path manipulations, or change your plugin manifest to say isDesktopOnly: true.

manifest changed

> Might want to consider using app.metadataCache.getFirstLinkpathDest instead of rolling your own here. It seems that your link resolution doesn't handle # fragments for subpaths. I think it also doesn't like the omission of the extension .md, and doesn't do case-insensitive matching.

The point of this method is to return a file if the link to it meets all the requirements of the markdown format. I want to ignore all proprietary solutions that have been added to the link format in Obidian. This is the meaning of the consistency of the notes, so that there is compatibility with all other programs that support the markdown format. 

The fact is that this is not just a plugin for moving attachments, as it might seem at first glance. If you use the approach to attachments that is used in Obsidian (when it finds files, even if the link leads to a neighboring folder), then in this case it makes no sense to move attachments. But I ran into a problem when I tried to upload my notes to git-pages, or tried to open them in another program (including mobile), because it requires more stringent matching of paths in links. Then I came up with the idea of ​​a plugin that will check links for compliance with the markdown format and will help keep links and files connected.

Probably, I did not very clearly explain this idea in the description for the plugin, I will supplement it.

I understand at the expense of performance. I think that it will be possible to optimize something in the future if there is such a request from the community. For now, I just want to get this to work as intended and hope to get feedback from other users if there are any bugs in the plugin's behavior.

derwish-pro:
reame updated

derwish-pro:
Is it possible to make "Update links" Obsidian dialog optional? Adding a "Don't ask" button to this dialog would be a great solution.

derwish-pro:
I renamed the plugin description in the manifest too, so that it doesn't seem that moving attachments is the main task of the plugin.

lishid:
> however in that case the type is TAbstractFile which does not have this field.

Using `file instanceof TFile` will automatically make TypeScript understand that `file` is a `TFile` in any subsequent code.

> Is it possible to make "Update links" Obsidian dialog optional? Adding a "Don't ask" button to this dialog would be a great solution.

There isn't a setting backing that I think, but you should be able to monkey-patch app.fileManager.renameFile and just run vault.rename instead of performing the link updating.

derwish-pro:
I couldn't get Obsidian to update the links correctly with maintaining the relative path, which is why I decided to update the links myself. Also, I constantly encounter the fact that obsidian works with metadataCache, which is updated with a delay, which also causes problems and forces me go for tricks.

derwish-pro:
Maybe there is a function that forces the metadataCache to be updated and lets me wait for the right time?

lishid:
There is an unexposed function `metadataCache.onCleanCache(callback)` that is fired when all metadata is resolved. Could consider using it with a `as any` cast.

derwish-pro:
I have added almost all the features I wanted. Please read the updated readme. Are there any further actions expected from me?

lishid:
That's fine. Let's have this merged when you fix the merge conflict.

derwish-pro:
complete

PR #197, opened 2021-03-14
include trashhalo/obsidian-extract-url in plugins

trashhalo:
[x] I am submitting a new Community Plugin

## Repo URL

https://github.com/trashhalo/obsidian-extract-url

## Release Checklist

- [x] tested on linux
- [x] tested on windows
  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


lishid:
Pretty cool!

Consider updating `.gitignore` to include your built main.js and main.js.map files, as well as removing the error.log from the repo.

https://github.com/trashhalo/obsidian-extract-url/blob/c2ca1296a79fede119000d9fe8cadfef9e337c33/index.js#L13
Don't need the async here.

https://github.com/trashhalo/obsidian-extract-url/blob/c2ca1296a79fede119000d9fe8cadfef9e337c33/index.js#L28
Don't use activeLeaf, especially not without null-checks. Here you're looking for something like:
```ts
let view = this.app.workspace.getActiveViewOfType(MarkdownView);
if(view) {
  let editor = view.sourceMode.cmEditor;
}
```

https://github.com/trashhalo/obsidian-extract-url/blob/c2ca1296a79fede119000d9fe8cadfef9e337c33/index.js#L30
Just use `getSelection` here?

https://github.com/trashhalo/obsidian-extract-url/blob/c2ca1296a79fede119000d9fe8cadfef9e337c33/index.js#L34
Consider promisifying this and using async await here.

https://github.com/trashhalo/obsidian-extract-url/blob/c2ca1296a79fede119000d9fe8cadfef9e337c33/html-readability/index.js#L1
Seems like this is quite import-heavy. I'd recommend trimming out some imports if you can to lean out your plugin. For example, it feels like underscore is pretty heavy for just checking `isUndefined` (`obj === undefined`), `isObject` (`typeof obj === 'object'`) and `isString` (`typeof obj === 'string'`).

You don't need node-fetch here either, because fetch is directly available (we're in a render process).

https://github.com/trashhalo/obsidian-extract-url/blob/c2ca1296a79fede119000d9fe8cadfef9e337c33/html-readability/index.js#L1
Wondering why you have a just-published personal fork of this library here.

https://github.com/trashhalo/obsidian-extract-url/blob/c2ca1296a79fede119000d9fe8cadfef9e337c33/html-readability/util.js#L4
You're using NodeJS library here, which means your plugin will only work on Obsidian's desktop app. If you wish to support potentially other platforms like mobile, you'll have to cut out node APIs and only use browser APIs. If not, you should specify that your plugin is desktop only in your manifest file.

trashhalo:
> Consider updating .gitignore to include your built main.js and main.js.map files, as well as removing the error.log from the repo.

It was a little unclear to me how the plugins work. Are you just pulling down a release.zip such that as long as its included in that zip its good to go?

> Don't need the async here.
> Consider promisifying this and using async await here.
> You don't need node-fetch here either, because fetch is directly available (we're in a render process).

👍 

>Don't use activeLeaf, especially not without null-checks. Here you're looking for something like:
>Just use getSelection here?

Interesting! I stole most of this code from this other community plugin. Happy to clean it up. https://github.com/FHachez/obsidian-convert-url-to-iframe/blob/master/main.ts#L24

> Seems like this is quite import-heavy. I'd recommend trimming out some imports if you can to lean out your plugin. For example, it feels like underscore is pretty heavy for just checking isUndefined (obj === undefined), isObject (typeof obj === 'object') and isString (typeof obj === 'string').
> You're using NodeJS library here, which means your plugin will only work on Obsidian's desktop app.

Ah yea. All of the code under html-readability is just this repo https://github.com/yangsibai/node-html-readability I brought it in to get it to be compatible with rollup. I'll see if I can refactor it some tonight, if not i'll at least flip the desktop only bool.

> Wondering why you have a just-published personal fork of this library here.

There is a circular dependency in the last published build of cheerio. The've since fixed in master but have not published a new build. When I tried to reference the master branch in my package.json my bundler a wrapper around rollup failed to inline the dependency. So I just threw up my hands and published under my account the latest commit to master.

Do you think including a note in the readme about this would be helpful?

lishid:
> It was a little unclear to me how the plugins work. Are you just pulling down a release.zip such that as long as its included in that zip its good to go?
There's some docs about that here: https://github.com/obsidianmd/obsidian-releases#how-community-plugins-are-pulled

> Interesting! I stole most of this code from this other community plugin. Happy to clean it up.
Yeah some older plugins are using the previous error-prone method of accessing activeLeaf directly without null checks, which could error out.

> There is a circular dependency in the last published build of cheerio. The've since fixed in master but have not published a new build. When I tried to reference the master branch in my package.json my bundler a wrapper around rollup failed to inline the dependency. So I just threw up my hands and published under my account the latest commit to master.
>
> Do you think including a note in the readme about this would be helpful?

Ah that makes sense. I'm not sure what the best solution is at this point, but it was just raising a little suspicion 😅 
Up to you to do what you think is best. Might be worth explaining it in readme just in case some users find it suspicious.

lishid:
Ah that is definitely a problem. I guess for now your only option would be to use node api, unfortunately.

trashhalo:
@lishid should be good to go now though it's been rewritten in rust using https://github.com/trashhalo/obsidian-rust-plugin as a base!

PR #198, opened 2021-03-15
Add "Pane Relief" plugin

pjeby:

## Repo URL

https://github.com/pjeby/pane-relief

## Release Checklist

- [x] I have tested this on Windows, ~~macOS, and Linux~~
  - [x] `main.js`
  - [x] `manifest.json`


pjeby:
Btw, if y'all like the way this handles history, feel free to grab the [history code](https://github.com/pjeby/pane-relief/blob/master/src/History.js) as an inspiration for Obsidian core.  It's only about 100 SLOC and a fair amount would be simplified by adding it directly to Obsidian instead of monkeypatching it in as this plugin does.  :wink:

PR #199, opened 2021-03-15
Add Dice Roller plugin

valentine195:
Adds a {dice: XdX} markdown post-processor to create re-rollable dice blocks.

## Repo URL

https://github.com/valentine195/obsidian-dice-roller

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
Repo URL: https://github.com/valentine195/obsidian-dice-roller

valentine195:
Whoops, my bad, grabbed the wrong link 😨

lishid:
Hmm would it make sense to ask the user to wrap the dice syntax into a inline code block, so that your search & replace can be easier?

That way you won't have to rely on `innerText.includes` and traversing DOM - instead you can just query selector for <code> and see if its contents match `{dice:....}`. Then you can replace the whole DOM element instead of doing an innerHTML replacement (which is always a security risk).

This has the added benefit of making it look highlighted/different in edit mode.


https://github.com/valentine195/obsidian-dice-roller/blob/69b01b5a16fd0ccad859d1771b5bd090c95f7ea9/src/main.ts#L159
Don't need to register this event - in fact registering it makes it into a memory leak.
You only need to register events that are related to global DOM elements that persists after your plugin disables. Anything that's on ephemeral/temporary DOM elements will be automatically garbage collected when the DOM element is detached.

https://github.com/valentine195/obsidian-dice-roller/blob/69b01b5a16fd0ccad859d1771b5bd090c95f7ea9/src/main.ts#L167
Probably don't need this split & join since you already did it in the initial parsing.

https://github.com/valentine195/obsidian-dice-roller/releases/tag/0.1.1
Seems like your release file is a whopping 1.17MB - Should probably add `sourcemapExcludeSources` to the rollup config.

valentine195:
> Hmm would it make sense to ask the user to wrap the dice syntax into a inline code block, so that your search & replace can be easier?
> 
> That way you won't have to rely on `innerText.includes` and traversing DOM - instead you can just query selector for `<code>` and see if its contents match `{dice:....}`. Then you can replace the whole DOM element instead of doing an innerHTML replacement (which is always a security risk).
> 
> This has the added benefit of making it look highlighted/different in edit mode.

That makes sense and definitely makes the element replacement easier/faster. Done.

> https://github.com/valentine195/obsidian-dice-roller/blob/69b01b5a16fd0ccad859d1771b5bd090c95f7ea9/src/main.ts#L159
> Don't need to register this event - in fact registering it makes it into a memory leak.
> You only need to register events that are related to global DOM elements that persists after your plugin disables. Anything that's on ephemeral/temporary DOM elements will be automatically garbage collected when the DOM element is detached.

Okay, didn't realize that. Thought it was something that needed to be done for any DOM event. Changed.

> https://github.com/valentine195/obsidian-dice-roller/blob/69b01b5a16fd0ccad859d1771b5bd090c95f7ea9/src/main.ts#L167
> Probably don't need this split & join since you already did it in the initial parsing.

Good point, done.
 
> https://github.com/valentine195/obsidian-dice-roller/releases/tag/0.1.1
> Seems like your release file is a whopping 1.17MB - Should probably add `sourcemapExcludeSources` to the rollup config.

Done.

Thanks for the feedback!

lishid:
Nice. One small issue: your docs say `dice 1d2` but you're testing for `dice:` in your code.

https://github.com/valentine195/obsidian-dice-roller/blob/f680a9e680f1a12f076a37e8b684f5cfb8019e58/src/main.ts#L74
Might want to rename your plugin class name.

https://github.com/valentine195/obsidian-dice-roller/blob/f680a9e680f1a12f076a37e8b684f5cfb8019e58/src/main.ts#L109
Actually, you can just use `dice` from the outer closure here without needing to set it on the data attribute!

valentine195:
Okay, updated. Thanks for catching these.

lishid:
https://github.com/valentine195/obsidian-dice-roller/blob/c18dd6e458058ae7eda7f6d9fe1745ccbfdce3bd/src/main.ts#L87
Hmm I don't recommend using the plain `.includes("dice")`. That seems way too likely to collide with something. Since you're running a regex anyway, why not just check if the regex matches successfully?

Also I'd recommend capping your regex with `^...$` just to make sure it doesn't pick up any accidental mention of somethingdice and now parses the whole thing as a dice block.

Also not sure why you decided to use triple backtick vs single backtick, but I guess both would work?

valentine195:
> https://github.com/valentine195/obsidian-dice-roller/blob/c18dd6e458058ae7eda7f6d9fe1745ccbfdce3bd/src/main.ts#L87
> Hmm I don't recommend using the plain `.includes("dice")`. That seems way too likely to collide with something. Since you're running a regex anyway, why not just check if the regex matches successfully?

Okay, I switched to a more explicit regex test. I also re-added the colon to the test to further prevent collisions.
 
> Also I'd recommend capping your regex with `^...$` just to make sure it doesn't pick up any accidental mention of somethingdice and now parses the whole thing as a dice block.

Done.
 
> Also not sure why you decided to use triple backtick vs single backtick, but I guess both would work?

This was just my misunderstanding on how markdown code blocks worked. I assumed they were all triple backticks, like the code block language processors. Changed to single.

Thanks!

lishid:
https://github.com/valentine195/obsidian-dice-roller/blob/c4eecc640cb65d25465e052a309d8d997180be41/src/main.ts#L91
Maybe just reuse the regex match result from the above?

Either way, we're good to merge. Will do it as soon as you fix the merge conflict.

valentine195:
Fixed the conflict. Thanks.

PR #201, opened 2021-03-15
Unique attachments

derwish-pro:

## Repo URL

## Repo URL

https://github.com/derwish-pro/obsidian-unique-attachments

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


derwish-pro:
I used the same links-handler.ts from my previous plugin (I assume that you have already checked its functions).

lishid:
I feel like it should be pretty easy to make your plugin work on all platforms, without using `path` and `crypto`.
For path, if you check for `if( file instanceof TFile)`, TypeScript will automatically infer that file is a TFile inside the scope, which means you'll be able to access file.extension and file.basename.
For crypto/md5, you can use the [SubtleCrypto](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto) API.

Again up to you if you want to make it cross-platform.

I'd also recommend specifying that your plugin requires and only works with markdown links, in relative path mode.

derwish-pro:
Ok, I will do it. Thank you.

derwish-pro:
Complete. Please, check it.

derwish-pro:
Because I am using a shared library of functions in my plugins, I also got rid of node path module in "Consistent attachments and links" plugin.

lishid:
I see that you've incorporated the md5 library - why not use a native built-in crypto library?

Here's what you could use out-of-the-box natively with no 3rd party library (it's SHA instead of MD5 since MD5 is nowadays considered deprecated).

```ts
let bytes = await window.crypto.subtle.digest('SHA-1', new Uint8Array(arrayBuffer));
let string = Array.prototype.map.call(new Uint8Array(bytes), x => ('00' + x.toString(16)).slice(-2)).join('');
```

Regarding the readme, I feel like a blockquote isn't obvious enough, but should rather be a big warning with bold text.
Most users won't read unless it's flashy. I think without an obvious warning, you will end up with github issues about "this plugin doesn't do anything" because the user didn't read the part where markdown links are required.

derwish-pro:
I used MD5 because I was transferring my notes to obsidian from Evernote and OneNote. I had to convert them to markdown via Joplin. This is the only way I have found. As a result, Joplin renamed all images from my notes using the MD5 function. I decided that this is a great solution and added the same functionality to Obisidan. Using MD5 instead of SHA at the moment will allow everyone to more painlessly switch to Obsidian from Joplin, and possibly from Evernote and OneNote, as in my case.

derwish-pro:
I also noticed that MD5 has less line length than SHA256 or even SHA1, which is better for filenames. I think that for the moment I will leave MD5. The issue of switching to SHA can be put to a vote by the community if the plugin gains popularity. Or maybe later I'll make it optional, but then I'll have to leave the third-party library.

lishid:
Ok fair enough. If you want to keep things like that then it's up to you.

> I also noticed that MD5 has less line length than SHA256 or even SHA1

You could trim it to a specific length if short name is desired.

LGTM otherwise. Will merge as soon as the conflict is fixed.

derwish-pro:
complete

PR #202, opened 2021-03-16
Add hotkey-helper plugin

pjeby:

## Repo URL
https://github.com/pjeby/hotkey-helper

## Release Checklist
- [x] I have tested this on Windows ~~macOS, and Linux~~
  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1013,5 +1013,12 @@
         "description": "Open or reveal the current note in DEVONthink.",
         "author": "ryanjamurphy",
         "repo": "ryanjamurphy/DEVONlink-obsidian"
+    },
+    {
+        "id": "hotkey-helper",
+        "name": "Hotkey Helper",
+        "author": "PJ Eby",
+        "description": "Easily see and access any plugin's settings or hotkey assignments (and conflicts) from the Community Plugins tab",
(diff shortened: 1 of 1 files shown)


pjeby:
Since opening this, I've added more functionality (options buttons) and so rewritten the description in the commit here.  I'm not 100% sure it should still be called hotkey helper since it does more than that now, but oh well.  :grinning:

lishid:
This will be broken by the upcoming insider release where addPluginTab and removePluginTab will be removed. Should we wait for that to land first?

pjeby:
What are they being replaced with?

I could certainly hook `Plugin.prototype.addSettingTab` and `.unload` instead; all those methods are hooked for is to detect the addition and removal of the setting tabs.

Really, all the monkeypatching in this plugin is to substitute for events that don't currently exist, then adding event triggers for those things (commands or setting tabs being added or removed) would suffice for an alternate implementation.

(Btw, If methods I'm hooking are removed, all that happens is they won't get called, so it won't so much "break" the plugin as it will mean the options icons won't come and go immediately upon toggling a plugin on or off; it'll end up waiting until the plugin page is re-rendered.)

PR #203, opened 2021-03-17
Add imdone-obsidian-plugin

piascikj:
# [x ] I am submitting a new Community Plugin

## Release Checklist

- [x] I have tested this on:
  - [x]  Windows
  - [x]  macOS
  - [/ ] Linux  _(working on it. okay to release without it)_
  - [x] `main.js`
  - [x] `manifest.json`


piascikj:
Tested on mac.  Still working on windows testing.

xnhp:
@piascikj Do you need help with windows testing?

piascikj:
> @piascikj Do you need help with windows testing?

@xnhp Will let you know.  I think I know what the issue might be.  It's definitely broken at the moment.  Thanks for the offer!

liamcain:
Yay!! 🎉  Very excited for this!

Just did a code review pass. Let me know if you have any questions!

---

## Potential issues

- https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L27 editor can be null here.
- https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L29 You're not specifying a radix for `parseInt`. This should be `parseInt(params.line, 10)`.
- What's does an imdone card link look like? My regex is a bit rusty so I'm trying to see understand what's getting matched by `hashRegex`.

## Code Style/Formatting

- https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L1 looks like some lines end with semicolons while others don't. Consider adding semicolons to avoid any nasty line-continuation bugs.
- https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L42 Use `const` instead of `var` for these
- https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L58 totally stylistic but these helper functions could live outside the `ImdonePlugin` class. That might make them easier to test.
- https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L9-L10 since these don't get reassigned, consider making them `const`s and pulling them out of the `ImdonePlugin` class.
- https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L104 You could make a Typescript `type` for all the values that `getImdoneURL` supports as a type `getImdoneURL(text: string, hash:string , type:'MARKDOWN' | 'HASHTAG') { ... }`.
- https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L52 Consider moving this `if` check into the onLoad so it doesn't run for each markdown block

## API Usage

- https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L3 since you're only using `fs` for checking the existence, you can instead use [`adapter.exists()`](https://github.com/obsidianmd/obsidian-api/blob/81ab85ade4552c9116c1e10d009127e62019c923/obsidian.d.ts#L519) from the API.
- https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L43 The API provides a more idiomatic way to do this with `getActiveViewOfType(MarkdownView)` - [full snippet here](https://liamca.in/Obsidian/API+FAQ/editor/get+the+active+CodeMirror+instance)
- https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L109 You can instead us [`getFullPath`](https://github.com/obsidianmd/obsidian-api/blob/81ab85ade4552c9116c1e10d009127e62019c923/obsidian.d.ts#L853) from the Obsidian API for this.

piascikj:
Thanks for the review @liamcain! You had some great advice! Now that I've made these changes, the last thing to do is create a new tag 1.0.1 and make the version change here.

> ## Potential issues
> * https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L27 editor can be null here.

**Fixed**

> * https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L29 You're not specifying a radix for `parseInt`. This should be `parseInt(params.line, 10)`.

**Fixed**

> * What's does an imdone card link look like? My regex is a bit rusty so I'm trying to see understand what's getting matched by `hashRegex`.

**[Card Syntax - Imdone docs](https://imdone.io/docs/#/cards?id=card-syntax)**
 
> ## Code Style/Formatting
> * https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L1 looks like some lines end with semicolons while others don't. Consider adding semicolons to avoid any nasty line-continuation bugs.

**Fixed**

> * https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L42 Use `const` instead of `var` for these

**Fixed**

> * https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L58 totally stylistic but these helper functions could live outside the `ImdonePlugin` class. That might make them easier to test.

**I'll get this done in the next release.**

> * https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L9-L10 since these don't get reassigned, consider making them `const`s and pulling them out of the `ImdonePlugin` class.

**Fixed**

> * https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L104 You could make a Typescript `type` for all the values that `getImdoneURL` supports as a type `getImdoneURL(text: string, hash:string , type:'MARKDOWN' | 'HASHTAG') { ... }`.

**Fixed**

> * https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L52 Consider moving this `if` check into the onLoad so it doesn't run for each markdown block

**This has to run for each markdown doc since we can't be sure how the user has nested their imdone projects.  They may have more than one in a vault and some docs might not be in an imdone project.**

> ## API Usage
> * https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L3 since you're only using `fs` for checking the existence, you can instead use [`adapter.exists()`](https://github.com/obsidianmd/obsidian-api/blob/81ab85ade4552c9116c1e10d009127e62019c923/obsidian.d.ts#L519) from the API.

**I tried this with the full path, but it doesn't seem to work.**

> * https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L43 The API provides a more idiomatic way to do this with `getActiveViewOfType(MarkdownView)` - [full snippet here](https://liamca.in/Obsidian/API+FAQ/editor/get+the+active+CodeMirror+instance)

**Fixed**

> * https://github.com/imdone/imdone-obsidian-plugin/blob/c36ce24762830ede85c13cf0a658133da3cc6d16/main.ts#L109 You can instead us [`getFullPath`](https://github.com/obsidianmd/obsidian-api/blob/81ab85ade4552c9116c1e10d009127e62019c923/obsidian.d.ts#L853) from the Obsidian API for this.

**Fixed**

liamcain:
Thanks for making those fixes, code looks good to me!

---

> Card Syntax - Imdone docs

Okay, I understand now. I'm assuming that `TODO` and `DOING` are the names of the configurable swimlanes (and could be anything)? My main concern was matching on false positives but maybe that's not a huge issue.

> This has to run for each markdown doc since we can't be sure how the user has nested their imdone projects. They may have more than one in a vault and some docs might not be in an imdone project.

Good call, hopefully this doesn't cause too much of a slowdown for larger files.

community-plugins.json
```diff
@@ -1013,5 +1013,12 @@
         "description": "Open or reveal the current note in DEVONthink.",
         "author": "ryanjamurphy",
         "repo": "ryanjamurphy/DEVONlink-obsidian"
+    },
+    {
+        "id": "imdone-obsidian-plugin",
+        "name": "Open cards in imdone from obsidian.",
```
liamcain, line 1019: Didn't want to just name it "imdone"? 🤔

liamcain (approved):

community-plugins.json
piascikj: Good point, but I didn't want confusion with the product and already have a repo named `imdone/imdone`.

lishid:
https://github.com/imdone/imdone-obsidian-plugin/blob/a8f96efbc9df13e3ea0fe8199ae9d078dbfe933f/main.ts#L27
openLinkText should have the file path in its first argument for proper link resolution - second parameter is for sourcePath which is from which file this link comes from, and is expected to be a fully normalized standard path instead of being sanitized and normalized like the first one. Also maybe use async/await here?

https://github.com/imdone/imdone-obsidian-plugin/blob/a8f96efbc9df13e3ea0fe8199ae9d078dbfe933f/main.ts#L28
Might be a good idea to make sure the current view has the same file correctly open before proceeding? Probably doesn't make too much of a difference though.

https://github.com/imdone/imdone-obsidian-plugin/blob/a8f96efbc9df13e3ea0fe8199ae9d078dbfe933f/main.ts#L3
Could just use fs.promises here instead of having to hand-wrap stat.

https://github.com/imdone/imdone-obsidian-plugin/blob/a8f96efbc9df13e3ea0fe8199ae9d078dbfe933f/main.ts#L51
This is quite inefficient because you're issuing a bunch of async fs calls for every single markdown block that's rendered in preview. Maybe there's a better way to mark which folder to process? Also the markdown post processor doesn't necessarily run for the current active focused file, it might be running for a different one because it's asynchronous. You'll want to use the second ctx parameter to find the `sourcePath` of the file being rendered. (It's not yet exposed so you'll have to cast to any for now)

https://github.com/imdone/imdone-obsidian-plugin/blob/a8f96efbc9df13e3ea0fe8199ae9d078dbfe933f/main.ts#L105
Might need escaping for path here?

piascikj:
@lishid Thanks for taking a look at the code!  I'll go through your comments this evening and follow up tomorrow.

piascikj:
Hi @lishid.  I think I addressed the major concerns, but one question remains.  See below.


> https://github.com/imdone/imdone-obsidian-plugin/blob/a8f96efbc9df13e3ea0fe8199ae9d078dbfe933f/main.ts#L27
> openLinkText should have the file path in its first argument for proper link resolution - second parameter is for sourcePath which is from which file this link comes from, and is expected to be a fully normalized standard path instead of being sanitized and normalized like the first one. Also maybe use async/await here?

**What I'm trying to accomplish here is to open the file in edit mode.  This approach seems buggy.  Is there another way to accomplish what I'm trying to do? Maybe open a new leaf?  Can you give me some pointers?**

> https://github.com/imdone/imdone-obsidian-plugin/blob/a8f96efbc9df13e3ea0fe8199ae9d078dbfe933f/main.ts#L28
> Might be a good idea to make sure the current view has the same file correctly open before proceeding? Probably doesn't make too much of a difference though.

**Related to above.**

> https://github.com/imdone/imdone-obsidian-plugin/blob/a8f96efbc9df13e3ea0fe8199ae9d078dbfe933f/main.ts#L3
> Could just use fs.promises here instead of having to hand-wrap stat.

** Done **

> https://github.com/imdone/imdone-obsidian-plugin/blob/a8f96efbc9df13e3ea0fe8199ae9d078dbfe933f/main.ts#L51
> This is quite inefficient because you're issuing a bunch of async fs calls for every single markdown block that's rendered in preview. Maybe there's a better way to mark which folder to process? Also the markdown post processor doesn't necessarily run for the current active focused file, it might be running for a different one because it's asynchronous. You'll want to use the second ctx parameter to find the `sourcePath` of the file being rendered. (It's not yet exposed so you'll have to cast to any for now)

**Fixed**

> https://github.com/imdone/imdone-obsidian-plugin/blob/a8f96efbc9df13e3ea0fe8199ae9d078dbfe933f/main.ts#L105
> Might need escaping for path here?

**Not needed**

lishid:
> What I'm trying to accomplish here is to open the file in edit mode. This approach seems buggy. Is there another way to accomplish what I'm trying to do? Maybe open a new leaf? Can you give me some pointers?

This should do it:
```ts
this.workspace.openLinkText(file, '', false, {state: {mode:'source'}})
```


> Not needed

https://github.com/imdone/imdone-obsidian-plugin/blob/6a82622a5dd4093a4ddf1e4930b2fdb4b12b9fa1/main.ts#L115
Are you sure?

Because on MacOS files can have `?` as part of their filename, and if a file was named `my?file.md` you'll end up with `imdone://my?file.md?text=....`

piascikj:
Thanks for the tip @lishid!  All items have been addressed and it works as expected.

lishid:
LGTM. Once you fix the merge conflict I'll have it merged!

piascikj:
On it

piascikj:
@lishid Merge conflict resolved.

PR #204, opened 2021-03-18
Obsidian Spotlight Plugin

Darakah:
# [ x] I am submitting a new Community Plugin

## Repo URL

https://github.com/Darakah/obsidian-spotlight

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


lishid:
https://github.com/Darakah/obsidian-spotlight/blob/09a979df542982ddea43f6ade42ed1da9ceafaf1/src/block.ts#L17
Typo "chooseRnadomNote" haha.

https://github.com/Darakah/obsidian-spotlight/blob/09a979df542982ddea43f6ade42ed1da9ceafaf1/src/block.ts#L27
Can probably use cachedRead here since this is just for display.

https://github.com/Darakah/obsidian-spotlight/blob/09a979df542982ddea43f6ade42ed1da9ceafaf1/src/block.ts#L35
This only exists if the "block" is identified by an ID.

https://github.com/Darakah/obsidian-spotlight/blob/09a979df542982ddea43f6ade42ed1da9ceafaf1/src/utils.ts#L57
Consider adding some line breaks in here.

https://github.com/Darakah/obsidian-spotlight/blob/09a979df542982ddea43f6ade42ed1da9ceafaf1/src/utils.ts#L84
Add a variable here for your block so you aren't repeating the same code twice.

Darakah:
@lishid , thanks for the useful remarks

1. Corrected ... thanks! 
2. Changed to cached Read
3. Not sure about this, 
You will reach this block only if `block = true` and `chooseRandomNote` returned a non-null note. and if block is true, one of the choose conditions is `metadataCache.getFileCache(file).blocks`. 

https://github.com/Darakah/obsidian-spotlight/blob/09a979df542982ddea43f6ade42ed1da9ceafaf1/src/utils.ts#L57

So, at the time this statement is reached, it will have a block no?

4. Added break lines
5. Saved as a variable

Darakah:
@lishid As suggested, changed the way the input arguments are passed / read. Let me know if this is ok. (You can find a detailed explanation in the updated readMe).

lishid:
https://github.com/Darakah/obsidian-spotlight/blob/bd9840d0e8b808d26c0bd8ed8e1b65105a1062be/src/block.ts#L26
This could be null, but more importantly, this could be a different file than the one being rendered in the block.

You'll want to use this one that's passed in through the post processor: https://github.com/obsidianmd/obsidian-api/blob/ca3cba04c46c8e7a1bbe69ce73b7d878bc1b6712/obsidian.d.ts#L1370

https://github.com/Darakah/obsidian-spotlight/blob/bd9840d0e8b808d26c0bd8ed8e1b65105a1062be/src/block.ts#L5
Doesn't seem used.

https://github.com/Darakah/obsidian-spotlight/blob/bd9840d0e8b808d26c0bd8ed8e1b65105a1062be/src/utils.ts#L23
Don't parse tags yourself, use this one that takes care of normalizing frontmatter tags that could have been a flat string instead of an array (like `"tag1, tag2"`): https://github.com/obsidianmd/obsidian-api/blob/ca3cba04c46c8e7a1bbe69ce73b7d878bc1b6712/obsidian.d.ts#L1096

---

General organization:

- You don't have to change this if you prefer it that way, but I don't really see the point of separating `types.ts` and `constants.ts`. Might as well just roll them all into `settings.ts`.
- Please add semicolons to your code base. Use an auto-formatter to add it for you automatically.

Darakah:
@lishid thanks for the helpful review as always.

1. Using `sourcePath` as suggested https://github.com/Darakah/obsidian-spotlight/blob/e88813c15ae9f24e40b701034a3b092e05ddaf63/src/block.ts#L25
2. Removed 
3. Using `getAllTags` as suggested https://github.com/Darakah/obsidian-spotlight/blob/e88813c15ae9f24e40b701034a3b092e05ddaf63/src/utils.ts#L17

----

- I prefer it this way if it is not a bother to you
- Added `;` to all statements using auto-formatter

PR #205, opened 2021-03-18
Obsidian Word Cloud Plugin

Darakah:

## Repo URL

https://github.com/Darakah/obsidian-word-cloud

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1013,5 +1013,13 @@
         "description": "Open or reveal the current note in DEVONthink.",
         "author": "ryanjamurphy",
         "repo": "ryanjamurphy/DEVONlink-obsidian"
+    },
+    {
+        "id": "obsidian-word-cloud",
+        "name": "Word Cloud",
+        "description": "Create word clouds of note tags or text of vault, specified project or notes.",
+        "author": "darakah",
+        "repo": "Darakah/obsidian-word-cloud",
+        "branch": "main"
     }
 ]


lishid:
https://github.com/Darakah/obsidian-word-cloud/blob/c5a00ce9f66f3d7a6046d3a988c86a705f166865/src/block.ts#L8

Actually, you can infer metadataCache and vault from app, so probably shouldn't be passing those in (same for your other plugins).

Consider also doing your argument parsing outside. Passing in an `args: string[]` is quite confusing - where'd you get it? what defines its format?

This is also not very good: `parseInt(args[1]) ? parseInt(args[1]) : settings.top`
- What does parseInt return on an empty string? Is it null, undefined, NaN?
- You're parsing the integer again once you confirmed its truthiness, extra work that's pretty unnecessary
Recommend adding a helper function to parse the integer or use the default value.

https://github.com/Darakah/obsidian-word-cloud/blob/c5a00ce9f66f3d7a6046d3a988c86a705f166865/src/utils.ts#L23
Again, same here. What is `args`'s format? Why is [2] filepath and [3] filename?
Remember, doing things like this also makes your life hard down-the-line when you want to change this format. If you need to insert a new field in between, you gotta go to every function that parses the args array and change the indices. Parse it into an keyword object first, and reuse it from there.

https://github.com/Darakah/obsidian-word-cloud/blob/c5a00ce9f66f3d7a6046d3a988c86a705f166865/src/utils.ts#L37
There's a function for doing all of this concatenation work which also normalizes frontmatter tags (which could be a string instead of a string[] by the way). The exposed function is called `getAllTags(cache)`.

https://github.com/Darakah/obsidian-word-cloud/blob/c5a00ce9f66f3d7a6046d3a988c86a705f166865/src/utils.ts#L123
This is a really bad idea. You'll end up creating a monster of a giant string (very memory intensive), and also for each step, you're creating a gigantic string containing the file + all the files before. Keep the file contents separate in an array of strings.

https://github.com/Darakah/obsidian-word-cloud/blob/c5a00ce9f66f3d7a6046d3a988c86a705f166865/src/utils.ts#L144
Your ignore list isn't regex-escaped, so things will not work properly or error out when you do this. For example, "ignore(this" will probably crash because it's a bad regex.

https://github.com/Darakah/obsidian-word-cloud/blob/c5a00ce9f66f3d7a6046d3a988c86a705f166865/src/types.ts#L12
Interfaces should probably be PascalCase instead of camelCase to avoid confusion when you use it.

https://github.com/Darakah/obsidian-word-cloud/blob/c5a00ce9f66f3d7a6046d3a988c86a705f166865/src/svelte/WordCloud.svelte#L14
You're not using anything from svelte here, you can just shove this into a regular javascript function and cut out svelte from the project altogether.

lishid:
Also Anychart is not freely licensed. Do you have a license to distribute it?
https://www.anychart.com/support/pages/faq/

Darakah:
@lishid Thanks for poingint that license thing out. No, i didn't know about this at all. Gonna take these down before I get myself into trouble I have no idea about. Thanks for the remarks.

PR #210, opened 2021-03-20
Update community-plugins.json

Clemens-E:

## Repo URL

https://github.com/Clemens-E/obsidian-languagetool-plugin

## Release Checklist

- [ ] I have tested this on macOS
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1072,5 +1072,12 @@
           "description": "Inline dice rolling for Obsidian.md",
           "author": "Jeremy Valentine",
           "repo": "valentine195/obsidian-dice-roller"
+    },
+    {
+        "id": "obsidian-languagetool-plugin",
+        "name": "LanguageTool Integration",
+        "description": "advanced spell/grammar checks with the help of language-tool",
+        "author": "Clemens Ertle",
+        "repo": "Clemens-E/obsidian-languagetool-plugin"
     }
 ]


lishid:
Cool plugin! Couple of suggestions:

https://github.com/Clemens-E/obsidian-languagetool-plugin/blob/6dffd3609a1046112a6a5d8b96739759ef9cbe46/src/index.ts#L77
Maybe also use await here instead of mixing promise callbacks?

https://github.com/Clemens-E/obsidian-languagetool-plugin/blob/6dffd3609a1046112a6a5d8b96739759ef9cbe46/src/index.ts#L126
Seems like you'll be attaching multiple copies of this event handler when you run detection multiple times.

Clemens-E:
Good catch with the event.
I fixed both issues, can you check again?

lishid:
I probably missed this in the first pass, but:

https://github.com/Clemens-E/obsidian-languagetool-plugin/blob/0ab9fdab5d8ba9b222689cd22179635dafeb7972/src/index.ts#L61
view could be empty here if the user is focused on a non-markdown pane. Probably a good idea to check for truthiness.

https://github.com/Clemens-E/obsidian-languagetool-plugin/blob/0ab9fdab5d8ba9b222689cd22179635dafeb7972/src/index.ts#L62
Actually, you're registering a click handler for the entire app - you should probably check that the click event's target is within the editor's bounds. I'd recommend using [`node.contains(otherNode)`](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains) to check that.

https://github.com/Clemens-E/obsidian-languagetool-plugin/blob/0ab9fdab5d8ba9b222689cd22179635dafeb7972/src/index.ts#L88
Same as above.

https://github.com/Clemens-E/obsidian-languagetool-plugin/blob/0ab9fdab5d8ba9b222689cd22179635dafeb7972/src/index.ts#L101
Same as above, view could be null.

Clemens-E:
updated again, lmk if there is anything else

lishid:
LGTM, will merge as soon as the merge conflict is resolved.

PR #211, opened 2021-03-21
Obsidian Commits Plugin Updated

Darakah:

## Repo URL

https://github.com/Darakah/obsidian-commits

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


lishid:
https://github.com/Darakah/obsidian-commits/blob/25c7fd77f6a3c8587f1c824a899ea50806afc47d/src/main.ts#L39
Why not just bind it on spot when it's used?

https://github.com/Darakah/obsidian-commits/blob/25c7fd77f6a3c8587f1c824a899ea50806afc47d/src/main.ts#L70
Use this instead:
```ts
if (file instanceof TFile && file.extension === 'md') {
  ...
}
```

https://github.com/Darakah/obsidian-commits/blob/25c7fd77f6a3c8587f1c824a899ea50806afc47d/src/utils.ts#L106
Use a proper type here instead of `number[]`.

https://github.com/Darakah/obsidian-commits/blob/25c7fd77f6a3c8587f1c824a899ea50806afc47d/src/utils.ts#L117
Use `for (let file of fileList)`.
Seriously, you've copy pasted `fileList[file]` 26 times!

https://github.com/Darakah/obsidian-commits/blob/25c7fd77f6a3c8587f1c824a899ea50806afc47d/src/utils.ts#L226
Same as above.

https://github.com/Darakah/obsidian-commits/blob/25c7fd77f6a3c8587f1c824a899ea50806afc47d/src/utils.ts#L163
Really this is unnecessary. If you have a path then you should use `vault.getAbstractFileByPath(path)`.

https://github.com/Darakah/obsidian-commits/blob/25c7fd77f6a3c8587f1c824a899ea50806afc47d/src/utils.ts#L186
Use `getAllTags`.

https://github.com/Darakah/obsidian-commits/blob/25c7fd77f6a3c8587f1c824a899ea50806afc47d/src/utils.ts#L309
Unsanitized/unescaped regex. This will error out if the file name contains special regex characters, like `(`.

https://github.com/Darakah/obsidian-commits/blob/25c7fd77f6a3c8587f1c824a899ea50806afc47d/src/utils.ts#L91
Same as above.

Darakah:
Hey @lishid , Thanks as always for the detailed review and suggestions.

- As suggested, binding it when used https://github.com/Darakah/obsidian-commits/blob/f919d4c6af364f47dcb5fdec292c3a8ddeacf98b/src/main.ts#L39
- As suggested, using the `istanceof` to access the TFile. https://github.com/Darakah/obsidian-commits/blob/f919d4c6af364f47dcb5fdec292c3a8ddeacf98b/src/main.ts#L67
- Using a defined type https://github.com/Darakah/obsidian-commits/blob/f919d4c6af364f47dcb5fdec292c3a8ddeacf98b/src/utils.ts#L30
- Removed
- As suggested using `getAllTags` https://github.com/Darakah/obsidian-commits/blob/f919d4c6af364f47dcb5fdec292c3a8ddeacf98b/src/utils.ts#L177
- Escaping project name before building the regex https://github.com/Darakah/obsidian-commits/blob/f919d4c6af364f47dcb5fdec292c3a8ddeacf98b/src/utils.ts#L93

Thanks a lot for your help Licat, greatly appreciated!

lishid:
https://github.com/Darakah/obsidian-commits/blob/f919d4c6af364f47dcb5fdec292c3a8ddeacf98b/src/utils.ts#L30

I meant more like a defined type that explains what the data structure is/does. `[number, number, number]` is better then `number[]` because you know there are three numbers instead of any amount of numbers, but it still doesn't say what the first/second/third number is.

Ideally you'll want an object/interface that says:
```
interface FileStats {
  fileSize: number
  numTags: number;
  numLinks: number;
}
```

Doing this ensures that someone looking at your code a year from now (could even be yourself) won't be confused what `stats[1]` is and what `stats[2]` is.

https://github.com/Darakah/obsidian-commits/blob/f919d4c6af364f47dcb5fdec292c3a8ddeacf98b/src/utils.ts#L130
Just a recommendation for more things to put into variables, try `let filePath = file.path` and `let fileName = file.name`. Also you already have `fileSize` as a variable, just continue to use that instead of `file.stat.size`.

PR #214, opened 2021-03-22
text-transformation-plugin

artur02:

## Repo URL

# [ X ] I am submitting a new Community Plugin

## Repo URL

https://github.com/artur02/obsidian-text

## Release Checklist

  - [X] `main.js`
  - [X] `manifest.json`
  - `styles.css` _(optional)_

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -993,5 +993,13 @@
         "author": "Jeremy Valentine",
         "repo": "valentine195/obsidian-leaflet-plugin",
         "branch": "master"
-  }
+  },
+    {
+        "id": "text-transformation-plugin",
+        "name": "Text Transformation",
+        "description": "This is a simple text transfromation plugin that supports UPPERCASE, lowercase, Title Case, Sentence case.",
+        "author": "Artur Herczeg",
+        "repo": "artur02/obsidian-text",
+        "branch": "master"
+    }
 ]


lishid:
Just a small organization perspective... I feel like the whole plugin could be rolled into a single file without so many functions that are one liner proxies for other things, which are all used just once.

- https://github.com/artur02/obsidian-text/blob/master/src/editor.ts This file is really just one function with 4 helper functions that are used only once. I'd suggest roll them into a single function (And also considering that the code implements the getActiveViewOfType check twice in different ways, that could be a bit of a problem)
- https://github.com/artur02/obsidian-text/blob/master/src/transform.ts Probably easiest to just move the code to where it's being used.
- https://github.com/artur02/obsidian-text/blob/master/test/case.spec.ts I'm not sure why you need tests for the `voca` library. [Seems like it's got a heavy test suite already](https://github.com/panzerdp/voca/tree/master/test)

Oh and consider renaming `MyPlugin` to something else 👍 

Otherwise, great plugin!

ericaxu:
Hi! Did you have time to address the comments @artur02 ? Thank you so much!

ericaxu:
Hi there, to keep things tidy, we're going to temporarily close plugin PRs after 10 days of no activity. Feel free to tell us to re-open it when you're ready, or submit a new PR. Thanks for your understanding!

PR #215, opened 2021-03-23
Add obsidian-tracker plugin

pyrochlore:

## Repo URL

https://github.com/pyrochlore/obsidian-tracker

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


pyrochlore:
Update the plugin to version 1.0.1
- Remove dependencies to Node.js modules
- Add example markdown files

lishid:
Seems like you're bundling moment.js in your plugin, which is making it quite large. We have moment.js bundled as well as `window.moment`, you should probably use that one instead of importing your own. 

Here's an example of what you can do: https://github.com/liamcain/obsidian-calendar-plugin/blob/b795a601ed6ed59d0970d0ed1883e207833eeb23/src/main.ts#L16

Also consider auto-formatting your code to normalize spaces vs tabs.

https://github.com/pyrochlore/obsidian-tracker/blob/d9b8795c34f185cd0e566a29b6f3ed1f2f8de88f/src/main.ts#L27
This doesn't seem used.

https://github.com/pyrochlore/obsidian-tracker/blob/d9b8795c34f185cd0e566a29b6f3ed1f2f8de88f/src/main.ts#L35
Here, you should use `this.registerMarkdownCodeBlockProcessor` instead, which will handle the code block automatically for you, as well as unregistering the post processor when your plugin unloads.

https://github.com/pyrochlore/obsidian-tracker/blob/d9b8795c34f185cd0e566a29b6f3ed1f2f8de88f/src/main.ts#L593
This doesn't seem used.

https://github.com/pyrochlore/obsidian-tracker/blob/d9b8795c34f185cd0e566a29b6f3ed1f2f8de88f/src/main.ts#L15
I'd recommend making these instance properties, and making your post processor also an instance function bound to the instance `registerXXX(this.callback.bind(this)`.

pyrochlore:
Thanks for the review.
Fixed in the master branch.
Will release after approval.

ericaxu:
@lishid Please take a look when you have time, thanks!

PR #216, opened 2021-03-23
Add Admonition Plugin

valentine195:
This plugin is based on this request in the forum: https://forum.obsidian.md/t/admonition-plugin/176/10

## Repo URL

https://github.com/valentine195/obsidian-admonition

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/valentine195/obsidian-admonition/blob/10f611c28cb3a9f6ca0a0f10c15f2da6e3eccf3e/src/main.ts#L80

Please use the API `Plugin.registerMarkdownCodeBlockProcessor` which takes care of handling the code block stuff for you, that should save you quite some work.

https://github.com/valentine195/obsidian-admonition/blob/10f611c28cb3a9f6ca0a0f10c15f2da6e3eccf3e/src/main.ts#L207
Don't use this - it might not always work. Instead, use `ctx.sourcePath`.

https://github.com/valentine195/obsidian-admonition/blob/10f611c28cb3a9f6ca0a0f10c15f2da6e3eccf3e/src/main.ts#L233
That's a rather obscure way of doing things. Perhaps just use a regular if/else and two blocks of code instead?

valentine195:
Okay, all updated.

lishid:
https://github.com/valentine195/obsidian-admonition/blob/997f5ca64c6a320e1fdb4c9754426797ba7823a2/src/main.ts#L76
Object.fromEntries is only available since mid 2019 - your plugin might not work on older mobile devices if you decide to use this. Might need a polyfill.

https://github.com/valentine195/obsidian-admonition/blob/997f5ca64c6a320e1fdb4c9754426797ba7823a2/src/main.ts#L117
Don't recommend using a notice here from a post processor - instead, render the error into your output DOM for the best experience.

https://github.com/valentine195/obsidian-admonition/blob/997f5ca64c6a320e1fdb4c9754426797ba7823a2/src/main.ts#L148
Maybe you're using an older version of obsidian's typescript definitions, but this field should be defined in the latest one.

>https://github.com/valentine195/obsidian-admonition/blob/10f611c28cb3a9f6ca0a0f10c15f2da6e3eccf3e/src/main.ts#L233
That's a rather obscure way of doing things. Perhaps just use a regular if/else and two blocks of code instead?

This didn't seem addressed, I can explain if my comment wasn't clear.

valentine195:
> https://github.com/valentine195/obsidian-admonition/blob/997f5ca64c6a320e1fdb4c9754426797ba7823a2/src/main.ts#L76
> Object.fromEntries is only available since mid 2019 - your plugin might not work on older mobile devices if you decide to use this. Might need a polyfill.

Okay, added the [es-shim polyfill](https://github.com/es-shims/Object.fromEntries#readme):
https://github.com/valentine195/obsidian-admonition/blob/0fbaffc7e7c08ff4e64d5971b98f335a1594a83e/src/main.ts#L10

> https://github.com/valentine195/obsidian-admonition/blob/997f5ca64c6a320e1fdb4c9754426797ba7823a2/src/main.ts#L117
> Don't recommend using a notice here from a post processor - instead, render the error into your output DOM for the best experience.

Okay, done:
https://github.com/valentine195/obsidian-admonition/blob/0fbaffc7e7c08ff4e64d5971b98f335a1594a83e/src/main.ts#L155

> https://github.com/valentine195/obsidian-admonition/blob/997f5ca64c6a320e1fdb4c9754426797ba7823a2/src/main.ts#L148
> Maybe you're using an older version of obsidian's typescript definitions, but this field should be defined in the latest one.

Correct, I was just using a later release. `npm update` fixed it.

> > https://github.com/valentine195/obsidian-admonition/blob/10f611c28cb3a9f6ca0a0f10c15f2da6e3eccf3e/src/main.ts#L233
> > That's a rather obscure way of doing things. Perhaps just use a regular if/else and two blocks of code instead?
> 
> This didn't seem addressed, I can explain if my comment wasn't clear.

Yeah, I guess I'm not sure what you were referring to. Thought it was just the fact that I was initializing a dynamic object key. The newest version is doing an explicit if/else statement to set the ["open" property](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#attributes).

https://github.com/valentine195/obsidian-admonition/blob/997f5ca64c6a320e1fdb4c9754426797ba7823a2/src/main.ts#L170

```javascript
if (collapse === "open") {
    attrs = { open: "open" };
} else {
    attrs = {};
}
```

lishid:
Looking great!

https://github.com/valentine195/obsidian-admonition/blob/0fbaffc7e7c08ff4e64d5971b98f335a1594a83e/src/main.ts#L168
One last thing I might suggest is to just use `color: var(--text-error)` here instead. The way you're getting the color requires re-flowing the entire DOM (to compute the style) which is bad for performance. `--text-error` is defined in our core theme anyway, so it should always be defined.

PR #218, opened 2021-03-24
add koncham-undefault-hotkeys plugin

manogna4:

## Repo URL

https://github.com/manogna4/obsidian-koncham-undefault-hotkeys

## Release Checklist

- [ ] I have tested this on
  - [x] Windows
  - [ ] macOS
  - [ ] Linux
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1072,5 +1072,15 @@
           "description": "Inline dice rolling for Obsidian.md",
           "author": "Jeremy Valentine",
           "repo": "valentine195/obsidian-dice-roller"
+    },
+    {
+
+          "id": "obsidian-koncham-undefault-hotkeys",
+          "name": "koncham undefault hotkeys",
+          "description": "unset some default CodeMirror hotkeys",
+          "author": "mano",
+          "repo": "manogna4/obsidian-koncham-undefault-hotkeys",
+          "branch": "main"
+            
     }
 ]


pjeby:
Just a heads up, but the functionality of this plugin can be more safely accomplished by [delegating CodeMirror keyboard events to Obsidian's Keymap](https://github.com/obsidianmd/obsidian-releases/pull/212#issuecomment-805117331), so that events for keys Obsidian recognizes are captured and not seen by CodeMirror.  This ensures that any hotkeys *not* overridden by Obsidian will still perform their default CodeMirror functionality.

(When something like that is incorporated into Obsidian itself, there also won't be any need to do it in a plugin, but in the meantime, it's okay to have plugins that provide the fix as a workaround.)

manogna4:
> Just a heads up, but the functionality of this plugin can be more safely accomplished by [delegating CodeMirror keyboard events to Obsidian's Keymap](https://github.com/obsidianmd/obsidian-releases/pull/212#issuecomment-805117331), so that events for keys Obsidian recognizes are captured and not seen by CodeMirror. This ensures that any hotkeys _not_ overridden by Obsidian will still perform their default CodeMirror functionality.

I see your point. But my intention is to unset some CodeMirror default hotkeys even if obsidian does not specifically capture them. For example, CodeMirror uses `Ctrl+T` to transpose characters on either side of the cursor. IMO, this is a rare hotkey. I did not know of it's existence till I read the documentation (and I am a keen hotkey user).

If a user hits `Ctrl+T` by mistake (when it's unassigned in obsidian) they would see an unexpected behavior. Searching in obsidian hotkeys to see what it does would yield no result. I would like cater to such use cases too.

lishid:
Sorry for the late response - we will actually be removing all of these shortcuts from the default in our next insider release.

After some investigation, it seems that all 6 shortcuts are in fact non-standard, so we have decided to remove them so users have more freedom of using these for hotkeys.

manogna4:
> Sorry for the late response

no worries. I am sure you have more pressing matters

>we will actually be removing all of these shortcuts from the default in our next insider release.
> After some investigation, it seems that all 6 shortcuts are in fact non-standard, so we have decided to remove them so users have more freedom of using these for hotkeys.

do you have a list of shortcuts you want to remove?
There are more unintuitive shortcuts in the CodeMirror defaults (`ctrl+T` for example) - that's my concern. I just started with these 6 to see how they work.

lishid:
In the upcoming release, I will remove all 6 of the shortcuts defined in your list, plus the Alt+D/F/B hotkeys on Mac that are used to input special symbols. I will also add an option to disable the emacsy navigation keys, which includes Ctrl+T on Mac.

This is the emacsy list for reference: https://github.com/codemirror/CodeMirror/blob/8bc57f76383e62e1a03c7d97c9eac74493fdbedc/src/input/keymap.js#L30

lishid:
On top of these, we will also make Obsidian hotkeys run at a higher priority than CodeMirror hotkeys, ensuring that user defined hotkeys will always override any editor ones, instead of triggering both.

manogna4:
> On top of these, we will also make Obsidian hotkeys run at a higher priority than CodeMirror hotkeys, ensuring that user defined hotkeys will always override any editor ones, instead of triggering both.

Perfect, I think this should do it. I was making a list of more hotkeys I'd like to override.

lishid:
Mind if we close this PR? Let me know if our core release fixes this issue.

manogna4:
> Mind if we close this PR? Let me know if our core release fixes this issue.

I believe the issue is fixed. Please go ahead and close the PR.

PR #221, opened 2021-03-28
add buttons to community plugins

shabegom:

## Repo URL

https://github.com/shabegom/buttons

## Release Checklist

- [ x] I have tested this on macOS
- [ x] Github release contains all required files
  - [ x] `main.js`
  - [ x] `manifest.json`
  - [x] `styles.css`
- [ x] Github release name matches the exact version number specified in your manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
- [ x] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [ x] README clearly describes the plugins purpose and provides clear usage instructions.


lishid:
https://github.com/shabegom/buttons/blob/b235371fffe4180b45cd619b3dcb4d43a30e6e12/src/index.ts#L123
This could return null or undefined.

https://github.com/shabegom/buttons/blob/b235371fffe4180b45cd619b3dcb4d43a30e6e12/src/index.ts#L124
Put the file into a var in case it changes, same for the next two functions.

https://github.com/shabegom/buttons/blob/b235371fffe4180b45cd619b3dcb4d43a30e6e12/src/index.ts#L61
This is just `window.open` right? Maybe write it like that just to be explicit.

shabegom:
@lishid  I've addressed those notes and updated the release. Thanks!

PR #222, opened 2021-03-28
Add "Hotkeys for templates" plugin

Vinzent03:
# [x]  I am submitting a new Community Plugin

## Repo URL

https://github.com/Vinzent03/obsidian-hotkeys-for-templates

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


liamcain:
https://github.com/Vinzent03/obsidian-hotkeys-for-templates/blob/324dae60a7fc6d3adb15335e874805873e0875e4/main.ts#L3 Maybe rename this?

https://github.com/Vinzent03/obsidian-hotkeys-for-templates/blob/324dae60a7fc6d3adb15335e874805873e0875e4/main.ts#L18 you might want to add a check here in case the internal plugin isn't loaded. Then maybe provide a notice or some messaging in the settings tab.

https://github.com/Vinzent03/obsidian-hotkeys-for-templates/blob/324dae60a7fc6d3adb15335e874805873e0875e4/main.ts#L22 you don't need to pass `this` in.

https://github.com/Vinzent03/obsidian-hotkeys-for-templates/blob/324dae60a7fc6d3adb15335e874805873e0875e4/main.ts#L47 I think you'll want to call [normalizePath](https://github.com/obsidianmd/obsidian-api/blob/55946e5a6259a28c416d2d6e600a7964a86a01dd/obsidian.d.ts#L1826) first. That will match the behavior of the internal `insertTemplate`

https://github.com/Vinzent03/obsidian-hotkeys-for-templates/blob/324dae60a7fc6d3adb15335e874805873e0875e4/main.ts#L68-L69 I don't think you need to make these instance variables.

https://github.com/Vinzent03/obsidian-hotkeys-for-templates/blob/324dae60a7fc6d3adb15335e874805873e0875e4/main.ts#L98 this shouldn't be necessary

Vinzent03:
Thanks for review.

1. good idea
2. good idea
3. Oh you are right. I thought I have to, because I use it in settings tab too. But never mind
4.  I think I don't need it, because I get the path always from `TFile`. I tested it with some cases and it didn't make a difference. But I can add it of course if you want
5. But then I will have to pass it every time to the `getTemplateFiles(file:TFile)` method. Why is this better?
6. I want to remove files, that no longer exist. If the user doesn't change something else, it's redundant to remove them from the settings.

liamcain:
> I think I don't need it, because I get the path always from TFile. I tested it with some cases and it didn't make a difference. But I can add it of course if you want

The issue is actually with the template folder - since it's user-defined, the format can be a little unpredictable. For example, someone might have a leading slash here 
image

> But then I will have to pass it every time to the getTemplateFiles(file:TFile) method. Why is this better?

I was actually thinking you would have `getTemplateFiles` return the TFiles. I didn't notice that you were using the `templateFolderPath` for trimming 🤔 . Maybe just leave it as is then.

> I want to remove files, that no longer exist. If the user doesn't change something else, it's redundant to remove them from the settings.

I see. My concern is just that `saveSettings` is synchronous and since it's doing a file write, it's not super fast. `display` gets called everytime the setting tab becomes active so ideally you don't slow down the render.

Vinzent03:
Should all be fixed in https://github.com/Vinzent03/obsidian-hotkeys-for-templates/commit/130927043efa70e4fbc4fddead2628d71ad6e526 . I am using `normalizePath()` now to format the template folder. The file existence check is now performed on adding the command and not on loading the settings page.

liamcain (approved):

lishid:
https://github.com/Vinzent03/obsidian-hotkeys-for-templates/blob/c601fbd2f82739bed80b8b0748ef2e32330223a7/main.ts#L135
Should be `file.name` here right?

https://github.com/Vinzent03/obsidian-hotkeys-for-templates/blob/c601fbd2f82739bed80b8b0748ef2e32330223a7/main.ts#L116
Maybe rename this function?

https://github.com/Vinzent03/obsidian-hotkeys-for-templates/blob/c601fbd2f82739bed80b8b0748ef2e32330223a7/main.ts#L18
Suggest adding a kill-switch if anything in this path isn't found, in case we change internal plugin structure one day.

lishid:
Just merged - last comment is all small nits so feel free to fix them later.

Vinzent03:
Thanks for merging.

> Should be file.name here right?

Actually not. I want to show the relative path to the template folder (which is set by the core plugin). The user could have two templates with the same name in a subdirectory.

> Suggest adding a kill-switch if anything in this path isn't found, in case we change internal plugin structure one day.

You mean just a check if the object exists?

lishid:
> Actually not. I want to show the relative path to the template folder (which is set by the core plugin). The user could have two templates with the same name in a subdirectory.

Oh you're right! Carry on.

> You mean just a check if the object exists?

Yeah kinda. In the future we might refactor to move `internalPlugins`, or anything down the chain (`internalPlugins.plugins`, internalPlugins.plugins.templates`, or `internalPlugins.plugins.templates.instance`). I think it's sufficient to just test for truthiness and return with a notice if that fails.

Vinzent03:
Sure thing. Will add it.

PR #223, opened 2021-03-30
Adds obsidian-style-settings plugin

mgmeyers:

## Repo URL

https://github.com/mgmeyers/obsidian-style-settings

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1155,7 +1155,7 @@
     {
         "id": "obsidian-style-settings",
         "name": "Style Settings",
-        "description": "Offers controls for adjusting theme, plugin, and CSS snippet styles",
+        "description": "Offers controls for adjusting theme, plugin, and snippet CSS variables",
         "author": "mgmeyers",
         "repo": "mgmeyers/obsidian-style-settings",
         "branch": "main"
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1125,7 +1125,7 @@
     },
     {
         "id": "obsidian-admonition",
-	      "name": "Admonition",
+        "name": "Admonition",
         "description": "Admonition block-styled content for Obsidian.md",
         "author": "Jeremy Valentine",
         "repo": "valentine195/obsidian-admonition"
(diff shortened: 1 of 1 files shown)


liamcain:
I'll take a look at this today. Also, thank you for finding a decent colorpicker! I've been searching everywhere for one.

liamcain:
https://github.com/mgmeyers/obsidian-style-settings/blob/7d963762c8db6fd69bcbcff46d768e61930d5d29/src/settingHandlers.ts#L153 you can use the [debounce](https://github.com/obsidianmd/obsidian-api/blob/55946e5a6259a28c416d2d6e600a7964a86a01dd/obsidian.d.ts#L607) function from the API here to simplify things a bit

https://github.com/mgmeyers/obsidian-style-settings/blob/7d963762c8db6fd69bcbcff46d768e61930d5d29/src/settingHandlers.ts#L148 is default guaranteed to be non-null?

https://github.com/mgmeyers/obsidian-style-settings/blob/7d963762c8db6fd69bcbcff46d768e61930d5d29/src/settingHandlers.ts#L290 you can use `createDiv` here instead

https://github.com/mgmeyers/obsidian-style-settings/blob/7d963762c8db6fd69bcbcff46d768e61930d5d29/src/settingHandlers.ts#L244 maybe cast this with `String(s)` instead?

---

Overall looks very cool. Is the idea that themes will define defaults in 2 places - in the YAML and then again in the CSS? I'm just curious if themes will be broken if a user doesn't have this plugin installed.

mgmeyers:
Thanks @liamcain! I've made some updates and update the release: https://github.com/mgmeyers/obsidian-style-settings/commit/eed829b011356933bf92e9f9d920ec378417c0eb

> I'm just curious if themes will be broken if a user doesn't have this plugin installed.

This won't be an issue. This plugin will only override the variable if the user modifies it from default. For example, If I wanted to control obsidian's default text color via this plugin, I could put this into a snippet:

```css

/* @settings

name: Color Overrides
id: my-color-overrides
settings:
    - 
        id: text-normal
        title: Text Color
        type: variable-color
        format: rgb
        default: '#000000'

*/

```

Without making any changes via the plugin pane, the plugin won't output anything. If I change the color to `#ff0000` via the plugin, however, it would then output:

```
body.css-settings-manager { --text-normal: rgb(255, 0, 0); }
```

So, theme authors MUST define CSS variables with default values in the CSS itself. Then, when they want to expose a configurable variable to the user, they must define it in the YAML.

Does this make sense?

liamcain:
I see, thanks for the explanation!

liamcain (approved):

PR #224, opened 2021-03-30
Adds obsidian-embedded-note-titles

mgmeyers:

## Repo URL

https://github.com/mgmeyers/obsidian-embedded-note-titles

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1159,5 +1159,13 @@
         "author": "mgmeyers",
         "repo": "mgmeyers/obsidian-style-settings",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-embedded-note-titles",
+        "name": "Embedded Note Titles",
+        "author": "mgmeyers",
+        "description": "Inserts the note file name as an H1 heading above each note",
+        "repo": "mgmeyers/obsidian-embedded-note-titles",
+        "branch": "main"
     }
 ]


liamcain:
https://github.com/mgmeyers/obsidian-embedded-note-titles/blob/62666d803fa27ac1502901e4df335b278fbf3963/src/main.ts#L31 not positive on the timing of things, but this might make more sense onLayoutReady

https://github.com/mgmeyers/obsidian-embedded-note-titles/blob/62666d803fa27ac1502901e4df335b278fbf3963/src/HeadingsManager.ts#L248 you can use the debounce function for this.

https://github.com/mgmeyers/obsidian-embedded-note-titles/blob/62666d803fa27ac1502901e4df335b278fbf3963/src/getMatchedCSSRules.ts#L90 small performance thing - since you are only using this in one spot, why not reverse the sort here since you're immediately calling `reverse()` on it currently

mgmeyers:
Thanks again, @liamcain ! Updates here: https://github.com/mgmeyers/obsidian-embedded-note-titles/commit/23f958b8d6c1b4803a13aebabdfaaccf7e0ed6d0

Re: `onLayoutReady` vs `layout-change`: I trigger `layout-change` on line 33 to invoke the handler I bound above on line 12. I use `layout-change` there, because it will tell me when the user has navigated to a new note, move a note into a different pane, etc. Basically, there are a handful of scenarios where the plugin needs to regenerate the H1 tags, and `layout-change` seemed to work better than the other events I tried.

liamcain:
> Re: onLayoutReady vs layout-change: I trigger layout-change on line 33 to invoke the handler I bound above on line 12.

Sorry, I realize I wasn't super clear with what I meant here. I was just referring to you triggering layout-change _before_ the layout is ready. It's probably fine, but I was just suggesting you wrap the `this.app.workspace.trigger("layout-change");` part in onLayoutReady is all.

---

Changes look good! I still can't get over how much work probably went into making this plugin though 😮 I looked at the main.ts hoping to find a psuedo-element injected, not a whole CSS parser!

mgmeyers:
Oh 💡 ! I see. I'll test out onLayoutReady, that seems like a sensible precaution to me.

mgmeyers:
Thanks @liamcain! onLayoutReady works just fine. I've updated the release.

liamcain (approved):

PR #225, opened 2021-03-31
add 'Advanced Obsidian URI' plugin

Vinzent03:

## Repo URL

https://github.com/Vinzent03/obsidian-advanced-uri

## Release Checklist

- [x] I have tested this on Windows, and Linux
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
https://github.com/Vinzent03/obsidian-advanced-uri/blob/aa446a22dcae9e261fd9486ddcfcbd901c0578d9/main.ts#L14 I think "overwrite" might be a more appropriate name here

https://github.com/Vinzent03/obsidian-advanced-uri/blob/aa446a22dcae9e261fd9486ddcfcbd901c0578d9/main.ts#L30 maybe add some optional chaining (`?.`) here in case anything here gets renamed in the future. From Licat's comment on the discord, I assume it's safe to assume plugins are loaded before the handler is called?

https://github.com/Vinzent03/obsidian-advanced-uri/blob/aa446a22dcae9e261fd9486ddcfcbd901c0578d9/main.ts#L84 I'm not quite sure what `getDisplayText()` returns but could you instead compare the `leaf.file.path` instead? I think it would be a safer check.

Vinzent03:
> I think "overwrite" might be a more appropriate name here

https://github.com/Vinzent03/obsidian-advanced-uri/commit/f6c5c6390de659cdc6ddc56d4a1f1797a2e5ece4

> maybe add some optional chaining (?.) here in case anything here gets renamed in the future. From Licat's comment on the discord, I assume it's safe to assume plugins are loaded before the handler is called?

Sure thing. I think I can assume this. https://github.com/Vinzent03/obsidian-advanced-uri/commit/f96ff2b3098f771928384d7e1df89513795bea8b

> I'm not quite sure what getDisplayText() returns but could you instead compare the leaf.file.path instead? I think it would be a safer check.

Good idea. (It's `leaf.view.file.path` 😉) https://github.com/Vinzent03/obsidian-advanced-uri/commit/98fb4fb37891a252bd7b931666c8eae04cc9e7d2

liamcain:
Small nit: Missed this before - 

https://github.com/Vinzent03/obsidian-advanced-uri/blob/3743890e2ae154c4037f52468f233a7c6d5db20e/main.ts#L52 this might be better as `endsWith(".md")`

---

otherwise looks good to me!

liamcain (approved):

Vinzent03:
Didn't know that exists. Thanks

PR #226, opened 2021-03-31
Update community-plugins.json to include "Blurry Edit Mode" plugin

jillalberts:
I followed the instructions carefully, but this is my first time trying anything like this! Any/all suggestions welcome.

## Repo URL

https://github.com/jillalberts/obsidian-blurry-edit-mode

## Release Checklist

  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1129,5 +1129,12 @@
         "description": "Admonition block-styled content for Obsidian.md",
         "author": "Jeremy Valentine",
         "repo": "valentine195/obsidian-admonition"
+    },
+    {
+        "id": "obsidian-blurry-edit-mode",
+	"name": "Blurry Edit Mode",
+        "description": "Provides a button and command to toggle blurred text for better privacy in Edit mode. No more 'shoulder surfing!'",
+        "author": "Jill Alberts",
+        "repo": "jillalberts/obsidian-blurry-edit-mode"
     }
 ]


ericaxu:
Very interesting plugin! And this is exactly what I see without my glasses on 🤣 .

https://github.com/jillalberts/obsidian-blurry-edit-mode/blob/1fefac6bcfc1e553dde057348857a9db82732d9e/main.ts#L34
Set the fill to `currentColor` instead of `#666666` should make it blend in with themes better. Although there might be reasons to hardcode a color, for example if you feel like your solid icon is too distracting.

https://github.com/jillalberts/obsidian-blurry-edit-mode/blob/1fefac6bcfc1e553dde057348857a9db82732d9e/styles.css#L7
Do you want to provide a setting for it? Or leave it like this because only power users will ever want to change this?

Otherwise looks good to me! 👍

jillalberts:
Hi Erica,

Thanks for the great feedback! 

I changed the icon fill to "currentColor" and it looks better now. Thanks for the guidance. (I haven't pushed the changes to github yet, but I will.)

I'll see what I can do with creating a setting for the "blurLevel" (and maybe another feature I've been thinking about that would require a settings menu)... I don't actually know any javascript, but I bet I can muddle through, the same way I have so far. I think I can come up with a version of this plugin that's a little more polished, so new users know from the very first release that it has settings they can adjust.

When I have a new version that I think is finally ready for release, what should I do? Publish a version "0.2" to my GitHub repo? And... comment here? What is the process? (This is my first plugin and indeed my first-ever pull request.) I'm happy to close this pull request and open a new one when "0.2" is ready if that works best.

Thanks again for your help.

liamcain:
Hi @jillalberts, you'll need to rename the Github Release to match the version number in your manifest.json file - so instead of "Initial Release" it should be `0.1`.

Also, not a requirement but you might want to follow "semantic versioning" convention for the release number. So instead of `0.1` it would be `0.1.0`.

ericaxu:
> I'll see what I can do with creating a setting for the "blurLevel" (and maybe another feature I've been thinking about that would require a settings menu)... I don't actually know any javascript, but I bet I can muddle through, the same way I have so far. I think I can come up with a version of this plugin that's a little more polished, so new users know from the very first release that it has settings they can adjust.

That's alright! I just had the idea when I saw your CSS comment. I think it's perfectly fine to leave it like so, as long as the default blur level is sensible.

jillalberts:
@liamcain Thank you for the feedback and guidance.

@ericaxu and @liamcain Thank you both. I will have a more polished "real" first version ready soon! It will return under a different name, though, so I'm closing this pull request and will open a new one when it's ready.

PR #227, opened 2021-03-31
Obsidian Fountain Plugin

Darakah:

## Repo URL

https://github.com/Darakah/obsidian-fountain

## Release Checklist

- [ ] I have tested this on  macOS only
  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


ericaxu:
Straightforward plugin, looks good to me, thanks for taking the initiative to make this!

I was wondering, would it make sense for this plugin to handle `.fountain` files too, like the CSV plugin? Or should that be the job of another plugin?

lishid:
https://github.com/Darakah/obsidian-fountain/blob/cd4481abfa86f73b510cef7bba3409bb8d258cb4/src/main.ts#L6
I don't see why having this new class is necessary - just move the 6 lines of code directly into the post processor callback?

Darakah:
added release `0.1.0`

@ericaxu , **rendering** and **editing** of `.fountain` files inside obsidian added.

@lishid , removed the processor class. 

----

@lishid Is there a way to register extensions to be considered capable of being embedded ?

ericaxu:
> @lishid Is there a way to register extensions to be considered capable of being embedded ?

Not sure if that's possible, but for advanced features like that I guess you can always add it later.

@lishid What do you think?

lishid:
> Is there a way to register extensions to be considered capable of being embedded ?

Not yet exposed. Will look into exposing it later.

https://github.com/Darakah/obsidian-fountain/blob/20879c0c50a4858b22a7bcfa01b78dc7b9d3263a/src/main.ts#L44
Should be `FountainView` instead of `fountainView`.

Are you sure this is a good idea? I think it's terrible to just extend MarkdownView like that, because you'll get a lot of conflicts with other plugins trying to do things in your view. That plus the MarkdownView is very complicated internally to support all the markdown stuff.

Plus you're registering 3 global hooks for a view which really shouldn't be watching for those events. https://github.com/Darakah/obsidian-fountain/blob/20879c0c50a4858b22a7bcfa01b78dc7b9d3263a/src/main.ts#L58

I think you should just inherit a regular FileView, and implement the `onLoadFile(file: TFile)` and `onUnloadFile` to render it straight into the `contentEl`.

Darakah:
@lishid pushed `0.2.0`

Extending `TextFileView` & changed view name to `FountainView`
https://github.com/Darakah/obsidian-fountain/blob/21747650c20cbd2fc36e9046a6413c1e127364e2/src/main.ts#L44

Using code mirror to edit the file and allowing to switch between the views by clicking a similar ob edit icon
https://github.com/Darakah/obsidian-fountain/blob/21747650c20cbd2fc36e9046a6413c1e127364e2/src/main.ts#L2
https://github.com/Darakah/obsidian-fountain/blob/21747650c20cbd2fc36e9046a6413c1e127364e2/src/main.ts#L86

Only using 1 global event
https://github.com/Darakah/obsidian-fountain/blob/21747650c20cbd2fc36e9046a6413c1e127364e2/src/main.ts#L74

lishid:
https://github.com/Darakah/obsidian-fountain/blob/0a3195baf1062ae4a1af8593a4f0745610e597f8/src/main.ts#L100
Use `requestSave` instead. It's a proper debounce and not just a setTimeout.

https://github.com/Darakah/obsidian-fountain/blob/0a3195baf1062ae4a1af8593a4f0745610e597f8/src/main.ts#L77
Use `setIcon(element, 'pencil', 17)` instead of hardcoding the icon here.

https://github.com/Darakah/obsidian-fountain/blob/0a3195baf1062ae4a1af8593a4f0745610e597f8/src/main.ts#L74
Shouldn't need this global event. That's what setViewData does (read the file and pass it to your class).

https://github.com/Darakah/obsidian-fountain/blob/0a3195baf1062ae4a1af8593a4f0745610e597f8/src/main.ts#L113
You should implement this to clear your editor and preview.

Darakah:
@lishid ,

Removed the setTimeout 
https://github.com/Darakah/obsidian-fountain/blob/9404164726bbff0f22e66bca3c9b17bd4c57e246/src/main.ts#L93

Using setIcon instead of hardcoding it
https://github.com/Darakah/obsidian-fountain/blob/9404164726bbff0f22e66bca3c9b17bd4c57e246/src/main.ts#L74

Removed the global event and the merged the initialization function with the `setViewData` as suggested
https://github.com/Darakah/obsidian-fountain/blob/9404164726bbff0f22e66bca3c9b17bd4c57e246/src/main.ts#L103

Clearing editor / preview
https://github.com/Darakah/obsidian-fountain/blob/9404164726bbff0f22e66bca3c9b17bd4c57e246/src/main.ts#L109

PR #229, opened 2021-04-02
Add cm-chs-patch and table-extended

AidenLx:
# [x] I am submitting a new Community Theme

## Repo URL

- [alx-plugins/cm-chs-patch](https://github.com/alx-plugins/cm-chs-patch)
- [alx-plugins/table-extended](https://github.com/alx-plugins/table-extended)

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_

Related Feature Request: <https://forum.obsidian.md/t/editor-cm5-doesnt-support-word-splitting-in-cjk-languages/15773>


lishid:
I'd suggest splitting your two unrelated plugins so that they can be reviewed and/or merged separately :)

Also please consider auto-formatting your code so they appear nicer when read from github.

https://github.com/alx-plugins/table-extended/blob/e830b0f04912ca0eecba60b25fa8ef458809f47b/src/main.ts#L42
That's pretty obscure, and also `.obsidian` is now configurable so this will definitely break. Use the undocumented API `Vault.getConfig` instead. It's still undocumented, but at least it won't break for now.

https://github.com/alx-plugins/table-extended/blob/e830b0f04912ca0eecba60b25fa8ef458809f47b/src/main.ts#L70
I don't think it's a good idea to work with unsanitary HTML - consider moving the DOM elements over directly?

lishid:
https://github.com/alx-plugins/cm-chs-patch/blob/6e7af09145d9a5709702de0747eb9c7440a98884/src/main.ts#L21
Here you can use `Workspace.iterateCodeMirrors` to restore the default findWordAt behavior. 

https://github.com/alx-plugins/cm-chs-patch/blob/6e7af09145d9a5709702de0747eb9c7440a98884/src/inject.js#L78
Here you are relying on `window.app` which is only there for debugging purposes and may not be always available.
I'd suggest you just pass in `segmentit`, or make the entire findWordAt function bound to your plugin.

AidenLx:
OK, I'll close this pull request and file new ones separately.

PR #230, opened 2021-04-03
chore: add highlightpublicnotes plugin

dennisseidel:

## Repo URL

https://github.com/dennisseidel/highlightpublicnotes-obsidian-plugin

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


liamcain:
https://github.com/dennisseidel/highlightpublicnotes-obsidian-plugin/blob/70461daaf4cc26aadf4c2ac540bdf06671484bc0/main.ts#L13 maybe rename the plugin class name

https://github.com/dennisseidel/highlightpublicnotes-obsidian-plugin/blob/master/README.md#compatibility this should probably use your plugin name rather than "calendar"

https://github.com/dennisseidel/highlightpublicnotes-obsidian-plugin/blob/70461daaf4cc26aadf4c2ac540bdf06671484bc0/main.ts#L64 you can remove this

---

This functionality is very similar to the built-in `cssclasses` that is supported by frontmatter currently. Was the issue that you can't style the title element with that? I'm curious since this has the disadvantage of toggling based only on the state of the last opened note. So if you have 2 open notes side by side, this plugin won't work as expected.

dennisseidel:
Thanks for the feedback, it's my first plugin, I very much appreciate it :) 
I fixed all your remarks.

About the `cssclasses`: I didn't know about the feature. I could realize my functionality "highlighting public notes" with `cssclasses` e.g., changing the note color to red. I don't really care what the indicator is as long as it is visible. 
About the plugin functionality: The titlebar has always the color of the data classification of the currently active focused note. For me that is the most important part, I can easily spot when I write to a public note.  

Trade-off of the two implementations starting with `cssclasses`:
+ functionality works (e.g. text color of note)
+ is built in (performance, maintenance) 
- data classification is harder to spot & maintain in the frontmatter (as one of the cssclasses)

`separate plugin`:
+ functionality works (titlebar)
- separate plugin (performance, maintenance)
+ data classification & functionality is very explicit

I would definitely add the cssclasses option into the plugin description - I guess for many this is the better solution.

lishid:
https://github.com/dennisseidel/highlightpublicnotes-obsidian-plugin/blob/78f892aa0ff5806ed19b3fd80d17c2aafd531093/main.ts#L26
This shouldn't need to await right?

PR #231, opened 2021-04-03
Add privacy-glasses

jillalberts:
"Privacy Glasses" is the new and much-improved version of the plugin formerly known as "Blurry Edit Mode"

## Repo Link

https://github.com/jillalberts/privacy-glasses

## Release Checklist

  - [X] `main.js`
  - [X] `manifest.json`


Darakah:
Cute plugin mate and very nice use case examples! I am 100% sold on this! 
Great work!

liamcain:
Hi @jillalberts, nice work on the plugin!

---

https://github.com/jillalberts/privacy-glasses/blob/c07c75ba88bcdd67110e0ffa5eea077ad1983b88/main.ts#L23 you could make this into a function in your class like `togglePrivacy()` that way you don't have to repeat it here and again for the command

https://github.com/jillalberts/privacy-glasses/blob/c07c75ba88bcdd67110e0ffa5eea077ad1983b88/main.ts#L236 The icon string shouldn't include the surrounding `<svg>` tag.

https://github.com/jillalberts/privacy-glasses/blob/c07c75ba88bcdd67110e0ffa5eea077ad1983b88/main.ts#L171 You could use a [slider](https://github.com/obsidianmd/obsidian-api/blob/dbfa19ad7aa6557f0ecff962065c3f540bc77e27/obsidian.d.ts#L2327) for this value

jillalberts:
@Darakah Thank you for the kind words. I think this plugin nicely straddles the line between "fun/ridiculous" and "could actually be useful"!

@liamcain Thank you, and thanks for the helpful suggestions! I have published Privacy Glasses release 0.4.1 to my github repo which includes the following changes:

- the plugin's toggle code is in its own new function toggleGlasses()

- removed the `<svg>` tag from the icon string

- changed the `blurLevel` selector from a dropdown to a slider and added auto-updating description text with its current value.

- added a "master switch" toggle to the settings menu (just another way to toggle the plugin)

- when unloaded, the plugin now properly removes the <style> element it created in <head>

- activating or deactivating the plugin now provides a more visible indication (via a `Notice`)

liamcain (approved):

lishid:
Super cool plugin by the way.

https://github.com/jillalberts/privacy-glasses/blob/85834b438e962ca44b4e42684c65a1f32170775a/main.ts#L9
I feel like this may be a bit unnecessary? You can just store a variable to the `statusBarEl: HTMLElement` and call setText directly.

https://github.com/jillalberts/privacy-glasses/blob/85834b438e962ca44b4e42684c65a1f32170775a/main.ts#L54
Could store this text in a variable to be used twice... so you don't have to type it twice and risk them being different by a single character 😂 

https://github.com/jillalberts/privacy-glasses/blob/85834b438e962ca44b4e42684c65a1f32170775a/main.ts#L84
You can just store a reference to this element in your plugin object to remove it later, instead of using an ID and querying it out of DOM.

https://github.com/jillalberts/privacy-glasses/blob/master/main.js
Add this file to your `.gitignore` so you don't have to commit it every time.

jillalberts:
Hi @lishid, thank you for the feedback.

I have published a new release to github, Privacy Glasses 0.4.2

- The statusbar code is now much cleaner
- Activation messages now match (and are stored in variables)
- The `style` element containing `blurLevel` variable is now referenced in a better way in the code, as you suggested
- The `style` element containing `blurLevel` is now only created when the plugin is actively on (not just when the plugin is loaded and available to use). It is cleanly removed when the plugin is toggled off (rather than waiting until the plugin is unloaded). Changes to the blur level slider value are still saved even when plugin is not on, but there's less clutter in the document's `head` tag. 
- The variable that stores whether or not the plugin is on is no longer saved among the plugin's settings, since I think having the plugin turn itself on upon loading a vault would probably be more annoying than useful. If users want the ability to save the plugin's state across sessions then I can do that in the future.
- Misc small code improvements

lishid:
https://github.com/jillalberts/privacy-glasses/blob/141a6827fab485c1726d72f22e4e9b07bba6cb81/main.ts#L92
https://github.com/jillalberts/privacy-glasses/blob/141a6827fab485c1726d72f22e4e9b07bba6cb81/main.ts#L116
https://github.com/jillalberts/privacy-glasses/blob/141a6827fab485c1726d72f22e4e9b07bba6cb81/main.ts#L124
https://github.com/jillalberts/privacy-glasses/blob/141a6827fab485c1726d72f22e4e9b07bba6cb81/main.ts#L129
https://github.com/jillalberts/privacy-glasses/blob/141a6827fab485c1726d72f22e4e9b07bba6cb81/main.ts#L136
All of these functions don't need to be async.

https://github.com/jillalberts/privacy-glasses/blob/141a6827fab485c1726d72f22e4e9b07bba6cb81/main.ts#L76
Consider moving this one to the very end so that in case any error happens you remove the css styles first.

jillalberts:
@lishid Thank you for the feedback. I have published a new release, Privacy Glasses 0.4.3, with the changes you suggested.

PR #233, opened 2021-04-06
Add cm-chs-patch

AidenLx:
## Repo URL

[alx-plugins/cm-chs-patch](https://github.com/alx-plugins/cm-chs-patch)

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_

Related Feature Request: <https://forum.obsidian.md/t/editor-cm5-doesnt-support-word-splitting-in-cjk-languages/15773>

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1183,5 +1183,12 @@
         "author": "darakah",
         "repo": "Darakah/obsidian-fountain",
         "branch": "main"
+    },
+    {
+        "id": "cm-chs-patch",
+        "name": "Word Splitting for Simplified Chinese in Edit Mode",
+        "author": "AidenLx",
+        "description": "A patch for Obsidian's built-in CodeMirror Editor to support Simplified Chinese word splitting",
+        "repo": "alx-plugins/cm-chs-patch"
     }
 ]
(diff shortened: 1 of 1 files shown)


AidenLx:
From [229#](https://github.com/obsidianmd/obsidian-releases/pull/229#issuecomment-813544149)
> I'd suggest splitting your two unrelated plugins so that they can be reviewed and/or merged separately :)
> 
> Also please consider auto-formatting your code so they appear nicer when read from github.
> 
> https://github.com/alx-plugins/table-extended/blob/e830b0f04912ca0eecba60b25fa8ef458809f47b/src/main.ts#L42
> That's pretty obscure, and also `.obsidian` is now configurable so this will definitely break. Use the undocumented API `Vault.getConfig` instead. It's still undocumented, but at least it won't break for now.
> 
> https://github.com/alx-plugins/table-extended/blob/e830b0f04912ca0eecba60b25fa8ef458809f47b/src/main.ts#L70
> I don't think it's a good idea to work with unsanitary HTML - consider moving the DOM elements over directly?

Resolved in v1.1.0 with improved inject method.

PR #234, opened 2021-04-06
Add table-extended

AidenLx:
# [x] I am submitting a new Community Theme

## Repo URL

[alx-plugins/table-extended](https://github.com/alx-plugins/table-extended)

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1183,5 +1183,12 @@
         "author": "darakah",
         "repo": "Darakah/obsidian-fountain",
         "branch": "main"
+    },
+    {
+        "id": "table-extended",
+        "name": "Table Extended",
+        "author": "AidenLx",
+        "description": "Enable extended table support with MultiMarkdown 6 syntax",
+        "repo": "alx-plugins/table-extended"
     }
 ]


AidenLx:
From [229#](https://github.com/obsidianmd/obsidian-releases/pull/229#issuecomment-813544149)

> https://github.com/alx-plugins/table-extended/blob/e830b0f04912ca0eecba60b25fa8ef458809f47b/src/main.ts#L42
> That's pretty obscure, and also `.obsidian` is now configurable so this will definitely break. Use the undocumented API `Vault.getConfig` instead. It's still undocumented, but at least it won't break for now.
> 
> https://github.com/alx-plugins/table-extended/blob/e830b0f04912ca0eecba60b25fa8ef458809f47b/src/main.ts#L70
> I don't think it's a good idea to work with unsanitary HTML - consider moving the DOM elements over directly?

Fixed in v1.0.1

lishid:
https://github.com/alx-plugins/table-extended/blob/8a381161fc5326ce137624ae6a997c0f8640231c/src/main.ts#L40
You can just cast vault to any for this, or use `// @ts-ignore` on the line before.

https://github.com/alx-plugins/table-extended/blob/8a381161fc5326ce137624ae6a997c0f8640231c/src/main.ts#L81
Don't recommend throwing errors here. Just try to cleanup as best as you can.

AidenLx:
Sorry for the inconvenience, I haven't spotted the mixed tab and space indents, it's fixed now.

All suggestions are [reflected in the code](https://github.com/alx-plugins/table-extended/commit/f5f91c4c40645e750f7d37a3e6abf21cdd534e29) now.

PR #238, opened 2021-04-07
Progress Plugin

Darakah:

## Repo URL

https://github.com/Darakah/obsidian-progress

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1198,5 +1198,13 @@
         "author": "Jill Alberts",
         "repo": "jillalberts/privacy-glasses",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-progress",
+        "name": "Progress",
+        "description": "Create progress displays in obsidian vault or project",
+        "author": "darakah",
+        "repo": "Darakah/obsidian-progress",
+        "branch": "main"
     }
 ]


lishid:
https://github.com/Darakah/obsidian-progress/blob/34ebe4bf156675ceea5ee5d6651bd38015e36a99/src/block.ts#L15
Same as your previous plugin, why create a class with a single function and then construct the class just to call its function once from a single place? Just move this code right into the post processor.

https://github.com/Darakah/obsidian-progress/blob/34ebe4bf156675ceea5ee5d6651bd38015e36a99/src/block.ts#L51
This looks like a good place for a util function to sanitize things.

https://github.com/Darakah/obsidian-progress/blob/34ebe4bf156675ceea5ee5d6651bd38015e36a99/src/utils.ts#L8
Put this right where it's being used.

https://github.com/Darakah/obsidian-progress/blob/34ebe4bf156675ceea5ee5d6651bd38015e36a99/src/utils.ts#L63
This is also a function that's used just once. Put this next to where it's used. It seems that in your plugins you do a lot of passing data objects around helper functions that you only use once.

https://github.com/Darakah/obsidian-progress/blob/34ebe4bf156675ceea5ee5d6651bd38015e36a99/src/svelte/Progress.svelte#L47
Any considerations for making width automatic? (i.e. 100%) Maybe just make it a string, ask the user to specify, and default to '100%' instead.

ericaxu:
Hi! Did you have time to address the comments @Darakah ? Thank you so much!

ericaxu:
Hi there, to keep things tidy, we're going to temporarily close plugin PRs after 10 days of no activity. Feel free to tell us to re-open it when you're ready, or submit a new PR. Thanks for your understanding!

PR #240, opened 2021-04-09
Add new obsidian-lab plugin

cristianvasquez:
## Repo URL

<https://github.com/cristianvasquez/obsidian-lab>

## Repo URL

<https://github.com/cristianvasquez/obsidian-lab>

## Release Checklist

- [X ] have tested this on Linux and Windows (I don't have access to a macOS)
  - [X] `main.js`
  - [X] `manifest.json`
  - `styles.css` _(optional)_
- [?] README clearly describes the plugins purpose and provides clear usage instructions. (Hopefuly)


lishid:
https://github.com/cristianvasquez/obsidian-lab/blob/2497bff7046cbee8dd1e3b436235f73696eacf53/src/main.ts#L41
I don't think making static classes are necessary, you can just declare your variables and functions right in the root of the file and use them directly.

https://github.com/cristianvasquez/obsidian-lab/blob/2497bff7046cbee8dd1e3b436235f73696eacf53/src/main.ts#L135
This is a bad idea to do here - you will block the app from loading and cause the app startup time to delay.
What you should do is run this asynchronously in another function, without `await` on the result.

Actually, ideally you can do this dynamically without having to reload the plugin

https://github.com/cristianvasquez/obsidian-lab/blob/2497bff7046cbee8dd1e3b436235f73696eacf53/src/main.ts#L138
It seems that you're looking for the `workspace.onLayoutReady` function.

https://github.com/cristianvasquez/obsidian-lab/blob/2497bff7046cbee8dd1e3b436235f73696eacf53/src/main.ts#L204
This function name is kinda misleading. 

https://github.com/cristianvasquez/obsidian-lab/blob/2497bff7046cbee8dd1e3b436235f73696eacf53/src/main.ts#L218
Here you should use `getActiveViewOfType(MarkdownView)` and test for truthiness instead of assuming that activeLeaf is not null.
Also, you should use `view.editor` instead of `view.sourceMode.cmEditor` as `cmEditor` is deprecated, unless you need specific APIs from CodeMirror 5 that's not available from `Editor`. Maybe you need to run `npm update` to get the latest definition files.

https://github.com/cristianvasquez/obsidian-lab/blob/2497bff7046cbee8dd1e3b436235f73696eacf53/src/main.ts#L237
Result isn't returned. Why not just use await for these, instead of mixing async/await and `.then()`?

https://github.com/cristianvasquez/obsidian-lab/blob/2497bff7046cbee8dd1e3b436235f73696eacf53/src/panel.ts#L29
This should just call redraw since every usage of it calls it right after.

https://github.com/cristianvasquez/obsidian-lab/blob/2497bff7046cbee8dd1e3b436235f73696eacf53/src/main.ts#L325
You're probably looking for `Workspace.detachLeavesOfType` (or `getLeavesOfType`) If not, you can just use `leaf.getViewType` instead of `leaf.getViewState().type`

cristianvasquez:
Hi Liam! Thanks for the review! It's appreciated.

I've done most of the changes, but I'm having problems adding and removing views when doing modifications in 'settings.'

I want that the user to make changes without reloading the plugin.

At some point, I do:
this.registerView(commandId, (leaf)) => ...

But I don't know how to refresh it when the user makes modifications.
Perhaps to 'unregister' and then 'register' it? How can I know when it's already registered? What is to 'register the view' after all? :D

lishid:
I see what you mean; indeed it's not very easy to unregister (they're automatically added to a queue to be unregistered when your plugin is deactivated). I guess for now you can leave it in as a restriction (having to disable/enable the plugin).

cristianvasquez:
Hi Liam, thanks for the support!

Would it be ok to merge?

cristianvasquez:
Hello @lishid ?

lishid:
Apologies for the late review, everything looks good! One last thing I want to check in - do you want to rename your plugin to something python related? Seems like in code you've named it `PythonLabPlugin`

cristianvasquez:
Hello, @lishid 

I renamed the plugin to python lab! :)

PR #241, opened 2021-04-09
Add Spaced Repetition Plugin

st3v3nmw:

## Repo URL

https://github.com/st3v3nmw/obsidian-spaced-repetition

## Release Checklist

- [x] I have tested this on Linux
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1200,10 +1200,10 @@
         "branch": "main"
     },
     {
-        "id": "concepts-review",
-        "name": "Concepts Review",
+        "id": "obsidian-spaced-repetition",
+        "name": "Spaced Repetition",
         "description": "Fight the forgetting curve & note aging by reviewing notes using spaced repetition on Obsidian.md",
         "author": "Stephen Mwangi",
-        "repo": "st3v3nmw/concepts-review"
+        "repo": "st3v3nmw/obsidian-spaced-repetition"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1198,5 +1198,12 @@
         "author": "Jill Alberts",
         "repo": "jillalberts/privacy-glasses",
         "branch": "main"
+    },
+    {
+        "id": "concepts-review",
+        "name": "Concepts Review",
+        "description": "Fight the forgetting curve & note aging by reviewing notes using spaced repetition on Obsidian.md",
+        "author": "Stephen Mwangi",
+        "repo": "st3v3nmw/concepts-review"
     }
 ]


lishid:
https://github.com/st3v3nmw/concepts-review/blob/f695aa2c0839bbfefe1719693d5e285fd6aab44b/main.ts#L84
You can use `workspace.onLayoutReady()` instead which does this check for you.

https://github.com/st3v3nmw/concepts-review/blob/f695aa2c0839bbfefe1719693d5e285fd6aab44b/main.ts#L90
You can do all three in a single file-menu event. Also you should call `Plugin.registerEvent` so that these events are unhooked when your plugin is disabled.

https://github.com/st3v3nmw/concepts-review/blob/f695aa2c0839bbfefe1719693d5e285fd6aab44b/main.ts#L130
Don't do this in your onload, as it will stop other plugins (and the rest of the app) from loading. Instead, do it in a setTimeout().

https://github.com/st3v3nmw/concepts-review/blob/f695aa2c0839bbfefe1719693d5e285fd6aab44b/main.ts#L148
`Date.now()` is probably a better choice.

https://github.com/st3v3nmw/concepts-review/blob/f695aa2c0839bbfefe1719693d5e285fd6aab44b/main.ts#L151
I would recommend doing this after you have verified the frontmatter entry just so that you aren't reading every single file in the vault. Also since you're only using the text for display, you can use `cachedRead` instead which might read the file from memory saving a trip to the disk.

https://github.com/st3v3nmw/concepts-review/blob/f695aa2c0839bbfefe1719693d5e285fd6aab44b/main.ts#L153
Use `getFileCache(note)` instead of `getCache(note.path)`.

https://github.com/st3v3nmw/concepts-review/blob/f695aa2c0839bbfefe1719693d5e285fd6aab44b/main.ts#L191
Prefer to use for...of instead of forEach unless you have special reasons to.

https://github.com/st3v3nmw/concepts-review/blob/f695aa2c0839bbfefe1719693d5e285fd6aab44b/main.ts#L195
This could get quite expensive, since you're running the a link resolution algorithm on every single link in every single file in the vault. We have all this information pre-computed in `metadataCache.resolvedLinks`, so you don't have to compute it again.

https://github.com/st3v3nmw/concepts-review/blob/f695aa2c0839bbfefe1719693d5e285fd6aab44b/main.ts#L740
There is a shortcut for creating elements with just a class name - just pass the classes as a string instead of an object.

lishid:
Oh yeah, considering that this is essentially spaced-repetition, I think using the words "spaced-repetition" somewhere in the name of the plugin would help more people find this plugin (instead of concepts-review)

st3v3nmw:
Hello @lishid , thanks for the review.
Per your suggestions, I have made the necessary changes to make the plugin more compliant to the API.

lishid:
Small nit: https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/a8fcdb46e6068698ef183717589ca2297423c919/main.ts#L159
You're constructing Object.keys 3 times, whereas you probably only need it once. Actually, you should just use frontmatter.hasOwnProperty(key).

https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/a8fcdb46e6068698ef183717589ca2297423c919/main.ts#L162
Actually, why are you storing a copy of every single note's contents in memory anyway? It seems that you never actually use the content you initially read. Can't you just read the file then? If so you can just turn `new_notes` into an array of TFile.

https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/a8fcdb46e6068698ef183717589ca2297423c919/main.ts#L53
These doesn't seem typed. I'm guessing it's due to how complicated this type is. I would recommend simplifying it and splitting it out, rather than putting all the information into a single complex untyped blob. For example, outgoing_links is just a copy of our resolvedLinks object with some additional information for the note, so I would argue that it's probably not even necessary to keep.

https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/a8fcdb46e6068698ef183717589ca2297423c919/main.ts#L182
Test for existence here - it's not guaranteed to exist.

https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/a8fcdb46e6068698ef183717589ca2297423c919/main.ts#L185
Don't forget to toLowerCase().

https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/a8fcdb46e6068698ef183717589ca2297423c919/main.ts#L195
Can probably do this in the same loop before, instead of looping again.

https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/a8fcdb46e6068698ef183717589ca2297423c919/main.ts#L253
Don't use cachedRead if you intend to write the contents back to the file; use read here instead for maximum data-integrity.

lishid:
It's coming along pretty well! Just a few more suggestions for general performance :)

ksandvik:
Could the documentation -- on github -- specify more how to use this plugin as end user, thx?

st3v3nmw:
Hello @ksandvik, I'll work on fixing that.

st3v3nmw:
@ksandvik, I've added a couple screenshots for the review process. I hope that helps you get started.

@lishid, thanks for the feedback. I've fixed a few of the problems. Though I have a chicken-and-egg situation with the types. I'll finish up later.

lishid:
https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/56deb8cfe5e21956e65121da20f7e9f9bb3618c6/main.ts#L155
This can now be removed right?

https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/56deb8cfe5e21956e65121da20f7e9f9bb3618c6/main.ts#L139
Based on how you're using these, I would recommend removing the "list" and "ease" from your structure, and changing `incoming_links` and `outgoing_links` like this:

- Add a new class field `easeByPath: Record<string, number>` to store the "ease" per file.
- Remove `outgoing_links` and just use `app.metadataCache.resolvedLinks` (It's identical other than that you have a "list" in between that isn't doing much).
- Flatten `incoming_links` to remove the "list", becoming a 2 level object instead of a 3 level.

st3v3nmw:
I think I've fixed all the identified problems.

lishid:
Alright, I think we're pretty close! If you'd like to have the plugin merged in this state, let me know and I can have it merged.

[Here](https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/39453d0012b7e7eddc4be818ce534123fa508521/main.ts#L60) you're using a javascript object to store order information, which I don't recommend using. Previously you were using an array.
While there is an iteration order for for javascript objects since ES2015, I think it's still much clearer to use an array.

https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/39453d0012b7e7eddc4be818ce534123fa508521/main.ts#L346
Consider flattening out statements like this to give yourself and the reviewer an easier time to see what's going on.
For example:
```ts
this.app.workspace.activeLeaf.openFile(
    Object.entries(this.scheduledNotes)[
        this.data.settings.openRandomNote
            ? Math.floor(Math.random() * this.dueNotesCount)
            : 0
    ][1].note
);
// can be rewritten as
let index = this.data.settings.openRandomNote ? Math.floor(Math.random() * this.dueNotesCount) : 0;
let scheduledNote = Object.entries(this.scheduledNotes)[index][1];
this.app.workspace.activeLeaf.openFile(scheduledNote.note);
```

(Actually you can use `Object.values` instead of `Object.entries`).

st3v3nmw:
> [Here](https://github.com/st3v3nmw/obsidian-spaced-repetition/blob/39453d0012b7e7eddc4be818ce534123fa508521/main.ts#L60) you're using a javascript object to store order information, which I don't recommend using. Previously you were using an array.
> While there is an iteration order for for javascript objects since ES2015, I think it's still much clearer to use an array.

Oh, yeah, an array and the `easeByPath` object you'd mentioned earlier makes more sense.

All done! 😃

PR #242, opened 2021-04-09
Add Juggl Plugin

HEmile:

## Repo URL
https://github.com/HEmile/obsidian-neo4j-graph-view/tree/juggl

(This is not the main branch to ensure people can still download Neo4j Graph View until this is pulled)

## Release Checklist
- [x] I have tested this on Windows, macOS and Linux
- [x] I have tested this on Android (it loads, but needs more work!)
- [x] Github release contains all required files (for the old version Neo4j Graph View Plugin)

Hi! I would like to remove the Neo4j Graph View plugin and replace it with the new plugin "Juggl". I will not support Neo4j Graph View anymore, which is why it replaces it. If I understand correctly, I do have to keep the id 'neo4j-graph-view' though to ensure it properly updates.

Although it is already on the plugin store, I much welcome a code review as the plugin has been completely rewritten :) Let me know if you have comments!


community-plugins.json
```diff
@@ -499,10 +499,10 @@
     },
     {
         "id": "neo4j-graph-view",
-        "name": "Neo4j Graph View",
+        "name": "Juggl",
         "author": "Emile van Krieken",
-        "description": "A plugin for advanced graph visualization and querying using Neo4j.",
-        "repo": "HEmile/obsidian-neo4j-graph-view",
+        "description": "Adds a completely interactive, stylable and expandable graph view to Obsidian",
+        "repo": "HEmile/juggl",
```
liamcain, line 505: Is the plan to make the plugin into its own repo or leave it in the neo4j one? This destination current doesn't exist.

HEmile: I will rename the current repo to juggl when this PR is accepted.

HEmile: I cannot change it before it is accepted, otherwise the old plugin won't appear on the community plugins section anymore

lishid:
So just making sure, you want people to auto-update their Neo4j plugin into your new plugin, right?
I'm afraid that might cause some confusion especially with the plugin ID being different than the name.

We could port over plugin stats for a new submission/changed github repo if that's what you need, but if you do wish to upgrade everyone to your new plugin directly from Neo4j then you have to keep the ID.

HEmile:
Hm, good points. I don't really like that it uses the neo4j-graph-view folder either. 
Would this be possible?
- use the juggl id instead, port over stats 
- Still keep neo4j-graph-view online
- I create an update for neo4j-graph-view with a notice to download juggl as ngv is deprecated
- remove neo4j-graph-view in a couple of weeks

HEmile:
Okay, current state:
- https://github.com/HEmile/juggl contains the new Juggl plugin (renamed the old repo)
- https://github.com/HEmile/obsidian-neo4j-graph-view Contains the Neo4j-Graph-View plugin

Both have their corresponding releases ready. I've moved the states of neo4j-graph-view to juggl. Is that the proper way to do this?

lishid:
That makes sense - but don't edit community-plugin-stats.json since that's generated by a script. Silver can help porting over the download stats once we merge the PR.

Pushing an update with a notice should work too.

HEmile:
Alright, I removed those changes. And indeed, I'll push an update with the notice later :) Thanks for the help.

lishid:
Cool, I'll pass by to review soon.

lishid:
https://github.com/HEmile/juggl/blob/main/setup.py
I think this isn't needed anymore since you don't need to setup python right?

https://github.com/HEmile/juggl/blob/a0cac0d2142aa6414903b9a3da04d03a510e65c1/juggl/main.ts#L145
Display text doesn't sound like what you'd want to use - instead consider using view.file.name, or path.

https://github.com/HEmile/juggl/blob/a0cac0d2142aa6414903b9a3da04d03a510e65c1/juggl/main.ts#L193
Use our exposed YAML parser instead of including another copy of 'yaml' in your distribution js.
https://github.com/obsidianmd/obsidian-api/blob/dbfa19ad7aa6557f0ecff962065c3f540bc77e27/obsidian.d.ts#L1926

https://github.com/HEmile/juggl/blob/a0cac0d2142aa6414903b9a3da04d03a510e65c1/juggl/main.ts#L209
Typically in js you would throw a new Error(message) and use e.message, instead of throwing a string (which you can but isn't conventional).

https://github.com/HEmile/juggl/blob/a0cac0d2142aa6414903b9a3da04d03a510e65c1/juggl/main.ts#L251
Don't await potentially slow promises in your plugin's onload as it will delay loading the app. You're probably looking for the workspace.onLayoutReady function too.

https://github.com/HEmile/juggl/blob/a0cac0d2142aa6414903b9a3da04d03a510e65c1/juggl/main.ts#L265
Why not just use raw? Just filter the event by path. Also you might want to take into account the fact that `.obsidian` can be configured to another folder now. https://github.com/obsidianmd/obsidian-api/blob/dbfa19ad7aa6557f0ecff962065c3f540bc77e27/obsidian.d.ts#L2726

https://github.com/HEmile/juggl/blob/a0cac0d2142aa6414903b9a3da04d03a510e65c1/juggl/main.ts#L301
I think you're looking for `Workspace.getLeaf(newLeaf)`

HEmile:
Thanks so much for your time and for the extensive and useful review! I incorporated all changes in the repo and build a new version.

lishid:
> https://github.com/HEmile/juggl/blob/a0cac0d2142aa6414903b9a3da04d03a510e65c1/juggl/main.ts#L251
Don't await potentially slow promises in your plugin's onload as it will delay loading the app. **You're probably looking for the workspace.onLayoutReady function too.**

You can probably use the onLayoutReady function, and also you don't need to call `.then()`, it should just work. (If you call an async function without `await` it will run just the function in parallel without waiting for its result, which is exactly what you want.
```ts
this.app.workspace.onLayoutReady(createPanes);
```

lishid:
@ericaxu can you port over the plugin stats from "obsidian-neo4j-graph-view" when you're available?

lishid:
@HEmile This is obviously a very large and impressive project - as such, I can't really do a full deep-dive review for the rest of the plugin. Hope you have it all under control, and keep up the great work!

HEmile:
Thanks @lishid ! It's quite a big one indeed so I hope I do :) 
And you're right, I overlooked that comment. Will look into that!

PR #243, opened 2021-04-10
Add obsidian-tabs

gitobsidiantutorial:

## Repo URL
[gitobsidiantutorial/obsidian-tabs](https://github.com/gitobsidiantutorial/obsidian-tabs)

## Release Checklist
* [x]  I have tested this on Windows
* [x]  Github release contains all required files
  
  * [x]  `main.js`
  * [x]  `manifest.json`
  * [x] `styles.css` 
* [x]  Github release name matches the exact version number specified in your manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
* [x]  README clearly describes the plugins purpose and provides clear usage instructions.


liamcain:
Very cool to see you turned this into a plugin!

---

I recommend using a code formatter (like [Prettier](https://prettier.io/)) since I got confused reading the indentation in your main.ts file.

https://github.com/gitobsidiantutorial/obsidian-tabs/blob/4839fce6eadda8332241d1f8c325aa79008f6816/src/main.ts#L81 don't use a class for your settings since and instead use a typescript Interface here. Then just use a plain JS object for your default values.

https://github.com/gitobsidiantutorial/obsidian-tabs/blob/4839fce6eadda8332241d1f8c325aa79008f6816/src/main.ts#L6 You should follow the pattern used in the sample plugin:
```ts
this.settings = Object.assign({}, DEFAULT_SETTINGS, await this.loadData());
```

This pattern for saving and loading settings will make it easier in the future when you need to add new settings. You'll be able to merge existing settings with new options without changing anything here.

https://github.com/gitobsidiantutorial/obsidian-tabs/blob/4839fce6eadda8332241d1f8c325aa79008f6816/src/main.ts#L65 instead of using `var` you can use a ternary for this lines of code:
```ts
const opentabs2 = modverticalselector ? modverticalselector.children.length : 0;
```

I'm still on my first pass but I'm not sure I know what `opentabs1` and `opentab2` are. I recommend choosing better names and avoiding variables that just end in a number.

https://github.com/gitobsidiantutorial/obsidian-tabs/blob/4839fce6eadda8332241d1f8c325aa79008f6816/src/main.ts#L15 it might make sense to store a variable pointing at `document.querySelector(':root')` on your plugin instance. That way you don't have to keep querying the DOM for it.

https://github.com/gitobsidiantutorial/obsidian-tabs/blob/4839fce6eadda8332241d1f8c325aa79008f6816/src/main.ts#L82 small nitpick - it's more common to use lowercase values for keys in javascript/typescript.

gitobsidiantutorial:
Thank you for the kind words and the much needed review!

Loading settings is now done in accordance with the sample plugin.

I've scrapped the usage of `mod:root` altogether, and DOM queries now only occur within the rootsplit. (This conveniently also gets rid of the hacky opentabs1 and opentab2

The casing for keys has also been followed.

gitobsidiantutorial:
> I recommend using a code formatter (like [Prettier](https://prettier.io/)) since I got confused reading the indentation in your main.ts file.

- [x]  Done!

> https://github.com/gitobsidiantutorial/obsidian-tabs/blob/4839fce6eadda8332241d1f8c325aa79008f6816/src/main.ts#L53 I think activeLeaf can be null, so I recommend adding a check for that at the start of this function.

Is checking for truthiness with an if () statement sufficient? https://github.com/gitobsidiantutorial/obsidian-tabs/blob/f6a855079cfda6895ec5dec55fc5cd8bd3d7caa1/src/main.ts#L112

> This pattern for saving and loading settings will make it easier in the future when you need to add new settings. You'll be able to merge existing settings with new options without changing anything here.

- [x] Done! [Default settings here](https://github.com/gitobsidiantutorial/obsidian-tabs/blob/f6a855079cfda6895ec5dec55fc5cd8bd3d7caa1/src/main.ts#L140), [loading settings here](https://github.com/gitobsidiantutorial/obsidian-tabs/blob/f6a855079cfda6895ec5dec55fc5cd8bd3d7caa1/src/main.ts#L18).

> [https://github.com/gitobsidiantutorial/obsidian-tabs/blob/4839fce6eadda8332241d1f8c325aa79008f6816/src/main.ts#L65](https://github.com/gitobsidiantutorial/obsidian-tabs/blob/4839fce6eadda8332241d1f8c325aa79008f6816/src/main.ts#L65) instead of using `var` you can use a ternary for this lines of code:

https://github.com/gitobsidiantutorial/obsidian-tabs/blob/f6a855079cfda6895ec5dec55fc5cd8bd3d7caa1/src/main.ts#L118
This was a workaround to count the number of panes open in a split, with an inaccurate and hacky way to support having multiple tabbed interfaces at the same time (picking the highest and praying for the best 😆). Now the vertical splits are given properties separately, for actual support.

> [https://github.com/gitobsidiantutorial/obsidian-tabs/blob/4839fce6eadda8332241d1f8c325aa79008f6816/src/main.ts#L15](https://github.com/gitobsidiantutorial/obsidian-tabs/blob/4839fce6eadda8332241d1f8c325aa79008f6816/src/main.ts#L15) it might make sense to store a variable pointing at `document.querySelector(':root')` on your plugin instance. That way you don't have to keep querying the DOM for it.

Whole document queries have been scrapped. As of now, no variables for saving query results are used, except the pre-existing `workspace.rootSplit`.

---

I've also added a [bit of code](https://github.com/gitobsidiantutorial/obsidian-tabs/blob/f6a855079cfda6895ec5dec55fc5cd8bd3d7caa1/src/main.ts#L22) on load that solves collapsed tabs on startup with multiple tabbed 'panes' in a workspace. As of now, it gives the first pane in a vertical split the `.stayopen` class. This happens *before* the refresh() function is called, which in turn calls a cleaner function to remove the `.stayopen` class from siblings of the active pane. If there's a cleaner or more lightweight way of doing this, I'd love to learn!

Due to time constraints and lack of knowledge, it is merely the first pane in a split that is set to open on loading, rather than writing down the leaf ID of open panes on unload, and applying the class to the leaf with a selected ID. (This should theoretically be possible right? The workspace file in .obsidian has an ID for leaves, but I can't seem to find anything in the API to interact with this.)

liamcain:
> Is checking for truthiness with an if () statement sufficient? https://github.com/gitobsidiantutorial/obsidian-tabs/blob/f6a855079cfda6895ec5dec55fc5cd8bd3d7caa1/src/main.ts#L112
An `if` statement works. BUT it doesn't really make sense where you put it since you're accessing activeLeaf a few lines ahead. I think you should wrap the whole function in the `if` statement.

> The workspace file in .obsidian has an ID for leaves, but I can't seem to find anything in the API to interact with this.)

You can use [getLeafById](https://github.com/obsidianmd/obsidian-api/blob/dbfa19ad7aa6557f0ecff962065c3f540bc77e27/obsidian.d.ts#L3066)

https://github.com/gitobsidiantutorial/obsidian-tabs/blob/f6a855079cfda6895ec5dec55fc5cd8bd3d7caa1/src/main.ts#L47 instead of making a function here, just move the contents of the function into the `forEach` block below.

https://github.com/gitobsidiantutorial/obsidian-tabs/blob/f6a855079cfda6895ec5dec55fc5cd8bd3d7caa1/src/main.ts#L97 the name `unloadCleaner` doesn't really make sense since this is a collection of nodes. Maybe call it something like `pinnedTabs`?

https://github.com/gitobsidiantutorial/obsidian-tabs/blob/f6a855079cfda6895ec5dec55fc5cd8bd3d7caa1/src/main.ts#L118 this function name isn't super helpful. Maybe something like `assignStylesToTab`?

gitobsidiantutorial:
> https://github.com/gitobsidiantutorial/obsidian-tabs/blob/f6a855079cfda6895ec5dec55fc5cd8bd3d7caa1/src/main.ts#L47 instead of making a function here, just move the contents of the function into the `forEach` block below.



> https://github.com/gitobsidiantutorial/obsidian-tabs/blob/f6a855079cfda6895ec5dec55fc5cd8bd3d7caa1/src/main.ts#L97 the name `unloadCleaner` doesn't really make sense since this is a collection of nodes. Maybe call it something like `pinnedTabs`?



> https://github.com/gitobsidiantutorial/obsidian-tabs/blob/f6a855079cfda6895ec5dec55fc5cd8bd3d7caa1/src/main.ts#L118 this function name isn't super helpful. Maybe something like `assignStylesToTab`?

liamcain:
Looks good, I think it's merge-able I just spotted a few more stylistic things (and sorry for the delay on my end!!)

---

https://github.com/gitobsidiantutorial/obsidian-tabs/blob/71fc79c66407f1b43f2cdad51ff6e7e6f26469d0/src/main.ts#L75 You could use the 29px from DEFAULT_SETTINGS here instead to avoid duplicating the value

https://github.com/gitobsidiantutorial/obsidian-tabs/blob/71fc79c66407f1b43f2cdad51ff6e7e6f26469d0/src/main.ts#L131, protip, you don't the inline function here, you could just write `childsplittabs.forEach(assignStylesToTab)`

https://github.com/gitobsidiantutorial/obsidian-tabs/blob/71fc79c66407f1b43f2cdad51ff6e7e6f26469d0/src/main.ts#L103
https://github.com/gitobsidiantutorial/obsidian-tabs/blob/71fc79c66407f1b43f2cdad51ff6e7e6f26469d0/src/main.ts#L24
https://github.com/gitobsidiantutorial/obsidian-tabs/blob/71fc79c66407f1b43f2cdad51ff6e7e6f26469d0/src/main.ts#L43-L45
You are accessing `app` throughout your plugin which is getting resolved to `global.app` I think. So it _works_ because Obsidian exposes the `app` object globally for debugging, but plugins shouldn't be accessing it. Instead, you should change this to `this.app`. This will also require binding `this` to the functions that you are passing to event handlers. For example, the `handleOpen` function. I recommend putting the following in your `onload` function:

```ts
this.handleOpen = this.handleOpen.bind(this);
...
```

https://github.com/gitobsidiantutorial/obsidian-tabs/blob/71fc79c66407f1b43f2cdad51ff6e7e6f26469d0/src/main.ts#L193 `parseInt` You should pass in a radix to the parseInt function to avoid it parsing the numbers as non-decimal numbers. So you'll want `parseInt(value.trim(), 10);` I think. Also, I think you want to handle `NaN` here instead of within `updateStyle`. You don't ever want to save bad data for the user. Instead you probably just want to save this as `null`.

gitobsidiantutorial:
> https://github.com/gitobsidiantutorial/obsidian-tabs/blob/71fc79c66407f1b43f2cdad51ff6e7e6f26469d0/src/main.ts#L75 You could use the 29px from DEFAULT_SETTINGS here instead to avoid duplicating the value

> https://github.com/gitobsidiantutorial/obsidian-tabs/blob/71fc79c66407f1b43f2cdad51ff6e7e6f26469d0/src/main.ts#L193 parseInt You should pass in a radix to the parseInt function to avoid it parsing the numbers as non-decimal numbers. So you'll want parseInt(value.trim(), 10); I think. Also, I think you want to handle NaN here instead of within updateStyle. You don't ever want to save bad data for the user. Instead you probably just want to save this as null.

- [x] [Done!](https://github.com/gitobsidiantutorial/obsidian-tabs/blob/aff1848fbe27a019b82ab34203e074e988172544/src/main.ts#L184) I decided to handle both NaN and null as the default value. Saving null and applying it collapses tabs, so I think it's better to handle it with the same bit of code.

> https://github.com/gitobsidiantutorial/obsidian-tabs/blob/71fc79c66407f1b43f2cdad51ff6e7e6f26469d0/src/main.ts#L131, protip, you don't the inline function here, you could just write childsplittabs.forEach(assignStylesToTab)

- [x] [Done!](https://github.com/gitobsidiantutorial/obsidian-tabs/blob/aff1848fbe27a019b82ab34203e074e988172544/src/main.ts#L120)

> https://github.com/gitobsidiantutorial/obsidian-tabs/blob/71fc79c66407f1b43f2cdad51ff6e7e6f26469d0/src/main.ts#L103
https://github.com/gitobsidiantutorial/obsidian-tabs/blob/71fc79c66407f1b43f2cdad51ff6e7e6f26469d0/src/main.ts#L24
https://github.com/gitobsidiantutorial/obsidian-tabs/blob/71fc79c66407f1b43f2cdad51ff6e7e6f26469d0/src/main.ts#L43-L45
You are accessing app throughout your plugin which is getting resolved to global.app I think. So it works because Obsidian exposes the app object globally for debugging, but plugins shouldn't be accessing it. Instead, you should change this to this.app. This will also require binding this to the functions that you are passing to event handlers. For example, the handleOpen function. I recommend putting the following in your onload function:

- [x] [Done!](https://github.com/gitobsidiantutorial/obsidian-tabs/blob/aff1848fbe27a019b82ab34203e074e988172544/src/main.ts#L8) I don't fully understand why the functions called by the event handler needed an explicit binding, though. Is it because the event handler calls back the functions in a different context?

liamcain:
Nice, looks good to me!

> I don't fully understand why the functions called by the event handler needed an explicit binding, though. Is it because the event handler calls back the functions in a different context?

Basically, `this` in javascript is weird and its value depends a lot on the context AND where it was called. When you register your class function as an event handler, it's being invoked from outside the current class instance, aka a different context.

```js
function invokeFunctionOutsideThisContext(someFunction) {
  someFunction();  // It's being invoked here, where `this` is undefined
}

class MyClass {
  constructor() {
    this.foo = 7;
    
    invokeFunctionOutsideThisContext(this.doSomething);
  }

  doSomething() {
    this.foo = 42;
  }
}
```

liamcain (approved):

PR #244, opened 2021-04-10
Add Kindle Highlights plugin

hadynz:
Adding my plugin which is very much still in an alpha release — only tested on Mac at the moment.

Confirming that the package can be successfully downloaded by Obsidian's community plugins flow is akin to confirming that I have a continuous release pipeline confirmed in place. I plan to rapidly iterate on the plugin once this is merged or tested by other members in the community.

## Repo URL

https://github.com/hadynz/obsidian-kindle-plugin

## Release Checklist

- [x] I have tested this on Windows, macOS, and Linux (Mac tested)
  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


lishid:
Hmm just passing by to take a look - could you please add some explanation to the README file as to what the plugin does, how to set it up, and how to use it. I think that would help users test your plugin better.

Given that your plugin's code base is fairly large and there are many components at play, I would recommend also giving a general architecture/design so users aren't surprised that the plugin runs an HTTP server in the background.

Thank you!

hadynz:
Hey @lishid. Thanks for your feedback. I noticed I never acknowledged that I read your message. I've gone back to the drawing table and reengineered the plugin to work in a very different way. I have a new working version that does not use an HTTP server. 

I'll close this Pull Request for now and reopen once I have productionised it, and also flesh out the plugin's Readme.

hadynz:
@lishid I have rewritten the plugin from scratch since you last looked at it. It is ready for a re-review, and to be used by public. 

Key changes made:

* Plugin no longer spins up a HTTP server and stored credentials. Instead it opens a Browser Window and asks users to login to their Amazon account
* Plugin simplified to simply screen scrape https://read.amazon.com to kindle information into Obsidian

lishid:
https://github.com/hadynz/obsidian-kindle-plugin/blob/c623ff2ce7308692261f43e5430afa1ee8182a17/src/scraper/getBookHighlights.ts#L26
You can actually make the code just `return document.body.innerHTML` and await on the result, instead of making an ipc call.

https://github.com/hadynz/obsidian-kindle-plugin/blob/c623ff2ce7308692261f43e5430afa1ee8182a17/src/scraper/getBookHighlights.ts#L31
Probably not needed?

https://github.com/hadynz/obsidian-kindle-plugin#readme
You should explain a bit more on how the login security works. While the user is logged in, the entire app, including other plugins, potentially other vaults, have full access to the user's Amazon account. Unfortunately that's just how the authorization has to work, so it's a trade-off for risk vs convenience.

https://github.com/hadynz/obsidian-kindle-plugin/blob/c623ff2ce7308692261f43e5430afa1ee8182a17/src/fileManager.ts#L25
Unimplemented? If it's intended, I'd suggest changing this to something like `new Notice('Plugin name: file XXX already exists, skipping');`.

Fantastic plugin by the way, the code is very well written and I really don't have much to say ❤️ .

hadynz:
@lishid I am amazed 😍 with your consistency in reviewing the code of every proposed plugin by looking at the Pull Requests of this repo. It is a lovely thing that you are doing and helps tremendously with knowledge sharing and general code craft. Thank you.

I've addressed your comments in release [0.0.6][1] as follows:

* > https://github.com/hadynz/obsidian-kindle-plugin/blob/c623ff2ce7308692261f43e5430afa1ee8182a17/src/scraper/getBookHighlights.ts#L26 
  > You can actually make the code just return document.body.innerHTML and await on the result, instead of making an ipc call.
  
   Thanks so much for the tip! I did not know that is possible. This has [greatly simplified the code][3] for me. I might refactor my scrapers into a base utility class and abstract the electron work later on too.

* Removed `window.webContents.openDevTools();` which was used for testing purposes only

* Added a [comment in README][2] for potential security risk you raised along with one method for a user to address it. I have also added a task in my roadmap to provide an option to always logout a user after ever sync session

* >https://github.com/hadynz/obsidian-kindle-plugin/blob/c623ff2ce7308692261f43e5430afa1ee8182a17/src/fileManager.ts#L25
  > Unimplemented? If it's intended, I'd suggest changing this to something like new Notice('Plugin name: file XXX already exists, skipping');.
 
  I have commented this out for now. I need to solve this scenario (handling syncing of existing files) properly. Raising a `Notice` can easily spam the user with many annoying notifications if they have hundreds of books in their library that they have synced once.

Is it possible that the plugin entry is merged and I can work on the above incrementally? It just means I can start getting user feedback to incorporate in the roadmap too 🙏 

[1]: https://github.com/hadynz/obsidian-kindle-plugin/releases/tag/0.0.6
[2]: https://github.com/hadynz/obsidian-kindle-plugin#how-the-sync-process-work
[3]: https://github.com/hadynz/obsidian-kindle-plugin/commit/2bcdd1f1d67559e28e0488a016770175b35367f0

PR #245, opened 2021-04-10
Adds obsidian-icon-swapper

mgmeyers:

## Repo URL

https://github.com/mgmeyers/obsidian-icon-swapper

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


liamcain:
Very cool! This would have saved you a bunch of time while making California Coast, huh?

---

https://github.com/mgmeyers/obsidian-icon-swapper/blob/6dbebb466b9b76cb3fa2a0f33be304d9f231bd9c/src/iconManager.ts#L69 Why are you doing this conversion when registering the icon instead of just once when you save it? Wouldn't it be simpler/faster to just validate on save, then assume all saved icons are valid, well-formatted SVGs?

https://github.com/mgmeyers/obsidian-icon-swapper/blob/6dbebb466b9b76cb3fa2a0f33be304d9f231bd9c/package.json#L29 Obsidian now exports the YAML package so you shouldn't have to bundle it yourself anymore

https://github.com/mgmeyers/obsidian-icon-swapper/blob/6dbebb466b9b76cb3fa2a0f33be304d9f231bd9c/src/iconManager.ts#L62 Nit: maybe instead of `skipSave` thoughout, invert the boolean to be `save`? Comes down to preference but I think it's a bit more readable without double negatives, i.e. `!skipSave`.

https://github.com/mgmeyers/obsidian-icon-swapper/blob/6dbebb466b9b76cb3fa2a0f33be304d9f231bd9c/src/createIconSetting.ts#L56 it might more readable to use the builder syntax here:

```ts
s.nameEl.createDiv("icon-swapper-container", container => {
  container.createDiv("icon-swapper-icon");
  container.createDiv("icon-swapper-name")
});
```

https://github.com/mgmeyers/obsidian-icon-swapper/blob/6dbebb466b9b76cb3fa2a0f33be304d9f231bd9c/src/createIconSetting.ts#L50 why do you need to call `setIcon` here?

https://github.com/mgmeyers/obsidian-icon-swapper/blob/6dbebb466b9b76cb3fa2a0f33be304d9f231bd9c/src/main.ts#L145 Should this also accept `.svg` or maybe use the MIME-type?

mgmeyers:
Thanks for the thorough review @liamcain! I appreciate it. Here are my updates: https://github.com/mgmeyers/obsidian-icon-swapper/commit/bd8298175feaea1d2ea88bab9ff225ad2dfc177d

> Very cool! This would have saved you a bunch of time while making California Coast, huh?

Yeah, definitely. I'm kicking myself for not figuring this out sooner!

> https://github.com/mgmeyers/obsidian-icon-swapper/blob/6dbebb466b9b76cb3fa2a0f33be304d9f231bd9c/src/iconManager.ts#L69 Why are you doing this conversion when registering the icon instead of just once when you save it? Wouldn't it be simpler/faster to just validate on save, then assume all saved icons are valid, well-formatted SVGs?

Good point! I still have to validate imported icons, but saved icons can be trusted.

> https://github.com/mgmeyers/obsidian-icon-swapper/blob/6dbebb466b9b76cb3fa2a0f33be304d9f231bd9c/package.json#L29 Obsidian now exports the YAML package so you shouldn't have to bundle it yourself anymore

From what I can tell, Obsidian only exports `parseYaml`, and I also need to convert JSON to YAML. https://github.com/obsidianmd/obsidian-api/blob/master/obsidian.d.ts#L1926

> https://github.com/mgmeyers/obsidian-icon-swapper/blob/6dbebb466b9b76cb3fa2a0f33be304d9f231bd9c/src/iconManager.ts#L62 Nit: maybe instead of `skipSave` thoughout, invert the boolean to be `save`? Comes down to preference but I think it's a bit more readable without double negatives, i.e. `!skipSave`.

✅ did some refactoring related to this.

> https://github.com/mgmeyers/obsidian-icon-swapper/blob/6dbebb466b9b76cb3fa2a0f33be304d9f231bd9c/src/createIconSetting.ts#L56 it might more readable to use the builder syntax here:
> 
> ```ts
> s.nameEl.createDiv("icon-swapper-container", container => {
>   container.createDiv("icon-swapper-icon");
>   container.createDiv("icon-swapper-name")
> });
> ```

Ah, yeah. I'm still getting used to these helper functions. I've refactored a ton of code to utilize these!

> https://github.com/mgmeyers/obsidian-icon-swapper/blob/6dbebb466b9b76cb3fa2a0f33be304d9f231bd9c/src/createIconSetting.ts#L50 why do you need to call `setIcon` here? **Edit:** nevermind, I think I see. But shouldn't this leverage the IconManager instead?

So, the `setIcon` used here is Obsidian's `setIcon` that injects the desired icon into the DOM. I use this to display the icon to the user in the settings pane.

> https://github.com/mgmeyers/obsidian-icon-swapper/blob/6dbebb466b9b76cb3fa2a0f33be304d9f231bd9c/src/main.ts#L145 Should this also accept `.svg` or maybe use the MIME-type?

So, this file is actually a YAML map of `{ [iconName: string]: string }`. I've changed the file type to `yml` to be more explicit what the file actually is. The importer doesn't support individual SVGs, but YAML file exported by this plugin.

liamcain:
> From what I can tell, Obsidian only exports parseYaml, and I also need to convert JSON to YAML. https://github.com/obsidianmd/obsidian-api/blob/master/obsidian.d.ts#L1926

I think in a recent release Licat added exports for the `yaml` package specifically, but I can't find any docs about it 🤔. Maybe he can shed some light on it when he has a chance. 

> So, the setIcon used here is Obsidian's setIcon that injects the desired icon into the DOM. I use this to display the icon to the user in the settings pane.

Ohh, I was actually mixing up `setIcon` with `addIcon` and assumed this was replacing the icon in Obsidian's icon collection 🤦  my bad.

> So, this file is actually a YAML map of { [iconName: string]: string }. I've changed the file type to yml to be more explicit what the file actually is. The importer doesn't support individual SVGs, but YAML file exported by this plugin.

I should have looked at the `examples` folder before; I see now.

---

Changes look good to me!

liamcain (approved):

PR #246, opened 2021-04-13
Add Tasks plugin

schemar:

Hey there, thank you for taking the time to review this. I am happy to make any required changes to the plugin before this can be merged. Just let me know. Thanks!

## Repo URL

https://github.com/schemar/obsidian-tasks

## Release Checklist

  - [x] Linux 
  - [x] Windows
  - [x] iOS
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1235,5 +1235,13 @@
         "description": "Fight the forgetting curve & note aging by reviewing notes using spaced repetition on Obsidian.md",
         "author": "Stephen Mwangi",
         "repo": "st3v3nmw/obsidian-spaced-repetition"
+    },
+    {
+        "id": "obsidian-tasks-plugin",
+        "name": "Tasks",
+        "author": "Martin Schenck",
+        "description": "Task management for Obsidian",
+        "repo": "schemar/obsidian-tasks",
+        "branch": "main"
     }
 ]


whitenoisedev:
Hi and thanks for your work.
I have two observations regarding this plugin.

1) In markdown, it's common to have tasks/checklists as
- [ ] taskname (for task todo-uncheked)
- [x] Task name (for a task done-checkek)
I understand that you probably come from an org-mode background, but I still think it would be nice if you supported the most common notation used.

2) @blacksmithgu 's dataview wanted to do something similar for tasks. It's already able to parse files and pull tasks by line but it hasn't any filtering capabilities. Something analogous of what you did I believe is dataview roadmap. I am not sure if there is room for collaboration or even simple non-overlapping of efforts.

Thanks for your work.

schemar:
Thank you for the review @whitenoisedev !

Regarding your points:

1. It was a deliberate design decision to not use the default checklist items like `- [ ]`. Not every checklist item I have is a task I want to track. My shopping list, for example, shouldn't be considered a list of todos. It is not trivial to support `- [ ]` in addition, as I would also need to remove the default checkbox in preview mode and replace it with my own. :thinking: In any case, let me think about making it a configurable setting. I need to think about the effort more :monocle_face: 
2. I will get in touch with @blacksmithgu. My initial gut feeling is that the dataview plugin's scope has only a very slim overlap with task management.

Thanks!

schemar:
After a discussion on discord it seems there is no way around supporting regular `- [ ]` syntax via a config toggle eventually. I would want to get the core task management done first, though (e.g. recurring tasks and task priorities). What do you think?

schemar:
1. I changed the task identifier from `- TODO` and `- DONE` to `- [ ] #task` and `- [x] #task`. The `#task` tag is a setting, but I don't expose it yet, as there is no settings menu, yet. After the discussion on dicsord and with other people, I now think this is the best solution. It is a common checkbox and at the same time it uses a tag. I pushed the changes, but haven't made a release, yet. I need to do more testing and bring my whole vault to the new system. Probably with `sed`, posting the `sed` command in the release notes as well so others can migrate easily. I also haven't updated the README, yet. What do you think?
2. I reached out to blacksmithgu in his dataview plugin. He also thinks that task management is not something that the dataview plugin should do, but he considers exposing his query engine to other plugins, which could be a nice way for me to do more complex queries if I need that.
3. I have the feeling I am using a very "hacky" approach to identifying tasks in the rendered markdown in the post processor. Also: my context child always immediately "unloads", so right now I don't properly unload as I don't do anything. These are probably reasons for you to not merge this plugin. I am working on that.

Thank you for all the time you put into this! :pray:

schemar:
I realized I don't have the time or energy to make this proper. I didn't yet learn the obsidian API properly, so things are not as efficient as they could be. At the same time some other things were done a little hacky.

I will use it as a personal replacement until a good tasks plugin comes around. I will close this PR. Thank you for the time you took.

whitenoisedev:
I was happy to talk with you this morning.
I want to thank you again for using Obsidian and we appreciate the time you put in this so far. 

 @lishid if possible, could you still have a look at the code of this plugin? @schemar would like to have some feedback, especially about parsing.

PR #247, opened 2021-04-15
Add obsidian-open-link-with

MamoruDS:

## Repo URL

https://github.com/MamoruDS/obsidian-open-link-with

## Release Checklist

-   [x] I have tested this on Windows, macOS and Linux
-   [x] Github release contains all required files
    -   [x] main.js
    -   [x] manifest.json
    -   styles.css (optional)
-   [x] Github release name matches the exact version number specified in your manifest.json _(**Note**: Use the exact version number, don't include a prefix `v`)_
-   [x] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
-   [x] README clearly describes the plugins purpose and provides clear usage instructions.


liamcain:
https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/main.ts#L21 maybe rename your plugin class?

https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/main.ts#L42 `!==` instead of `!=`

https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L54 no need for a ternary, could just be `failed = code === 0;`

https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L57 why the setTimeout here?

https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L63 the spread here looks unnecessary. You can just do `cmd.map(() ...` on line 65 instead. Also consider using more verbose variable names. It's hard keeping track of all the line letter var names.

https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L109 spread looks unnecessary here

https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L110 your `test` function is async but all the usages are synchronous. Maybe just make the function synchronous instead?

https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L122 style-nit: consider wrapping these `if` blocks in parentheses. Having so many one-line `if` blocks like this can be pretty dangerous and it's easy to introduce a bug by accident.

https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L11 you can use the built-in nodejs [`os`](https://nodejs.org/api/os.html) package for this instead

https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L40 I wonder if it would be simpler to use something like the [`open`](https://github.com/sindresorhus/open) package instead of handling this functionality yourself.

https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/main.ts#L139 what's happening here?

MamoruDS:
Hello @liamcain, thanks for the review! I appreciate it, [here](https://github.com/MamoruDS/obsidian-open-link-with/commit/de7eaf6ba3bbb1ed8d6fc1fc36e1a1c7c95ada50) are my updates:

> https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/main.ts#L21 maybe rename your plugin class?

I've changed plugin class name to `OpenLinkPlugin`.

> https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/main.ts#L42 `!==` instead of `!=`

I've updated this.

> https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L54 no need for a ternary, could just be `failed = code === 0;`

I've updated this.

> https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L57 why the setTimeout here?

The reason for setting a `setTimeout` here is: if the browser was running, `url` will be opened in the existing session then `child_process` will exit; if the browser was not running, `child_process` will not exit until the quit the browser. So if `child_process` exits with non-zero or still running after a little delay, I treat these two cases as not failed.

> https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L63 the spread here looks unnecessary. You can just do `cmd.map(() ...` on line 65 instead. Also consider using more verbose variable names. It's hard keeping track of all the line letter var names.

I've fixed this and using more meaningful variable names instead of letters.

> https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L109 spread looks unnecessary here

Sure, it is pointless to shallow copy here, I've fixed it.

> https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L110 your `test` function is async but all the usages are synchronous. Maybe just make the function synchronous instead?

For now, all `test` functions are synchronous, but I was thinking `test` return `Promised<boolean>` could leave more flexibility.

> https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L122 style-nit: consider wrapping these `if` blocks in parentheses. Having so many one-line `if` blocks like this can be pretty dangerous and it's easy to introduce a bug by accident.

I've replaced all these one-line `if`.

> https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L11 you can use the built-in nodejs [os](https://nodejs.org/api/os.html) package for this instead

I've removed related functions and type defines and using built-in `os` for getting platform.

> https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/open.ts#L40 I wonder if it would be simpler to use something like the [open](https://github.com/sindresorhus/open) package instead of handling this functionality yourself.

Since this is a relatively simple implementation, and I haven't encountered any issues on several tested platforms yet, it makes sense to handle these utilities by myself with zero package dependencies.

> https://github.com/MamoruDS/obsidian-open-link-with/blob/9f8600ba641367f961169cb773d6ec5b53525368/src/main.ts#L139 what's happening here?

So, this is actually a simple implementation of the debounce function, `ck` is the id of the timer, I've removed this and using obsidian built-in debouncer instead.

ericaxu:
@lishid Can you take another look please? Thanks!

lishid:
https://github.com/MamoruDS/obsidian-open-link-with/blob/221e9b38f8cfe973156d8c3458124bfb8d0499ce/src/main.ts#L34
I'd suggest avoiding long-running tasks in your plugin's onload, because it will block the entire app's loading sequence. You should probably do this without awaiting (run it in parallel), or defer to running it later.

https://github.com/MamoruDS/obsidian-open-link-with/blob/221e9b38f8cfe973156d8c3458124bfb8d0499ce/src/main.ts#L45
I'm not sure about the semantics of preventDefault in an async function, but I'd recommend keeping your function non-async and wrap the async parts in an anonymous async IIFE.

PR #248, opened 2021-04-15
Add Related Notes Finder plugin to releases

lifegems:

## Repo URL

## Repo URL

https://github.com/lifegems/obsidian-related-notes-finder

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


pjeby:
A few comments:

* The random note functionality appears identical to Obsidian's built-in random note plugin (which can be activated from the "Core Plugins" settings page)
* The path functionality seems to be a duplicate of the existing "Show Current File Path" and "Active note to window title" community plugins
* The reading time appears to be a duplicate of the "Reading Time" plugin

Given all this, it might be better to rename this plugin to "Related Notes Link Finder" or something of that sort (and drop all the other functionality), so that potential users searching for a plugin to find related notes, or similar keywords would be able to find this plugin.  A name like "Obsidian Plus" is unlikely to let anybody find your plugin in order to use it -- entirely aside from any trademark issues or making it sound like an official product of the Obsidian developers.

lifegems:
@pjeby Thanks for those pointers and for the referrals for the related plugins, didn't even know those existed.

Will update and send back shortly. Thanks!

lifegems:
@pjeby The related repo and this PR have been updated. Thanks again for your suggestions. 👍

lishid:
https://github.com/lifegems/obsidian-related-notes-finder/blob/5bf0a6e55e8bcd6100aa41b6db68d8a64754252b/main.ts#L17
Here I would recommend renaming MyPlugin, MyPluginSettings, etc to your actual plugin name :)

https://github.com/lifegems/obsidian-related-notes-finder/blob/5bf0a6e55e8bcd6100aa41b6db68d8a64754252b/main.ts#L128
You can use our helper function `createEl/createDiv` here to make things look cleaner:
```ts
parentEl.createDiv({cls: 'classes', text: 'text'});
```

https://github.com/lifegems/obsidian-related-notes-finder/blob/5bf0a6e55e8bcd6100aa41b6db68d8a64754252b/main.ts#L28
https://github.com/lifegems/obsidian-related-notes-finder/blob/5bf0a6e55e8bcd6100aa41b6db68d8a64754252b/main.ts#L31
Here you're getting the active file and also the active leaf, without verifying that you're working with a markdown file.

I'd recommend you use the latest API for working with markdown views:

- Workspace.getActiveViewOfType(MarkdownView), which will return a MarkdownView or null if the current view isn't markdown.
- MarkdownView.editor, which is our new editor replacement for `cmEditor`, which also works on the mobile app. It doesn't have everything on CodeMirror but should be pretty close.

lishid:
https://github.com/lifegems/obsidian-related-notes-finder/blob/c5dad30754b2288996d9a402d262ce0b1b8c4a26/main.ts#L119
You should create them from the correct parent in the first place, like `let noteLink = noteContainer.createEl('p')`, then it will be automatically appended to the parent.

https://github.com/lifegems/obsidian-related-notes-finder/blob/c5dad30754b2288996d9a402d262ce0b1b8c4a26/main.ts#L27
This should be called `activeView`.

https://github.com/lifegems/obsidian-related-notes-finder/blob/c5dad30754b2288996d9a402d262ce0b1b8c4a26/main.ts#L30
You should use `cachedRead` here instead of `read` for performance.

lifegems:
Thanks for the feedback @lishid, I've now added these changes in.

PR #249, opened 2021-04-15
Add Auto Link Title

zolrath:

https://github.com/zolrath/obsidian-auto-link-title

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


lishid:
https://github.com/zolrath/obsidian-auto-link-title/blob/987b49d94c38d05a71e9bfa352935ea8b86c3438/main.ts#L163
Here you should be using `this.app.workspace.getActiveViewOfType(MarkdownView)` instead of casting to any. Make sure you test for truthyness in case the active view is null or isn't a markdown view.

https://github.com/zolrath/obsidian-auto-link-title/blob/987b49d94c38d05a71e9bfa352935ea8b86c3438/main.ts#L164
Here you should be using `view.editor` instead of `view.sourceMode.cmEditor`.

https://github.com/zolrath/obsidian-auto-link-title/blob/987b49d94c38d05a71e9bfa352935ea8b86c3438/main.ts#L2
Your plugin will error out here on the mobile apps, causing it to not load. I think while the clipboard API isn't available on mobile, the rest of your plugin can still work. What you can do is use `window.require('electron').clipboard.readText` right in the function it's used, and that will just error out on mobile without breaking the entire plugin.

https://github.com/zolrath/obsidian-auto-link-title/blob/987b49d94c38d05a71e9bfa352935ea8b86c3438/main.ts#L110
Please add a warning to your README.md to indicate that you are using this method to bypass CORS (and that your plugin will actually download the webpage to find the title). This helps privacy sensitive users to know what to expect.

lishid:
Ah indeed it doesn't support markers. I guess if you don't want to try and support mobile that's fine. There are probably other ways to make placeholder links that you can replace easily later, such as using a unique random ID and searching for it after.

zolrath:
Ok! I've made the changes aside from moving from `cmEditor`.
I've swapped it to `isDesktopOnly` for now and will reassess once `Editor` expands in features.
I've also swapped to `checkCallback` to allow the default paste to take over when the clipboard isn't a URL or is a URL that links to an image.

lishid:
I took a look at the way you're attempting to find the URL - I would highly recommend to use a regex on the line text instead of using the token information and findWordAt - the token information is tied to the CodeMirror 5 implementation and won't be available on Mobile and/or web, probably ever, because CodeMirror 6 uses a completely different parse & token system.

I am also not sure if we will ever attempt to make Marker compatible with CodeMirror 6. It does have a concept of marker, but is mostly for internal uses and would need an actual view extension to use markers.

Editor is meant to be a common API to unify CM5 and CM6 for the common things such as getting and setting selection, getting and replacing text, etc. It is not meant to include the full CM5's API for all the advanced use cases, most of which don't have an exact equivalent in CM6 anyway.

In the end, it's up to you. Restricting your plugin for desktop use only is an ok solution too.

lishid:
Let me know if you want it merged it the current state. The code looks fine at this point 👍

zolrath:
I'm up for a merge at this point! 
I'll look towards adapting to Editor in the future as I'm sure I'll end up wanting it on Mobile but I'll leave that as a task for future me.

PR #250, opened 2021-04-17
Add tokuhirom/obsidian-stopwatch-plugin

tokuhirom:

## Repo URL

## Repo URL

https://github.com/tokuhirom/obsidian-stopwatch-plugin

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/97c0d008e333404f17c55e64c82cd65e04af9a9d/main.ts#L15
https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/97c0d008e333404f17c55e64c82cd65e04af9a9d/main.ts#L3 
Maybe a better name than `MyPlugin`?

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/97c0d008e333404f17c55e64c82cd65e04af9a9d/main.ts#L26 I don't think you meant to leave this in, right?

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/97c0d008e333404f17c55e64c82cd65e04af9a9d/main.ts#L57 looks like this can be removed

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/97c0d008e333404f17c55e64c82cd65e04af9a9d/main.ts#L37 The commands will fail if `this.view` doesn't exist (user closes the pane, for example). Maybe use `checkCallback` and hide these commands if the view doesn't exist?

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/97c0d008e333404f17c55e64c82cd65e04af9a9d/main.ts#L55-L60 Instead of doing this, you can use `this.app.workspace.onLayoutReady(() => this.initLeaf());` to accomplish the same thing.

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/97c0d008e333404f17c55e64c82cd65e04af9a9d/main.ts#L86 Maybe a better name than "sampleplugin" for this?

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/97c0d008e333404f17c55e64c82cd65e04af9a9d/main.ts#L153 you can remove this function completely

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/97c0d008e333404f17c55e64c82cd65e04af9a9d/main.ts#L235 might be easier to use `momentjs` to get the timestamp for this.

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/97c0d008e333404f17c55e64c82cd65e04af9a9d/main.ts#L107 you should put a few safeguards here to avoid saving the value as `NaN` or `0`. Maybe set a minimum refresh to something sensible.

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/97c0d008e333404f17c55e64c82cd65e04af9a9d/main.ts#L158 you shouldn't need the casting here.

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/97c0d008e333404f17c55e64c82cd65e04af9a9d/main.ts#L169 you can do this within the `createDiv()` function by passing in `{ cls: 'stopwatch-time'}` in the config options.

tokuhirom:
@liamcain 
Thank you for your review!! (I'm not familiar with obsidian plugin development...  This is my first plugin for obsidian.  🙇 )
I applied all of your reviews.

liamcain:
Nice looks good! Just a few small things I spotted now.

---

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/6834fc9d3d093e80336edb7c789612541acceaf3/package.json#L25 Obsidian exports `moment` so you don't need to include it in your JS bundle. There are some instructions for how to set it up and and keep typescript happy here: <https://liamca.in/Obsidian/API+FAQ/third-party/use+momentjs>

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/6834fc9d3d093e80336edb7c789612541acceaf3/main.ts#L37 checkCallback is a little weird. You don't want to specify both a callback and a checkcallback. You want to move your execution into the checkcallback function. There's some explanation and sample code you can follow here: <https://liamca.in/Obsidian/API+FAQ/commands/what+does+checking+mean+in+checkCallback>

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/6834fc9d3d093e80336edb7c789612541acceaf3/main.ts#L66 there's probably no need to pass the model in from the plugin. You could just instantiate in the view constructor.

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/6834fc9d3d093e80336edb7c789612541acceaf3/main.ts#L119 from a user perspective, this is a bit confusing since their input will not be saved for a variety of reasons and they won't know that. Maybe try to display an error message when the input is invalid here.

tokuhirom:
> https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/6834fc9d3d093e80336edb7c789612541acceaf3/package.json#L25 Obsidian exports moment so you don't need to include it in your JS bundle. There are some instructions for how to set it up and and keep typescript happy here: https://liamca.in/Obsidian/API+FAQ/third-party/use+momentjs
I'm using `moment-duraiton-format` for formatting duration. Since moment does not support duration formatting. But `moment-duration-format` change the `moment` object itself. I guess, this plugin shouldn't change the global `moment` object. 

> https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/6834fc9d3d093e80336edb7c789612541acceaf3/main.ts#L37 checkCallback is a little weird. You don't want to specify both a callback and a checkcallback. You want to move your execution into the checkcallback function. There's some explanation and sample code you can follow here: https://liamca.in/Obsidian/API+FAQ/commands/what+does+checking+mean+in+checkCallback

> https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/6834fc9d3d093e80336edb7c789612541acceaf3/main.ts#L66 there's probably no need to pass the model in from the plugin. You could just instantiate in the view constructor.
Thanks. applied.

> https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/6834fc9d3d093e80336edb7c789612541acceaf3/main.ts#L119 from a user perspective, this is a bit confusing since their input will not be saved for a variety of reasons and they won't know that. Maybe try to display an error message when the input is invalid here.
Ah, I forgot it. I implemented the notification.

liamcain:
Hi! Sorry for the long delay.

> I'm using moment-duraiton-format for formatting duration. Since moment does not support duration formatting. But moment-duration-format change the moment object itself. I guess, this plugin shouldn't change the global moment object

I think you're probably safe to modify the global `window.moment` but it doesn't matter too much either way.

Nice job on the clean-up. Code looks great!

liamcain (approved):

community-plugins.json
```diff
@@ -1281,13 +1281,21 @@
         "description": "An interface to experiment with python scripts and more.",
         "author": "Cristian Vasquez",
         "repo": "cristianvasquez/obsidian-lab"
+        "branch": "main"
```
liamcain, line 1284: Looks like an issue with the rebase. Can you remove this? It should be on the `kanban` section instead.

liamcain (changes requested):

liamcain (approved):

lishid:
Apologies on the late review.

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/036af986d12a1dc0a5077c3b6560044d477bc6e7/main.ts#L30
You shouldn't keep a copy of views on your plugin. The view could have been detached/removed, and this will cause your plugin to hold onto an unused copy. Instead, you should always use the result of `getLeavesOfType(MY_VIEW_TYPE)`.

https://github.com/tokuhirom/obsidian-stopwatch-plugin/blob/036af986d12a1dc0a5077c3b6560044d477bc6e7/main.ts#L60
You should always return whether your callback can perform an action, even if `checking` is false. Same for the other command,

tokuhirom:
@lishid 
Thanks. I fixed the issues!

PR #252, opened 2021-04-19
Add Tidy Footnotes

charliecm:

## Repo URL

https://github.com/charliecm/obsidian-tidy-footnotes

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
Looks like you're using CodeMirror 5's API, which means your plugin can only run on desktop. As such, you should mark your plugin as `isDesktopOnly: true` https://github.com/charliecm/obsidian-tidy-footnotes/blob/df7a3d432b739b64e6a1575c94c78c95bc32b661/manifest.json#L9

charliecm:
@lishid Thanks for reviewing. I'm using CodeMirror for [tests only](https://github.com/charliecm/obsidian-tidy-footnotes/blob/main/src/tidyFootnotes.test.ts), so it's not included in `main.js` (in run-time, it's using Obsidian's global `CodeMirror` instance). Would you recommend a different approach in using the API for testing without needing to set `isDesktopOnly` to true?

lishid:
https://github.com/charliecm/obsidian-tidy-footnotes/blob/df7a3d432b739b64e6a1575c94c78c95bc32b661/src/main.ts#L14
Here you are using `sourceMode.cmEditor` which is CodeMirror 5. You are also using it here: https://github.com/charliecm/obsidian-tidy-footnotes/blob/df7a3d432b739b64e6a1575c94c78c95bc32b661/src/tidyFootnotes.ts#L41

If you are able to use our Editor interface, which bridges CM5 and CM6 functions together, your plugin will work on mobile as well: https://github.com/obsidianmd/obsidian-api/blob/068ff3052f0641baff90856dcb420dc2721680c9/obsidian.d.ts#L682
To start, you'll need to use `MarkdownView.editor` instead of `MarkdownView.sourceMode.cmEditor`.

charliecm:
I'll certainly try to use the new Editor interface when it's ready. Is there an API in Obsidian for getting the indentation of a line? I'm currently using CodeMirror's `StringStream.indentation()`. Otherwise, I may need to know the user's tab preference to check for indents reliably.

lishid:
Pretty sure you can just use a regex to find indentations; you aren't even using the number, just testing that it's larger than 0.

charliecm:
Thanks for the suggestion. I'm using the [regex](https://github.com/charliecm/obsidian-tidy-footnotes/blob/a0dc20aa6bd351f540f401b0d66aed70f955a5dc/src/tidyFootnotes.ts#L41) so the code no longer references CodeMirror. Also, changed `isDesktopOnly` to `true` for now. Will change it to `false` once it's been tested on mobile and is using the Editor API.

lishid:
Cool - the Editor API is just a bridge to CodeMirror 5 on desktop, so you can just swap to use it right away since all the functions you're currently using are already supported (they end up calling cmEditor anyway). Let me know if you want the plugin merged now or after you make the change.

charliecm:
Ah, I thought API v0.12.0 is mapped to the latest Obsidian stable build. I've updated the code to use the Editor API, so it's ready to go 👍

PR #254, opened 2021-04-19
Add Metatable plugin

arnau:

## Repo URL

https://github.com/arnau/obsidian-metatable/

## Release Checklist

- [x] I have tested this on macOS, and Linux
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
> At this time the default theme doesn't look as good.
I think most of the people trying out the plugin would be on some kind of default theme, so I'd recommend getting that improved.

https://github.com/arnau/obsidian-metatable/blob/cb10432169af9c58e9186b56a61174a9dcc11023/src/plugin.ts#L20
Please don't use innerHTML, especially with user inputs. It's susceptible to DOM injection attacks.

https://github.com/arnau/obsidian-metatable/blob/cb10432169af9c58e9186b56a61174a9dcc11023/src/plugin.ts#L32
Don't need to await here.

https://github.com/arnau/obsidian-metatable/blob/cb10432169af9c58e9186b56a61174a9dcc11023/src/plugin.ts#L31
This will fail if your plugin was disabled then enabled a second time. You should check that it's not defined first.

Actually, I don't think the use of a custom element is warranted since this just need to render a simple table for a frontmatter object. Is there any reason to prefer custom elements instead of using a simple js class and generating some DOM elements inside, like we do with `Modal`, `FuzzySuggestModal`, `Notice`, etc?

To me, the advantage of using a simple class (to manage internal DOM elements) is that it's self contained and doesn't pollute the global space (You can't even un-define or re-define custom elements). You can pass the parameters directly to the constructor rather than having to serialize them into HTML attributes, and you can clean up after them cleanly.

arnau:
> > At this time the default theme doesn't look as good.
> I think most of the people trying out the plugin would be on some kind of default theme, so I'd recommend getting that improved.

Agreed. It's on my list. I'll prioritise it.

> https://github.com/arnau/obsidian-metatable/blob/cb10432169af9c58e9186b56a61174a9dcc11023/src/plugin.ts#L20
> Please don't use innerHTML, especially with user inputs. It's susceptible to DOM injection attacks.

Sure, will do.

> https://github.com/arnau/obsidian-metatable/blob/cb10432169af9c58e9186b56a61174a9dcc11023/src/plugin.ts#L32
> Don't need to await here.

Thanks 👍 

> https://github.com/arnau/obsidian-metatable/blob/cb10432169af9c58e9186b56a61174a9dcc11023/src/plugin.ts#L31
> This will fail if your plugin was disabled then enabled a second time. You should check that it's not defined first.

Oh of course, thanks!

> Actually, I don't think the use of a custom element is warranted since this just need to render a simple table for a frontmatter object. Is there any reason to prefer custom elements instead of using a simple js class and generating some DOM elements inside, like we do with `Modal`, `FuzzySuggestModal`, `Notice`, etc?

The reason is isolation. The fact that I'm using a table now is an implementation detail. It also signals to the user the expected boundaries for customsiation, it keeps the plugin JavaScript contained within the shadowDom scope, etc. It is true that a `class` would suffice but I'd rather stick to a custom element.

> To me, the advantage of using a simple class (to manage internal DOM elements) is that it's self contained and doesn't pollute the global space (You can't even un-define or re-define custom elements). You can pass the parameters directly to the constructor rather than having to serialize them into HTML attributes, and you can clean up after them cleanly.

It is indeed a limitation that you can't un-define a custom element. On the parameters point though, you have 'attributes' which yes, you have to serialise/deserialise like with any other element and you have object methods, like the one I use to add the forntmatter data.

Are you asking me to remove the use of custom elements as an Obsidian policy? Otherwise I'd like to stick to it.

arnau:
Hi, I released [version 0.3.0](https://github.com/arnau/obsidian-metatable/releases/tag/0.3.0) where:

- [ ] better default theme
- [ ] removes use of `innerHTML`

lishid:
Oh yeah, shadows are fine if that's what you intend to limit user customizations. I will review the changes in a bit.

arnau:
Hi, I released [version 0.5.0](https://github.com/arnau/obsidian-metatable/releases/tag/0.5.0). Hopefully everything checks out now.

PR #257, opened 2021-04-20
new plugin `koncham-workspace` : workspace enhancements, including vertical tabs

manogna4:

## Repo URL

https://github.com/manogna4/obsidian-koncham-workspace/

## Release Checklist

- [x] I have tested this on 
  - [x] Windows
  - [ ] macOS
  - [ ] Linux
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
Apologies for the late review.

https://github.com/manogna4/obsidian-koncham-workspace/blob/main/main.js
Might want to add this to the .gitignore file to avoid committing it out.

https://github.com/manogna4/obsidian-koncham-workspace/blob/a80fa4504c37a28da45fd72c270a7ab4588763ac/main.ts#L8
Don't keep a reference to views - you could end up holding a dead unloaded view. See next point
https://github.com/manogna4/obsidian-koncham-workspace/blob/a80fa4504c37a28da45fd72c270a7ab4588763ac/main.ts#L22
Move all these event handlers onto your view's onload() function, including the handleChange and handleDelete functions.

https://github.com/manogna4/obsidian-koncham-workspace/blob/a80fa4504c37a28da45fd72c270a7ab4588763ac/main.ts#L28
Sounds like you're looking for `Workspace.onLayoutReady` here.

https://github.com/manogna4/obsidian-koncham-workspace/blob/a80fa4504c37a28da45fd72c270a7ab4588763ac/main.ts#L84
Don't use getViewState, which causes the entire view to be serialized (it's inefficient). Instead, use `leaf.view.getViewType()`

https://github.com/manogna4/obsidian-koncham-workspace/blob/a80fa4504c37a28da45fd72c270a7ab4588763ac/main.ts#L139
Same as the above.

https://github.com/manogna4/obsidian-koncham-workspace/blob/a80fa4504c37a28da45fd72c270a7ab4588763ac/main.ts#L71
https://github.com/manogna4/obsidian-koncham-workspace/blob/a80fa4504c37a28da45fd72c270a7ab4588763ac/main.ts#L77
https://github.com/manogna4/obsidian-koncham-workspace/blob/a80fa4504c37a28da45fd72c270a7ab4588763ac/main.ts#L83
https://github.com/manogna4/obsidian-koncham-workspace/blob/a80fa4504c37a28da45fd72c270a7ab4588763ac/main.ts#L131
Don't cast this to `any`, it should be typed `WorkspaceLeaf`.

https://github.com/manogna4/obsidian-koncham-workspace/blob/a80fa4504c37a28da45fd72c270a7ab4588763ac/main.ts#L150
Here you don't need to store this value as an attribute. You can simply define a inline variable `let leafIndex = n;` and use it in the onClick event.

manogna4:
thanks for the detailed reply!

I have done everything and updated the release except:

#### 1. leaf index as attribute
> https://github.com/manogna4/obsidian-koncham- 
workspace/blob/a80fa4504c37a28da45fd72c270a7ab4588763ac/main.ts#L150
Here you don't need to store this value as an attribute. You can simply define a inline variable let leafIndex = n; and use it in the onClick event.

For future features in this plugin, the order of the leafs in the sidebar may be different than the order of leaves in the main area. For example, alphabetical in the view vs however the notes were opened in the main area. This is why I am storing the leaf-index on the html element. Is there a better way of doing this?

#### 2. add main.js to .gitignore
>https://github.com/manogna4/obsidian-koncham-workspace/blob/main/main.js
Might want to add this to the .gitignore file to avoid committing it out.
can you explain this a bit further, or point to a link that does? 

My understanding is that if I include `main.js` in gitignore, it doesn't get committed to the repo? I thought someone who didn't want to install the plugin through the community portal (or couldn't because it's not accepted yet) could always copy the `main.js` from my plugin into the proper folder inside their vault folder, and the plugin would still work. Won't excluding `main.js`prevent this?

lishid:
> This is why I am storing the leaf-index on the html element. Is there a better way of doing this?

Create a variable for it with `let` or `const` and it will be bound to the closure of that block. Since you're hooking the click even directly, the variable will always remain the one you assigned along with the HTML element.

> My understanding is that if I include main.js in gitignore, it doesn't get committed to the repo? I thought someone who didn't want to install the plugin through the community portal (or couldn't because it's not accepted yet) could always copy the main.js from my plugin into the proper folder inside their vault folder, and the plugin would still work. Won't excluding main.jsprevent this?

You should send those users to find the download from your releases instead of grabbing a potentially work-in-progress version from your git repo.

manogna4:
> Create a variable for it with let or const and it will be bound to the closure of that block. Since you're hooking the click even directly, the variable will always remain the one you assigned along with the HTML element.

I will look into this approach. If you have the time, can you point me to a plugin that uses this approach? 
The thing is, I would like to access the leaf that the item in the view represents from other blocks too. For instance, through a right-click on the item in the view. I might add options to close the file, or  move it to the sidebars. I am not sure how I can do this if leaf the item in the view represents is limited to that block. I am also thinking of accessing that information from other plugins. does the variable approach allow that?

> You should send those users to find the download from your releases instead of grabbing a potentially work-in-progress version from your git repo.

Ah, I understand what you mean now. It's certainly a better approach for other users.
But, if it's the same to you, I'd like like leave it there just in case I want to make some quick edits on another computer for myself without installing npm etc...

lishid:
> I will look into this approach. If you have the time, can you point me to a plugin that uses this approach?
The thing is, I would like to access the leaf that the item in the view represents from other blocks too. For instance, through a right-click on the item in the view. I might add options to close the file, or move it to the sidebars. I am not sure how I can do this if leaf the item in the view represents is limited to that block. I am also thinking of accessing that information from other plugins. does the variable approach allow that?

Sure, that approach will not affect whether or not you can do those things. I'm literally talking about assigning it to an inline variable instead of assigning it on the HTML element attribute and having to parse it out of the attribute afterwards.

```ts

for (let i = 0; i < array.length; i++) {
  let el = createEl('div');

  // Store it in an inline variable
  let n = i;
  // Instead of on the element
  // el.setAttribute('test', i);
  el.addEventListener('click', () => {
    // Use it in an enclosure that's on the same or in a sub level as the location you declared the variable.
    console.log(n);
    // Instead of
    // console.log(parseInt(el.getAttribute('test')));
  });
}

```

> But, if it's the same to you, I'd like like leave it there just in case I want to make some quick edits on another computer for myself without installing npm etc...

That's a really bad approach - By editing the js file, you now make the js and the ts diverge from each other, and next time you build a release, if you forget to "backport" your changes you will lose them. Also every time you commit, your commit will contain the changes to the generated code, which adds noise and makes it more difficult for others to review changes.

manogna4:
> Sure, that approach will not affect whether or not you can do those things. I'm literally talking about assigning it to an inline variable instead of assigning it on the HTML element attribute and having to parse it out of the attribute afterwards.
> 
> ```ts
> for (let i = 0; i < array.length; i++) {
>   let el = createEl('div');
> 
>   // Store it in an inline variable
>   let n = i;
>   // Instead of on the element
>   // el.setAttribute('test', i);
>   el.addEventListener('click', () => {
>     // Use it in an enclosure that's on the same or in a sub level as the location you declared the variable.
>     console.log(n);
>     // Instead of
>     // console.log(parseInt(el.getAttribute('test')));
>   });
> }
> ```

correct me if I am wrong. With this approach the 2nd div will always refer to the 2nd leaf.

although this is not a feature right now, I might want the 2nd element in the view to represent the fifth leaf. For instance, the elements in the view could be in alphabetic order but those in the main note area would be in whatever order they've been opened (and inside splits etc). How do I encode information about which leaf the element (in the view) refers to without encoding the leaf-number in the html? 

> That's a really bad approach - By editing the js file, you now make the js and the ts diverge from each other, and next time you build a release, if you forget to "backport" your changes you will lose them. Also every time you commit, your commit will contain the changes to the generated code, which adds noise and makes it more difficult for others to review changes.

I understand that the js and ts file diverge - I wouldn't have contributed the changes back to the repo anyway. But I suppose I can always use the `main.js` file from the release if I want to. I'll add main.js to gitignore when the above point is settled.

lishid:
> correct me if I am wrong. With this approach the 2nd div will always refer to the 2nd leaf.
although this is not a feature right now, I might want the 2nd element in the view to represent the fifth leaf. For instance, the elements in the view could be in alphabetic order but those in the main note area would be in whatever order they've been opened (and inside splits etc). How do I encode information about which leaf the element (in the view) refers to without encoding the leaf-number in the html?

If you are always re-constructing the elements, then you can always set the proper indices when you construct them. Actually, since your div displays the pane name - it should always be associated with the pane itself instead of refer to some index (which may be changed since you rendered the pane information).

See this line: it should actually just activate the `leaf` that's passed in to the function.
https://github.com/manogna4/obsidian-koncham-workspace/blob/a80fa4504c37a28da45fd72c270a7ab4588763ac/main.ts#L157

ericaxu:
@manogna4 Hi there! Did you get a chance to take a look & resolve the comments? Thanks!

manogna4:
Hi. I've taken a look but I am learning this as I go, so it's taking me time to understand and implement the leaf suggestion by @lishid in his last comment.
I'll try to resolve it in the next 48 hours. If not, I'll close the PR for now and resubmit it once I have implemented the suggestion. Is that ok?

manogna4:
@lishid, I believe I've addressed all your concerns now:
+ main.js has been ignored
+ as suggested, I am using `leaf` to activate the leaf when clicked (much simpler, thanks!)

the changes are in a new release [0.0.3](https://github.com/manogna4/obsidian-koncham-workspace/releases/tag/0.0.3)

I've also made the following small changes that shouldn't cause more problems (I hope):
+ changed the workspace-leaf header icon
+ cleaned up code

lishid:
Ok, that's some good progress. just a few final small nits:

https://github.com/manogna4/obsidian-koncham-workspace/blob/7038b9760c48461d6ebbde3086f2199108682305/main.ts#L121
These 3 lines of code shouldn't be here, it should be outside this iteration. Also you can directly use `this.contentEl` instead of doing `this.containerEl.children[1]`.

https://github.com/manogna4/obsidian-koncham-workspace/blob/7038b9760c48461d6ebbde3086f2199108682305/main.ts#L22
Should probably use upper case for the first letter in your command names.

https://github.com/manogna4/obsidian-koncham-workspace/blob/7038b9760c48461d6ebbde3086f2199108682305/main.ts#L40
This says "refresh open panes" but it actually create a new pane for your plugin, or doesn't do anything if one already exists. Maybe rename the command?

https://github.com/manogna4/obsidian-koncham-workspace/blob/7038b9760c48461d6ebbde3086f2199108682305/main.ts#L52
Sidebar panes can't be set to active, so this line doesn't do anything. I'd recommend removing it.

https://github.com/manogna4/obsidian-koncham-workspace/blob/7038b9760c48461d6ebbde3086f2199108682305/main.ts#L7
Class names should be PascalCase (First letter should be upper case).

manogna4:
I've resolved everything you've raised.
Could you take a look at my question on nomenclature before you approve?

>These 3 lines of code shouldn't be here, it should be outside this iteration. Also you can directly use this.contentEl instead of doing this.containerEl.children[1].

done

> Should probably use upper case for the first letter in your command names

done
I tend to just use lower case everywhere (unless it breaks something) just to make it simpler. Is this a common convention?

>This says "refresh open panes" but it actually create a new pane for your plugin, or doesn't do anything if one already exists. Maybe rename the command?

I've changed the name to `Start open-panes` - that should explain what the command does better.
Do you have a suggestion on what to call the panes in the `root` area? `open-panes` seems like a misnomer since you can have a pane open in the sidebars too. These are those I've considered:
+ `main panes`: too generic?
+ `root panes`: too technical?
+ `open panes`: misleading?

>Sidebar panes can't be set to active, so this line doesn't do anything. I'd recommend removing it.

done

>Class names should be PascalCase (First letter should be upper case).

done

lishid:
> I tend to just use lower case everywhere (unless it breaks something) just to make it simpler. Is this a common convention?

It's like that for all command names (this is shown to the user in command palette and hotkeys)

> I've changed the name to Start open-panes - that should explain what the command does better.

To be honest, I'm not even sure having this command is necessary - your plugin always creates one on startup/enable anyway. If you still insist having one, I'd recommend naming it `Open pane for [plugin name]` to avoid making it confusing for the user.

> Do you have a suggestion on what to call the panes in the root area?

Hmm I'm not too sure either. Perhaps something like "Panes in the center area"?

manogna4:
> To be honest, I'm not even sure having this command is necessary - your plugin always creates one on startup/enable anyway

I understand your point. I didn't create this command initially either. But I closed the sidebar-pane by mistake a couple of times, and this was the only way to get it back without disabling+enabling the entire plug in. Do you have a better suggestion on how to implement getting the pane back if it's closed by mistake?

>I'd recommend naming it Open pane for [plugin name] to avoid making it confusing for the user.

I plan to have at least another sidebar pane as part of this plugin. So, `Open pane for [plugin name]` won't do. I like your suggestion and I am leaning towards `center-panes` at the moment.

Does this take care of all technical issues for now?
If so, I'd like to think on the nomenclature for a few hours before you accept the PR. I'll update this conversation when I am ready. Thanks...

manogna4:
@lishid 

changes since our previous conversation:
+ I went with `center-panes`
+ the command we discussed is now called *re*open center-panes to emphasize that it is only of use if it's been closed
+ I've updated the readme
+ refactoring some code (should be inconsequential)

Is there anything else you want me to work on?

pjeby:
Minor bit of feedback -- why not call it a "panes view"?  This would be consistent with other Obsidian terminology and what it does.

I personally didn't understand what "center panes" meant without very careful study of your README, and I don't think it's actually necessary to distinguish them from panes in the sidebars, since those are usually referred to as "views" rather than panes (e.g. "file explorer view", "backlinks view", "graph view", etc.

I might say the plugin provides something like, "a pane switcher that shows you a list of all panes open in the workspace and allows you to jump to any of them with a click.  This can be useful when some panes are off-screen (as when using plugins like Sliding Panes or Maximize Active Pane)."

(I didn't find the current animated screenshot to be much help in understanding, either: the combination of the theme and some of the horizontal lines in the documents made it look like there were vertical splits in effect and so I spent a lot of time trying to figure out how these stacked panes were being "centered" by this "center pane" function.  :wink:)

You will also probably get more users if you called your plugin "Pane Switcher" or "Switch and Pin Panes" as it will then be obvious what benefit(s) the plugin provides.  As a user, I don't know if I want to "enhance my workspace", but I do know that I'm always looking for Obsidian pane relievers.  :smile:

manogna4:
@pjeby  Thanks for taking the time out to think and talk about this.

I plan to have many related features in this plug in. I am new to programming in js, and for the public; so I am just trying to understand the approval process and get the hygiene factors right with some very basic stuff first before rolling out more features. What I've built so far is maybe about 20% of what I want to build in ultimately. (I have some of this working already). This should explain a lot of the decisions I have made regarding nomenclature.

>Minor bit of feedback -- why not call it a "panes view"? This would be consistent with other Obsidian terminology and what it does.
>I personally didn't understand what "center panes" meant without very careful study of your README, and I don't think it's actually necessary to distinguish them from panes in the sidebars, since those are usually referred to as "views" rather than panes (e.g. "file explorer view", "backlinks view", "graph view", etc.

I see your point, and largely agree. But obsidian does have the ability to have notes in the sidebars. At times, I use notes there - most often, this is to ensure a specific width for that note independent of the number of other panes I have open. But there are other use cases too. This is why I hesitate to call them `panes`. But you are right, it's confusing whether the `center` is a verb or noun. hm... would changing it to `centre` make it clearer? I know some people use `centre` exclusively as a noun. But it's British English from what I understand, so I don't know how well understood this will be.

>I might say the plugin provides something like, "a pane switcher that shows you a list of all panes open in the workspace and allows you to jump to any of them with a click. This can be useful when some panes are off-screen (as when using plugins like Sliding Panes or Maximize Active Pane)."

There's a lot of helpful language here. I'll work most of it in.

> I didn't find the current animated screenshot to be much help in understanding, either: the combination of the theme and some of the horizontal lines in the documents made it look like there were vertical splits in effect...

great point. I've been using my theme for so long that I guess I don't even notice these things anymore. I'll upload a recording with the default theme, and show unmaximized panes first to make it clear.

>You will also probably get more users if you called your plugin "Pane Switcher" or "Switch and Pin Panes" as it will then be obvious what benefit(s) the plugin provides. As a user, I don't know if I want to "enhance my workspace", but I do know that I'm always looking for Obsidian pane relievers. 😄

Pane switching will only be a small part of the plugin's ability. Functions that mess with the default look of the workspace will be toggle-able (with many being off by default), so a user should be able to pick what they want. I do want the `koncham` prefix since I have other plugins in the works, and some will make more sense together. So there's `koncham-workspace`, `koncham-notespace`, `koncham-outline` etc. The exact names are yet to be finalized, but I want to make it clear that they could go together. I'll get the keywords in the description so that the newer search function will surface my plugins for 'pane switcher' type searches.

PR #258, opened 2021-04-20
Publishing obsidian-excalidraw-plugin

zsviczian:

## Repo URL

[https://github.com/zsviczian/obsidian-excalidraw-plugin](https://github.com/zsviczian/obsidian-excalidraw-plugin)

## Release Checklist
  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


lishid:
Yay it's finally here!

Ok so first, please auto-format your code to normalize spaces vs tabs.

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/f8b8dffb94c0123527fa7a5699da0b54127a1f0a/src/main.ts#L74
Don't cast to TFile here, instead in your next line, check for `file && file instanceof TFile`, which will automatically make typescript cast it to TFile inside the if condition.

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/f8b8dffb94c0123527fa7a5699da0b54127a1f0a/src/main.ts#L88
Here I would recommend you to reverse all your if/else statements to avoid heavy nesting:

```ts
if (!cond1) {
  parseError("error1");
  return;
}
if (!cond2) {
  parseError("error2");
  return;
}
if (!cond3) {
  parseError("error3");
  return;
}
// All checks passed
// Do something
```

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/f8b8dffb94c0123527fa7a5699da0b54127a1f0a/src/main.ts#L133
No need to `getDoc()` here - the editor itself has everything.

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/f8b8dffb94c0123527fa7a5699da0b54127a1f0a/src/main.ts#L154
`activeLeaf` isn't guaranteed to be non-null, but it is most of the time. You could use `getLeavesOfType()` first and if that returns empty then `getLeaf()` to get a guaranteed leaf to fill in.

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/f8b8dffb94c0123527fa7a5699da0b54127a1f0a/src/main.ts#L167
Should await this.

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/f8b8dffb94c0123527fa7a5699da0b54127a1f0a/src/main.ts#L170
Same as above.

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/f8b8dffb94c0123527fa7a5699da0b54127a1f0a/src/main.ts#L172
Use await here instead of `.then`.

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/f8b8dffb94c0123527fa7a5699da0b54127a1f0a/src/main.ts#L166
https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/f8b8dffb94c0123527fa7a5699da0b54127a1f0a/src/main.ts#L170
I suggest passing any user-defined strings through normalizePath before using it with the vault API.
https://github.com/obsidianmd/obsidian-api/blob/068ff3052f0641baff90856dcb420dc2721680c9/obsidian.d.ts#L1941

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/f8b8dffb94c0123527fa7a5699da0b54127a1f0a/src/openDrawing.ts#L38
Can squash these two lines together.

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/f8b8dffb94c0123527fa7a5699da0b54127a1f0a/src/openDrawing.ts#L71
What does this do? Are you expecting an error here?

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/f8b8dffb94c0123527fa7a5699da0b54127a1f0a/src/utils.ts#L1
Here you can use window.moment instead.

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/f8b8dffb94c0123527fa7a5699da0b54127a1f0a/src/ExcalidrawView.ts#L69
This should refer to your constant.

zsviczian:
Awesome! Thank you! I never expected to receive so helpful comments. I've updated the code and addressed all the feedback.

lishid:
That's a great first pass.

A couple more if you don't mind:

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/39085bc962e6d10d07928443080faa6938521a21/src/main.ts#L95
Here I'd recommend using await on the read instead of then. Make your entire post processor an async function so that the renderer can await on it.

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/39085bc962e6d10d07928443080faa6938521a21/src/main.ts#L171
Ok I missed this last time, but you really shouldn't keep a copy of `this.view` in your plugin because it may have been removed (and you have no idea). You should use the API `Workspace.getLeavesOfType()` to see if there's an existing leaf of your type instead.

https://github.com/zsviczian/obsidian-excalidraw-plugin/blob/39085bc962e6d10d07928443080faa6938521a21/src/openDrawing.ts#L39
Pretty sure getFiles always return an array, but just in case you can always write `(files || []).filter...`

Looking good!

zsviczian:
Thanks! 
I address all the points raised.

zsviczian:
I need to hold off with releasing due to https://github.com/zsviczian/obsidian-excalidraw-plugin/issues/12 raised recently.

zsviczian:
I will need to hold off for a few days while I wait for the next Excalidraw release which is expected to solve the keyboard event handling issue
https://github.com/excalidraw/excalidraw/blob/master/src/packages/excalidraw/CHANGELOG.md#unreleased

lishid:
Sounds good. The rest looks good to me, I'm ready to merge as soon as you incorporate the excalidraw changes.

zsviczian:
@lishid Excalidraw 0.7.0 was just published. The Excalidraw plugin is ready to be released in Community Plugins

PR #260, opened 2021-04-21
Adds obsidian-kanban

mgmeyers:

## Repo URL

https://github.com/mgmeyers/obsidian-kanban

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1281,5 +1281,13 @@
         "description": "An interface to experiment with python scripts and more.",
         "author": "Cristian Vasquez",
         "repo": "cristianvasquez/obsidian-lab"
+    },
+    {
+        "id": "obsidian-kanban",
+        "name": "Kanban",
+        "author": "mgmeyers",
+        "description": "Create markdown-backed Kanban boards in Obsidian",
+        "repo": "mgmeyers/obsidian-kanban",
+        "branch": "main"
     }
 ]


lishid:
Did you want to try booting the view back to markdown mode when the file doesn't have the frontmatter key?

mgmeyers:
@lishid So, I was able to get every case working but one with this. Can you see anything amiss?

```typescript
this.registerEvent(
  this.app.workspace.on("file-open", (file) => {
    const activeLeaf = this.app.workspace.activeLeaf;
    const activeViewType = activeLeaf.view.getViewType();
    const cache = file ? this.app.metadataCache.getFileCache(file) : null;

    console.log('file-open', activeViewType, file)

    if (
      activeViewType === "kanban" &&
      (!cache?.frontmatter || !cache.frontmatter[frontMatterKey])
    ) {
      // We're in kanban mode without the kanban frontmatter key
      return this.setMarkdownView(activeLeaf);
    }

    if (
      // No frontmatter
      !cache?.frontmatter ||

      // In a special view (eg. localgraph)
      activeViewType !== "markdown" ||
      
      // The default view for kanban is set to markdown
      defaultViewType === "markdown"
    ) {
      return;
    }

    if (cache.frontmatter[frontMatterKey]) {
      return this.setKanbanView(activeLeaf);
    }
  })
);
```

What works:

- Backlinks, localgraph, outline views
- Navigating backwards from `markdown` to `kanban`
- Navigating forward from `markdown` to `kanban`
- Navigating forward from `kanban` to `markdown`

What doesn't work:

- Navigating backwards from `kanban` to `markdown`

Here's what happens within the `file-open` event when I do this last one:

Screen Shot 2021-04-22 at 4 18 29 PM

Noticing there are some `null` files after I trigger:

```typescript
async setKanbanView(leaf: WorkspaceLeaf) {
    await leaf.setViewState({
      type: kanbanViewType,
      state: leaf.view.getState(),
    });
  }
```

Do you think it's trying to "go back" to a null file? Is there another way to change the view type?

lishid:
I think what's happening here is that you're attempting to re-set the view type while it's setting the view type & loading the file, causing a race condition in the async loading progress. I guess there isn't a great way to override the markdown view loading from the API directly.

If you still want to do things like this, I think the most safe thing to do is to actually monkey patch `WorkspaceLeaf.prototype.setViewState`. In your patched version, you can do all your checks and if you want to override the view, change the type before passing it to the real function. @pjeby has a library designed for monkey-patching Obsidian methods at https://www.npmjs.com/package/monkey-around

mgmeyers:
That works really well! It seems to work seamlessly now.

https://github.com/mgmeyers/obsidian-kanban/blob/main/src/main.ts#L38-L65

pjeby:
The issue was re-entrance, specifically that of setViewState(), since the markdown view's loadFile() was triggering the file-open while setViewState() was still running, and then the Kanban plugin started a second setViewState() which then closed the first view and then added a history entry...  before returning to the *original* setViewState(), which then added a *second* history entry.

To do this without monkeypatching, Obsidian's setViewState() would need to be serialized against re-entry, and the Kanban plugin would need to include `popstate: true` in the state in order to prevent two history entries from being added.

Alternately, if setViewState triggered an event before loading to ask, "any views want this file?", then plugins could hook that to override the view type selection.  (It should not be triggered on popstate though.)  Potentially there could also then be UI for selecting one of those view types other than the default.

But yeah, after digging through the code a bit, without monkeypatching setViewState it's not currently possible to override view selection.  (Well, technically you could override the markdown view itself, but since the factory doesn't know what file it's being created to serve, it's not much help.)

mgmeyers:
Thanks @pjeby, it's helpful to understand what's going on here.

PR #261, opened 2021-04-23
- Added: incremental writing

bjsi:

## Repo URL

https://github.com/bjsi/incremental-writing

## Release Checklist

- [ ] I have tested this on ~~Windows, macOS~~, and Linux (I only have a Linux computer)
  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


lishid:
Please auto-format your code to normalize tabs and spaces.

https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/main.ts#L30
Should update this to use the latest recommended way of loading settings. This way might lead to errors that are difficult to debug for your users. See https://github.com/obsidianmd/obsidian-sample-plugin/blob/3b38a36a07a121484a31f9b3fcbc59c2e7f892e6/main.ts#L60

https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/main.ts#L31
This doesn't sound like it belongs to loadConfig - should just be in onload()

https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/main.ts#L50
This might return null.

https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/main.ts#L51
What's this for? `open` is not a public API and by calling it directly you've broken its internal state management (open should never be called twice without first closing).

https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/main.ts#L66
Don't need to setDisabled(false) because that's the default.

Given you've used a lot of private API of the search pane, I'd advise you to add some warning to your README.md that you'll be hijacking the search panel, which may break when Obsidian updates.

https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/main.ts#L153
You should probably use `checkCallback` here and pass the actual value to your FuzzyNoteAdder modal. Also probably best to use `Workspace.getActiveViewOfType(MarkdownView)` instead of just checking for activeLeaf, because the user might be focused on a different type of view.

https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/queue.ts#L59
Recommend using `metadataCache.getFileCache(tfile)?.frontmatter` instead of importing your own frontmatter & YAML parser here.

https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/queue.ts#L150
https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/helpers/file-utils.ts#L22
Second parameter should be `""` here, or whichever file you're linking from, not the target file itself.

https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/queue.ts#L181
https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/helpers/file-utils.ts#L50
https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/views/queue-modal.ts#L19
Don't cast to TFile. Do an instanceof check (and a null check) instead before it's being used.

https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/settings.ts#L10
This should be a plain object instead of a class: `export const DefaultSettings: IWSettings = {...};`

https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/hashset.ts#L1
Use a `Set`?

https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/helpers/functools.ts#L2
This seems quite similar to our `debounce` from the API. Feel free to use that one which is slightly more optimal (it doesn't constantly register new timers when called multiple times).

https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/helpers/file-utils.ts#L56
This has many ways it can fail - activeLeaf can be null, and activeLeaf.view can be a different type. Please use `Workspace.getActiveViewOfType(MarkdownView)` and null check the result.

https://github.com/bjsi/incremental-writing/blob/64c4e6596b34bc65ea8880a1b79857ccd14d11b4/src/views/queue-modal.ts#L23
file name here but "queueFilePath" two lines after may not be compatible.

bjsi:
I really appreciate the detailed code review. Thanks for your efforts! I made all of the requested changes apart from the YAML parser - I think I was having some issues with the cache returning null when there was YAML present in the file.

lishid:
https://github.com/bjsi/incremental-writing/blob/e538525fbca9633ea4d613b096d7f4e2876ffb4a/src/main.ts#L158
checkCallback should always return the proper result even if checking is false - you should code it like this:

```ts
let condition = ...;
if (condition) {
  if (!checking) {
    // do something, preferably using the results of the condition check, such as the current leaf/view/file.
  }
  return true;
}
```

https://github.com/bjsi/incremental-writing/blob/e538525fbca9633ea4d613b096d7f4e2876ffb4a/src/views/queue-modal.ts#L20
https://github.com/bjsi/incremental-writing/blob/e538525fbca9633ea4d613b096d7f4e2876ffb4a/src/helpers/file-utils.ts#L55
Same as before - use instanceof instead of casting.

https://github.com/bjsi/incremental-writing/blob/e538525fbca9633ea4d613b096d7f4e2876ffb4a/src/helpers/functools.ts#L1
This file is probably no longer needed right?

Everything else is clear. Once these are addressed we can get it merged.

bjsi:
Ok! I just pushed the updates you suggested. Thanks again.

PR #262, opened 2021-04-25
Image in Editor Plugin

ozntel:

## Repo URL

https://github.com/ozntel/oz-image-in-editor-obsidian

## Release Checklist

    - [x] macOS
    - [x] Windows
    - [ ] Linux

  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css`
- [x] Github release name matches the exact version number specified in your manifest.json


lishid:
https://github.com/ozntel/oz-image-in-editor-obsidian/blob/5ab74f7233bb7fcaa74f46fcb11a88bc2be598c9/main.js#L30
https://github.com/ozntel/oz-image-in-editor-obsidian/blob/5ab74f7233bb7fcaa74f46fcb11a88bc2be598c9/main.js#L4
This won't work on mobile. Don't deal with the file system if you don't have to.

https://github.com/ozntel/oz-image-in-editor-obsidian/blob/5ab74f7233bb7fcaa74f46fcb11a88bc2be598c9/main.js#L34
Don't need to use `'path'`, just concatenate them with '/' and then use `normalizePath(path)`. Also you are using a private API for vault.config which you shouldn't be using either way - use `vault.getConfig('key')` which will return you a default value instead of possibly undefined.

https://github.com/ozntel/oz-image-in-editor-obsidian/blob/5ab74f7233bb7fcaa74f46fcb11a88bc2be598c9/main.js#L40
This looks incredibly inefficient. You are parsing the entire file for each and every change event, which means that this will introduce a ton of lag for large files and/or slow machines. Use `debounce` (from `'obsidian'`) with a good timeout (like 500-1000) to re-scan after changes.

https://github.com/ozntel/oz-image-in-editor-obsidian/blob/5ab74f7233bb7fcaa74f46fcb11a88bc2be598c9/main.js#L122
You really don't need to resolve those paths manually. Use `DataAdapter.getResourcePath(normalizedPath)` and let the app figure out the proper paths for you.

You seems to have committed the `.js` file instead of the real `.ts` file. Might want to fix that.

ozntel:
Hi, @lishid  

Thank you for your valuable feedback. I have written the plugin originally in JS but I am converting it right now into `.ts`. Meanwhile, I have a few questions, though:

1. `this.app.vault.getResourcePath` causes quite a delay for images to load and discomfort during writing. Due to delay, widgets sit on the text written for a second and then pushed below. Is there any other solution to get the full path, which would be fully compatible with both computers and mobile?

2. For `debounce`, is it ok to use as below:
`cm.on("change", () => debounce(this.check_lines(cm), 1000, false))`

3. I changed a little bit way of working to be able to support all kinds of attachment settings. I firstly get the files `this.app.vault.getFiles()` and then filter using the Image name and use the path of the filtered image if exists. Is this approach ok?

You can find the initial version of `.ts` here:
https://github.com/ozntel/oz-image-in-editor-obsidian/blob/c2a4d2a7e0c2bca8a42bab5f27c2a05e1f6dcb2c/main.ts

Btw sorry for the headache. I am normally an HR Specialist. I am just coding as a hobby during my free time and I appreciate your patience 🙂

lishid:
1. `getResourcePath` should return exactly what you're trying to do here with `'app://local' + img_path`. The problem here is that you're rewriting an almost identical way to generate app://local links, which will stop working should we decide to rename `app://local` to something else. If you use `getResourcePath` which now returns `app://local` paths, but will be properly updated if we decide to change the protocol so that your plugin will continue to function. The other thing is that 'app://local' does not work on mobile at all. On Android, it's `content://something/emulated/0/Documents/Obsidian/VaultName/...` and on iOS it's something else I don't even know. Our `adapter` correctly calculates the paths, so you should just use that instead of trying to come up with the path yourself.

2. Use `cm.on("change", debounce(() => this.check_lines(cm), 1000, false))` instead.

3. Don't use that, instead, you should use `app.metadataCache.getFirstLinkpathDest(imagePath, "")` which will find the exact file a path links to.

> You can find the initial version of .ts here:

Now that you committed the ts file, you can now add the .js to your `.gitignore` and remove it from the repo - you only need to upload the .js file when releasing the plugin.

ozntel:
I already implemented the `getResourcePath` in `.ts` file:
https://github.com/ozntel/oz-image-in-editor-obsidian/blob/c2a4d2a7e0c2bca8a42bab5f27c2a05e1f6dcb2c/main.ts#L89

However, it causes an issue as I recorded in the video I am sharing:
https://youtu.be/1-4qfWX0BgM

When I handle the links manually, it doesn't happen. Do I maybe use the `getResourcePath ` function incorrectly?

pjeby:
A few other issues:

* You're not unloading your handler when the plugin is disabled, which means that it can't be safely upgraded (or even unloaded properly) without restarting Obsidian.  Your `onunload()` method should `iterateCodeMirrors()` and remove the `change` event handlers, as well as the widgets themselves.
* It isn't necessary to rescan the entire file on change events; change events contain information about the range of lines affected by the change (see [this part of the CodeMirror manual](https://codemirror.net/doc/manual.html#events)), so you can loop over only the changed lines.  This isn't a requirement but would considerably improve responsiveness.  Likewise, there would be less flickering and redrawing as it would only replace images when a line containing one is edited.  (Checking the existing widgets against the image links you find would also allow you to skip debouncing, since the DOM would not be updated on every keystroke, even when the line being edited contains images.)
* Don't search through all the files to find the image; as lishid says, use `app.metadataCache.getFirstLinkpathDest(url, currentFilename)` to obtain the file, then get its resource path.  Don't trim the image URL with getFirstIndexOfImageFileName, because you should instead let `getFirstLinkpathDest()` use the full URL as a relative link, passing in the current file's filename to use as the base URL.  This will ensure that you actually find the correct file to use as an image.  (After all, a vault might contain 10 `main.png` files, only one of which is the intended image.)

ozntel:
Hi, 

- I included a function to unload my handlers

- I changed the way of scanning as you suggested. Now I only check single line change rather than iterating through all lines for each change. Now I need to scan all lines only once the markdown file is fired.  

   **Question**: Is there any better way than `focus` for this purpose?

- I changed the way of finding the file path as you suggested to `getFirstLinkpathDest()`  and resigned from filtering from available files. 

- I still have a problem with the speed of loading Images and it seems like the only problem is caused by `app.vault.getSourcePath()`. As a comparison I recorded 10 seconds video, where you can see the difference it causes:

    Without `getSourcePath()`:  https://youtu.be/Oizrgay5bxg
    With `getSourcePath()`:  https://youtu.be/1-4qfWX0BgM
  
    Due to the delay caused by this method, the cursor stays behind the image once I try to go to the new line. Is there any reason why this method specifically causes such a problem and if there is any alternative way?

pjeby:
The issue with `getResourcePath()` is that it's adding the current timestamp, in order to force the file to reload each time.  @lishid: maybe instead of using Date.now() it should use the stat.mtime, so it's keyed by the version of the file?  If not, then a workaround would be for the plugin to cache the resource paths, or to just generate its own and ignore the API (since this plugin cannot and will never run on mobile, so it should always be seeing the filesystem adapter).

@ozntel: Regarding the focus event, I'm not sure why you're doing that.  Is the problem that it's not updating when the editor loads a new file?

Anyway, the plugin is definitely looking much cleaner!

I do see one other remaining problem, which is that the images are being looked up based on the *active* view, but a non-active view can be updated (by being a mirror of another view, edited eternally, etc.), causing change events to fire.  So you need to not merely get the active view's file (which your current implementation will get an error if the active view isn't a file), but the file of the view the codemirror belongs to.  @lishid is there a way to go from a codemirror instance to its owning view or otherwise determine the file?  It is needed for this sort of thing, unless one wants to iterate all views of type markdown looking for the one with the matching codemirror.

ozntel:
Hi @pjeby, yes, that's right. When I remove the focus event, only the lines of the markdown file, which is currently open, are looped through. If I don't have any open file when the plugin is activated or I want to switch to a new markdown file, the plugin only checks the changes in certain lines and doesn't run `this.check_lines(cm)`

lishid:
> The issue with getResourcePath() is that it's adding the current timestamp, in order to force the file to reload each time. @lishid: maybe instead of using Date.now() it should use the stat.mtime, so it's keyed by the version of the file? If not, then a workaround would be for the plugin to cache the resource paths, or to just generate its own and ignore the API (since this plugin cannot and will never run on mobile, so it should always be seeing the filesystem adapter).

I have made this adjust for the upcoming release, thank you for the proposal.

ozntel:
> I do see one other remaining problem, which is that the images are being looked up based on the active view, but a non-active view can be updated (by being a mirror of another view, edited eternally, etc.), causing change events to fire. So you need to not merely get the active view's file (which your current implementation will get an error if the active view isn't a file), but the file of the view the codemirror belongs to. @lishid is there a way to go from a codemirror instance to its owning view or otherwise determine the file? It is needed for this sort of thing, unless one wants to iterate all views of type markdown looking for the one with the matching codemirror.

@pjeby It returns null when it can not find the file:
`metadataCache.getFirstLinkpathDest(decodeURIComponent(filename), this.getActiveNoteFile().path)`
Hence, I included an additional check to make sure that I don't look for the path of image when it is null to make sure that it is not giving any error in the console. 
I also tried to check logs during my tests for the `changes` event and it was never fired unexpectedly:
if(image != null) img.src =  this.app.vault.getResourcePath(image); 

Is there maybe any change required for the first release of the plugin? I understand that I can leave `getResourcePath()` at the moment since it will be improved in the next release as lishid confirmed.

ozntel:
> I do see one other remaining problem, which is that the images are being looked up based on the _active_ view, but a non-active view can be updated (by being a mirror of another view, edited eternally, etc.), causing change events to fire. So you need to not merely get the active view's file (which your current implementation will get an error if the active view isn't a file), but the file of the view the codemirror belongs to. @lishid is there a way to go from a codemirror instance to its owning view or otherwise determine the file? It is needed for this sort of thing, unless one wants to iterate all views of type markdown looking for the one with the matching codemirror.

Sorry, I just understood what you mean. I missed the point that there might be more than 1 md files open. I have just changed my event to `file-open`. Any other CodeMirror won't be scanned again if I can find a solution to go from file to cm here or from cm to file:

https://github.com/ozntel/oz-image-in-editor-obsidian/blob/a6eb2648ba1598628f59e824c2a7a9db459740da/main.ts#L172

If I manage to achieve, only the active leaf will be always scanned, hence, images won't be looked up on wrong view.

pjeby:
From the file-open event, use `cm = this.app.workspace.getActiveViewOfType(MarkdownView)?.sourceMode?.cmEditor`, and then you can pass it to your refresh handler.

If you then also set, say, a `target_file` property on the code mirror, then your handlers could get the path from it.  If you use the file object (rather than just its path string), then if the file is renamed or moved it will still have the right `.path`.

Alternately, you can loop over `this.app.workspace.getLeavesOfType("markdown")` and take each one's `.view.sourceMode?.cmEditor` to get the code mirror, and its `.view.file` to get the file.

This is an area where it would be nice if Obsidian had an official way to go from a codeMirror object to the view or leaf associated with it -- preferably as a property on the cmEditor, rather than needing to loop over all the workspace's views every time.  I've run into this issue with codemirror-enhancing plugins before myself, and worked around it by attaching an object to each codemirror that included both my enhancement-related state info and a reference to the editor's view for just this type of thing.

ozntel:
> From the file-open event, use `cm = this.app.workspace.getActiveViewOfType(MarkdownView)?.sourceMode?.cmEditor`, and then you can pass it to your refresh handler.

I wanted to use this, however, it says that it is deprecated and won't be supported on Mobile and also once WYSIWYG is released. As an alternative way, I tried first to get Editor using `this.app.workspace.getActiveViewOfType(MarkdownView)?.editor` and, since there is no public API to get cmEditor from Editor, I used for the moment `editor.cm` to retrieve CodeMirror. Is there maybe any chance to add also a public API for this purpose?

pjeby:
This plugin will never work on mobile, because it's written for CodeMirror 5.  So using cmEditor is the correct (documented) way to get a hold of the deprecated CodeMirror 5, which *everything else in your plugin depends upon*.

Likewise, when the WYSIWYG editor is released, there will also be no need for this plugin, as the WYSIWYG will presumably have its own image handling built-in.

tl;dr: ignore the deprecation warning, and use the documented APIs for accessing CodeMirror, as for the expected useful life of this plugin, the deprecation will have no effect.  From lishid's comments on Discord, there will also be a crossover period where both the existing editor and the new editor will be selectable via a setting, before the old editor is phased out entirely.

ozntel:
Thanks for the confirmation, I used `this.app.workspace.getActiveViewOfType(MarkdownView)?.sourceMode?.cmEditor` as you suggested. I am also using `targetFile `inside of `check_line` to make sure that I look up correct file. 

Since this plugin works only on Desktop I changed the plugin to `isDesktopOnly` in `manifest`, and also switched back the way of path obtaining to the old way to make sure that the images load properly, and not with a delay. I will correct it once the 'getResourcePath()` method is improved. 

Is there anything else you would like to change or is it ok to release in this format?

pjeby:
[This line](https://github.com/ozntel/oz-image-in-editor-obsidian/blob/master/src/main.ts#L20) is unnecessary: the registerEvent in onload means that Obsidian will unregister the event for you at unload time.

Your file-open event will currently refresh all open files, every time a user moves between panes.  This should probably be changed to only update the file that is being opened, using the active view's codemirror and passing the targetFile to check_lines so it doesn't have to getFileCmBelongsTo.

You have a getActiveNoteFile function that assumes the active view is a markdown view,  Don't do this, since there are many other kinds of views that could be active.  Use `getActiveLeafOfType(MarkdownView)` instead.

Last, I should note that your URL detection has a bug: if someone has an image whose name begins with `http` (e.g. `http-transport-diagram.png`, or `http-presentation/part-1.png`) it will be considered a URL.  `data:` URLs are also not supported.  The technically correct way to detect an absolute url is a string of letters, numbers, plus or minus, followed by a `:`, e.g. a regex of `/^[a-z][a-z0-9+\-.]*:/i`.

These are the only issues I see remaining from my review.  Once those things are done, it's up to lishid to make the final approval.  But I believe we've addressed everything in his previous comments.

ozntel:
> This line is unnecessary: the registerEvent in onload means that Obsidian will unregister the event for you at unload time.

Removed

> Your file-open event will currently refresh all open files, every time a user moves between panes. This should probably be changed to only update the file that is being opened, using the active view's codemirror and passing the targetFile to check_lines so it doesn't have to getFileCmBelongsTo.

Unfortunately, I couldn't find any other way to loop through all the files once the plugin is loaded. `iterateCodeMirrors()` is not triggered within `onload()`.  `registerCodeMirror()` works only for the active file and the future files, however, it doesn't work on other open files during the load. Is there is any way to check all files only during the load?

>You have a getActiveNoteFile function that assumes the active view is a markdown view, Don't do this, since there are many other kinds of views that could be active. Use getActiveLeafOfType(MarkdownView) instead (or just getActiveFile()).

Changed

>Last, I should note that your URL detection has a bug: if someone has an image whose name begins with http (e.g. http-transport-diagram.png, or http-presentation/part-1.png) it will be considered a URL. data: URLs are also not supported. The technically correct way to detect an absolute url is a string of letters, numbers, plus or minus, followed by a :, e.g. a regex of /^[a-z][a-z0-9+\-.]+:/i.

Changed the regex as you suggested

pjeby:
`registerCodeMirror()` should invoke its callback for all editors that exist when the plugin starts, and it should also be called for every editor created thereafter.  So I don't understand what you mean by it not working on "other open files during the load".  Could you explain your test methodology for that?  i.e., are you reloading Obsidian or hot reloading the plugin, what panes do you have open, what state are they in, etc.?

pjeby:
Ah, I think I see what you mean now...  when the callback to registerCodeMirror is called for a new editor, it won't have any contents yet.  And when the contents do get set, your handler is only checking the *last line* of the change, so it doesn't process most of the file.  (That is, the change event for the file being loaded will only cause the last line of the file to be checked.)

The bug here is that `codemirrorLineChanges` should loop over the `from.line` through the `to.line`, instead of only processing the `to.line`.  I think you were assuming here that the changes array will be individual lines, but in actuality each change can be a range that can potentially be the entire file (e.g. when loaded).  There is another complication here as well: if a change is *adding* lines, then the added lines will be missed by the scan as well.  In essence, you need to scan from the `from.line` up through `to.line + added_lines`, and you need to use the `change` event, not the `changes` event.

In effect, these issues with your change processing are why you currently need the file-open to refresh the whole thing.  If your change handler were correct, the only time you would need to refresh a whole file would be in an iterateCodeMirrors call at plugin start, and you might not need a file-open handler at all.

ozntel:
I suppose it still doesn't solve the issue since `cm.on("change", (cm, change) => console.log(change))` returns from and to 0 during the initial load:

console

Or did you mean something else?

pjeby:
That's what I meant by the `+ added_lines` part: i.e., that you need to take (count of lines in text) - (count of lines in removed) and extend the `to.line` accordingly.  Or I suppose you could scan from `from.line` through `from.line + text.length - 1` (inclusive), which would accomplish the same thing, I think.  The point is to loop only over the new lines, since any old lines that were removed would already have their widgets removed (I believe.)

ozntel:
You are totally right. I just made the change as you mentioned and it works like a charm 🙂 
Is there anything else I should include & delete so that it can be released for desktop?

pjeby:
To handle the case where your plugin is loaded after Obsidian (i.e. going into settings and enabling it), I believe your registerCodeMirror handler should check all lines currently in that editor.  Otherwise, if files are open that contain images, they will not be rendered when somebody first installs and enables the plugin.

Only other suggestion I can think of is that rather than forcing the style props on the image, that you instead define them in your styles.css.  Otherwise, users and themes cannot override them.  (Speaking of which, your styles should probably be using `img.oz-image-widget[alt...]` so that those styles are only applied to the images this plugin generates.)

Finally, is there a reason you're not allowing svg images?

Apart from these comments, I can't think of anything else that would produce bugs, conflict with other plugins, or anything else relevant.  @lishid's the one who has to actually approve and merge, though, I'm just a volunteer helping out.  (It'll make it easier on him, though, if you also go ahead and resolve the conflicts on the PR.)

Congrats on what looks like an awesome plugin, which I will likely give a try myself once it's in the plugin shop!

ozntel:
> To handle the case where your plugin is loaded after Obsidian (i.e. going into settings and enabling it), I believe your registerCodeMirror handler should check all lines currently in that editor. Otherwise, if files are open that contain images, they will not be rendered when somebody first installs and enables the plugin.

Yes, you are right. I included `initialLoadHandler` for the plugin.

> Only other suggestion I can think of is that rather than forcing the style props on the image, that you instead define them in your styles.css. Otherwise, users and themes cannot override them. (Speaking of which, your styles should probably be using img.oz-image-widget[alt...] so that those styles are only applied to the images this plugin generates.)

I moved to the CSS file.

> Finally, is there a reason you're not allowing svg images?

Actually there is no reason. I only had these formats in my repo, that's the reason for the limitation. I included svg, as well

> Apart from these comments, I can't think of anything else that would produce bugs, conflict with other plugins, or anything else relevant. @lishid's the one who has to actually approve and merge, though, I'm just a volunteer helping out. (It'll make it easier on him, though, if you also go ahead and resolve the conflicts on the PR.)

Just resolved.

> Congrats on what looks like an awesome plugin, which I will likely give a try myself once it's in the plugin shop!

Thank you. I received some good feedback. At least, it should help until the WYSIWYG is ready for usage.

pjeby:
You've got a typo in your onunload; it needs to use `cm.off` not `cm.on` -- apparently I missed that on my last look-through.  You should probably test enabling and disabling your plugin from the settings to be sure it doesn't do anything weird; it looks like the current code will take off all the widgets when disabled but will put them back if the currently-open file(s) are edited, despite the plugin being "unloaded" at that point.

ozntel:
Yes, I just corrected and tested. It behaves normally now when I unload the plugin

ozntel:
@lishid is there anything else I should address for the release of the plugin for desktop? Thanks

lishid:
https://github.com/ozntel/oz-image-in-editor-obsidian/blob/45d09629ad8cbb3408792327af6ba8307b524f70/src/utils.ts#L55
Your comment says active markdown file but it may not be a markdown file.

https://github.com/ozntel/oz-image-in-editor-obsidian/blob/45d09629ad8cbb3408792327af6ba8307b524f70/src/utils.ts#L65
https://github.com/ozntel/oz-image-in-editor-obsidian/blob/45d09629ad8cbb3408792327af6ba8307b524f70/src/utils.ts#L71
You should probably URI encode the path here. There are also additional filename normalizations performed with the adapter - I will spare you the details here since you're going to scrap this once we release the patch anyway.

lishid:
Oh and maybe add an instanceof MarkdownView check here just in case someone registers an alternate markdown leaf type?
https://github.com/ozntel/oz-image-in-editor-obsidian/blob/45d09629ad8cbb3408792327af6ba8307b524f70/src/utils.ts#L79

ozntel:
> https://github.com/ozntel/oz-image-in-editor-obsidian/blob/45d09629ad8cbb3408792327af6ba8307b524f70/src/utils.ts#L65
https://github.com/ozntel/oz-image-in-editor-obsidian/blob/45d09629ad8cbb3408792327af6ba8307b524f70/src/utils.ts#L71
You should probably URI encode the path here. There are also additional filename normalizations performed with the adapter - I will spare you the details here since you're going to scrap this once we release the patch anyway.

Sure, I will revert back to `getResourcePath()` once the release is live.

> https://github.com/ozntel/oz-image-in-editor-obsidian/blob/45d09629ad8cbb3408792327af6ba8307b524f70/src/utils.ts#L55
> Your comment says active markdown file but it may not be a markdown file.

I tried to address the concern pjeby raised as below and used `getActiveFile()` instead:

> You have a getActiveNoteFile function that assumes the active view is a markdown view, Don't do this, since there are many other kinds of views that could be active. Use getActiveLeafOfType(MarkdownView) instead (or just getActiveFile()).

However, I can change this to `getActiveLeafOfType(MarkdownView)` within the next release together with `getResourcePath()` if it is ok.

> Oh and maybe add an instanceof MarkdownView check here just in case someone registers an alternate markdown leaf type?
https://github.com/ozntel/oz-image-in-editor-obsidian/blob/45d09629ad8cbb3408792327af6ba8307b524f70/src/utils.ts#L79

I just tried but all markdown files return false for `leafs[i] instanceof MarkdownView`, hence, no image is loaded after. Or maybe I understood wrong?

lishid:
> I just tried but all markdown files return false for leafs[i] instanceof MarkdownView, hence, no image is loaded after. Or maybe I understood wrong?

Should be `leafs[i].view instanceof MarkdowView`

pjeby:
> Your comment says active markdown file but it may not be a markdown file.

Note that the code is fine, it's just that the name and comment are misleading.  In truth, that function cannot be called except on a codemirror whose view has a null file...  which should be impossible.

PR #268, opened 2021-05-01
Add Chisel to community plugins

eliyap:

## Repo URL

## Repo URL

https://github.com/eliyap/Chisel/releases/tag/1.0.0

## Release Checklist

- [x] I have tested this on ~~Windows~~, macOS, and ~~Linux~~ _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1311,5 +1311,12 @@
         "description": "An Obsidian plugin for syncing your Kindle book highlights using your Amazon login",
         "author": "Hady Osman",
         "repo": "hadynz/obsidian-kindle-plugin"
+    },
+    {
+        "id": "chisel",
+        "name": "Chisel",
+        "description": "Refinements to the vanilla Obsidian experience",
+        "author": "Elijah Yap",
+        "repo": "eliyap/Chisel"
     }
 ]


whitenoisedev:
For atomicity purposes, this plugin should be split in 3.

lishid:
Yeah... so this is more of a CSS snippet sort of thing - we don't yet have a marketplace for distributing snippets but it's definitely more of a snippet.

https://github.com/eliyap/Chisel/blob/f9e014a9b6b2fd5b735fc86f2a709e9d4dd7d3d9/styles/todo.scss#L5
Might want to make it fallback to `monospaced` on other operating systems. Actually - I'm inclined to make this monospaced in core too.

eliyap:
Yeah, I'm not sure if `@font-face` allows you to set the generic `monospaced` style.

Would love if this tweak could be incorporated into core Obsidian!

lishid:
Should be in the `font-family` in the next block with your font first, then monospaced, then the defaults. But yeah I will add this to core as well, maybe not your font but our monospaced variable definition.

eliyap:
Sounds good, though that would cause the square brackets to go monospace as well, which might be a little irritating.

With regards to the image alt-text styling,
- is this something that should be discouraged (e.g. it messes with alt text in obsidian publish)
- I'd argue that's not really "themeing" per se, so could this still be added to the plugin list?

lishid:
It's not a theme, I agree with that, but it certainly fall under CSS snippets, which is used to quickly toggle on small pieces of CSS to achieve some functionality.

eliyap:
Totally understand.

I'd suggest that users be able to add plugins / snippets by pasting a repo link (similar to Swift Package Manager), and letting Obsidian clone the files for them, instead of manually cloning to a directory. Of course, that feature request is out of the scope of this discussion, but it might be useful way to address this middle ground between theme and full blown plugin.

Thanks for the help!

PR #270, opened 2021-05-02
Add a new plugin : Internal Links

mdelobelle:

## Repo URL

https://github.com/mdelobelle/obsidian_supercharged_links

## Release Checklist

- [x] main.js
- [x] manifest.json
- [ ] styles.css (optional)
- [x] Github release name matches the exact version number specified in your manifest.json


ericaxu:
Very interesting plugin!

https://github.com/mdelobelle/obsidian_supercharged_links/blob/master/main.ts#L1
Is `privateDecrypt` used somewhere?

Otherwise it looks good but I have tested it.

It looks like it would do some potentially heavy DOM querying on each metadata change though. Have you had a chance to test whether it lags the UI when making changes in a big vault? Just curious.

mdelobelle:
Hello @ericaxu , thank you for your comment, and thank you very much for your work, Obsidian is GOLD!

I've removed the useless imports (it was a copy from the sample plugin)
I've tested it on a vault with 10k files. When I display 5 notes at the same time, each containing 500+ links to a note with frontmatter attributes, the change of a metadata value in the source note is immediately displayed. It doesn't seem to lag the UI.

To be honest I find it a bit hacky right now: I have the feeling that the way i'm catching changes is not efficient. 
For example:
I don't know how to catch the fact that a note has been displayed
I don't know how to catch the fact that the content of the active leaf has changed
So this introduces limitations in the expected behaviour of the plugin: the a.internal-link are altered only when the active-leaf changes or when the metadataCache changes.
If you had some advice that could point me in the right direction, it would be really helpful. 

Don't hesitate to include this feature as a core capability of Obsidian if you think that it could benefit to everyone ;) I think this kind of feature could increase the adoption of frontmatter among users

mdelobelle:
Hi @lishid thank you for your review. This was a bit challenging but I think that i managed to go through it

> mdelobelle/obsidian_supercharged_links@19b7753
You should probably add back the gitignore entry for the js file and remove it from the repo.

Done

> https://github.com/mdelobelle/obsidian_supercharged_links/blob/265bfc5b5681da54ede4c4d21c3c94ea0a55d8a2/main.ts#L12
Please rename your plugin name as well as the settings and other placeholder names.

Done

> https://github.com/mdelobelle/obsidian_supercharged_links/blob/265bfc5b5681da54ede4c4d21c3c94ea0a55d8a2/main.ts#L29
Leaves aren't guaranteed to have files. You should test leaf.view instanceof MarkdownView before attempting to do anything here.

Done

> https://github.com/mdelobelle/obsidian_supercharged_links/blob/265bfc5b5681da54ede4c4d21c3c94ea0a55d8a2/main.ts#L34
Should use file.path here instead for relative paths to resolve properly.

Done

>https://github.com/mdelobelle/obsidian_supercharged_links/blob/265bfc5b5681da54ede4c4d21c3c94ea0a55d8a2/main.ts#L47
Actually I just realized what you're trying to do - you're much better off using a markdown post-processor instead of scanning the dom the whole time - the app has optimizations to remove DOM elements that aren't in the viewport range by scroll, which means you will unable to target elements that are further down a large document.

Done: that's exactly what I needed, thanks 🙏 . I was a bit puzzled by the philosophy of processor at first but thanks to examples (Templater was a good one) I've managed to understand how to use it


> https://github.com/mdelobelle/obsidian_supercharged_links/blob/265bfc5b5681da54ede4c4d21c3c94ea0a55d8a2/main.ts#L40
Also I highly recommend using a prefix from the frontmatter key instead of just putting every single frontmatter element into link properties. Also make sure to check that the frontmatter entry is a string - you don't want to be setting other types directly into properties.

Done. I've included a required setting so that only the attributes that are included in the settings will be added as attributes of the internal link. Also checking that frontmatter values are typeof string

The release 0.0.1 is updated accordingly

lishid:
https://github.com/mdelobelle/obsidian_supercharged_links/blob/b3ea089fde391bda8a9cb6008d0dff2fd2661792/main.ts#L1
Seems like you're still importing privateDecrypt.

https://github.com/mdelobelle/obsidian_supercharged_links/blob/b3ea089fde391bda8a9cb6008d0dff2fd2661792/main.ts#L33
I'm thinking here - your code here would basically prevent Obsidian or another plugin from adding any other attributes to links, except the ones hardcoded in your list. I think that's rather inflexible.

I would propose to require a prefix on your link attributes, like `data-link-XXX` from both the frontmatter entry, and to the actual DOM element attribute. Using this, you can filter out other frontmatter entries, as well as only removing the DOM attributes your plugin adds.

https://github.com/mdelobelle/obsidian_supercharged_links/blob/b3ea089fde391bda8a9cb6008d0dff2fd2661792/main.ts#L39
Why not just make this a plain object, like `Record<string, string>`?
(in case you don't know the shorthand, it's equivalent to `{[key: string]: string}`)

lishid:
Not sure what's going on here, but there's now yet another unused import:
https://github.com/mdelobelle/obsidian_supercharged_links/blob/17fa2585c352a47f0e394b5a2785dadd1617c4e5/main.ts#L2

mdelobelle:
Oh, neither do I...
I suspect my VSCode is autoimporting stuff when I'm calling specific functions ?
Anyway, I've removed this import and checked before committing main.ts + updated the release accordingly: this looks ok on Github.
Sorry for the noise 😞

lishid:
Ok, one last thing I think you should change is probably the plugin name. "Internal links" is a really un-descriptive name for your plugin and I think it's quite confusing for the user as to what it does.

Your repo is named "supercharged links", perhaps use that, or a variant of that?

mdelobelle:
Hello @lishid. Thanks for the advice. I've changed the plugin name to Supercharged Links

mdelobelle:
@lishid, Thanks for the merge 😀
this plugin can't be enabled through the plugin community.
main.js is missing in .obsidian/plugins/supercharged-links-obsidian (it only contains manifest.json)

Did I miss something in my release?

mdelobelle:
Ok I see, i've uploaded the main.ts instead of main.js in the release package 🤦‍♂️. Sorry about that

PR #271, opened 2021-05-02
Clear Unused Images Plugin

ozntel:
# [ X ] I am submitting a new Community Plugin

## Repo URL

https://github.com/ozntel/oz-clear-unused-images-obsidian

## Release Checklist

  - [X] `main.js`
  - [X] `manifest.json`
- [X] Github release name matches the exact version number specified in your manifest.json

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1289,5 +1289,12 @@
         "description": "Create markdown-backed Kanban boards in Obsidian",
         "repo": "mgmeyers/obsidian-kanban",
         "branch": "main"
+    },
+    {
+        "id": "oz-clear-unused-images",
+        "name": "Clear Unused Images",
+        "author": "Ozan Tellioglu",
+        "description": "Clear the images that you are not using anymore in your markdown notes to save space.",
+        "repo": "ozntel/oz-clear-unused-images-obsidian"
     }
 ]


ericaxu:
Thanks for the plugin! I'm sure many people will find it useful to keep their vaults clean.

I have a question with the implementation: instead of iterating each file and match its content against image regex (like so: https://github.com/ozntel/oz-clear-unused-images-obsidian/blob/master/src/utils.ts#L11), have you considered going through the metadata cache of each Markdown file, and see if their `embeds` field contains an image? I think that would be more performant because the image parsing has already been done when building the metadata cache.

This is not a huge concern though, since I imagine this clean up command is only run manually and won't be run very often. Still, curious what your thoughts are on this.

ozntel:
Hi @ericaxu thank you very much for your feedback.

As you mentioned, it will clean up only if someone uses the command, however, why not optimize if we have a chance 🙂

I completed the change and the plugin now loops through the embeds in the metadata cache in this function: 

https://github.com/ozntel/oz-clear-unused-images-obsidian/blob/165d9fdf0d629395e4f13f150606b2c062e8c4c7/src/utils.ts#L6

At least, it reduces the concern of finding the match for the markdown/wikilinks patterns at the beginning and I can concentrate on only the image pattern.

Please let me know if there is anything else I should consider.

Thanks!

ericaxu:
Looks good to me! I haven't tested it albeit, just looking at the code. Thanks for taking the initiatives to refactor it, really appreciate it! I hope in addition to performance, it can help reduce risk of mis-parsing the content. Great work!

lishid:
A couple of suggestions:

https://github.com/ozntel/oz-clear-unused-images-obsidian/blob/master/src/main.ts
Since your entire plugin is just mostly one function, I'd suggest merging it all in your main.ts file, instead of having most of your plugin's code in "util.ts" which is kinda misleading 😂 .

Also, if you move all the "util" functions into class functions on your plugin, then they'd all be able to use `this.app` instead of having to pass `app` around.

https://github.com/ozntel/oz-clear-unused-images-obsidian/blob/b2c028f8c61a3e7a9419e4a03037e72e4e0986fc/src/utils.ts#L10
Instead of using embeds (what if someone links to an image instead of embedding it?), you should use the `resolvedLinks` from the metadataCache: https://github.com/obsidianmd/obsidian-api/blob/48bd2b91f66aa7d3768629be48c52e5351ec6696/obsidian.d.ts#L1833

This API gives you all known resolved links and embeds, with the proper resolved file paths of source and destination files.

https://github.com/ozntel/oz-clear-unused-images-obsidian/blob/b2c028f8c61a3e7a9419e4a03037e72e4e0986fc/src/utils.ts#L27
You can just match by the file's `extension` property instead of running a regex on the path.

https://github.com/ozntel/oz-clear-unused-images-obsidian/blob/b2c028f8c61a3e7a9419e4a03037e72e4e0986fc/src/utils.ts#L35
This is quite inefficient, especially with large vaults. You can use a `Set<string>` that's the set of all files that have been linked to, then just check for each image whether its path is in the set.

https://github.com/ozntel/oz-clear-unused-images-obsidian/blob/b2c028f8c61a3e7a9419e4a03037e72e4e0986fc/src/utils.ts#L43
Don't do this in parallel. Make this an async function and use then `for (let file of list) { await app.vault.delete(file); }`

ozntel:
Hi @lishid thanks for all of the feedback 🙂 

- I merged all of the code into main.ts
- I converted `embeds` to `resolvedLinks`
- I created an Extension Set to compare rather than running a regex on the path.
- I changed the Image List into Image Path Set to check if image in the list
- I made the delete an async function to not do in parallels.

Please let me know if you have any more suggestions.
Thanks

ozntel:
Hi @lishid, I corrected the async function as you suggested and run auto-formatter to normalize the spacing.

ozntel:
Hi @lishid, thanks for merging the pull request. 

I just wanted to implement an auto-clear function to the plugin in case someone wants to use it. (As per a request from user) Therefore, created a settings tab for users' preferences. I implemented these changes in a new branch for this function just to make sure that it is safe before releasing:

https://github.com/ozntel/oz-clear-unused-images-obsidian/tree/auto-clear-feature

I have one question, though. For the cleanup during the load of the workspace, I included a time interval (3 seconds) just to make sure that the internal links are created before the plugin is run. I tried without an interval but it runs the plugin before internal links are set up. Hence, wanted to make sure if it is ok with these 3 seconds, or is there maybe any other option I can implement to make sure that the plugin waits for the internal links first?

Thanks again! 🙂

PR #280, opened 2021-05-09
Add React Components Plugin

elias-sundqvist:

## Repo URL

https://github.com/elias-sundqvist/obsidian-react-components

## Release Checklist

> Not tested on macOS, but the plugin does not use any platform specific features
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/elias-sundqvist/obsidian-react-components/blob/ea35758895a809832e9b4b4946c4d60cdb4192f2/main.tsx#L146
You should wrap these 3 events in `this.registerEvent()` to have it properly unhooked when your plugin is disabled.

https://github.com/elias-sundqvist/obsidian-react-components/blob/ea35758895a809832e9b4b4946c4d60cdb4192f2/main.tsx#L80
Might want to add an extension check here in case you get binary files like images fed in.

https://github.com/elias-sundqvist/obsidian-react-components/blob/ea35758895a809832e9b4b4946c4d60cdb4192f2/main.tsx#L39
This isn't properly typed - activeLeaf isn't guaranteed to be the leaf you're rendering, and it's not even guaranteed to contain a MarkdownView. If you need the active view, then you should use `workspace.getActiveViewOfType(MarkdownView)`, and null-check on that.

If I'm understanding the purpose, then I think this is probably unnecessary. Obsidian doesn't render the preview DOM until the user is in preview mode anyway.

lishid:
Pretty sure you want to use `getMode` instead of `getState` here: https://github.com/elias-sundqvist/obsidian-react-components/commit/dd99f5ac148c079d1e835a1434093f831e2371d1#diff-61ee70fcb1659759aa75fd34ebdeb68a2164aa59817e5be66419a100f15788ddR40

elias-sundqvist:
Thanks for the feedback!
The mentioned issues should now be resolved. 

Regarding `isPreviewMode`. I'm not entirely sure if it's needed either, but @lucasew wanted to add it (see [this PR](https://github.com/elias-sundqvist/obsidian-react-components/pull/5)), and I don't really see any harm doing so. I made your suggested modification however.

lishid:
> I've sent that PR because it was rerendering the components everytime I type something and if you have a useEffect requesting to an API you get quickly ratelimited.
> 
> A rule for react hooks is that at every rerender you must call all the same hooks in the same order so early return will break this rule so I added a function to the scope that makes easy a user check if the note is in preview mode.
> 
> Maybe that is not the optimal way to implement this but I tried and it solved for me.

I think the only problem with this is that it will not behave properly in a multi pane setup. Renderers aren't necessarily called from the currently focused/active pane, so if you're typing in an edit mode pane, then a preview-mode pane on the side will be unable to update since the active pane is an edit pane.

elias-sundqvist:
I have now changed `isPreviewMode` to `useIsPreview`. 

See code [here](https://github.com/elias-sundqvist/obsidian-react-components)

It basically does this:

```js
markdownPostProcessorContext.containerEl
                    .closest('.workspace-leaf-content')
                    .getAttribute('data-mode') === 'preview'
```

It should work fine as long as the markdown renderer is always somewhere inside a DOM element with the `'.workspace-leaf-content'` class, and with a `'data-mode'` attribute.  

Is this a safe assumption?

lishid:
As far as I can tell, because we only render while the preview view is actually attached to the DOM and actively displayed, you will find that condition to be always true. The renderer component stops rendering new sections as soon as it switches out of preview mode and don't do anything until it's switched back.

elias-sundqvist:
It can actually be false. If you render the components in preview mode and switch to edit mode they are still there in the background. Not sure if that's a problem or not.

elias-sundqvist:
isPreviewExample

lishid:
Ohhh you're not calling it directly in the rendering phase, but afterwards! Ok yeah that looks good then.

PR #284, opened 2021-05-11
Update community-plugins.json

flip-md:
## Repo URL

## Repo URL
https://github.com/flip-md/obsidian-inline-variables/

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1409,5 +1409,12 @@
         "author": "Tokuhiro Matsuno",
         "description": "This plugin will display the page linked 2hop ahead.",
         "repo": "tokuhirom/obsidian-2hop-links-plugin"
+    },
+        {
+        "id": "obsidian-inline-variables",
+        "name": "Inline Variables",
+        "author": "flip",
+        "description": "Enables setting, autocompleting and rendering inline variables.",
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/flip-md/obsidian-inline-variables/blob/d70b79600c32023cb0df0df61b746f225da0826d/main.ts#L15

Please unregister your CodeMirror event in the onunload function using `workspace.iterateCodeMirrors`.

https://github.com/flip-md/obsidian-inline-variables/blob/d70b79600c32023cb0df0df61b746f225da0826d/main.ts#L22

This seems rather unnecessary. The easiest way I can think of to achieve the same thing is to create an instance of your object inside the closure right before you register the event, so that you have a copy of it for each CodeMirror instance.

lishid:
Since you're using the CodeMirror 5 API, you should also mark your plugin as desktop only here: https://github.com/flip-md/obsidian-inline-variables/blob/d70b79600c32023cb0df0df61b746f225da0826d/manifest.json#L9

ericaxu:
Will be reopened when the comments get resolved.

PR #285, opened 2021-05-11
Add Tasks plugin

schemar:

Thank you for taking the time to consider this.
This is a complete rewrite since my last PR for obsidian tasks. It now uses API features of obsidian 0.12.

When I opened the previous PR, obsidian tasks was a proof of concept that I wasn't very confident in.
The rewrite has a better foundation. There are already some users, so I am confident it isn't fundamentally broken :sweat_smile: 

Happy to help if anything is unclear.

## Repo URL
https://github.com/schemar/obsidian-tasks

## Release Checklist
- [x] I have tested this on Linux and iOS
  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css`


lishid:
https://github.com/schemar/obsidian-tasks/blob/f5f0432e8e5ddff9e75dbe4e32f38524c906fc7f/src/QueryRenderer.ts#L152
Don't use innerHTML - it's insecure and could lead to DOM injection attacks. It's also inefficient the way you're using innerHTML because you're asking the engine to repeatedly parse some HTML (although not really much)


https://github.com/schemar/obsidian-tasks/blob/f5f0432e8e5ddff9e75dbe4e32f38524c906fc7f/src/QueryRenderer.ts#L87
Same as above.

https://github.com/schemar/obsidian-tasks/blob/d8ae9de5dee301fbef7dae4d41fc7116dfe631ba/src/Task.ts#L185
Same, here you'll probably want to use something like
```ts
while (oldParent.firstChild) newParent.appendChild(oldParent.firstChild);
```

https://github.com/schemar/obsidian-tasks/blob/f5f0432e8e5ddff9e75dbe4e32f38524c906fc7f/src/Query.ts#L2
Here you should probably use Obsidian's copy of Moment.js rather than importing a copy of your own, which should reduce your bundle size. (Use window.monent, see the calendar plugin on how to use the global moment.js).

https://github.com/schemar/obsidian-tasks/blob/f5f0432e8e5ddff9e75dbe4e32f38524c906fc7f/src/Cache.ts#L137
Here you aren't properly unloading these next 3 events because your unload script does `metadataCache.offref` instead of `vault.offref`

https://github.com/schemar/obsidian-tasks/blob/f5f0432e8e5ddff9e75dbe4e32f38524c906fc7f/src/Commands.ts#L154
You're probably looking for `workspace.getActiveViewOfType(MarkdownView)` here.

schemar:
Thank you for the quick response and review, Shida. I addressed all your points in the latest release `0.8.2`.

- [x] Removed all usages of `innerHTML`
- [x] Use obsidian's moment to reduce bundle size
- [x] Properly deref vault and cache event references (and fix some bugs I found on the way :see_no_evil: )
- [x] Switched to `editorCheckCallback`, so no more need for my own `getActiveMarkdownView`

I forgot I also had some questions. If you have the time, please help me by shedding some light on those :slightly_smiling_face: 

1. Is it ok to add a normal `onclick` listener to the checkboxes of the `li` elements of a task? I saw that other code uses specific methods to create listeners to ensure that they are properly removed. https://github.com/schemar/obsidian-tasks/blob/f5f0432e8e5ddff9e75dbe4e32f38524c906fc7f/src/Task.ts#L196
2. Is it ok to pass `null` as the `Component` to the `MarkdownRenderer`? The type definition explicitly asks for a `Component`. https://github.com/schemar/obsidian-tasks/blob/f5f0432e8e5ddff9e75dbe4e32f38524c906fc7f/src/Task.ts#L179
4. Is it correct that `metadataCache.getFileCache(file)` sometimes returns`undefined` even though `file` is guaranteed to be of type `TFile`? That's not documented in the type definitions. I don't have it myself, but one of my users reported it. I built a "retry" as a workaround which mitigates the issue, as the return value will be the cache instead of `undefined` in later tries. Is this intended? Documented? https://github.com/schemar/obsidian-tasks/blob/main/src/File.ts#L98
5. Is it already possible to set the icon of a command in the mobile toolbar? I added tasks' commands, but the icon is a generic question mark.

Thank you so much for your time :pray:

pjeby:
The cache issue could be one of two things: either the file is being updated, or it's not a markdown file.  If it is a markdown file, then the issue is that you need to wait for the cache to be clean; it has a `onCleanCache(callback)` method that will do this for you; the callback will be called either *immediately* (if the cache is up to date) or it will happen later when the cache's `resolved` event is fired.  If you don't want to use the undocumented API, you could replace your retry logic with a promise that waits for the resolved event, e.g:

```js
await new Promise(res => { const ref = metadataCache.on("resolved", () => metadataCache.offref(ref); res(); });
```

This promise will resolve when the cache is clean: if getFileCache() still doesn't return anything then, it probably never will.

(Note, however, that you should only perform this await if getFileCache() didn't return something the first time, otherwise you will be waiting until some other file gets changed!)

lishid:
> Is it ok to add a normal onclick listener to the checkboxes of the li elements of a task? I saw that other obsidian code provides specific methods to create listeners to ensure that they are properly removed. https://github.com/schemar/obsidian-tasks/blob/f5f0432e8e5ddff9e75dbe4e32f38524c906fc7f/src/Task.ts#L196

That should be fine. Attaching listeners to the specific element should have it GC'd when it's removed.

> Is it ok to pass null as the Component to the MarkdownRenderer? The type definition explicitly asks for a Component. https://github.com/schemar/obsidian-tasks/blob/f5f0432e8e5ddff9e75dbe4e32f38524c906fc7f/src/Task.ts#L179

It's coded defensively to not fail without one, so you should be fine. There might be some event listener runoffs if you don't provide one that properly unloads when your component does, but I can take care of that later.

> Is it correct that metadataCache.getFileCache(file) sometimes returnsundefined even though file is guaranteed to be of type TFile? That's not documented in the type definitions. I don't have it myself, but one of my users reported it. I built a "retry" as a workaround which mitigates the issue, as the return value will be the cache instead of undefined in later tries. Is this intended? Documented? https://github.com/schemar/obsidian-tasks/blob/main/src/File.ts#L98

pjeby has a great explanation already, but yeah you should check for `tfile.extension === 'md'` first to avoid running into non-indexable files. You may want to go one step further and create a full lifecycle-managed component (`ctx.addChild(yourComponent)`) which then uses event listeners to re-render/update your view when metadata cache detects changes to the file (`resolve` event). This would be similar to how search queries work, but it is certainly going to make things more complicated.


> Is it already possible to set the icon of a command in the mobile toolbar? I added tasks' commands, but the icon is a generic question mark.

You should add the icon svg first along with an ID to our registry (search for the function `addIcon`), then you can use that ID here.

schemar:
Thank you both for the detailed explanations!

Would it be ok to merge “as-is”?

I still want to change the mobile toolbar icon (to be the same as the existing “toggle task” one), but I can do that with a later update.

Regarding the handling of the metadata cache: I would prefer to leave it as is, because
a) it is simple and works and
b) I cannot test an alternative solution as I can not reproduce the issue locally

All other topics are fine, if I am not overlooking something.

Thanks!

schemar:
I added the file extension check in the latest release `0.8.3`

PR #289, opened 2021-05-14
add better-fn

AidenLx:

## Repo URL

<https://github.com/alx-plugins/better-fn>

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css`


AidenLx:
In addition, I wonder if it is possible to re-render the page entirely in preview mode, similar to reopen the page. I'm trying to put this in [here](https://github.com/alx-plugins/better-fn/blob/301ba543c1aa3966d651365d6ca3ed1b3d08f1d4/src/main.ts#L61) and run it when loading and unloading to update changes in MarkdownPostProcessor to opened files. `MarkdownPreviewView.rerender()` does not seem to work.

lishid:
I'd suggest giving a better name (or at least expanding the `fn`) part because many people were confused thinking it had something to do with "function".

https://github.com/alx-plugins/better-fn/blob/bb046561b5936cf74ec5f39b14236be7a10e64e8/src/main.ts#L43
Probably easier with `src.call(view, file)`

https://github.com/alx-plugins/better-fn/blob/bb046561b5936cf74ec5f39b14236be7a10e64e8/src/main.ts#L33
I think this may not work if there is ever another view that inherits MarkdownView. (I know some plugin hackily does this).

What you could do is to store the original function's backup onto your new function (like `modified`), and then use that to revert.

https://github.com/alx-plugins/better-fn/blob/bb046561b5936cf74ec5f39b14236be7a10e64e8/src/main.ts#L68
Probably better to run the for loop outside?

https://github.com/alx-plugins/better-fn/blob/bb046561b5936cf74ec5f39b14236be7a10e64e8/src/main.ts#L94
This is not needed since registerMarkdownPostProcessor manages the unhook automatically.

https://github.com/alx-plugins/better-fn/blob/bb046561b5936cf74ec5f39b14236be7a10e64e8/src/modules/tools.ts#L2
We have a similar util on `Node.empty()` you can call directly on the element.

AidenLx:
Thanks! I've done basically all the changes you suggested, except: 
- `better-fn` is merely the internal id, the name in `manifest.json` is actually `Better Footnote`, so it should not be a problem.
- For [doAllLeaves](https://github.com/alx-plugins/better-fn/blob/bb046561b5936cf74ec5f39b14236be7a10e64e8/src/main.ts#L68): The purpose for the loop to be wrapped is to get the callback function of [event](https://github.com/alx-plugins/better-fn/blob/bb046561b5936cf74ec5f39b14236be7a10e64e8/src/main.ts#L84). To avoid confusion, I've renamed it to `getLoopAllLeavesFunc`

Also, is there a way to achieve this? 

> In addition, I wonder if it is possible to re-render the page entirely in preview mode, similar to reopen the page. I'm trying to put this in [here](https://github.com/alx-plugins/better-fn/blob/301ba543c1aa3966d651365d6ca3ed1b3d08f1d4/src/main.ts#L61) and run it when loading and unloading to update changes in MarkdownPostProcessor to opened files. `MarkdownPreviewView.rerender()` does not seem to work.

lishid:
- For the ID, you're right, `better-fn` is fine if you want to leave it like that.
- For doAllLeaves, I understood what you wanted it to do, I just meant swapping the order of the for loop with the iterate loop to make things simpler:
```ts
for (const action of actions) {
    this.app.workspace.iterateAllLeaves(action);
}
```
- `MarkdownPreviewView.rerender()` should do the trick, but you may want to wrap it in a `setTimeout` so that it runs after your plugin fully enables/disables. I'm not sure why it wouldn't, but based on my source code it should force a full re-render as long as the preview pane is in view.

AidenLx:
> * For doAllLeaves, I understood what you wanted it to do, I just meant swapping the order of the for loop with the iterate loop to make things simpler:

Got it! I've done it in the latest commit. 

---

> * `MarkdownPreviewView.rerender()` should do the trick, but you may want to wrap it in a `setTimeout` so that it runs after your plugin fully enables/disables. I'm not sure why it wouldn't, but based on my source code it should force a full re-render as long as the preview pane is in view.

I add the following code to [here](https://github.com/alx-plugins/better-fn/blob/301ba543c1aa3966d651365d6ca3ed1b3d08f1d4/src/main.ts#L61): 

```ts
  refresh: leafAction = (leaf) => {
    setTimeout(() => {
      if (leaf.view instanceof MarkdownView){
        console.log(leaf.view)
        leaf.view.previewMode.rerender();
      }  
    }, 1000);
  };
```

And then, I've disabled the plugin and do a full restart of Obsidian before the recording, Here is what I've got so far: 

https://user-images.githubusercontent.com/31102694/119297697-12b28780-bc8e-11eb-8263-ad9a4348f3f9.mp4

lishid:
Ah, I see what's going on here. `rerender` does a full re-parse, but still reuses existing DOM sections if they were found to be identical.

I will add a boolean switch to `rerender` on the next release to allow for a full flush.

lishid:
I don't have commit permission to your fork to fix the conflict resolution for you - I'll have it merged as soon as we fix that,

AidenLx:
Got it, it's done now.

PR #292, opened 2021-05-15
Add MetaEdit

chhoumann:

## Repo URL

https://github.com/chhoumann/MetaEdit

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


lishid:
Please modify your plugin ID to use lowercase with dashes to stay consistent.

https://github.com/chhoumann/MetaEdit/blob/e18525219ca81623ab2c98747110f5211819da6e/src/main.ts#L37

Don't think you'll want this to be included in the final export, especially the `process` part, since does not exist on mobile.

https://github.com/chhoumann/MetaEdit/blob/e18525219ca81623ab2c98747110f5211819da6e/src/main.ts#L85
You should make sure that the active file is a markdown file `(tfile.extension === 'md')` so you do not accidentally work on an image or pdf.

https://github.com/chhoumann/MetaEdit/blob/e18525219ca81623ab2c98747110f5211819da6e/src/main.ts#L120
This might error out if, for instance, the metadataCache hasn't yet cached the file. You should add a null check since getFileCache may return null.

https://github.com/chhoumann/MetaEdit/blob/e18525219ca81623ab2c98747110f5211819da6e/src/main.ts#L137
You should use `cachedRead` instead of `read` for performance reasons if you are only displaying information from the file - keep using `read` if you intend to write the data back to disk.

https://github.com/chhoumann/MetaEdit/blob/e18525219ca81623ab2c98747110f5211819da6e/src/main.ts#L20
Seems like the second parameter (file contents) is being ignored, and kanbanHelper is reading the file contents again.

chhoumann:
Thank you, @lishid, for the great feedback! I really appreciate it.
I've addressed the issues below. Most of them have been fixed in [1.6.0](https://github.com/chhoumann/MetaEdit/releases/tag/1.6.0).

- [x] Please modify your plugin ID to use lowercase with dashes to stay consistent.

- [x] Don't think you'll want this to be included in the final export, especially the `process` part, since does not exist on mobile.

- [x] You should make sure that the active file is a markdown file `(tfile.extension === 'md')` so you do not accidentally work on an image or pdf.

- [x] This might error out if, for instance, the metadataCache hasn't yet cached the file. You should add a null check since getFileCache may return null.
 
> [chhoumann/MetaEdit@`e185252`/src/main.ts#L137](https://github.com/chhoumann/MetaEdit/blob/e18525219ca81623ab2c98747110f5211819da6e/src/main.ts#L137)
> You should use `cachedRead` instead of `read` for performance reasons if you are only displaying information from the file - keep using `read` if you intend to write the data back to disk.

This is intentional. 

- [x] Seems like the second parameter (file contents) is being ignored, and kanbanHelper is reading the file contents again.

lishid:
Looking good, just a few more things I noticed:

https://github.com/chhoumann/MetaEdit/blob/0dbb13f345c2365185e6a7c9b6fe3b9de5a83e5b/src/main.ts#L163
It says `five_minutes` here but the number is actually 18 seconds 🤣 

https://github.com/chhoumann/MetaEdit/blob/0dbb13f345c2365185e6a7c9b6fe3b9de5a83e5b/src/main.ts#L150
Not sure if it's a good idea to cache files based on file name

> This is intentional.
Can you explain a bit more? `cachedRead` uses the last known file contents to Obsidian, which is also updated when the file is changed on disk. I don't see that you're writing the file back to disk here either.

https://github.com/chhoumann/MetaEdit/blob/0dbb13f345c2365185e6a7c9b6fe3b9de5a83e5b/src/main.ts#L18
Actually I don't think this works properly - debounce only lets through one call, but the modify event can be called for many different files. In case that you've made modifications to multiple files within your 4s delay, you will lose all but one of them.

I think your best bet here is to use debounce to call some kind of update function that iterates through your file cache, and use the `modify` event to invalidate the cache.

chhoumann:
> Looking good, just a few more things I noticed:

> [chhoumann/MetaEdit@`0dbb13f`/src/main.ts#L163](https://github.com/chhoumann/MetaEdit/blob/0dbb13f345c2365185e6a7c9b6fe3b9de5a83e5b/src/main.ts#L163)
> It says `five_minutes` here but the number is actually 18 seconds 🤣

Omg😂😂


> [chhoumann/MetaEdit@`0dbb13f`/src/main.ts#L150](https://github.com/chhoumann/MetaEdit/blob/0dbb13f345c2365185e6a7c9b6fe3b9de5a83e5b/src/main.ts#L150)
> Not sure if it's a good idea to cache files based on file name

I've changed it to the file path, instead. I know that the path may be prone to change, however, it shouldn't pose much of a problem. The cache is only there to prevent an endless `onModify` loop.

> > This is intentional.
> > Can you explain a bit more? `cachedRead` uses the last known file contents to Obsidian, which is also updated when the file is changed on disk. I don't see that you're writing the file back to disk here either. 

I've changed how this is handled in the new version. The `onModifyProxy` is no longer used.

The functions I want to call when modify events occur should nevertheless read files, otherwise conflicts can occur. It cannot be a cached file either, because they all have a chance to write the contents to the file again, in case it detects the need to.

> [chhoumann/MetaEdit@`0dbb13f`/src/main.ts#L18](https://github.com/chhoumann/MetaEdit/blob/0dbb13f345c2365185e6a7c9b6fe3b9de5a83e5b/src/main.ts#L18)
> Actually I don't think this works properly - debounce only lets through one call, but the modify event can be called for many different files. In case that you've made modifications to multiple files within your 4s delay, you will lose all but one of them.
> 
> I think your best bet here is to use debounce to call some kind of update function that iterates through your file cache, and use the `modify` event to invalidate the cache.


You're right. I've opted to use both the cache and a queue.

https://github.com/chhoumann/MetaEdit/blob/3706703803415bd879c6c162ba0e831db968b734/src/main.ts#L151


Thank you again for the feedback. I appreciate you taking the time! :)

lishid:
> The functions I want to call when modify events occur should nevertheless read files, otherwise conflicts can occur. It cannot be a cached file either, because they all have a chance to write the contents to the file again, in case it detects the need to.

I think there's a misunderstanding of what `cachedRead` does. It does not give you an old/outdated copy of the file - it's as recent as we know it. If a file is written through the `Vault` or its underlying `Adapter` interface, then `cachedRead` will always give you the latest file contents.

The only time you get different results when using `cachedRead` is briefly when another process (Sync tools like Dropbox, or Git) makes the change to the file, but before the OS fires the file changed event to let Obsidian know the file has changed externally. Changes from Obsidian and plugins are always fully reflected from `cachedRead`.

If you're monitoring the `modify` event, then you are guaranteed to read the latest version of the file using `cachedRead` (or else another `modify` event would be triggered afterwards).

Does this help?

chhoumann:
> > The functions I want to call when modify events occur should nevertheless read files, otherwise conflicts can occur. It cannot be a cached file either, because they all have a chance to write the contents to the file again, in case it detects the need to.
> 
> I think there's a misunderstanding of what `cachedRead` does. It does not give you an old/outdated copy of the file - it's as recent as we know it. If a file is written through the `Vault` or its underlying `Adapter` interface, then `cachedRead` will always give you the latest file contents.
> 
> The only time you get different results when using `cachedRead` is briefly when another process (Sync tools like Dropbox, or Git) makes the change to the file, but before the OS fires the file changed event to let Obsidian know the file has changed externally. Changes from Obsidian and plugins are always fully reflected from `cachedRead`.
> 
> If you're monitoring the `modify` event, then you are guaranteed to read the latest version of the file using `cachedRead` (or else another `modify` event would be triggered afterwards).
> 
> Does this help?

Thank you! That does clarify things. The primary reason why I've been using `read` is that I heard that 'if you only need to read, use `cachedRead`, but if you need to write, use `read`.'

Nevertheless! I'll use `cachedRead` instead. It does make more sense now that you've clarified it.

Thanks again :)

lishid:
Hmm I think I still haven't gotten it across properly - you should still use `read` if you intend to write the data back. My original comment was specifically towards this read which seemed to be used only for displaying information: https://github.com/chhoumann/MetaEdit/blob/e18525219ca81623ab2c98747110f5211819da6e/src/main.ts#L137

In the end, I think my best judgement is to revert to using `read` for the usages in metaController.ts but keep `cachedRead` in parser.ts

chhoumann:
Ah, I think I understand you now. I've updated accordingly: https://github.com/chhoumann/MetaEdit/commit/7924814d1b66bd7ac14348fe7bef09c73a345cd5

MetaController will use `read` as it writes back to the file. `cachedRead` will be used in other places, where the read data is merely displayed.

Thank you for your patience :)

PR #293, opened 2021-05-17
Add argdown plugin

amdecker:

## Repo URL

https://github.com/amdecker/obsidian-argdown-plugin

## Release Checklist

- [x] I have tested this on Windows, macOS,
  - [x] `main.js`
  - [x] `manifest.json`


ericaxu:
Great plugin! I remember seeing requests for this :)

https://github.com/amdecker/obsidian-argdown-plugin/blob/master/main.ts#L35
Would you consider renaming this, assuming `testing` was a temporary name?

https://github.com/amdecker/obsidian-argdown-plugin/blob/master/main.ts#L42
What is this hardcoded timeout for? Do we need to change it somehow for slower environments? (Just brainstorming)

amdecker:
Oops! good catch. Renamed that function. 

The timeout isn't strictly necessary, but I put it in so that the argdown processor would not rerun on every keypress, but only after 350ms of no typing. That should make things smoother on slower machines.

ericaxu:
Got it!

@lishid Could you take a look please?

lishid:
https://github.com/amdecker/obsidian-argdown-plugin/blob/7e3efcf56bc37cdf8a43417570a8bc7628f5b007/main.ts#L23
Wondering what this does? I suppose it makes CodeMirror re-render everything? Suggest adding a comment for it.

https://github.com/amdecker/obsidian-argdown-plugin/blob/7e3efcf56bc37cdf8a43417570a8bc7628f5b007/main.ts#L34

You can type this to `number` and use `window.setTimeout` instead of `setTimeout` to fix the typing error, which are the proper types in this case.

https://github.com/amdecker/obsidian-argdown-plugin/blob/7e3efcf56bc37cdf8a43417570a8bc7628f5b007/util.ts#L3
I don't think this is a good idea - the app's folder may not permit writing to. Best way is to use the temp folder, or a `.temp` folder in the vault's dir. On some linux distros, it's not guaranteed that `__dirname` is where you expect it to be.

https://github.com/amdecker/obsidian-argdown-plugin/blob/7e3efcf56bc37cdf8a43417570a8bc7628f5b007/util.ts#L24
Use `await fs.promises.writeFile(file, fileContent)` - your promise is never resolved if an error happens because your throw happens in a callback which goes unhandled.

https://github.com/amdecker/obsidian-argdown-plugin/blob/7e3efcf56bc37cdf8a43417570a8bc7628f5b007/util.ts#L9
https://github.com/amdecker/obsidian-argdown-plugin/blob/7e3efcf56bc37cdf8a43417570a8bc7628f5b007/util.ts#L37
You should probably sanitize the input parameters to avoid command injection here.

https://github.com/amdecker/obsidian-argdown-plugin/blob/7e3efcf56bc37cdf8a43417570a8bc7628f5b007/main.ts#L38
I think this won't work if you have multiple code blocks, or multiple panes each with an argdown block. I'd suggest adding a button so the user can manually force a re-render, and remove it when you finish rendering.

amdecker:
Thanks so much for the feedback it definitely makes my plugin better!!

I made those changes. 

To avoid command injection I wrapped all my variable commands with quotes as per https://stackoverflow.com/questions/4273074/sanitize-user-input-in-bash-for-security-purposes. 

I reconfigured how I do the rerender to allow for multiple code blocks automatically.

I also added a Notice that pops up when there is an argdown syntax error and it cannot process the codeblock.


If there is anything else let me know. I tested these changes on windows + mac and updated my existing release.

lishid:
> To avoid command injection I wrapped all my variable commands with quotes as per https://stackoverflow.com/questions/4273074/sanitize-user-input-in-bash-for-security-purposes.

What you're doing doesn't actually work - The SO answer explains that they're quoting a (bash) variable, which causes the shell to automatically escape the string. You're injecting the user provided text directly into the javascript string before it's being ran, which means one can easily get out of your escaped string by using `"` inside the filename or path name.

amdecker:
Ok I took care of single and double quotes in file paths.

lishid:
You might want to make a utility to do that since there's still a bunch of places that aren't sanitized:

https://github.com/amdecker/obsidian-argdown-plugin/blob/45cee935d938db2d3c8c35c87bbd68fba300823b/util.ts#L22
https://github.com/amdecker/obsidian-argdown-plugin/blob/45cee935d938db2d3c8c35c87bbd68fba300823b/util.ts#L76
https://github.com/amdecker/obsidian-argdown-plugin/blob/45cee935d938db2d3c8c35c87bbd68fba300823b/util.ts#L78
https://github.com/amdecker/obsidian-argdown-plugin/blob/45cee935d938db2d3c8c35c87bbd68fba300823b/util.ts#L82
https://github.com/amdecker/obsidian-argdown-plugin/blob/45cee935d938db2d3c8c35c87bbd68fba300823b/util.ts#L84

I think you might also want to make sure to escape additional characters like `$` on unix and `%` on windows so it doesn't do something unintentional.

https://unix.stackexchange.com/questions/270977/what-characters-are-required-to-be-escaped-in-command-line-arguments

amdecker:
Updated code + edited the release with the sanitization function

amdecker:
Thanks for all the help! I noticed that my demo gif doesn't show up when I view it in the browse community plugin area -- it is ~2mb. Is there a size limit or is something else preventing it from showing up?

lishid:
I think you need to link to it using a raw.githubusercontent.com link instead of a relative link.

PR #294, opened 2021-05-18
Add Pandoc export plugin

OliverBalfour:

## Repo URL

https://github.com/OliverBalfour/obsidian-pandoc

## Release Checklist

- [x] I have tested this on Windows, macOS, and Linux
  - [x] `main.js`
  - [x]  `manifest.json`
  - [x] `styles.css`

This plugin has been around for a while and has a few users, so even though there are some rough edges (eg lots of ways to get ugly exports) I thought I'd submit it. I intend to keep maintaining it after my uni semester is over and I have some semblance of free time 😛.


ryanpcmcquen:
I think you'll want to use `this.app.vault.adapter` for file operations, rather than importing `fs` directly.

lishid:
https://github.com/OliverBalfour/obsidian-pandoc/blob/master/main.js
Please add this to your gitignore and remove it from the repo.

https://github.com/OliverBalfour/obsidian-pandoc/blob/7eaacb949cecf2b34670f7cbb0e206622c0977c2/main.ts#L29
Don't await for this result while in the onload of your plugin, because that will delay the initialization of the entire app and other plugins, causing extra loading time. Just run it without await and let it process in the background.

https://github.com/OliverBalfour/obsidian-pandoc/blob/7eaacb949cecf2b34670f7cbb0e206622c0977c2/main.ts#L63
This is incorrect and will result in failed path conversions. Obsidian's normalized file paths may be different than the file paths in the file system, due to some conversion and normalization steps like non-breaking spaces becoming regular spaces, and unicode normalization differences. You can use `filesystemAdapter.getFullPath(normalizedPath)` to get the actual path, rather than getting the base path and performing the join yourself.

https://github.com/OliverBalfour/obsidian-pandoc/blob/7eaacb949cecf2b34670f7cbb0e206622c0977c2/main.ts#L92
Don't cast this to any. Perform instanceof checks whenever unsure. Here you probably want `workspace.getActiveViewOfType(MarkdownView)` instead of using activeLeaf directly (which might be null!)

Also this seems to only work for markdown, but your other functions seems to suggest that the command can run for other formats: https://github.com/OliverBalfour/obsidian-pandoc/blob/7eaacb949cecf2b34670f7cbb0e206622c0977c2/pandoc.ts#L34

https://github.com/OliverBalfour/obsidian-pandoc/blob/7eaacb949cecf2b34670f7cbb0e206622c0977c2/renderer.ts#L29
You should rather pass null instead of `{} as Component` because this will definitely error out with other plugins that adds component lifecycles and expecting the parameter to be an actual Component.

OliverBalfour:
Thanks for the feedback!
I'll fix these issues in 2-3 weeks time (it's exam period at uni currently lol) and let you know when I'm done.

OliverBalfour:
Hi,
Everything above has been fixed, except the note about how the command can run for other formats, this is a feature I'll take advantage of down the line.

My plugin is ready to be reviewed again.

PR #295, opened 2021-05-18
Add obsidian-collapse-all plugin

OfficerHalf:

## Repo URL

https://github.com/OfficerHalf/obsidian-collapse-all

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1510,5 +1510,13 @@
         "author": "Viacheslav Slinko",
         "repo": "vslinko/obsidian-zoom",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-collapse-all-plugin",
+        "name": "Collapse All",
+        "description": "This plugin adds a button to collapse all folders in the file explorer.",
+        "author": "OfficerHalf",
+        "repo": "OfficerHalf/obsidian-collapse-all",
+        "branch": "main"
     }
 ]


OfficerHalf:
Updated to `main` branch, removed auto-format changes.

pjeby:
Rather than relying on timing, you may wish to manipulate the view directly; as an experiment, I added the following to a plugin I'm working on, and it appears to instantly and correctly collapse or expand all folders:

```js
collapseAll(state=true) {
    for(const leaf of this.app.workspace.getLeavesOfType("file-explorer")) {
        for(const item of Object.values(leaf.view.fileItems)) {
            if (item.file.children && item.file.path != "/" && item.collapsed !== state) item.setCollapsed(state);
        }
    }
}
```
I did not extensively test this, and certainly not with the note folders plugin, so you may want to investigate a bit.

While this does require using certain internal APIs of the explorer view, the UI approach is likely more fragile since an Obsidian update could break either approach, but the click-based approach suffers from timing issues that this approach does not.

(Note: the above snippet is JS, not Typescript, so you'll proably need to use `(leaf.view as any).fileItems` to get access to that property.)

OfficerHalf:
I think that would both remove the need for both settings parameters _and_ fix the incompatibility. I was trying to stay away from undocumented APIs (hence the `TODO:` comment where I used a property of the file-explorer leaf that wasn't on the interface), but it's likely worth it in this case.

@pjeby you mentioned adding that in a plugin you're working on - would it supersede this one? If so I can just abandon this entirely.

OfficerHalf:
@pjeby Updated with that strategy. It greatly simplifies the whole thing, and is what I really wanted to do in the first place. Until those properties are publicly documented, I've created my own types. 🙂

pjeby:
No, that plugin would not supersede this one, especially since it's rather far from release.  (It's called Folder Focus, and it's a tool for hiding folders other than the one you pin as a "focus", along with some commands for e.g. focusing the folder of the current file, picking a folder to focus by name, and things like that.  But it does have code that uses file-explorer internals and so it was a handy place to sketch a command like this.  The final version might include a collapse-all command, but probably not a button.)

Oh, also, rather than using the `active-leaf-change` event, you might want to instead use onLayoutReady (for initialization) and register a `layout-change` handler (for when panes are added or removed, or a new workspace is loaded).  That will ensure you get a chance to modify every file explorer view.  (Since the view can be closed or opened, and different workspaces could have one either open or not open.  At present, you are only modifying once after load, so a close and reopen would result in an explorer with no collapse button.)

Last, but not least, you should remove your button in the plugin's `onunload()` method, so if your plugin is deactivated it won't leave the old button behind.  (This is also important for upgrades, which deactivate and reactivate the plugin.)

OfficerHalf:
@pjeby Updated:

- Initialize `onLayoutReady`
- Listen for `layout-change` event
- Register `layout-change` event to be automatically unloaded
- Remove icon on unload.

pjeby:
Looks good to me!

OfficerHalf:
@pjeby thanks for all the help. Better docs would be great, but there's no substitute for someone who is very familiar with how things work. 🙂

lishid:
https://github.com/OfficerHalf/obsidian-collapse-all/blob/70ba0fecb8bb55c98275ce0ae4fb94651dab68bc/src/plugin.ts#L49
Maybe `leaf.view.containerEl`?

https://github.com/OfficerHalf/obsidian-collapse-all/blob/70ba0fecb8bb55c98275ce0ae4fb94651dab68bc/src/plugin.ts#L50
Might want a null check on this.

https://github.com/OfficerHalf/obsidian-collapse-all/blob/70ba0fecb8bb55c98275ce0ae4fb94651dab68bc/src/plugin.ts#L63
This collapse all button should only work on this specific leaf (even though it's unlikely there are multiple explorer panes).

https://github.com/OfficerHalf/obsidian-collapse-all/blob/70ba0fecb8bb55c98275ce0ae4fb94651dab68bc/src/plugin.ts#L121
That's not a great way haha - probably better to use `item.file instanceof TFolder`.

https://github.com/OfficerHalf/obsidian-collapse-all/blob/70ba0fecb8bb55c98275ce0ae4fb94651dab68bc/src/plugin.ts#L53
I'd suggest using a more unique css name in case it'd conflict with future updates. Something like `plugin-collapse-all-button`.

OfficerHalf:
@lishid addressed those comments. Each button now affects only the matching file explorer, though the command still closes all folders in all file explorers. Also updated to resolve conflicts with `community-plugins.json`

lishid:
Great work! I'm sure it won't take long for you to receive a feature request for the ability to expand all 😂

PR #303, opened 2021-05-21
Add Pluck plugin

kevboh:

## Repo URL

https://github.com/kevboh/obsidian-pluck

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1467,5 +1467,13 @@
         "description": "obsidian workspace management",
         "repo": "manogna4/obsidian-koncham-workspace",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-pluck",
+        "name": "Pluck",
+        "author": "Kevin Barrett",
+        "description": "Quickly create notes in Obsidian from web pages.",
+        "repo": "kevboh/obsidian-pluck",
+        "branch": "main"
     }
 ]


Vinzent03:
Why are you using `editor.getDoc()`? Isn't `editor` already what you need? 
https://github.com/kevboh/obsidian-pluck/blob/master/main.ts#L68

kevboh:
> Why are you using `editor.getDoc()`? Isn't `editor` already what you need? 
> 
> https://github.com/kevboh/obsidian-pluck/blob/master/main.ts#L68

I copied this pattern from Templater; happy to use `editor` directly if that's the recommended approach.

lishid:
https://github.com/kevboh/obsidian-pluck/blob/4879e8484ed891b6b01f084102ffe9eef130a73d/main.ts#L18

Don't do this. This would modify the behavior of everything inside Obsidian, including making the use of `<iframe>`s insecure. Instead, you can look into using electron's `net.request` (https://www.electronjs.org/docs/api/net) (which uses chromium's connection mechanism, including using the OS proxy settings) or `node-fetch` (which uses node's `http`/`https` libs underneath, ignoring proxy settings).

https://github.com/kevboh/obsidian-pluck/blob/4879e8484ed891b6b01f084102ffe9eef130a73d/main.ts#L61
You should probably test this condition first before attempting to download the file.

https://github.com/kevboh/obsidian-pluck/blob/4879e8484ed891b6b01f084102ffe9eef130a73d/main.ts#L68
`getDoc` is unnecessary as it's just `getDoc() { return this }`.

https://github.com/kevboh/obsidian-pluck/blob/4879e8484ed891b6b01f084102ffe9eef130a73d/main.ts#L73
Seems like you just need a regular `Modal` with a textbox? `FuzzySuggestModal` is definitely not what you're looking for.

kevboh:
> Instead, you can look into using electron's net.request (https://www.electronjs.org/docs/api/net) (which uses chromium's connection mechanism, including using the OS proxy settings) or node-fetch (which uses node's http/https libs underneath, ignoring proxy settings).

Oddly, I tried node-fetch first, and it was also hitting CORS errors. It's possible I misconfigured it.

> You should probably test this condition first before attempting to download the file.

Great idea, thanks.

> getDoc is unnecessary as it's just getDoc() { return this }.

Good to know. As I said above, I copied that code straight [from Templater](https://github.com/SilentVoid13/Templater/blob/291d3baeaf4cbf4225079f00dcea356d6e9532ed/src/Templater.ts#L89-L91).

> Seems like you just need a regular Modal with a textbox? FuzzySuggestModal is definitely not what you're looking for.

Fair enough. I really like the UI of `FuzzySuggestModal`, it feels faster and more vs-code like to me. But I can switch it over.

Thanks for the review!

kevboh:
I've updated this plugin to address the issues raised here and [opened a new PR](https://github.com/obsidianmd/obsidian-releases/pull/336) for the new version.

PR #304, opened 2021-05-22
Add obsidian static file server

elias-sundqvist:

https://github.com/elias-sundqvist/obsidian-static-file-server

## Release Checklist

- I have tested this on Windows, macOS, and Linux _(if applicable)_
  - [x] Windows
  - [ ] macOS
  - [ ] Linux 
  - I can only test Windows, but the code does not contain anything platform-specific. 
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
Given this plugin only works on desktop as an electron app, you should set isDesktopOnly to true: https://github.com/elias-sundqvist/obsidian-static-file-server/blob/35a574f69f90baf2e8f25cf51cbf2d842bc25501/manifest.json#L9

https://github.com/elias-sundqvist/obsidian-static-file-server/blob/35a574f69f90baf2e8f25cf51cbf2d842bc25501/static-server.ts#L32
Here you should test for `instanceof TFile` instead of blindly assuming it's a TFile. Also should check for null.

https://github.com/elias-sundqvist/obsidian-static-file-server/blob/35a574f69f90baf2e8f25cf51cbf2d842bc25501/static-server.ts#L35
Did you intend to use `mtime` instead? `ctime` doesn't get updated normally.

https://github.com/elias-sundqvist/obsidian-static-file-server/blob/35a574f69f90baf2e8f25cf51cbf2d842bc25501/static-server.ts#L32
You should run `normalizePath` on the input here to normalize any path delimiters.

elias-sundqvist:
Thanks for your feedback.
The mentioned things should be fixed now. 

Regarding`isDesktopOnly`. I believe you that it doesn't work on mobile, but what is the reason? Is the HTTP module not supported?

lishid:
On mobile there is no NodeJS support - all `require('package')` for node packages will fail and return undefined.

Unfortunately, I don't think there's even a way to host an http server on mobile from javascript at all.

elias-sundqvist:
Ok, that's a shame. 
How does the built-in PDF reader work? Is that mobile compatible?

I just want to be able to create iframes that can resolve relative links (in style/script tags and fetch/XMLHttpRequest calls) to some folder in the vault. 
I can't think of a way for doing that without a web server. (*maybe using a serviceWorker, but I haven't figured out how to do that yet.)

I'll try to come up with a better solution, but I have changed `isDesktopOnly` to true for now.

lishid:
The built-in PDF viewer uses electron's native PDF renderer if it's available and falls back to a PDF.js (canvas+webworker) based renderer.

I can't think of a good way to do that on mobile unfortunately, because the path resolutions are pretty weird and subject to restrictions from Android and/or iOS.

lishid:
https://github.com/elias-sundqvist/obsidian-static-file-server/blob/bc389051ae573b07e56afbb89a0dec83af3af901/static-server.ts#L31
This should be named `file` instead of `abstractPath`.

lishid:
I've also updated your plugin description to `Static File Server` instead of `React Components`

PR #309, opened 2021-05-25
Add 'Dropbox Backups' plugin.

ryanpcmcquen:
- [x] I am submitting a new Community Plugin

## Repo URL

https://github.com/ryanpcmcquen/obsidian-dropbox-backups
## Release Checklist

- [x] I have tested this on Linux (The 0.12.3 Snap works on Kubuntu 20.04, the AppImage doesn't seem to understand how to open Obsidian with the `obsidian://` protocol)
  - [x] `main.js`
  - [x] `manifest.json`


ryanpcmcquen:
I _think_ this will work on mobile too (that's my main motivation, honestly). I have tested it with `app.emulateMobile(true)`, but I don't have a way to test it natively until it gets merged ...

ryanpcmcquen:
@lishid:
I wasn't sure whether `window.location.assign` or `window.open` would be better here. It seems that `window.open` would be more accurate for what I am doing, but I'm also not sure what is better for mobile:
https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/b747c8e3811f7c7f138b0147ddc1006c561f25f6/main.ts#L106

I've went back and forth on `read` vs. `cachedRead`:
https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/b747c8e3811f7c7f138b0147ddc1006c561f25f6/main.ts#L24

Since this is background/asynchronous/non-blocking I figured `read` would be a better fit, but I am open to changing it and did read your comment here (which definitely clarified some things!):
https://github.com/obsidianmd/obsidian-releases/pull/292#issuecomment-848068970

lishid:
Apologies for the late review.

> I wasn't sure whether window.location.assign or window.open would be better here. It seems that window.open would be more accurate for what I am doing, but I'm also not sure what is better for mobile (I ended up switching to window.open in a later commit):

`window.open` is the right API to use. It will open the browser correctly.

> I've went back and forth on read vs. cachedRead:

I think the right API to use is `read` since you want a guaranteed latest version.

https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/95a7e556de0f871768dc8c57ee3f325de0717408/main.ts#L61
I don't think you need to import Bluebird for this. JS promises can easily do that - the replacement is just `await Promises.all(array.map(async (obj) => {}));` but be really careful here! You'll be processing everything in parallel which may not be ideal - imagine concurrently uploading thousand of files.

https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/95a7e556de0f871768dc8c57ee3f325de0717408/main.ts#L28
You may not want to call `read` on non-utf-8 files - especially images and other types.

https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/95a7e556de0f871768dc8c57ee3f325de0717408/main.ts#L201
You'll probably want the user to be able to customize this.

https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/95a7e556de0f871768dc8c57ee3f325de0717408/main.ts#L38
I don't think this is the right approach here. Reading all files into memory, and then uploading them all concurrently, isn't a great way to scale. I think your best bet is to rather come up with a list of files, then upload them one-by-one. The only thing that you'll have to be mindful of is if files are deleted while the upload process happens. In that case, you can catch the exception, and verify with `vault.exists(file.path)` to ignore the error if the file doesn't exist anymore.

https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/95a7e556de0f871768dc8c57ee3f325de0717408/main.ts#L50
You can get a reference to your ribbon icon returned from when it was created here: https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/95a7e556de0f871768dc8c57ee3f325de0717408/main.ts#L190

https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/95a7e556de0f871768dc8c57ee3f325de0717408/main.ts#L194
Don't await on long-running operations in your plugin's onload. That will delay the entire app's loading sequence, including other plugins. Do this asynchronously, or just skip doing this until it's needed (lazy).

ryanpcmcquen:
Will make the adjustments soon, thank you! I imported Bluebird to limit concurrency because otherwise Dropbox will throw 429 errors. I know this is possible without Bluebird but I've used Bluebird a lot for this specific reason.

I think I'll add a Dropdown with a few more settings besides 10 minutes. Maybe 30 and 60 minutes as well and 'Never', meaning ad hoc backups only.

lishid:
Ah I realize now you were using `concurrency: 1`. Why not just a plain and simple for loop then?
```ts
for (let file of files) {
  await upload(file);
}
```

ryanpcmcquen:
That's a great idea! It's funny, I feel like `for` loops have become so non-idiomatic in JavaScript that I forget about them, although I write them in C all the time ...

lishid:
I've avoided using `forEach` and opted for `for-of` loops ever since it was available in ES6. Some discussions about it in another plugin review (towards the end of the comment https://github.com/obsidianmd/obsidian-releases/pull/317#issuecomment-853553720)

ryanpcmcquen:
> [ryanpcmcquen/obsidian-dropbox-backups@`95a7e55`/main.ts#L61](https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/95a7e556de0f871768dc8c57ee3f325de0717408/main.ts#L61)
> I don't think you need to import Bluebird for this. JS promises can easily do that - the replacement is just `await Promises.all(array.map(async (obj) => {}));` but be really careful here! You'll be processing everything in parallel which may not be ideal - imagine concurrently uploading thousand of files.

Fixed with the for loop!

> [ryanpcmcquen/obsidian-dropbox-backups@`95a7e55`/main.ts#L28](https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/95a7e556de0f871768dc8c57ee3f325de0717408/main.ts#L28)
> You may not want to call `read` on non-utf-8 files - especially images and other types.

I would actually like to backup everything, is there a different method I should use?
 
> [ryanpcmcquen/obsidian-dropbox-backups@`95a7e55`/main.ts#L201](https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/95a7e556de0f871768dc8c57ee3f325de0717408/main.ts#L201)
> You'll probably want the user to be able to customize this.

So I know this is a _me_ thing, but settings totally mess with my OCD, I'd rather launch with a sensible default and no configuration, and then add it down the road if enough users really want it. Settings are just crazy distracting to me.

> [ryanpcmcquen/obsidian-dropbox-backups@`95a7e55`/main.ts#L38](https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/95a7e556de0f871768dc8c57ee3f325de0717408/main.ts#L38)
> I don't think this is the right approach here. Reading all files into memory, and then uploading them all concurrently, isn't a great way to scale. I think your best bet is to rather come up with a list of files, then upload them one-by-one. The only thing that you'll have to be mindful of is if files are deleted while the upload process happens. In that case, you can catch the exception, and verify with `vault.exists(file.path)` to ignore the error if the file doesn't exist anymore.

This has been fixed, and we removed a ton of JavaScript (698kb vs 1.09mb) from `main.js` by getting rid of Bluebird. Thanks!

Screen Shot 2021-06-03 at 6 57 09 AM

> [ryanpcmcquen/obsidian-dropbox-backups@`95a7e55`/main.ts#L50](https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/95a7e556de0f871768dc8c57ee3f325de0717408/main.ts#L50)
> You can get a reference to your ribbon icon returned from when it was created here: [ryanpcmcquen/obsidian-dropbox-backups@`95a7e55`/main.ts#L190](https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/95a7e556de0f871768dc8c57ee3f325de0717408/main.ts#L190)

Fixed!

> [ryanpcmcquen/obsidian-dropbox-backups@`95a7e55`/main.ts#L194](https://github.com/ryanpcmcquen/obsidian-dropbox-backups/blob/95a7e556de0f871768dc8c57ee3f325de0717408/main.ts#L194)
> Don't await on long-running operations in your plugin's onload. That will delay the entire app's loading sequence, including other plugins. Do this asynchronously, or just skip doing this until it's needed (lazy).

I've removed the `await`.

ryanpcmcquen:
I removed some in 0.7.1, the only ones I have are for missing props:
Screen Shot 2021-06-03 at 8 02 41 AM

Is there a better way around this?

ryanpcmcquen:
Also, I fixed your other comment on the commit in 0.7.2.

ryanpcmcquen:
Binary files should be read correctly, images and pdfs are working as I would expect now!

lishid:
I think it's `vault.adapter.exists`.

ryanpcmcquen:
Nice! Looks like they both exist, but that one has a type definition:
Screen Shot 2021-06-03 at 8 32 01 AM

Not sure how to work around `ariaLabel` missing from `HTMLElement`:
Screen Shot 2021-06-03 at 8 33 29 AM

The other `@ts-ignore`s are for missing type definitions from the Dropbox SDK ... I am almost using their examples verbatim.

ryanpcmcquen:
I've removed more `@ts-ignore`s in 0.7.4.

lishid:
Easiest to just use `element.setAttribute('aria-label', text)`. Maybe upgrade your typescript engine to see if the definition has been added? https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaLabel

ryanpcmcquen:
I upgrade all that stuff obsessively already. `.setAttribute` is better and fixed that. I also fixed more of the typing issues. There are still a few in there but they are for missing type definitions from the Dropbox SDK, I'll try to create an issue over there for them to add type definitions.

ryanpcmcquen:
0.7.5 is out now which has less `@ts-ignore` statements.

ryanpcmcquen:
0.7.6 is out and now there are no `@ts-ignore` statements! It's basically the best thing ever. NOT!

lishid:
Great work!

ryanpcmcquen:
Thank you for the solid review!

ryanpcmcquen:
Oh man, @lishid! It is failing to load on mobile and I don't really know why. Is there a way to debug this?
IMG_6074

ryanpcmcquen:
IMG_6074

lishid:
I don't think it'll work on mobile at all. I'm pretty sure the dropbox module uses NodeJS's API, which isn't available on mobile.

ryanpcmcquen:
Dang, I guess I could use the HTTP api with `fetch` on mobile ...

lishid:
Yeah, I'm not sure how Dropbox's API internally work, but if you can override its request functionality with fetch/XMLHttpRequest, that might make it work on mobile.

lishid:
Though I'd still suggest just making your plugin desktop only. On mobile, the app gets backgrounded/killed often. If you do still want to make it mobile friendly, then perhaps making it default to manual sync only, because the user will have to wait for it to finish before leaving the app.

ryanpcmcquen:
Cool. I know they offer a non-Node version of the library, will try that and disable auto backups on mobile. I will see how that goes. Thank you!

lishid:
Good luck!

ryanpcmcquen:
@lishid I managed to manually browserify the Dropbox SDK. And now the plugin enables successfully on mobile, but it appears that `window.open` does not open the browser. Is there a different command for opening the browser on mobile?

ryanpcmcquen:
I see that the help icon does it successfully ...

lishid:
Yeah... window.open should work, not sure why it's not in your case.

ryanpcmcquen:
@lishid, I did it! I did it! A delayed `window.open` call gets blocked on mobile, so I switched to `window.location.assign` and now everything works on mobile too!

PR #311, opened 2021-05-27
Add 'Sidebar Expand on Hover' plugin.

toiq:

## Repo URL

[https://github.com/toiq/obsidian-sidebar-expand-on-hover](https://github.com/toiq/obsidian-sidebar-expand-on-hover)
## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


toiq:
@lishid Can you check if everything is alright? 
Also is there any better way to access the ribbons and sidebars? Currently I'm accessing these using getElementsByClassName() but I don't think this is a good approach.

toiq:
Can anyone check if everything is okay? I believe I removed all the bugs and the plugin is ready.

lishid:

https://github.com/toiq/obsidian-sidebar-expand-on-hover/blob/dd78b0f30a5e6ccbd91570ac414cafc2952bfdf4/src/main.ts#L2
Please rename your classes from `MyPlugin` to your plugin name.

https://github.com/toiq/obsidian-sidebar-expand-on-hover/blob/dd78b0f30a5e6ccbd91570ac414cafc2952bfdf4/src/main.ts#L42
These aren't of type Workspace, cast them to `any` instead.

https://github.com/toiq/obsidian-sidebar-expand-on-hover/blob/dd78b0f30a5e6ccbd91570ac414cafc2952bfdf4/src/main.ts#L144
Don't do this - load settings once in your plugin's onload. This would cause your plugin to constantly read the JSON file on disk.

https://github.com/toiq/obsidian-sidebar-expand-on-hover/blob/dd78b0f30a5e6ccbd91570ac414cafc2952bfdf4/src/main.ts#L122
I don't really understand why there needs to be such complicated code to manage the sidebar's width. The app already provide a persistent way to store the sidebar width using the drag resizer, no?

toiq:
@lishid Done! 😁

lishid:
https://github.com/toiq/obsidian-sidebar-expand-on-hover/blob/7274543bdece546e8bb93b319ebb485c823d6410/src/main.ts#L2
https://github.com/toiq/obsidian-sidebar-expand-on-hover/blob/7274543bdece546e8bb93b319ebb485c823d6410/src/main.ts#L166
Should still rename these two hehe.

> https://github.com/toiq/obsidian-sidebar-expand-on-hover/blob/dd78b0f30a5e6ccbd91570ac414cafc2952bfdf4/src/main.ts#L122
I don't really understand why there needs to be such complicated code to manage the sidebar's width. The app already provide a persistent way to store the sidebar width using the drag resizer, no?

I think you might have missed this question.

toiq:
Well, 
The plugin doesn't really allow you to resize the sidebars by dragging when it's unpinned. I tried allowing that and it makes it feel glitchy. So I only allowed it when the sidebar isn't pinned(when the expand on hover is disabled). 
If the user still wants to change the size in unpinned mode then he/she can go to plugin settings and change that. And I needed to keep that settings. At least, that was the original motivation.
That is why I added those lines of code to manage sidebar width and apply it before calling expand(). Otherwise it always uses size set by dragging I think? I don't remember but this is probably the reason. 

And...I renamed the identifiere. 😅

toiq:
@lishid Is there anything else I should fix?

lishid:
Oh and I suggest removing main.js from your repo and adding it to the .gitignore - it's a lot of noise for your commits and you'll be releasing it into Github Releases anyway.

PR #315, opened 2021-05-29
Add Zoottelkeeper plugin

akosbalasko:

## Repo URL

https://github.com/akosbalasko/zoottelkeeper-obsidian-plugin/

## Release Checklist

  - [x] Windows
  - [x] macOS
  - [ ] Linux
  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


lishid:
There seems to be some formatting issues with your code - you should run an auto-formatter to normalize whitespace characters.

https://github.com/akosbalasko/zoottelkeeper-obsidian-plugin/blob/c2c08cc9081bbcf215c17c810b0d46a65b61547b/main.ts#L2
Importing this from `os` would disqualify your plugin for mobile as NodeJS APIs do not exist in Android or iOS. I'd suggest you simply use `\n` as that's the standard end of line we use in Obsidian.

https://github.com/akosbalasko/zoottelkeeper-obsidian-plugin/blob/c2c08cc9081bbcf215c17c810b0d46a65b61547b/main.ts#L25
You should register your events in `onload` just in case there was any order of operation issues between loading/unloading your plugin and onLayoutReady. 

https://github.com/akosbalasko/zoottelkeeper-obsidian-plugin/blob/c2c08cc9081bbcf215c17c810b0d46a65b61547b/main.ts#L45
A better idea is to add all your index files to a set, then iterate the set.

https://github.com/akosbalasko/zoottelkeeper-obsidian-plugin/blob/c2c08cc9081bbcf215c17c810b0d46a65b61547b/main.ts#L135
This should return a TFile instead of a TAbstractFile. You can verify that an abstract file is a TFile by using `file instanceof TFile`, after which TypeScript will be able to infer that file is a TFile. In general, you'll want to consistently work with TFile and TFolders instead of using TAbstractFile.

https://github.com/akosbalasko/zoottelkeeper-obsidian-plugin/blob/c2c08cc9081bbcf215c17c810b0d46a65b61547b/main.ts#L99
This doesn't seem like it performs action on a folder. You should also properly type the `func` to be `func: (file: TFile) => Promise<any>`

https://github.com/akosbalasko/zoottelkeeper-obsidian-plugin/blob/c2c08cc9081bbcf215c17c810b0d46a65b61547b/main.ts#L50
This seems problematic... Why does it go up exactly 2 parent steps?

I think in general a big problem here is that your leaf functions take an arbitrary path, and tries to get an index file from that path - This ends up getting done repeatedly for the same file. For example, `updateIndexContent` calls `removeIndexFile` and `generateIndexContents`, and each of that calls getIndexFile separately on the same identical path.

I'd recommend making your code such that you identify all the index files you want to update first, then directly update those files.

https://github.com/akosbalasko/zoottelkeeper-obsidian-plugin/blob/c2c08cc9081bbcf215c17c810b0d46a65b61547b/main.ts#L123
You may want to use `fileManager.generateMarkdownLink` instead, which will respect the user's wikilink vs markdown link settings.

https://github.com/akosbalasko/zoottelkeeper-obsidian-plugin/blob/c2c08cc9081bbcf215c17c810b0d46a65b61547b/main.ts#L129
Don't cast, test with instanceof and bail/error if it fails.

https://github.com/akosbalasko/zoottelkeeper-obsidian-plugin/blob/c2c08cc9081bbcf215c17c810b0d46a65b61547b/main.ts#L90
This function will just create an empty file, then delete it, if the file doesn't exist... That's not very elegant :)

akosbalasko:
Hi @lishid , 
Thanks for the review I'll improve the code according to your suggestions soon.

akosbalasko:
Hi @lishid ,

I reshaped the code according to your suggestions (thank you for them). It is released as 0.5.3, so I think it's ready for the next turn of the review.
Thanks a lot!

lishid:
> https://github.com/akosbalasko/zoottelkeeper-obsidian-plugin/blob/c2c08cc9081bbcf215c17c810b0d46a65b61547b/main.ts#L123
You may want to use fileManager.generateMarkdownLink instead, which will respect the user's wikilink vs markdown link settings.

Looks like you missed this one. Otherwise, everything else looks great!

I will merge your PR and you can continue working on it after.

PR #317, opened 2021-05-30
Add Block Reference Counter to Community Plugins

shabegom:

## Repo URL

https://github.com/shabegom/obsidian-reference-count

## Release Checklist

- [X] I have tested this on Windows, macOS
  - [X] `main.js`
  - [X] `manifest.json`


shabegom:
I seem to have brought along a few other commits... lmk if that's an issue and I can resubmit.

lishid:
https://github.com/shabegom/obsidian-reference-count/blob/e26111a32483e33b1e03017c824b7531ada56b4b/src/main.ts#L10
This import doesn't seem used.

https://github.com/shabegom/obsidian-reference-count/blob/e26111a32483e33b1e03017c824b7531ada56b4b/src/indexer.ts#L33
You definitely need a null check as it's not guaranteed to give you a result back, when the file hasn't been indexed yet.

https://github.com/shabegom/obsidian-reference-count/blob/e26111a32483e33b1e03017c824b7531ada56b4b/src/indexer.ts#L32
Why not a simple `for (let file of files)` loop?

https://github.com/shabegom/obsidian-reference-count/blob/e26111a32483e33b1e03017c824b7531ada56b4b/src/indexer.ts#L34
Why not just pass the cache object here? You can pass the file as a first param, then the cache object in the second.

https://github.com/shabegom/obsidian-reference-count/blob/e26111a32483e33b1e03017c824b7531ada56b4b/src/indexer.ts#L56
You probably want to use `embeds = embeds || []` instead of that. Destructuring into another array will make a copy of the array, which doesn't seem necessary here.

https://github.com/shabegom/obsidian-reference-count/blob/e26111a32483e33b1e03017c824b7531ada56b4b/src/indexer.ts#L214
You should use Obsidian's parseLinktext function  (https://github.com/obsidianmd/obsidian-api/blob/48bd2b91f66aa7d3768629be48c52e5351ec6696/obsidian.d.ts#L2026) which is what we use to split the link. This will give you a `linkpath` which is the path that will resolve to a note, and a `subpath` which starts with `#`.

https://github.com/shabegom/obsidian-reference-count/blob/e26111a32483e33b1e03017c824b7531ada56b4b/src/indexer.ts#L130
Probably don't need pages to be passed in from the global object?

https://github.com/shabegom/obsidian-reference-count/blob/e26111a32483e33b1e03017c824b7531ada56b4b/src/main.ts#L21
You can use `this.registerEvent(something.on(...))` to automatically register an event to be unloaded when your plugin unloads.

https://github.com/shabegom/obsidian-reference-count/blob/e26111a32483e33b1e03017c824b7531ada56b4b/src/main.ts#L40
I find it weird that you keep using keyword args as function params. Why not just call it normally like `indexBlockReferences(this.app)`?

https://github.com/shabegom/obsidian-reference-count/blob/e26111a32483e33b1e03017c824b7531ada56b4b/src/main.ts#L60
You should probably debounce the calls from these events as they may fire very frequently.

https://github.com/shabegom/obsidian-reference-count/blob/e26111a32483e33b1e03017c824b7531ada56b4b/src/main.ts#L71
Use `workspace.getActiveViewOfType(MarkdownView)` and test for null on the result. I'd also recommend adding a try-catch since this is undocumented API.

Your code might have unintentional consequences if the active leaf isn't viewing a markdown file - for example, a backlink pane, or a local graph, or an image/video/PDF. This is true for all the [other situations](https://github.com/shabegom/obsidian-reference-count/blob/e26111a32483e33b1e03017c824b7531ada56b4b/src/main.ts#L113) where you're accessing `activeView` directly, and assuming that it has `view.previewMode`. Test for `instanceof MarkdownView` and bail if it's not the right type.

https://github.com/shabegom/obsidian-reference-count/blob/e26111a32483e33b1e03017c824b7531ada56b4b/src/main.ts#L343
Why not just use `tempLeaf` directly? Would appreciate a bit more comment here to explain the wizardry that's happening to the search-ref panes.

---

It seems that this code is using a lot of `array.forEach` - I would highly recommend using `for (let obj of array)` instead as it doesn't create an anonymous function, so it's more performant and easier to debug. 

Also, it (ab)uses the optional chaining `?` operator a lot for things it should rather rely on proper type checks (like instanceof) quite often. I'd argue that it makes code harder to read - and recommend properly handling null/undefined scenarios when they are expected.

lishid:
> it's more performant

Actually, performance doesn't matter anymore, at least in V8/Chromium. But there are still a lot of great reasons to use ES6 for-of loops, such as working with iterables, and supporting `await`, `continue`, `break`, etc: https://stackoverflow.com/a/50844413

shabegom:
> I find it weird that you keep using keyword args as function params. Why not just call it normally like indexBlockReferences(this.app)?

It's just a pattern I was playing with in this plugin

I read somewhere that using objects is a better practice because:
- order doesn't matter
- can omit a key from an object and the function still works

I guess it doesn't really matter if the function is taking in only a single argument...

lishid:
> - order doesn't matter
> - can omit a key from an object and the function still works

That's true - normally this isn't a problem for up to 3~4 parameters, but if you have a "kitchen sink" function that has many optional parameters, then it makes sense. [An example of how we use those in createEl, which takes many parameters like css class, innerText, attributes, etc](https://github.com/obsidianmd/obsidian-api/blob/48bd2b91f66aa7d3768629be48c52e5351ec6696/obsidian.d.ts#L115)

But in general, normal function calls are still the idiomatic ways. There are still several advantages to using regular parameters:

- Performance. The JIT usually infers parameter types and optimize your function if you call it consistently with the same types. For example, `(a: number, b: number) => a + b` can be optimized once the compiler knows that `a` and `b` are always numbers. If you pass it as `{a: number, b: number}`, then it can't perform this optimization step, and so your function might be left as two object reads, followed by a complicated type check (because the compiler doesn't know that `a` and `b` are numbers, they might be strings, booleans, etc). In practice this may not always be true because of things like hidden classes.
- IDE integration. Most IDEs have great parameter type prompting, when you're typing a function parameters, the IDE can give you an idea what type it's expected and any js-doc associated with the param. This is more difficult and less-well-supported for object types.
- You incur a chance to run into a null exception if the caller doesn't provide the object properly - usually not an issue if all your code is all self-written, but more of a problem for API design. Generally we have to test the options parameter for null before attempting to destructure it.

With normal function calls, normally the order has a natural preference order - in my experience, typically the more "global" or "state" or "context" objects go in first, and ephemeral information goes in last. For example, `app` would go in before `positionX`. Most of the time this is not a problem. Omissions can still happen, they just have to be the last few parameters.

ericaxu:
Hi @shabegom ! Are you still interesting in submitting this PR?

shabegom:
@ericaxu @lishid I'm going to close this PR for now, work on the requested changes over time and resubmit when the plugin is in better shape. I've unfortunately been super busy launching a new project at the job job and haven't made progress here. Apologies for leaving this sitting for so long without any updates.

PR #318, opened 2021-05-31
Add Obsidian Tweaks

JeppeKlitgaard:

## Repo URL

https://github.com/JeppeKlitgaard/ObsidianTweaks

## Release Checklist

  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1546,5 +1546,12 @@
         "description": "MetaEdit helps you manage your metadata.",
         "author": "Christian B. B. Houmann",
         "repo": "chhoumann/MetaEdit"
+    },
+    {
+        "id": "obsidian-tweaks",
+        "name": "Obsidian Tweaks",
+        "description": "A small plugin that implements some sorely missed features.",
+        "author": "Jeppe Klitgaard",
+        "repo": "JeppeKlitgaard/ObsidianTweaks"
     }
 ]


whitenoisedev:
> I initally planned to contribute to the Hotkeys++ plugin, but since I don't necessarily want ObsidianTweaks to be limited to hotkeys in the future, I decided to make my own.

If you want to have your own plugins is okay. However, I do not like the idea of kitchen sink plugins. I prefer single-purpose plugin.
If users install this plugin for X, it shouldn't start doing Y & Z at some point in the future.

JeppeKlitgaard:
> > I initally planned to contribute to the Hotkeys++ plugin, but since I don't necessarily want ObsidianTweaks to be limited to hotkeys in the future, I decided to make my own.
> 
> If you want to have your own plugins is okay. However, I do not like the idea of kitchen sink plugins. I prefer single-purpose plugin.
> If users install this plugin for X, it shouldn't start doing Y & Z at some point in the future.

I absolutely agree. I should perhaps make the scope of this plugin clearer: It is meant to implement **small** fixes to the Obsidian UX experience that are not large enough to necessarily warrant a plugin of their own. Further, I want to stress that this plugin will never have any default behaviour. Any functionality it implements will need to be toggled on individually by users by either assigning a hotkey to the commands exposed by the plugin or by a toggle in the settings of the plugin.

I made this consideration explicitly to avoid the scenario of someone updating the plugin and their user experience suddenly changing in some way.

TLDR: It is kind of a kitchen sink plugin, but only to the extent that it only holds cutlery (and only cutlery put there explicitly by the... Person washing up. The analogy is falling apart)

Let me know whether this is sufficiently strict. The name is inspired by Gnome Tweaks, which does a similar thing for the Gnome shell.

lishid:
https://github.com/JeppeKlitgaard/ObsidianTweaks/blob/a93b890d54831078d37094cd1a2052db46902ec2/src/SelectLine.ts#L19
https://github.com/JeppeKlitgaard/ObsidianTweaks/blob/a93b890d54831078d37094cd1a2052db46902ec2/src/DirectionalCopy.ts#L27
https://github.com/JeppeKlitgaard/ObsidianTweaks/blob/a93b890d54831078d37094cd1a2052db46902ec2/src/ToggleHeading.ts#L34
You should put `activeView.editor` into its own variable `editor` so that you don't have to constantly call `activeView.editor`.

I don't see any coding issues with the plugin, it's pretty simple, but I am not too sure with the naming.

I know you have plans to add additional tweaks, but given that you're only adding hotkeys for some editing shortcuts, I think improving the name would help more users understand what your plugin is and does right off the bat, rather than having to read through your README.md and realize it's just a bunch of hotkeys.

JeppeKlitgaard:
@lishid Thanks for having a look.

I'll have to reconsider whether I want to just abandon the plugin entirely and submit these as an addition to Hotkeys++ instead and then submit any other tweaks as separate plugins.

I am currently a bit caught up with exams, so it might have to wait until after that before I get around to this again. Would it be alright if I leave this hanging here and then ping you once I get a chance to take a look?

lishid:
Sure, that's fine!

JeppeKlitgaard:
@lishid 

I have given it some thought, and I do think that I would like to continue with Obsidian Tweaks as a separate project to, say, Hotkeys++. I added some functionality today, which seeks to improve upon the built-in toggles for bold, italics, etc, which I think would be beyond the declared scope of Hotkeys++.

Since last time I have implemented your suggestion and added:
1. Copy left/right
2. Move selection left/right
3. Reimplementation of built-in toggles for bold, italics, strikethrough, comment, code. 
4. Bold and italics toggles using underscores instead of asterisks (in addition to asterisks version). 
5. Inline and block math toggles

1 and 2 are useful when doing LaTeX stuff, I find, but for non-STEM users this is likely not very useful.

3: The current implementation of toggles in Obsidian is not super consistent. I based mine on the principle that toggling twice should be the same as doing nothing (for both selection and content) - this is not hard, but also not as trivial as one could've hoped for in CodeMirror.

4: I personally prefer underscores, and some people prefer having one asterisks and one underscore. This is now possible. I chose this approach over a config option because it seems simpler and easier for people to use.

5: Just nice to have.

In the near future I plan to add features such as:

6. Duplicate file functionality
7. Autocomplete in frontmatter (maybe)
8. Path of current file in vault in status bar
9. Whatever else crops up.

---

The toggles wrap whole words by default, since I find it rare that I want formatting in the middle of a word - if I do, I will do this manually. The advantage of this is that it is a lot quicker to use - just put the cursor in a world (or across a few words) and execute the toggle to, for example, highlight the word(s) touched by the selection.

---

Hopefully this explains roughly the scope of the plugin. I also think the name has grown on me slightly, and would serve the project well if it continues in the direction of adding small tweaks to Obsidian.

Let me know your thoughts!

PR #320, opened 2021-06-03
File Explorer Note Count

ozntel:

## Repo URL

https://github.com/ozntel/file-explorer-note-count

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css`


ericaxu:
Great plugin! I'm sure many will find it useful.

https://github.com/ozntel/file-explorer-note-count/blob/main/src/main.ts#L25
What's this interval for?

Everything else looks great except for one ts-ignore which I don't know if can be fixed. Maybe we didn't expose those types.

@lishid Please take a look when you have time!

lishid:
https://github.com/ozntel/file-explorer-note-count/blob/9a9a64b81b5abf1403ef230019ea5258b49f17d0/src/main.ts#L5

Importing this would cause your plugin to fail on mobile, because the nodejs environment doesn't exist there. Maybe just write a small helper function for this?

https://github.com/ozntel/file-explorer-note-count/blob/9a9a64b81b5abf1403ef230019ea5258b49f17d0/src/main.ts#L69
You should use workspace.onLayoutReady() which will replace this entire logic.

https://github.com/ozntel/file-explorer-note-count/blob/9a9a64b81b5abf1403ef230019ea5258b49f17d0/src/main.ts#L101
I think you're just looking for `!list.contains(target)` and `list.contains(target)` in these two conditions.

https://github.com/ozntel/file-explorer-note-count/blob/9a9a64b81b5abf1403ef230019ea5258b49f17d0/src/folder-count.ts#L73
This can just be `array.some((item) => condition)` rather than `array.findIndex((item) => condition) !== -1`

AidenLx:
Done! https://github.com/ozntel/file-explorer-note-count/commit/6983e3ac234378fdb814e94f33eb502030a14a8d

>https://github.com/ozntel/file-explorer-note-count/blob/9a9a64b81b5abf1403ef230019ea5258b49f17d0/src/main.ts#L5
>Importing this would cause your plugin to fail on mobile, because the nodejs environment doesn't exist there. Maybe just write a small helper function for this?

I've used rollup-plugin-polyfill-node to handle this so no more `require('path')` in main.js now

AidenLx:
Also, could you also take a look at this branch to see if there is any obvious error? : https://github.com/ozntel/file-explorer-note-count/pull/3

AidenLx:
And is there a way currently to add a menu item to the explorer item context menu? I'm having a hard time trying to use `workspace.trigger('file-menu',menu,file,'file-explorer-context-menu',null)`

lishid:
> And is there a way currently to add a menu item to the explorer item context menu

Register for the event `workspace.on('file-menu', callback)` and you'll have the opportunity to add menu items to any file menu, including ones that come from the more-options three dot menu, etc. If you just want to add to the file explorer menu then filter the source by checking it against 'file-explorer-context-menu'.

AidenLx:
> Should I review your branch or the main branch, I'm confused...

I'm one of the collaborators of this plugin and basically, I maintain both branches. I've kept `debounce` branch based on `main` branch and it's only one commit ahead of `main`. Once it's checked, I'll merge it to `main`.

AidenLx:
@lishid Any updates? I’ve merged debounce to main.

lishid:
https://github.com/ozntel/file-explorer-note-count/blob/4a7f482218a1bfc6e8413892e145d7384ae0d516/src/main.ts#L34
This class isn't removed when plugin unloads. I think it'd make more sense to use `toggleClass('class', showAllNumbers)` here.

Otherwise I think there isn't much else to be concerned about.

I didn't see any code that does something with `toggleCollapsed`, did you guys want to get that reviewed or do you want to work on it afterwards?

AidenLx:
The function that needs to modify `toggleCollapsed` is moved to another plugin, so no need to review for now. And thanks again for all the help! :)

PR #321, opened 2021-06-03
new plugin Text Format

Benature:

## Repo URL
https://github.com/Benature/obsidian-text-format

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1560,5 +1560,12 @@
         "description": "Auto format when typing",
         "author": "yaozhuwa",
         "repo": "Yaozhuwa/easy-typing-obsidian"
+    },
+    {
+        "id": "obsidian-text-format",
+        "name": "Text Format",
+        "author": "Benature",
+        "description": "Format selected text upper/lower/capitalize",
+        "repo": "Benature/obsidian-text-format",
     }
 ]


sowmenappd (approved):

lishid:
https://github.com/Benature/obsidian-text-format/blob/ba06d9a4537347d7c4dd2e623b001e2beee0ded1/main.ts#L59
This seems like it does the same thing as "blanks"? Maybe you should also change the name of the command to say that it removes newline characters instead of "enters" because I don't think "enters" is something people understand.

https://github.com/Benature/obsidian-text-format/blob/ba06d9a4537347d7c4dd2e623b001e2beee0ded1/main.ts#L22
"spaces" instead of "blanks"?

https://github.com/Benature/obsidian-text-format/blob/ba06d9a4537347d7c4dd2e623b001e2beee0ded1/main.ts#L7
"Lowercase text" instead of "Lower text"? Same for uppercase?

https://github.com/Benature/obsidian-text-format/blob/ba06d9a4537347d7c4dd2e623b001e2beee0ded1/main.ts#L67
I'd suggest testing that the replaced text indeed changed before applying it.

Benature:
Thanks for your time for reviewing, I've modified the code and the command names according to your suggestions.

lishid:
https://github.com/Benature/obsidian-text-format/blob/28920a9856e108ed15fafbeecf4ca22bce05bc45/main.ts#L52
I'm not sure users would agree to convert 4 spaces to 1 tab when triggering the command "Remove redundant spaces".
I think you could separate that into its own command to switch between 1 tab vs 4 spaces?

https://github.com/Benature/obsidian-text-format/blob/28920a9856e108ed15fafbeecf4ca22bce05bc45/main.ts#L54
This is rather inefficient, because you keep re-scanning starting from the beginning of the string here.
The easiest way to accomplish this would be to use `text.replace(/ +/g, ' ')`, which replaces any consecutive spaces with a single space, and you only have to run this once.

https://github.com/Benature/obsidian-text-format/blob/28920a9856e108ed15fafbeecf4ca22bce05bc45/main.ts#L57
This is also something I think users wouldn't expect. It's trimming off a single space from the beginning of every line. Perhaps you meant the regex to be `/\n $/`? I'd take it one step further and do `/\n +$` which trims out any line that's purely spaces.

https://github.com/Benature/obsidian-text-format/blob/28920a9856e108ed15fafbeecf4ca22bce05bc45/main.ts#L61
Similar to the above, users probably don't expect a command that "Remove all newline characters in selection" to also remove consecutive spaces...

Benature:
Thank you for your suggestion, I am not quite good at regex and `/ +/g` is indeed much better.

Sometimes I copy text from pdf, there are redundant spaces and newline characters, so I merged these two actions in one command. But after your suggestion, to satisfy common users, I would separate them into two commands for now.

PR #322, opened 2021-06-06
Add OCR plugin

schlundd:

## Repo URL

https://github.com/schlundd/obsidian-ocr-plugin

## Release Checklist

- [ ] I have tested this on
  - [ ] Windows
  - [ ] macOS
  - [x] Linux
  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


lishid:
https://github.com/schlundd/obsidian-ocr-plugin/blob/57b90d51ed3942a02951fe51f7437db15b0f832b/main.ts#L80
This isn't a good idea. Use `vault.getAbstractFileForPath` instead. Test the result for `instanceof TFile` as well.

https://github.com/schlundd/obsidian-ocr-plugin/blob/57b90d51ed3942a02951fe51f7437db15b0f832b/main.ts#L141
Actually - you can make this modal inherit the type `SuggestModal<TFile>`, so you are always dealing with Obsidian files instead of having to constantly convert filepath to TFile.

https://github.com/schlundd/obsidian-ocr-plugin/blob/57b90d51ed3942a02951fe51f7437db15b0f832b/main.ts#L117
Might want to make sure it doesn't exist before calling `create`.

schlundd:
Thanks for the quick review and feedback.

> https://github.com/schlundd/obsidian-ocr-plugin/blob/57b90d51ed3942a02951fe51f7437db15b0f832b/main.ts#L80
> This isn't a good idea. Use `vault.getAbstractFileForPath` instead. Test the result for `instanceof TFile` as well.

I agree. That looks wrong :-) IIRC I saw that in another plugin and just copied it, thinking this may be the right/only way. I can't remember where, unfortunately.

> https://github.com/schlundd/obsidian-ocr-plugin/blob/57b90d51ed3942a02951fe51f7437db15b0f832b/main.ts#L141
> Actually - you can make this modal inherit the type `SuggestModal<TFile>`, so you are always dealing with Obsidian files instead of having to constantly convert filepath to TFile.

Done. Thanks.

> https://github.com/schlundd/obsidian-ocr-plugin/blob/57b90d51ed3942a02951fe51f7437db15b0f832b/main.ts#L117
> Might want to make sure it doesn't exist before calling `create`.

This "shouldn't happen" as it is checked earlier in the command. There are edge cases where it might happen, though - if the file is created while the command is running. So I wrapped it in a try/catch for now and will handle some of these cases properly in the next version. I hope thats sufficient for you for now.

PR #324, opened 2021-06-07
Add Grandfather plugin

noatpad:

## Repo URL

https://github.com/noatpad/obsidian-grandfather

## Release Checklist

- [x] I have tested this on ~~Windows~~, **macOS**, and ~~Linux~~ _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


ericaxu:
Interesting plugin! Looks good to me.

Just curious, why did you name it "Grandfather"?

lishid:
https://github.com/noatpad/obsidian-grandfather/blob/ce2045641d8712321c0b52eab718485fa25ca904/src/main.ts#L11
Not sure where you picked up this code, the recommended way to load settings is now `Object.assign` (see https://github.com/obsidianmd/obsidian-sample-plugin/blob/c228a7022301e4a74614a67570d38ed2e2e05a71/main.ts#L60)

Also, yeah, very curious about the plugin name.

noatpad:
The line for loading settings was just an alternate way I thought of doing it, haha. Just changed it to the recommended way, so that should do it.

And the name is actually just a lil' wordplay from me, with "Grandfather" just being a lil' nod to a grandfather clock. I like naming things a bit cryptically that way, haha

lishid:
> The line for loading settings was just an alternate way I thought of doing it, haha. Just changed it to the recommended way, so that should do it.

We used to do it the old way, but it's actually a bit different in that the default settings wouldn't be properly upgraded when you add new items to the object:
```ts
// If loadData returns an object with {firstSetting}, but your plugin has two 
// settings {firstSetting, secondSetting}, you will now run into secondSetting is undefined
this.settings = await this.loadData() || DEFAULT_SETTINGS;
```

PR #326, opened 2021-06-07
Add Obsidian iCal

mdelobelle:

## Theme checklist

- [] Your repo contains all required files (please *do not* add them to our `obsidian-releases` repo)
  - [ ] `obsidian.css`
  - [ ] The screenshot file you pointed to
- [ ] You have indicated which modes (dark, light, or both) are compatible with your theme
- [ ] If you're not using the `master` branch (legacy), please indicate

## Repo URL

https://github.com/mdelobelle/obsidian-ical

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1,5 +1,12 @@
 [
     {
+        "id": "obsidian-ical",
+        "name": "iCal",
+        "author": "mdelobelle",
+        "description": "Include your macOSX calendar events in daily notes.",
+        "repo": "mdelobelle/obsidian-ical"
+    },
+    {
         "id": "nldates-obsidian",
         "name": "Natural Language Dates",
         "author": "Argentina Ortega Sainz",


lishid:
https://github.com/mdelobelle/obsidian-ical/blob/master/.DS_Store
Might want to add this to the gitignore.

https://github.com/mdelobelle/obsidian-ical/blob/8b27824d613a76f9c726b8f24b4b90758dc694e4/main.ts#L20
Use `vault.getAbstractFileByPath` and check for `file instanceof TFile`. Also, why not just make this an async function rather than using `.then()`?

https://github.com/mdelobelle/obsidian-ical/blob/8b27824d613a76f9c726b8f24b4b90758dc694e4/main.ts#L45
Here you should use `workspace.getActiveViewOfType(MarkdownView)` and check for null.

Also, same here, why not just use an async function rather then several layers of nested `.then()`?

https://github.com/mdelobelle/obsidian-ical/blob/8b27824d613a76f9c726b8f24b4b90758dc694e4/src/ICalEvent/ICalEvent.ts#L3
I think this package is causing your plugin to swell up to 3MB! in size - you might want to consider using its sibling project ical.js (https://github.com/peterbraden/ical.js) instead of node-ical since it imports a bunch of other libraries like node-fetch to support things like `ical.async.fromURL` which you don't need anyway.

---


So your woes with not having calendar is likely because Obsidian does not enlist NSCalendarsUsageDescription in the app's Info.plist (because we don't need it ourselves). MacOS take that as a rejection without even prompting the user for allowing the access.

I can include a description in our next public release, but it'll only take effect for users who reinstall Obsidian with the latest installer/dmg. At least, you can prompt the user to upgrade the app if you get a rejection when accessing that folder.

Would you like to wait until this is in place, so people don't have to go through the hoops to rsync their calendar data?

mdelobelle:
Hi @lishid thank you for your fast review and your proposal.
It's definitely a good idea to wait for this feature to be available in the next public release. 
In the meantime I'll make the adjustment that you have mentioned

lishid:
Perfect.

> https://github.com/mdelobelle/obsidian-ical/blob/master/.DS_Store
Might want to add this to the gitignore.

Probably also remove it from the repo :)

https://github.com/mdelobelle/obsidian-ical/blob/7faf21e302f1e54cffba6db6ae513e97561142f4/main.ts#L21

I don't think this is right - you should probably raise the error or return null. You should probably define the function as `async getTemplate (): Promise<string | null> {` instead, and make sure you return something everywhere.

https://github.com/mdelobelle/obsidian-ical/blob/7faf21e302f1e54cffba6db6ae513e97561142f4/main.ts#L41

MarkdownView is always a FileView, so you don't need to check here.

https://github.com/mdelobelle/obsidian-ical/blob/7faf21e302f1e54cffba6db6ae513e97561142f4/main.ts#L43
https://github.com/mdelobelle/obsidian-ical/blob/7faf21e302f1e54cffba6db6ae513e97561142f4/src/ICalEvent/ICalEvent.ts#L85
Convert these callbacks here to async/await? It seems like you're trying to process all the files in parallel, which isn't what I'd recommend. Iterate through them with a for-of loop will unroll the async/await into a serial process.

https://github.com/mdelobelle/obsidian-ical/blob/7faf21e302f1e54cffba6db6ae513e97561142f4/src/ICalEvent/chooseSectionModal.ts#L5
Might want to rename this to TitleCase.

mdelobelle:
> https://github.com/mdelobelle/obsidian-ical/blob/master/.DS_Store
Might want to add this to the gitignore.
Probably also remove it from the repo :)

oops! done 

> https://github.com/mdelobelle/obsidian-ical/blob/7faf21e302f1e54cffba6db6ae513e97561142f4/main.ts#L21
I don't think this is right - you should probably raise the error or return null. You should probably define the function as async getTemplate (): Promise<string | null> { instead, and make sure you return something everywhere.

done

> https://github.com/mdelobelle/obsidian-ical/blob/7faf21e302f1e54cffba6db6ae513e97561142f4/main.ts#L41
MarkdownView is always a FileView, so you don't need to check here.

done

> https://github.com/mdelobelle/obsidian-ical/blob/7faf21e302f1e54cffba6db6ae513e97561142f4/main.ts#L43
https://github.com/mdelobelle/obsidian-ical/blob/7faf21e302f1e54cffba6db6ae513e97561142f4/src/ICalEvent/ICalEvent.ts#L85
Convert these callbacks here to async/await? It seems like you're trying to process all the files in parallel, which isn't what I'd recommend. Iterate through them with a for-of loop will unroll the async/await into a serial process.

done
thank you for your advices, I'm quite new to learning JS and I'm still struggling to understand how to choose the best approach to use these async/await/Promise protocols

> https://github.com/mdelobelle/obsidian-ical/blob/7faf21e302f1e54cffba6db6ae513e97561142f4/src/ICalEvent/chooseSectionModal.ts#L5
Might want to rename this to TitleCase.

done

lishid:
Looking good! I'm ok for a merge if you want it in asap.

> thank you for your advices, I'm quite new to learning JS and I'm still struggling to understand how to choose the best approach to use these async/await/Promise protocols

My experience has been generally to stick with async/await as much as possible, and only use `new Promise` when you have no other choice (for example, to deal with a browser/node API that only does callbacks or events). I also find that you never really have to use `.then()`.

https://github.com/mdelobelle/obsidian-ical/blob/5ee0c28a2a333df0ad1ff138db24d1147919a3bf/src/ICalEvent/ICalEvent.ts#L86
I think you can simplify this further with `let result = await plugin.app.vault.cachedRead(file);`

lishid:
Oh and I would highly suggest putting your code through an auto-formatter, typically available in almost every modern IDEs. This will improve whitespace consistency and help others read your code.

mdelobelle:
>My experience has been generally to stick with async/await as much as possible, and only use new Promise when you have no other choice (for example, to deal with a browser/node API that only does callbacks or events). I also find that you never really have to use .then().

👍

>https://github.com/mdelobelle/obsidian-ical/blob/5ee0c28a2a333df0ad1ff138db24d1147919a3bf/src/ICalEvent/ICalEvent.ts#L86
I think you can simplify this further with let result = await plugin.app.vault.cachedRead(file);

Done

>Oh and I would highly suggest putting your code through an auto-formatter, typically available in almost every modern IDEs. This will improve whitespace consistency and help others read your code.

oooh that's cool, I've found the settings in VSCode -> applied

>Looking good! I'm ok for a merge if you want it in asap.

I think I'll wait for the capability to read "Calendars" folder on Mac to be available to provide a better user experience (automating the rsync can really be overwhelming for non tech people). 
So, I stay tuned :)

Thanks and congrats again for the great work on Obsidian 👍

lishid:
Alright, I think we're good with the review for now. Let's pick it up once the next installer ships.

mdelobelle:
Hi @lishid , do you have some news about an installer for mac enlisting NSCalendarsUsageDescription in the app's Info.plist?

lishid:
Should have been shipped in our latest installer, sorry I forgot to update here!

mdelobelle:
Thanks!
I have installed 0.12.10 installer 

image

still I can't see Obsidian in the list of application that can be granted an access to my calendar

image

Is there anything else that I should do to setup this capability?

lishid:
Have you tried using your plugin to request access to the ical folder? I think it only shows up once the app tries to access calendar data and it prompts the user.

mdelobelle:
Here is what i've tried so far:

```js
const fs = require('fs');
const testFolder = '/Users/mathieudelobelle/Library/Calendars/';
fs.readdirSync(testFolder).forEach((file: string) => {
	console.log(file);
});
```

this is the result in the console, without having been prompted a request to grant access to Calendar from Obsidian

```
fs.js:1021 Uncaught (in promise) Error: EPERM: operation not permitted, scandir '/Users/mathieudelobelle/Library/Calendars/'
    at Object.readdirSync (fs.js:1021)
    at Object.e.readdirSync (electron/js2c/asar_bundle.js:5)
    at ICal.eval (eval at <anonymous> (app.js:1), <anonymous>:10919:16)
    at Generator.next (<anonymous>)
    at fulfilled (eval at <anonymous> (app.js:1), <anonymous>:33:58)
```

If I change `'/Users/mathieudelobelle/Library/Calendars/'` by `'/Users/mathieudelobelle/Library/'` everything works fine.
(installer 0.12.10, obsidian restarted, system restarted)

lishid:
Hmm maybe I'm wrong about how macOS is controlling access to the calendars folder then? I based my assumptions off this thread: https://github.com/electron/electron/issues/14833

Unfortunately, my knowledge of macOS is fairly limited, and I don't have a proper setup to test all of these. Let me know if you have any more hypothesis on why it's not making a permission prompt and why it's denying the permission outright.

Perhaps it'd be worth testing with some simple snippets in Node.JS.

Creling:
Well, I am not familiar  with MacOS, however, according to the [document](https://developer.apple.com/documentation/eventkit/accessing_the_event_store), you may not modify the calendar database directly. Maybe _sync_ is a kind of _modify_ somewhat.

mdelobelle:
Hi @lishid I've made some research and tests and unfortunately I've not managed to have Obsidian be granted access to the folder.... I guess better not publish a plugin that is depends on external tools to sync one's calendar with obsidian. I'll close this pull request and open a new one in the future if I find a solution to this problem. Thank you for your review.

PR #327, opened 2021-06-08
File Tree Alternative Plugin

ozntel:

## Repo URL

https://github.com/ozntel/file-tree-alternative

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1588,6 +1588,7 @@
         "name": "File Tree Alternative Plugin",
         "author": "Ozan Tellioglu",
         "description": "This plugin allows you to have an alternative file tree view.",
-        "repo": "ozntel/file-tree-alternative"
+        "repo": "ozntel/file-tree-alternative",
+        "branch": "main"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1582,5 +1582,12 @@
         "description": "It maintains index files in all of your folders in your vault: if you create/delete/move a note, the index files will be updated automatically. It can be used to show folders in Graph View",
         "author": "Akos Balasko",
         "repo": "akosbalasko/zoottelkeeper-obsidian-plugin"
+    },
+    {
+        "id": "file-tree-alternative",
+        "name": "File Tree Alternative Plugin",
+        "author": "Ozan Tellioglu",
+        "description": "This plugin allows you to have an alternative file tree view.",
+        "repo": "ozntel/file-tree-alternative"
     }
 ]


lishid:
https://github.com/ozntel/file-tree-alternative/blob/0902e02bdf28bf07a2d2158c5678e86091966cca/src/main.ts#L16
Use `workspace.onLayoutReady` intead.

https://github.com/ozntel/file-tree-alternative/blob/0902e02bdf28bf07a2d2158c5678e86091966cca/src/utils.ts#L46
This doesn't really work - you need to call `off` with the exact reference to the function that was originally called with `on`, not a new copy of the function.

https://github.com/ozntel/file-tree-alternative/blob/0902e02bdf28bf07a2d2158c5678e86091966cca/src/utils.ts#L84
You should keep a reference to your leaf so you don't have to attempt to get it again after.

ozntel:
> https://github.com/ozntel/file-tree-alternative/blob/0902e02bdf28bf07a2d2158c5678e86091966cca/src/main.ts#L16
Use workspace.onLayoutReady intead.

Here `if` helps me to detect the reload of plugin during the usage. Since `layout` is already ready during the reload, it won't trigger the load of leaf and `vault events` I would like to register. That's why if `layout` is already ready, they are triggered directly. If not, waiting for the `layout` to be ready.

> https://github.com/ozntel/file-tree-alternative/blob/0902e02bdf28bf07a2d2158c5678e86091966cca/src/utils.ts#L46
This doesn't really work - you need to call off with the exact reference to the function that was originally called with on, not a new copy of the function.

It is exactly the same as initiated in `addEventListenerForFolders` above. 
I see that the line is pushed to the second line in `removeEventListenerForFolders` after `"click"` but the code is exactly the same. 

> https://github.com/ozntel/file-tree-alternative/blob/0902e02bdf28bf07a2d2158c5678e86091966cca/src/utils.ts#L84
You should keep a reference to your leaf so you don't have to attempt to get it again after.

Since I need to always keep the current `leaf` reference somewhere and `leaf` is also coming from Promise, I didn't want to create more complexity. For the moment, I load `leaf` if there is no active `leaf` in the workspace, and remove any `leaf` if there is one. Hence, there shouldn't be more than one `leaf`. Is it ok to keep this way for the moment?

lishid:
> Here if helps me to detect the reload of plugin during the usage. Since layout is already ready during the reload, it won't trigger the load of leaf and vault events I would like to register. That's why if layout is already ready, they are triggered directly. If not, waiting for the layout to be ready.

That's exactly what the `workspace.onLayoutReady` function does. It replaces the `on('layout-ready')` event.

> It is exactly the same as initiated in addEventListenerForFolders above.
I see that the line is pushed to the second line in removeEventListenerForFolders after "click" but the code is exactly the same.

Using the "exact" same code still creates a new function which does the exact same thing, but the js engine considers it to be a completely different function. You need to store a reference to your original function to be used with `off`.

>Since I need to always keep the current leaf reference somewhere and leaf is also coming from Promise, I didn't want to create more complexity. For the moment, I load leaf if there is no active leaf in the workspace, and remove any leaf if there is one. Hence, there shouldn't be more than one leaf. Is it ok to keep this way for the moment?

I think you misunderstood - I meant to change your code like this:
```ts
// Before
await app.workspace.getLeftLeaf(true).setViewState({ type: VIEW_TYPE });
app.workspace.revealLeaf(app.workspace.getLeavesOfType(VIEW_TYPE).first());
// After
let leaf = await app.workspace.getLeftLeaf(true);
await leaf.setViewState({ type: VIEW_TYPE });
app.workspace.revealLeaf(leaf);
```

ozntel:
> That's exactly what the workspace.onLayoutReady function does. It replaces the on('layout-ready') event.

Corrected

> Using the "exact" same code still creates a new function which does the exact same thing, but the js engine considers it to be a completely different function. You need to store a reference to your original function to be used with off.

Can you maybe help me with how I can do this? I already use a function inside but I don't have much idea how can I create a reference

> I think you misunderstood - I meant to change your code like this

I see I misunderstood. I just made the correction as you suggested.

ozntel:
> It usually goes something like this

Ok, I just completed the change as you suggested. 
I also updated the `main.js` within the release

lishid:
That looks great! You can go one step further to simplify your code and only use `folderClickEvent` instead of having to assign it to `FolderClickEventHolder`.

ozntel:
> That looks great! You can go one step further to simplify your code and only use folderClickEvent instead of having to assign it to FolderClickEventHolder.

Sure I will do it in the following release. I need a lot of corrections anyway :-)

If you don't mind, I wanted to ask you 2 topics. The first one I asked in the `plugin-dev` channel, but haven't received a response from anyone. Do you maybe have an idea of how I can achieve it?

https://discord.com/channels/686053708261228577/840286264964022302/851444023276208139

Secondly, I couldn't manage to add `delete` option for the context menu. Here you can find the part:

https://github.com/ozntel/file-tree-alternative/blob/a1ea2172858f9fb98f9e54404b85fffa020d4797/src/components/FileTreeComponent.tsx#L24

Is there a way to add `delete` option, as well?

lishid:
For 1) let's keep it on discord since it's not related to this plugin's code review.

For 2) You shouldn't be using onLinkContextMenu, because the user is right clicking on a file, not a link. You should create your own Menu instead, for which you can add the delete menu item, and then trigger the file-menu event on Workspace to have plugins populate your menu.
`workspace.trigger('file-menu', menu: Menu, file: TAbstractFile, source: string)`.

AidenLx:
> For 2) You shouldn't be using onLinkContextMenu, because the user is right clicking on a file, not a link. You should create your own Menu instead, for which you can add the delete menu item, and then trigger the file-menu event on Workspace to have plugins populate your menu.
> `workspace.trigger('file-menu', menu: Menu, file: TAbstractFile, source: string)`.

Is it possible to add a menu item to the default context menu of the folder/file in explorer without creating a completely new menu?

PR #329, opened 2021-06-10
Add title index plugin

renmu123:

## Repo URL
https://github.com/renmu123/obsidian-markdown-index

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


renmu123:
thanks for your review.l have fixed them

lishid:
https://github.com/renmu123/obsidian-markdown-index/blob/9ab57c9998119129c61f9fc3832271e53fb158e8/main.ts#L4

This should be TitleCase like `TitleIndexPlugin`. Also, any reason why it isn't `MarkdownIndexPlugin`?

https://github.com/renmu123/obsidian-markdown-index/blob/9ab57c9998119129c61f9fc3832271e53fb158e8/markdownIndex.ts#L8

Since your code isn't a lot, why not just move all of these functions into your main plugin class? I don't see the need to instantiate a new copy of MarkdownIndex with `new MarkdownIndex()` every time the command executes either way.

lishid:
Ah my bad, it seems that your repo is named obsidian-markdown-index but the plugin is actually title-index.

lishid:
Let's merge it like this - class renames shouldn't matter, so you can do that later if you wish.

PR #331, opened 2021-06-11
Add `extended-context-menu` Plugin

phibr0:

## Repo URL

https://github.com/phibr0/obsidian-extended-context-menu

## Release Checklist

- [x] I have tested this on Windows, Linux _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


mgmeyers (commented):
I'm super excited for this, @phibr0! A couple of things I noticed:

---

I see you're unregistering the context menu handler now. One thing you'll need to update to get this to work is: https://github.com/phibr0/obsidian-extended-context-menu/blob/master/src/main.ts#L16-L18

Should be

```
cm.on('contextmenu', this.handleContextMenu);
```

also, since `handleContextMenu` references `this`, you may need to convert it to an arrow funtion:

```
private handleContextMenu = (instance: CodeMirror.Editor, e: MouseEvent) => {
```

---

Not a big deal, but you could improve perf here by making `setting.states` an object instead of an Array. https://github.com/phibr0/obsidian-extended-context-menu/blob/master/src/main.ts#L105

That way you could do a direct lookup `this.settings.states[getID(entry.pluginName, entry.name)]`. You may also want to handle the case where the state for a given entry doesn't exist in settings.

phibr0:
> You may also want to handle the case where the state for a given entry doesn't exist in settings.

Hmm, but if it doesn't exist yet it will be created while it's being registered in the `registerCommand` Method.

phibr0:
I fixed the unloading, it should now work correctly

mgmeyers:
> Hmm, but if it doesn't exist yet it will be created while it's being registered in the `registerCommand` Method.

Ah! Didn't notice this. 👍

mgmeyers (approved):

lishid:
LGTM. We will be implementing similar functionality shortly, but I can have this merged meanwhile.

PR #333, opened 2021-06-14
Add QuickAdd plugin

chhoumann:

## Repo URL

[QuickAdd](https://github.com/chhoumann/quickadd)

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/chhoumann/quickadd/blob/e419724f7bc4cc31d10c6e9ec7c0439df91f457f/src/main.ts#L76

Might not be a good idea to await this because it will prevent the app and other plugins from initializing, and prolong the loading time of the app. Or maybe this is intentional?

https://github.com/chhoumann/quickadd/blob/e419724f7bc4cc31d10c6e9ec7c0439df91f457f/src/utility.ts#L42
`app` the global reference is only there for debugging purpose. I suggest you pass the reference from your plugin in case the global reference is changed or removed in the future.

https://github.com/chhoumann/quickadd/blob/e419724f7bc4cc31d10c6e9ec7c0439df91f457f/src/utility.ts#L31
Should use `workspace.getActiveViewOfType(MarkdownView)` and check for null before using the editor.

chhoumann:
> [chhoumann/quickadd@`e419724`/src/main.ts#L76](https://github.com/chhoumann/quickadd/blob/e419724f7bc4cc31d10c6e9ec7c0439df91f457f/src/main.ts#L76)
> 
> Might not be a good idea to await this because it will prevent the app and other plugins from initializing, and prolong the loading time of the app. Or maybe this is intentional?
> 
> [chhoumann/quickadd@`e419724`/src/utility.ts#L42](https://github.com/chhoumann/quickadd/blob/e419724f7bc4cc31d10c6e9ec7c0439df91f457f/src/utility.ts#L42)
> `app` the global reference is only there for debugging purpose. I suggest you pass the reference from your plugin in case the global reference is changed or removed in the future.
> 
> [chhoumann/quickadd@`e419724`/src/utility.ts#L31](https://github.com/chhoumann/quickadd/blob/e419724f7bc4cc31d10c6e9ec7c0439df91f457f/src/utility.ts#L31)
> Should use `workspace.getActiveViewOfType(MarkdownView)` and check for null before using the editor.

Thank you for your review, @lishid. It's very appreciated.

I have addressed the feedback in this commit:
https://github.com/chhoumann/quickadd/commit/89860b89d738e713806789e90c2542c97f03fba1

lishid:
https://github.com/chhoumann/quickadd/blob/3cf2a7e6485e98d7f941dc5081e0be9ace198be8/src/main.ts#L47
This should probably be `workspace.onLayoutReady` instead (which runs your callback if layout is already ready).

lishid:
I'll merge it now, you can fix it once you have time.

PR #334, opened 2021-06-17
Require licenses

JeppeKlitgaard:
Require plugins to have a license in the LICENSE file to avoid any legal snags / too many abandoned projects being lost to the litigious ether down the line.

Typically the MIT license will be a good choice for most projects who want to be very permissive (which might often be the case for plugins that are unlikely to ever be monetized or independent of Obsidian - I am not a lawyer).

Perhaps a comment containing a link to https://choosealicense.com/ could be added to the template, though I didn't want to add 'noise'.

Per discussion in #plugin-dev channel: https://discord.com/channels/686053708261228577/840286264964022302/854099720505786409

Feel free to close/discard this if this isn't how you want it - I just wanted to make sure it wasn't forgotten!

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -10,6 +10,7 @@
   - [ ] `obsidian.css`
   - [ ] The screenshot file you pointed to
 - [ ] You have indicated which modes (dark, light, or both) are compatible with your theme
+- [ ] You have added a license in the LICENSE file
 - [ ] If you're not using the `master` branch (legacy), please indicate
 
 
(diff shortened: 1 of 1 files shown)


shabegom:
I don't want to add a license and I don't believe a license should be required.

JeppeKlitgaard:
> I don't want to add a license and I don't believe a license should be required.

Not adding a license would still be allowed, I guess (A license file stating nothing more than "all rights reserved", perhaps) - this will mean that people cannot reasonably contribute to your code without you doing some sort of CLA. If you leave the community, no one will be able to pick up where you left off and any code you wrote couldn't be used by others.

In this case you'd still want to explicitly state that your code is not open for other's to use. See: https://choosealicense.com/no-permission/

I am not a lawyer, but if you take contributions to your project you enter murky legal territory, since you'd not have the rights to all the code at that point (afaik).

If you see the discussion on Discord, a "license" does not necessarily mean an open source or permissive license.

ryanpcmcquen:
You're automatically adopting 'some' license by having the code on GitHub. Specifically, you are agreeing to their terms (implicitly). I think MPL is a great license that balances the best of MIT and GPL, but we should leave it up to the author.

PR #335, opened 2021-06-19
Adds 'obsidian-habit-tracker' to community plugin list.

duoani:

## Repo URL

[duoani/obsidian-habit-tracker: This plguin for Obsidian creates a simple month view for visualizing your punch records.](https://github.com/duoani/obsidian-habit-tracker)

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/duoani/obsidian-habit-tracker/blob/47939b7f3baef059eec3babdd1fdf86d162e5400/main.ts#L51

Since your renderer doesn't actually add anything that needs to be cleaned up, you can probably do all of this inside the callback from `registerMarkdownCodeBlockProcessor` directly, instead of adding a child with a tracked lifecycle and not doing anything in `onunload`. You can leave it like this if you want, it'll just cost a tad bit of extra performance to keep track of your rendered block's class.

Everything else looks great!

duoani:
Thanks for your review. l have refactor my code and removed render child.

lishid:
Ok, that looks good. One more thing I missed in the last review:

https://github.com/duoani/obsidian-habit-tracker/blob/47939b7f3baef059eec3babdd1fdf86d162e5400/main.ts#L103
While using innerHTML is very convenient, it's highly insecure and can lead to DOM injection attacks because it is easy to forget to sanitize user content.

I would recommend refactoring the code here to build up your DOM elements, instead of using a string HTML. There are some convenience methods that Obsidian provides, which will help with this. You can also use something like svelte to build up your DOM.

```ts
// Create detached element
let tableEl = createEl('table');
// Create element and append to parent
let childEl = parentEl.createEl('p', {cls: 'some-css-class', text: 'some inner text'});
// Shortcuts for div and span
let childDiv = parentElement.createDiv({cls: 'css-class', text: 'some inner text'});
let childSpan = parentElement.createSpan({cls: 'css-class', text: 'some inner text'});
```

duoani:
You're right, I'll find more detail of these APIs.

duoani:
> Ok, that looks good. One more thing I missed in the last review:
> 
> https://github.com/duoani/obsidian-habit-tracker/blob/47939b7f3baef059eec3babdd1fdf86d162e5400/main.ts#L103
> While using innerHTML is very convenient, it's highly insecure and can lead to DOM injection attacks because it is easy to forget to sanitize user content.
> 
> I would recommend refactoring the code here to build up your DOM elements, instead of using a string HTML. There are some convenience methods that Obsidian provides, which will help with this. You can also use something like svelte to build up your DOM.
> 
> ```ts
> // Create detached element
> let tableEl = createEl('table');
> // Create element and append to parent
> let childEl = parentEl.createEl('p', {cls: 'some-css-class', text: 'some inner text'});
> // Shortcuts for div and span
> let childDiv = parentElement.createDiv({cls: 'css-class', text: 'some inner text'});
> let childSpan = parentElement.createSpan({cls: 'css-class', text: 'some inner text'});
> ```

I'd done that.

lishid:
Hmm did you have the change on a dev branch? I'm not seeing changes on master.

duoani:
> Hmm did you have the change on a dev branch? I'm not seeing changes on master.

I am **SO SORRY** for waiting you time. I just `git push --tags` but I didn't know that it just push tags. I had pushed to master already.

lishid:
Looks good!

PR #337, opened 2021-06-20
Add tq plugin

tgrosinger:
New Community Plugin

## Repo URL

https://github.com/tgrosinger/tq-obsidian

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - `styles.css` _(optional)_


lishid:
https://github.com/tgrosinger/tq-obsidian/blob/e1f420a57c9b052d4831df41376e3e244c975254/rollup.config.js#L11
You can make your plugin's distribution file much much smaller by making this disabled (or just disabling source code inclusion) on production builds (it's >4mb right now). See how our sample plugin handles it: https://github.com/obsidianmd/obsidian-sample-plugin/blob/c228a7022301e4a74614a67570d38ed2e2e05a71/rollup.config.js#L5

https://github.com/tgrosinger/tq-obsidian/blob/e1f420a57c9b052d4831df41376e3e244c975254/src/main.ts#L32
https://github.com/tgrosinger/tq-obsidian/blob/e1f420a57c9b052d4831df41376e3e244c975254/src/main.ts#L36
You shouldn't be keeping references to your views on your plugin global. You can get your views using `workspace.getLeavesOfType` instead if you need references to your leaves.

https://github.com/tgrosinger/tq-obsidian/blob/e1f420a57c9b052d4831df41376e3e244c975254/src/main.ts#L41
activeLeaf may be null, and shouldn't be accessed directly like this. Use `workspace.getLeaf` instead, which coincidentally also does what your TODO comment says.

https://github.com/tgrosinger/tq-obsidian/blob/e1f420a57c9b052d4831df41376e3e244c975254/src/main.ts#L60
Same here, you may also want to double check that a the active view if of the correct type (which you do inside the function), but the easiest way here is to use `workspace.getActiveViewOfType(MarkdownView)` and pass the view instead of the leaf.

https://github.com/tgrosinger/tq-obsidian/blob/e1f420a57c9b052d4831df41376e3e244c975254/src/file-interface.ts#L3
https://github.com/tgrosinger/tq-obsidian/blob/e1f420a57c9b052d4831df41376e3e244c975254/src/modals.ts#L5
Use the `moment` from Obsidian to avoid packing a copy of moment.js into your plugin.

tgrosinger:
Thank you for the feedback @lishid! I have made those changes except for the last one about importing `moment`. I have not had luck finding the right way to do that. Do you have an example I could follow?

lishid:
You can either use `window.moment` or `import moment from 'obsidian'`. Make sure you have the latest API definitions to see it :)

tgrosinger:
I did a little research and experimentation. I'm pretty sure moment is not being packaged. The definition there is for typing only (`import type {Moment} from 'moment'`, and in my `package.json` moment is in the dev dependencies.

I experimented with making the change you suggested and did not see any impact on the file size of `main.js`.

tgrosinger:
@lishid, when you have a minute please let me know if there's anything else before merging this.
Thank you.

tgrosinger:
Bump

lishid:
Oh ok I see what you mean. If it's not getting into the final .js then that's fine.

PR #338, opened 2021-06-23
Add obsidian-kroki plugin

gregzuro:

## Repo URL

https://github.com/gregzuro/obsidian-kroki

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
> Github release name matches the exact version number specified in my manifest.json

You did not do this; your repo does not have releases.

https://github.com/gregzuro/obsidian-kroki/blob/61cf5390123e89c5be934c4c94fc386641486ab4/main.ts#L1
This seems unused.

https://github.com/gregzuro/obsidian-kroki/blob/61cf5390123e89c5be934c4c94fc386641486ab4/main.ts#L59
Don't use NodeJS code if you want your plugin to work on mobile as well. Otherwise, change your manifest.json to indicate that your plugin only functions on deskstop.

(Use `ArrayBuffer` instead of `Buffer`)

https://github.com/gregzuro/obsidian-kroki/blob/61cf5390123e89c5be934c4c94fc386641486ab4/main.ts#L172
That's a rather odd way of doing it. You can use `let diagramType = this.plugin.settings.diagramTypes[i]` and always use the same reference to it in your onChange event. Using `let` will cause the variable to be scoped to the inside of the for loop, which means it will not change from one loop iteration to the next. It will also make your code look a lot cleaner by reducing the number of repeated `this.plugin.settings.diagramTypes[i]`.

https://github.com/gregzuro/obsidian-kroki/blob/61cf5390123e89c5be934c4c94fc386641486ab4/main.ts#L83
You should either await this async function, or not mark your function as async.

gregzuro:
https://github.com/gregzuro/obsidian-kroki/commit/96ceefdc74f753556b5e482ec125a0b9211fa6ae

> > Github release name matches the exact version number specified in my manifest.json
> 
> You did not do this; your repo does not have releases.

My releases are being reverted to draft status for some unknown reason.  I'm mirroring this repo from Gitlab (the master) to Github, and suspect that this may be the cause.
There is a release out there right now, however.

> https://github.com/gregzuro/obsidian-kroki/blob/61cf5390123e89c5be934c4c94fc386641486ab4/main.ts#L1
> This seems unused.

Done: Removed.

> https://github.com/gregzuro/obsidian-kroki/blob/61cf5390123e89c5be934c4c94fc386641486ab4/main.ts#L59
> Don't use NodeJS code if you want your plugin to work on mobile as well. Otherwise, change your manifest.json to indicate that your plugin only functions on deskstop.
> 
> (Use `ArrayBuffer` instead of `Buffer`)

Done: (Uint8Array.)

> https://github.com/gregzuro/obsidian-kroki/blob/61cf5390123e89c5be934c4c94fc386641486ab4/main.ts#L172
> That's a rather odd way of doing it. 

If you mean looping through all the diagram types to find the one that has changed, then: (I'm new to javascript.) I'm not sure how to tie the my settings to the processor so I can find the changed element directly.  This was the first thing I came up with that worked... I'd welcome any hints re: a more direct approach.

> You can use `let diagramType = this.plugin.settings.diagramTypes[i]` and always use the same reference to it in your onChange event. Using `let` will cause the variable to be scoped to the inside of the for loop, which means it will not change from one loop iteration to the next. It will also make your code look a lot cleaner by reducing the number of repeated `this.plugin.settings.diagramTypes[i]`.

Done: (2x.)

> https://github.com/gregzuro/obsidian-kroki/blob/61cf5390123e89c5be934c4c94fc386641486ab4/main.ts#L83
> You should either await this async function, or not mark your function as async.

Done: Not marked.

Thanks!

lishid:
> My releases are being reverted to draft status for some unknown reason. I'm mirroring this repo from Gitlab (the master) to Github, and suspect that this may be the cause.
There is a release out there right now, however.

Ah I see, I'm not familiar for mirroring from Gitlab but that may be a bit difficult. Your release still doesn't have the right files uploaded to it for Obsidian to pick it up correctly (manifest.json, main.js).

https://github.com/gregzuro/obsidian-kroki/blob/96ceefdc74f753556b5e482ec125a0b9211fa6ae/main.ts#L49
Definitely use a TextEncoder here instead. It's available for all platforms Obsidian deploys to (and in fact we use it ourselves too).

> If you mean looping through all the diagram types to find the one that has changed, then: (I'm new to javascript.) I'm not sure how to tie the my settings to the processor so I can find the changed element directly. This was the first thing I came up with that worked... I'd welcome any hints re: a more direct approach.

Something like this should do it:
```ts
for (var i=0; i<array.length; i++){
  let item = array[i]; // this will bind the item value to this scope (within the curly brackets of the for loop)
  /* ... */element.onChange(async (value) => {
         item.enabled = value;
         await this.plugin.saveSettings();
  }
}
```

gregzuro:
https://github.com/gregzuro/obsidian-kroki/commit/63465c564bf6915ea21d767d52e70b0d8fe2e6e2

> Your release still doesn't have the right files uploaded to it for Obsidian to pick it up correctly (manifest.json, main.js).

Done: https://github.com/gregzuro/obsidian-kroki/releases/tag/1.1.0

> https://github.com/gregzuro/obsidian-kroki/blob/96ceefdc74f753556b5e482ec125a0b9211fa6ae/main.ts#L49
> Definitely use a TextEncoder here instead. It's available for all platforms Obsidian deploys to (and in fact we use it ourselves too).

Done: https://github.com/gregzuro/obsidian-kroki/commit/63465c564bf6915ea21d767d52e70b0d8fe2e6e2#diff-564c860307257bfac9176f554a35aa858c350a59d75c0dcd0a5d63a480805bd0R58

> Something like this should do it:
That seems to work.
Thanks for the guidance.

Done: https://github.com/gregzuro/obsidian-kroki/commit/63465c564bf6915ea21d767d52e70b0d8fe2e6e2#diff-564c860307257bfac9176f554a35aa858c350a59d75c0dcd0a5d63a480805bd0R172

Thanks,

PR #339, opened 2021-06-24
Add wavedrom plugin

kingsquirrel152:

## Repo URL

https://github.com/kingsquirrel152/obsidian-wavedrom

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/kingsquirrel152/obsidian-wavedrom/releases/tag/0.0.3
You did not attach the release files here.
image

https://github.com/kingsquirrel152/obsidian-wavedrom/blob/main/main.js
You should add this to your gitignore and remove it from the repo. See a sample gitignore here: https://github.com/obsidianmd/obsidian-sample-plugin/blob/master/.gitignore

https://github.com/kingsquirrel152/obsidian-wavedrom/blob/c4f0b808cfa563b70550a62c9e6e4d60c4a8eb5c/main.ts#L1
This seems unused.

https://github.com/kingsquirrel152/obsidian-wavedrom/blob/c4f0b808cfa563b70550a62c9e6e4d60c4a8eb5c/main.ts#L38
You'll want to make sure you don't load this twice. Maybe check for the global variable `WaveDrom` first, and only load the js files if they don't exist?

Also consider just bundling that file into your plugin, that way your plugin users without internet will be able to use the plugin as well.

https://github.com/kingsquirrel152/obsidian-wavedrom/blob/c4f0b808cfa563b70550a62c9e6e4d60c4a8eb5c/main.ts#L49
This doesn't do anything.

https://github.com/kingsquirrel152/obsidian-wavedrom/blob/c4f0b808cfa563b70550a62c9e6e4d60c4a8eb5c/main.ts#L27
Hmmm Does WaveDrom only process elements on the page? Because Obsidian's renderer is highly optimized and it doesn't actually append every HTML element onto the page (You can test this with a long page). Ideally your renderer should expose some API that makes it render a select HTML element, rather than asking it to go through everything on the app page.

kingsquirrel152:
Thank you very much for the feedback!

Apologies for miss understanding the instructions re: adding files to the release - in hindsight I see what is required. Ill cut a better release once the other issues I have are resolved.
> https://github.com/kingsquirrel152/obsidian-wavedrom/releases/tag/0.0.3
> You did not attach the release files here.
> image


> https://github.com/kingsquirrel152/obsidian-wavedrom/blob/main/main.js
> You should add this to your gitignore and remove it from the repo. See a sample gitignore here: https://github.com/obsidianmd/obsidian-sample-plugin/blob/master/.gitignore

Fixed.


> https://github.com/kingsquirrel152/obsidian-wavedrom/blob/c4f0b808cfa563b70550a62c9e6e4d60c4a8eb5c/main.ts#L1
> This seems unused.

Fixed


> https://github.com/kingsquirrel152/obsidian-wavedrom/blob/c4f0b808cfa563b70550a62c9e6e4d60c4a8eb5c/main.ts#L38
> You'll want to make sure you don't load this twice. Maybe check for the global variable `WaveDrom` first, and only load the js files if they don't exist?

Nice catch. Global check guard added.

> Also consider just bundling that file into your plugin, that way your plugin users without internet will be able to use the plugin as well.
> 
Total transparency, I initially tried that and failed to configure the bundle correctly. I am not a JS programmer. Any suggestions on a good way to bundle this correctly?


> https://github.com/kingsquirrel152/obsidian-wavedrom/blob/c4f0b808cfa563b70550a62c9e6e4d60c4a8eb5c/main.ts#L49
> This doesn't do anything.

Fixed. (Removed).


> https://github.com/kingsquirrel152/obsidian-wavedrom/blob/c4f0b808cfa563b70550a62c9e6e4d60c4a8eb5c/main.ts#L27
> Hmmm Does WaveDrom only process elements on the page? Because Obsidian's renderer is highly optimized and it doesn't actually append every HTML element onto the page (You can test this with a long page). Ideally your renderer should expose some API that makes it render a select HTML element, rather than asking it to go through everything on the app page.

I agree. Sadly the WaveDrom API only appears to expose the one render path: https://github.com/wavedrom/wavedrom#html-pages short of forking it and improving it, this was the quick solution. Open to ideas if you have an alternate suggestion?

kingsquirrel152:
@lishid any thoughts?

lishid:
Apologies for the late reply.

> Total transparency, I initially tried that and failed to configure the bundle correctly. I am not a JS programmer. Any suggestions on a good way to bundle this correctly?

Easiest way is to download the file right into your plugin sources folder, and `import 'wavedrom.min.js';` in your plugin.
Though this might be problematic with importing the code twice.

Alternatively, base-64 encode the minified file into a string field, then use a `src='data:'` directive instead. That way you have some control, but the sources isn't directly inlined into your plugin's code.

> I agree. Sadly the WaveDrom API only appears to expose the one render path: https://github.com/wavedrom/wavedrom#html-pages short of forking it and improving it, this was the quick solution. Open to ideas if you have an alternate suggestion?

I took a quick look at WaveDrom's code and it seems that `processAll()` does the querying in the document and calls the [`renderWaveForm()` function](https://github.com/wavedrom/wavedrom/blob/2413e2e4e4219bcf8d7275b955658aa6ab8600ea/lib/wave-drom.js#L12)

You can see the source code here: https://github.com/wavedrom/wavedrom/blob/2413e2e4e4219bcf8d7275b955658aa6ab8600ea/lib/process-all.js

This should help explain how to use the specific renderer function I think.

kingsquirrel152:
@lishid Thank you again for all the good suggestions. *I think* I have fixed all the issues you called out:
- WaveDrom is now wrapped as a base64 encoded blob included in the plugin.
- I forked it while I was wrapping and exposed some more API calls to let the plugin render json -> direct to HTML and not have to refire the whole page.

Let me know if I missed anything, happy to scrub/clean more.

lishid:
https://github.com/kingsquirrel152/obsidian-wavedrom/blob/d3843225bfe4ee6c76dfae1c7c083d19fa159fc5/rollup.config.js#L9
Consider using no sourcemaps (or at least strip the source code) for production/distribution.

HiWQ:
can wavedrom just integrated into OB master release like mermaid ?

PR #342, opened 2021-06-25
Rename Dropbox Backups plugin.

ryanpcmcquen:
I have to rename the plugin to meet Dropbox's branding guidelines, and the app has already surpassed the limit of users for a development app.

https://github.com/ryanpcmcquen/obsidian-dropbox-backups


ericaxu:
Sorry to hear that!

Luckily the plugin marketplace search will search descriptions, so people looking for Dropbox backups can still find it.

PR #344, opened 2021-06-28
Submit Breadcrumbs plugin

SkepticMystic:

## Repo URL

https://github.com/SkepticMystic/breadcrumbs

## Release Checklist

  - Windows and Linux (thanks Koala :D)
  - [x] `main.js`
  - [x] `manifest.json`


lishid:

https://github.com/SkepticMystic/breadcrumbs/blob/a650c88bd9467bb2331e557f31cb9ee9509a7512/src/main.ts#L45

Don't suggest keeping a copy of your view from your the plugin - this could be pointed to a closed/detached view. If you need access to your views, get the latest ones from `workspace.getLeavesOfType`, or alternatively, hook your events from the views themselves.

https://github.com/SkepticMystic/breadcrumbs/blob/a650c88bd9467bb2331e557f31cb9ee9509a7512/src/main.ts#L163
This may error out if `unresolvedLinks[currFile.path]` doesn't exist.

https://github.com/SkepticMystic/breadcrumbs/blob/a650c88bd9467bb2331e557f31cb9ee9509a7512/src/main.ts#L219
`getActiveFile` may return null. It may also return a non-markdown file, just to keep in mind, for something later like here: https://github.com/SkepticMystic/breadcrumbs/blob/a650c88bd9467bb2331e557f31cb9ee9509a7512/src/main.ts#L269

https://github.com/SkepticMystic/breadcrumbs/blob/a650c88bd9467bb2331e557f31cb9ee9509a7512/src/main.ts#L278
Yeah... instead of doing this, get the active view (`workspace.getActiveViewOfType(MarkdownView)`), test for null, and query its `contentEl`.

SkepticMystic:
Thanks for the feedback Lishid :) I'll fix these issues when I can and ping you when I have

SkepticMystic:
Thanks for the feedback, Lishid :)

See my changes here: https://github.com/SkepticMystic/breadcrumbs/commit/cda862baf42016976661277e675b0961548d87aa

1. I'm not sure if I've done it correctly, but now I use `getLeavesOfType` to grab the view instead.

2. I now check if `unresolvedLinks[currFile.path]` exists

3. Do you recommend using this `workspace.getActiveViewOfType(MarkdownView)` to only get md files?

4. I used `workspace.getActiveViewOfType(MarkdownView)` to get the actice view instead. But I get a bug where, when changing notes, it prepends the breadcrumbs trail in a non-visibile area above the note content. You'll see then that I now `querySelect` from the active markdown view's `contentEl` instead

lishid:
https://github.com/SkepticMystic/breadcrumbs/commit/cda862baf42016976661277e675b0961548d87aa#diff-9936cb357bae853e59c54fe2eae8abeeab607a3332b9b7f82843a5b1326a38bfR61

- First, don't explicitly check for undefined, just check for a falsy value (in case it's null or something).
- Second, I think if it doesn't exist, you should assume that such a link doesn't exist, instead of completely erroring out.

> I'm not sure if I've done it correctly, but now I use getLeavesOfType to grab the view instead.

Yeah so in general, there seem to be a lot of trouble working with null or nullable types (I know it's confusing!). Here there's a few places that I can see potentially problematic:

https://github.com/SkepticMystic/breadcrumbs/commit/cda862baf42016976661277e675b0961548d87aa#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR77
Missing null check before you call `draw`.

https://github.com/SkepticMystic/breadcrumbs/commit/cda862baf42016976661277e675b0961548d87aa#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR91
This null check is more expensive than it needs to be - `getActiveView`, which calls `getLeavesOfType` actually loops through all the views, which means you're performing this computation twice. Instead, you should cache the result into a variable, or use `?` before calling draw.

https://github.com/SkepticMystic/breadcrumbs/commit/cda862baf42016976661277e675b0961548d87aa#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR120

Here's my attempt at a very explicit way:

```ts
getActiveView(): MatrixView | null {
  let views = this.app.workspace.getLeavesOfType(VIEW_TYPE_BREADCRUMBS_MATRIX);
  if (views && views.length > 1) {
    let view = views[0];
    if (view instanceof MatrixView) {
      return view;
    }
  }
  return null;
}
````

https://github.com/SkepticMystic/breadcrumbs/blob/57229f9d51045b367bf7a552ea6077f6fd7683fd/src/main.ts#L282
Maybe run this check first? You can get the active file from this view's `.file` too without having to query it from workspace again.

https://github.com/SkepticMystic/breadcrumbs/blob/57229f9d51045b367bf7a552ea6077f6fd7683fd/src/main.ts#L291
That's a little bit weird? I'd expect `respectReadableLineLength` to toggle the `is-readable-line-width` css class...

> Do you recommend using this workspace.getActiveViewOfType(MarkdownView) to only get md files?

It will only get you markdown files, but you can also just check `file.extension === 'md'` if you are working with a file not inside a view.

> I used workspace.getActiveViewOfType(MarkdownView) to get the active view instead. But I get a bug where, when changing notes, it prepends the breadcrumbs trail in a non-visibile area above the note content. You'll see then that I now querySelect from the active markdown view's contentEl instead of document

I'm not sure I fully understand, does using querySelect fix the issue?

SkepticMystic:
@lishid see my changes here: https://github.com/SkepticMystic/breadcrumbs/commit/10b2810cbbe563caa9573bb7ffca715d633c7ba5

1. Alright, I now check for falsey values, and assume a link doesn't exist if that's the case.

2. I check if activeView is truthy before calling `draw`

3. Fixed inefficient null check in two places

4. Implemented your improved `getActiveView()` (I think there was a mistake. Shouldn't the third line be `if (leaves && leaves.length >= 1) {`? not `> 1`

5.1 I now check if the activeMDView is truthy before running other computations
5.2 And I get the active file from the view instead of going to the workspace

6. haha it is weird... It seems that readable-line-width only takes effect if the element (or its parent) has `.markdown-preview-sizer.markdown-preview-section`. So the way I had it technically works, but I've still changed it

7. I've added better checking when getting the active file

8. Yes, so now I use `app.workspace.getActiveViewOfType(MarkdownView)` to get the active view, and then I query it's `contentEl` to get it's `..markdown-preview-view`.
So it works

lishid:
For 4. you are right, that was my mistake!

All good, merged now!

PR #345, opened 2021-06-29
Added a new Map View plugin

esm7:

## Repo URL

https://github.com/esm7/obsidian-map-view

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


HEmile:
Hi! This looks like an excellent plugin and I cannot wait to play with it :) Here's a quick review:

I am not seeing main.js and manifest.json in your release: https://github.com/esm7/obsidian-map-view/releases/tag/0.0.1

What are you going to do with this event? https://github.com/esm7/obsidian-map-view/blob/f1bbd3a84c13fca3c84a583e106065eadc4f04f2/src/main.ts#L72

It looks like you give the option for Google maps tiles: https://github.com/esm7/obsidian-map-view/blob/f1bbd3a84c13fca3c84a583e106065eadc4f04f2/src/consts.ts#L7  I see on this site that this might be against the TOS https://gis.stackexchange.com/questions/325925/adding-detailed-google-maps-baselayer-on-leaflet , that might be relevant for the review.

esm7:
1. Missing files in the release: oops, looks like my upload failed and I didn't notice! Fixed.

2. Empty event: removed, but actually if I knew you guys will review the code, I'd clean up before posting :) I still want to do lots of improvements in the upcoming weeks and one of the first steps, now that the plugin is mostly feature-complete, is to restructure it better. But I wanted to go ahead and release an initial version anyway, so I can get a sense of how useful the current feature set is for people.

3. Google Maps tiles: I reviewed the TOS and to my best understanding, this usage in a plugin doesn't violate anything. However there's a proper required copyright notice that's missing and I'll add that right away.

esm7:
I stand corrected, complying to the Google Maps attribution guidelines is actually very tricky.
I'm removing it, OpenStreetMap is a perfectly fine alternative.

esm7:
Done with everything, including some refactoring, and released 0.0.2.
Looking forward for the next step :)

esm7:
What's the next action here? Any chance to get this merged soon? :)

HEmile:
You'll need to be merged by the Obsidian developers (Licat in particular). Some plugin devs (like me) sometimes chime in to review plugins to help Licat out a bit. You'll have to wait for him, I'm afraid!

lishid:

https://github.com/esm7/obsidian-map-view/blob/34c0d9cda1ac02695e78f5c3d873195e261ee9fd/src/main.ts#L66
This isn't the right way to load views. You should use `workspace.getLeaf().setViewState({type: MY_VIEW_TYPE});` to get hold of a leaf and let the leaf take care of loading the view. `activeLeaf` can also be null.

https://github.com/esm7/obsidian-map-view/blob/34c0d9cda1ac02695e78f5c3d873195e261ee9fd/src/main.ts#L68
You can also pass this information when setting the view state, and in your view, you can override the `setState` and `getState` methods to have your view information persisted.

https://github.com/esm7/obsidian-map-view/blob/34c0d9cda1ac02695e78f5c3d873195e261ee9fd/src/mapView.ts#L362
Same here, don't do this. Instead, you should use `workspace.getActiveViewOfType(MarkdownView)`. You should also use `view.editor` instead of `view.sourceMode.cmEditor` unless you specifically needs CodeMirror 5 - otherwise you will have to mark your plugin as desktop-only in your manifest.json because it will crash on mobile which does not use CodeMirror 5.

https://github.com/esm7/obsidian-map-view/releases/tag/0.0.5
Your release file is quite big at 5.15mb - over half of that is the sourcemap. I think you can remove the sourcemap in your build options for the production build to save some memory for users :)

esm7:
Fixed all and fixed the merge conflict.
I didn't make a new release in case you'll have more comments, but the release file is now 2.2MB.

lishid:
Looks good!

https://github.com/esm7/obsidian-map-view/blob/9774d2f56c35c971fa202e43059609b8c292f475/src/mapView.ts#L358
`getActiveViewOfType` will convert the return type to MarkdownView so I think you don't need to check `getViewType` and also don't need to convert `markdownView = view as MarkdownView`. `view.editor` should just work, but you'll still need to `if(view)` check.

PR #346, opened 2021-06-30
Adding obsidian-pocket to the community plugins list

nybbles:

## Repo URL

https://github.com/nybbles/obsidian-pocket/

## Release Checklist

  - (tested on macOS, not sure about what plugin devs do to test on other platforms that they do not have machines for)
  - [x] `main.js`
  - [x] `manifest.json`


nybbles:
Still need to test on Windows and Linux

nybbles:
Please LMK if it is imperative to test on Windows/Linux and what are some good solutions for that. I've looked into Windows/Linux VMs on Azure, but would rather not create an account and enter CC details unless really needed.

kometenstaub:
It's not imperative, it's just to let the plugin reviewer know on which platforms you have tested it.

lishid:
Sorry for the late review! One good news I have is that we will be releasing a request API that bypasses CORS. This will be soon available on both desktop versions (expecting v0.12.11+) and mobile versions (v1.0.1+).

Perhaps it'd be best to wait until that's released so you can clean out the CORS bypass, to review the plugin?

nybbles:
That sounds like a reasonable plan, so that the CORS proxy stuff here wouldn't even need to be reviewed.

I hope CORS handling in Obsidian lands soon, as I'd love to make this easily installable for users.

Thanks @lishid.

nybbles:
@lishid hey, just following up on the CORS proxy issue. I have removed the dependence on CORS proxy stuff, in favor of Obsidian's request API, as of `obsidian-pocket` 0.3.0.

Please advise on next steps!

lishid:
Ok, here's the first pass review:

https://github.com/nybbles/obsidian-pocket/blob/7763cded25029184f34ed3f14ee60a72653daa20/src/main.ts#L66
Don't run any async long-running (or networking) code in your plugin's onload, since that will block Obsidian from loading and adds a significant amount of delay to starting up the app. Instead, run it asynchronously (without await).

nybbles:
> https://github.com/nybbles/obsidian-pocket/blob/7763cded25029184f34ed3f14ee60a72653daa20/src/main.ts#L66
> Don't run any async long-running (or networking) code in your plugin's onload, since that will block Obsidian from loading and adds a significant amount of delay to starting up the app. Instead, run it asynchronously (without await).

The plugin only sends out network requests (to the Pocket API) from clicking on buttons in the Settings page (the "Connect your Pocket account" button and the "Sync Pocket items" button), and as a result of the Obsidian protocol handler registered here: https://github.com/nybbles/obsidian-pocket/blob/7763cded25029184f34ed3f14ee60a72653daa20/src/main.ts#L71.

The function you pointed out eventually just accesses `plugin.app.vault`: https://github.com/nybbles/obsidian-pocket/blob/7763cded25029184f34ed3f14ee60a72653daa20/src/PocketAuth.ts#L47. It does not result in any long-running calls or involve network requests.

`await`-ing on `plugin.app.vault` accesses should be okay, right?

lishid:
Oops my bad. I didn't really read through all the inside code and just assumed things based on the function and error message.

I don't think it would make a big difference, but I would still suggest running it in an IIFE without awaiting the results like this:
```ts
(async() => {
    const accessInfo = await loadPocketAccessInfo(this);
    if (!accessInfo) {
      console.info(`Not authenticated to Pocket`);
    }
    this.pocketAuthenticated = !!accessInfo;
    this.pocketUsername = accessInfo?.username;
})();
```

nybbles:
alright, addressed that feedback in https://github.com/nybbles/obsidian-pocket/pull/31. I just implemented it based exactly on your code, so thanks for that.

Keep the code feedback coming!

PR #353, opened 2021-07-03
Remove extended-context-menu + Update my Plugin Info

phibr0:
- I removed mentions of my old discord name (rythm)
- I removed the extended-context-menu Plugin since its not needed anymore, for my knowledge no Plugin relies on it
- On Discord I was told that using "Obsidian" in a Plugin Name is more or less bad, since of course its for Obsidian and it messes with the sorting


phibr0:
Since I didn't change the ID nothing should break right?

PR #357, opened 2021-07-11
Add Obsimian to community plugins

orlade:
Obsimian is an Obsidian simulation framework for integration testing Obsidian plugins without an actual Obsidian `App` instance.

Still early days, but it will evolve with my other plugins. Tested and working for my simple test vault on Mac and Wnidows.

## Repo URL

https://github.com/motif-software/obsimian

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
Very cool!

Given the nature of your framework, may I suggest renaming the plugin to something more like "Obsimian Exporter" to avoid confusion.

https://github.com/motif-software/obsimian/blob/e7152860f1fd02c4acd2bada2ec262004564a4dc/src/plugin/main.ts#L30
I see that you have an option to use ~ as the out dir, but I'd still suggest not depending on the desktop app dependencies if possible (`import 'path'`, `FileSystemAdapter`, and basePath), and write directly using the `vault` or `adapter` API.

https://github.com/motif-software/obsimian/releases/tag/0.1.0
I'd highly suggest keeping out `lodash` and `fs-extra` - lodash is actually taking up 90% of your bundle size, and fs-extra is also chewing up a fair bit.

Your plugin, without these two dependencies, only weighs 5kb, instead of 1.1mb!

orlade:
Thanks lishid, really great feedback. I'm new to the NPM publishing game, so I'll make those changes today.

orlade:
Done, release 0.2.0 @ https://github.com/motif-software/obsimian/releases/tag/0.2.0.

PR #359, opened 2021-07-13
add Copy URL in preview

NomarCub:

## Repo URL

https://github.com/NomarCub/obsidian-copy-url-in-preview

## Release Checklist

- [ ] I have tested this on Windows, macOS, and Linux _(if applicable)_ - only Windows and Android was available to me to test on
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
Pretty nice! I want to let you know that this is indeed a functionality we'll add into the app eventually. Until then, I'm sure users would be delighted to use this plugin.

https://github.com/NomarCub/obsidian-copy-url-in-preview/blob/968c66df45eab9f445d4fd60383fd932288dbd47/main.ts#L32
Any reason to use capture mode here? I'd suggest not using it if necessary. Otherwise, consider putting a comment in code explaining.

https://github.com/NomarCub/obsidian-copy-url-in-preview/blob/968c66df45eab9f445d4fd60383fd932288dbd47/main.ts#L8
Any reason to re-declare all this? I think we declare the on/off functions in our .d.ts already, but if it doesn't do what you need then please document the discrepancy in a comment to explain why this is needed.

https://github.com/NomarCub/obsidian-copy-url-in-preview/blob/968c66df45eab9f445d4fd60383fd932288dbd47/main.ts#L30
Consider making this `"a.external-link"` to make sure you get an anchor element

https://github.com/NomarCub/obsidian-copy-url-in-preview/blob/968c66df45eab9f445d4fd60383fd932288dbd47/main.ts#L41
Instead of augmenting the event with `target: {href: string}`, the best practice here would be to do the following:

```ts
let element = event.target;
if(!(element instanceof HTMLAnchorElement)) {
  return;
}
// TypeScript will now know that element is an anchor element and `.href` will work!
let href = element.href;
//...
```

NomarCub:
I couldn't find what exactly capture mode does in the API. Should I also not use it [here](https://github.com/NomarCub/obsidian-copy-url-in-preview/blob/968c66df45eab9f445d4fd60383fd932288dbd47/main.ts#L63)?

I used the re-declaration for other reasons, but now I'm not sure how to refactor. `document` is not a HTMLElement, but I'd rather use HTMLElement. What do you suggest? Is there some HTMLElement object other than document that I could use here?

lishid:
Ah, if you don't need capture mode, then don't use it. Should be safe to remove on both spots as long as it works without that flag.

(You can find a tutorial on capture mode here: https://javascript.info/bubbling-and-capturing )

> I used the re-declaration for other reasons, but now I'm not sure how to refactor. document is not a HTMLElement, but I'd rather use HTMLElement. What do you suggest? Is there some HTMLElement object other than document that I could use here?

Weird, does it not work when you remove the declaration? We declare a version for `document` and a version for `HTMLElement` here:
https://github.com/obsidianmd/obsidian-api/blob/b0aa06eab03d8c39c0733949da5aa1edc2084b01/obsidian.d.ts#L176

NomarCub:
It works if I use `Document` in my interfaces.

When I use `HTMLElement` it says on use of `document`: `Argument of type 'Document' is not assignable to parameter of type 'HTMLElement'. Type 'Document' is missing the following properties from type 'HTMLElement': ...`.

When I use `HTMLElement | Document` it says on use of `el.on(`: `The 'this' context of type 'HTMLElement | Document' is not assignable to method's 'this' of type 'HTMLElement & Document'. Type 'HTMLElement' is not assignable to type 'HTMLElement & Document'. Type 'HTMLElement' is missing the following properties from type 'Document': ...`.

I'm not super familiar with TypeScript typing. I hoped for an elegant solution where I can use HTMLElement on interfaces and don't have to cast Document to HTMLElement. I'll use Document if this is not practical.

lishid:
Ah yes I see the issue now - you're trying to merge the two types into a single function. Just write two separate functions instead of forcing both into one  `onElement`?

NomarCub:
I only need one for Document for now, so I'll leave it at that. Thank you for your time! I published the fixes.

lishid:
Ok that works.

I'd propose that instead of wrestling with TypeScript, you should just write the code right there anyway:

```ts
	onload() {
		let fn = this.onClick.bind(this);
		document.on("contextmenu", "a.external-link", fn);
		this.register(() => el.off("contextmenu", "a.external-link", fn));
	}
```

PR #362, opened 2021-07-15
Add Obsidian Encryption

ehsan-keshavarzian:
Added obsidian-encryption

- [x] I am submitting a new Community Plugin

## Repo URL

https://github.com/ehsan-keshavarzian/obsidian-encryption

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1735,5 +1735,12 @@
         "author": "Greg Zuro",
         "description": "Render Kroki diagrams.",
         "repo": "gregzuro/obsidian-kroki"
+    },
+    {
+        "id": "obsidian-encryption",
+        "name": "Obsidian Encryption",
+        "author": "Ehsan Keshavarzian",
+        "description": "Encrypt/Decrypt part or entire document.",
+        "repo": "ehsan-keshavarzian/obsidian-encryption"
     }
 ]


lishid:
Hi thank you for the plugin - have you checked out this plugin? https://github.com/meld-cp/obsidian-encrypt

https://github.com/ehsan-keshavarzian/obsidian-encryption/blob/08c61241d719f46adf23283aa2e39d4e29c48bf9/main.ts#L2
Either way I'd highly suggest not using `crypto` as it's only available on the desktop app. You can use `subtleCrypto` API as that's widely available across all platforms (https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto).

https://github.com/ehsan-keshavarzian/obsidian-encryption/blob/08c61241d719f46adf23283aa2e39d4e29c48bf9/main.ts#L27
You should also avoid using `view.sourceMode.cmEditor` and instead use `view.editor`.

https://github.com/ehsan-keshavarzian/obsidian-encryption/blob/08c61241d719f46adf23283aa2e39d4e29c48bf9/main.ts#L35
This would make it quite easy to accidentally encrypt twice, or decrypting an already decrypted block. Perhaps make a distinction between the two?

ehsan-keshavarzian:
Thanks for the reply. I made the changes and made a new version.
I did check out obsidian-encrypt, but I really don't want to run anything other than my own code even if I can see the source code. And I don't want to insert the password every time.
BTW, it's kind of weird that we need to publish the plugin to be able to test it. I mean I couldn't find any info about how to test it on my own obsidian without / before publishing the plugin.

lishid:
I'm just suggesting that you two might want to consider collaborating instead of creating a separate plugin that's almost identical with minor differences (like saving the password).

As for testing the plugin, you can definitely compile and run it locally: https://github.com/obsidianmd/obsidian-sample-plugin#first-time-developing-plugins

Seems like your repo isn't forked from our template - that's fine, but you've probably missed the instructions and sample setups. Feel free to browse around that repo for more examples.

PR #366, opened 2021-07-15
add Open vault in VSCode

NomarCub:
# [x ] I am submitting a new Community Plugin

## Repo URL

https://github.com/NomarCub/obsidian-open-vscode

## Release Checklist

- [ ]  I have tested this on Windows, macOS, and Linux (if applicable) - only Windows was available to me to test on
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1751,6 +1751,13 @@
         "repo": "NomarCub/obsidian-copy-url-in-preview"
     },
     {
+        "id": "open-vscode",
+        "name": "Open vault in VSCode",
+        "author": "NomarCub",
+        "description": "Ribbon button and command to open vault as a Visual Studio Code workspace",
+        "repo": "NomarCub/obsidian-open-vscode"
+    },
+    {
         "id": "initiative-tracker",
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/NomarCub/obsidian-open-vscode/blob/5088638dd236c88cf332940fba73959de6e99944/main.ts#L11
From the documentation of addIcon:
> the content of the SVG, without the <svg>. Must fit viewBox="0 0 100 100".

Here you just need the part starting from `<path ...>`

https://github.com/NomarCub/obsidian-open-vscode/blob/5088638dd236c88cf332940fba73959de6e99944/main.ts#L29
Don't cast, do an `instanceof` check instead and do nothing otherwise.

NomarCub:
I fixed these in 1.0.1.

liamcain:
> Passing addIcon the whole works, is it still unsupported? Do I have any other way to set the viewbox?

Sorry, only seeing this message now. We now support custom viewboxes if you pass in the full svg string. So we removed the API note about requiring the 100x100 viewbox.

PR #367, opened 2021-07-16
Add LogSeq format plugin

ruivieira:

## Repo URL

https://github.com/ruivieira/obsidian-plugin-logseq

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1785,5 +1785,13 @@
         "author": "NomarCub",
         "description": "Ribbon button and command to open vault as a Visual Studio Code workspace",
         "repo": "NomarCub/obsidian-open-vscode"
+    },
+    {
+        "id": "logseq-format",
+        "name": "LogSeq markdown formatting",
+        "author": "ruivieira",
+        "description": "Render LogSeq-specific markdown",
+        "repo": "ruivieira/obsidian-plugin-logseq",
+        "branch": "main"
     }
 ]


lishid:
You should remove main.js from your repo and add it to .gitignore - keeping the compiled versions to your GitHub releases should be sufficient.

I'd suggest a different plugin name - using just "logseq" makes it sound like this would bring Logseq features to Obsidian, or makes Logseq run inside Obsidian. Instead, I'd suggest something like "logseq-compat" (short for compatibility) or "logseq-markdown-render" to make it less likely to confuse users.

https://github.com/ruivieira/obsidian-plugin-logseq/blob/fe67cbb92327755944e9a61b262f956890ca544b/main.ts#L105
This isn't a good idea - apart from possible vulnerabilities of DOM injection/DOM clobbering, you are also disrupting any event handlers and object references. This means that your plugin will likely disrupt other plugins' functionality if they use pos-processors and dynamically update the content after it's been rendered.

https://github.com/ruivieira/obsidian-plugin-logseq/blob/fe67cbb92327755944e9a61b262f956890ca544b/main.ts#L139
You can use `this.registerMarkdownPostProcessor` to automatically un-register it.

https://github.com/ruivieira/obsidian-plugin-logseq/blob/fe67cbb92327755944e9a61b262f956890ca544b/main.ts#L141
You should find a way to remove this overlay when you unload your plugin.

ruivieira:
Hi @lishid, thanks for the feedback.
Some of the changes implemented in https://github.com/ruivieira/obsidian-plugin-logseq/releases/tag/0.0.4:

> You should remove main.js from your repo and add it to .gitignore - keeping the compiled versions to your GitHub releases should be sufficient.

This is now removed.

> I'd suggest a different plugin name - using just "logseq" makes it sound like this would bring Logseq features to Obsidian, or makes Logseq run inside Obsidian. Instead, I'd suggest something like "logseq-compat" (short for compatibility) or "logseq-markdown-render" to make it less likely to confuse users.

Name changed to `logseq-compat` on `manifest.json` and `package.json`.

> https://github.com/ruivieira/obsidian-plugin-logseq/blob/fe67cbb92327755944e9a61b262f956890ca544b/main.ts#L105
> This isn't a good idea - apart from possible vulnerabilities of DOM injection/DOM clobbering, you are also disrupting any event handlers and object references. This means that your plugin will likely disrupt other plugins' functionality if they use pos-processors and dynamically update the content after it's been rendered.

Add new elements and/or styling where possible.

> https://github.com/ruivieira/obsidian-plugin-logseq/blob/fe67cbb92327755944e9a61b262f956890ca544b/main.ts#L139
> You can use `this.registerMarkdownPostProcessor` to automatically un-register it.
> 

Using `this.registerMarkdownPostProcessor`.

> https://github.com/ruivieira/obsidian-plugin-logseq/blob/fe67cbb92327755944e9a61b262f956890ca544b/main.ts#L141
> You should find a way to remove this overlay when you unload your plugin.

CodeMirror overlay removed when unloading plugin.

Let me know if there's still any further change that needs to be done.

lishid:
https://github.com/ruivieira/obsidian-plugin-logseq/blob/7146dfb4952762b3c1b22aa91ee2101feacd8a1b/main.ts#L154
Would it be possible to have this replaced with DOM manipulations as well?

https://github.com/ruivieira/obsidian-plugin-logseq/blob/7146dfb4952762b3c1b22aa91ee2101feacd8a1b/main.ts#L208
I think the proper way here should be using `workspace.iterateCodeMirrors` rather than `registerCodeMirror`.

PR #373, opened 2021-07-19
Re-add obsidian-filename-heading-sync

dvcrn:
The plugin got disabled due to a reported data loss issue. The plugin got hardened and a few bugs fixed so I'd like to re-submit it - https://github.com/dvcrn/obsidian-filename-heading-sync/issues/18


lishid:
Great work!

PR #374, opened 2021-07-19
feat: add obsidian-vim-im-switch-plugin

yuanotes:

## Repo URL

https://github.com/yuanotes/obsidian-vim-im-switch-plugin

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1792,5 +1792,12 @@
         "author": "NomarCub",
         "description": "Ribbon button and command to open vault as a Visual Studio Code workspace",
         "repo": "NomarCub/obsidian-open-vscode"
+    },
+    {
+        "id": "obsidian-vim-im-switch-plugin",
+        "name": "Vim Input Method Switch",
+        "author": "Yuan Chen",
+        "description": "Switch input method with fcitx-remote when Vim keymap is enabled.",
+        "repo": "yuanotes/obsidian-vim-im-switch-plugin"
     }
 ]


lishid:
https://github.com/yuanotes/obsidian-vim-im-switch-plugin/blob/039be88cbdda20f2ef70155cc1ef75d5755f4d3d/main.ts#L39
Your plugin does not clean up this event on exit - if a user enables and disables your plugin, it will continue to run; if a user enables your plugin twice, it will run twice every time.

To cleanup, you should use `cm.off('event', callback)` and give it the exact reference of the callback that was passed in the `on`.
You can use `Workspace.iterateCodeMirrors((cm) => {/*...*/});` to go through all active CodeMirror instances in your onunload function.

I also suggest not keeping a reference to `cm` from your plugin, to avoid memory leaks when that instance of CodeMirror goes out of use.

yuanotes:
> https://github.com/yuanotes/obsidian-vim-im-switch-plugin/blob/039be88cbdda20f2ef70155cc1ef75d5755f4d3d/main.ts#L39
> Your plugin does not clean up this event on exit - if a user enables and disables your plugin, it will continue to run; if a user enables your plugin twice, it will run twice every time.
> 
> To cleanup, you should use `cm.off('event', callback)` and give it the exact reference of the callback that was passed in the `on`.
> You can use `Workspace.iterateCodeMirrors((cm) => {/*...*/});` to go through all active CodeMirror instances in your onunload function.
> 
> I also suggest not keeping a reference to `cm` from your plugin, to avoid memory leaks when that instance of CodeMirror goes out of use.

Thanks for reviewing. I will fix that.

PR #375, opened 2021-07-19
Add Advanced Mobile Toolbar Plugin

phibr0:

## Repo URL

https://github.com/phibr0/obsidian-advanced-toolbar
## Release Checklist

	- Tested on Android
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1792,5 +1792,13 @@
         "author": "NomarCub",
         "description": "Ribbon button and command to open vault as a Visual Studio Code workspace",
         "repo": "NomarCub/obsidian-open-vscode"
+    },
+    {
+        "id": "advanced-toolbar",
+        "name": "Advanced Mobile Toolbar",
+        "author": "phibr0",
+        "description": "Enhances the Toolbar for Obsidian Mobile",
+        "repo": "phibr0/obsidian-advanced-toolbar",
+        "branch": "master"
     }
 ]


lishid:
https://github.com/phibr0/obsidian-advanced-toolbar/blob/b8ce78d671ff832a26fccaee2fad31b77c4ea10a/src/main.ts#L12
This check should be changed to `w.getActiveLeafOfType(MarkdownView)?.getMode() === 'source'`

https://github.com/phibr0/obsidian-advanced-toolbar/blob/b8ce78d671ff832a26fccaee2fad31b77c4ea10a/src/main.ts#L34
Should wrap this in a `this.registerEvent` to have it unloaded properly.

https://github.com/phibr0/obsidian-advanced-toolbar/blob/b8ce78d671ff832a26fccaee2fad31b77c4ea10a/src/ui/icons.ts#L6
The API here requires you to pass in the svg's contents, without the `svg` tag (usually just a `<path ... >`).

phibr0:
> The API here requires you to pass in the svg's contents, without the svg tag (usually just a <path ... >).

Hmm, it does in fact work fine even though the svg's are nested.

lishid:
Yeah it does seem to work as the DOM's HTML parser is rather forgiving, but I still suggest making the change.

phibr0:
> This check should be changed to w.getActiveLeafOfType(MarkdownView)?.getMode() === 'source'

You mean `getActiveViewOfType()` right?

If so I implemented the suggestions except the one for the Icons, I have a regex ready to remove the svg tags without parsing every single icon (`/(?<=>).*(?=<\/svg>)/`). But removing the Tag with all the attributes (`fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"`) the Icon won't look right. 
How would I go about that?

lishid:
> You mean getActiveViewOfType() right?

Right, mistyped there, sorry!

>If so I implemented the suggestions except the one for the Icons, I have a regex ready to remove the svg tags without parsing every single icon (/(?<=>).*(?=<\/svg>)/). But removing the Tag with all the attributes (fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round") the Icon won't look right.
How would I go about that?

This is an example of what we use, I think you can put them on the path:

```
let data = '<path fill="currentColor" stroke="currentColor" d="....."/>';
```

phibr0:
Some Feather Icons actually don't only use a `path`, instead there are also tags like `polygon`. So there are multiple tags inside the SVG which I would need to parse and edit for every Icon. Its definitely doable but I would need to do it for more than 250 Icons.
Also it does work on both Desktop (I actually have done it that way for all my Plugins..) and on at least Android (iOS wasn't tested but I guess it works also), so even though it's not as clean it doesn't create any Problems so why not just leave it as that?

lishid:
Sure I'm fine with that.

phibr0:
Okay! I think its ready then

lishid:
I don't have commit permission to your fork to fix the merge conflict for you - I'll merge once you fix the conflict.

phibr0:
Seems like I cant fix the Conflicts since I deleted the Fork already on my Account. Was able to open this PR via the GitHub CLI but I cant push my commits since its trying to push to this Repo instead.

I will open a new fixed PR shortly.

PR #382, opened 2021-07-24
Update community-plugins.json

phibr0:

## Repo URL

https://github.com/phibr0/obsidian-plugin-changelogs

## Release Checklist

    - Tested on Windows, Linux
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1810,7 +1810,7 @@
     {
         "id": "plugin-changelogs",
         "name": "Plugin Changelogs",
-        "author:" "phibr0",
+        "author": "phibr0",
         "description": "Read Plugin Changelogs inside Obsidian",
         "repo": "phibr0/obsidian-plugin-changelogs"
     }
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1806,5 +1806,12 @@
         "author": "kurakart",
         "description": "Garbling text in Obsidian turns all content in the entire app (notes, sidebar, etc) into lines so you can take screenshots without exposing sensitive data.",
         "repo": "kurakart/garble-text"
+    },
+    {
+        "id": "plugin-changelogs",
+        "name": "Plugin Changelogs",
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/phibr0/obsidian-plugin-changelogs/blob/0af81b3247027b9682668dae548a351acdfc22c5/src/main.ts#L27
Don't do this in your onload, because it will block the execution of the app and make the loading period longer. Run this lazily (only when needed).

https://github.com/phibr0/obsidian-plugin-changelogs/blob/0af81b3247027b9682668dae548a351acdfc22c5/src/main.ts#L50
Nice touch!

https://github.com/phibr0/obsidian-plugin-changelogs/blob/0af81b3247027b9682668dae548a351acdfc22c5/src/main.ts#L56
Be very careful here. Github's API is highly throttled (I think 50/hour), whereas `raw.githubusercontent.com` isn't (like 5k/hour).
That's why our plugin delivery mechanism isn't checking `api.github.com` for the latest release, and we rely on manifest.json instead.
I don't think there's a better way apart from parsing the HTML from the releases page, but at the very least you should cache the results.

phibr0:
Implemented the Lazy Loading. 

> Be very careful here. Github's API is highly throttled (I think 50/hour), whereas raw.githubusercontent.com isn't (like 5k/hour).
That's why our plugin delivery mechanism isn't checking api.github.com for the latest release, and we rely on manifest.json instead.
I don't think there's a better way apart from parsing the HTML from the releases page, but at the very least you should cache the results.

Hm I am not sure if caching would make a lot of sense here. I would still need to do a API call for every plugin of which changelogs are looked at and who is ever going to look at the changelog of a single plugin 50 times in one hour?  
Remember its only called once per Plugin and only on viewing the Changelog

Update: Still implemented somewhat of a cache for every Plugin. Will still get throttled when you look at changelogs of so many Plugins in one hour xD

PR #384, opened 2021-07-25
Add Random To-Do Plugin

NatiAris:

## Repo URL

Link to my plugin: https://github.com/NatiAris/obsidian-random-todo

## Release Checklist

  - I don't have access to a Windows machine
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1806,5 +1806,12 @@
         "author": "kurakart",
         "description": "Garbling text in Obsidian turns all content in the entire app (notes, sidebar, etc) into lines so you can take screenshots without exposing sensitive data.",
         "repo": "kurakart/garble-text"
+    },
+    {
+        "id": "obsidian-random-todo",
+        "name": "Random To-Do",
+        "author": "NatiAris",
+        "description": "Open a random file containing your custom to-do marker, or a random marker at its position.",
+        "repo": "NatiAris/obsidian-random-todo"
     }
 ]


lishid:
> I used allegedly mobile-incompatible adapter.getBasePath() so it likely won't work on mobile

I don't see any reason why your plugin needs that. `await vault.cachedRead(file)` will get you the file contents and is mobile-compatible.
You also don't really need to read the contents of the file at all. If you're ok to stick with standards, you can find markdown checklists in our `metadataCache` that's the parsed cached contents of each markdown file. This should make your plugin a lot quicker and less resource-intensive.
Here's the reference:
https://github.com/obsidianmd/obsidian-api/blob/bde556afa033e909ebfb9fcee8f5ef288276f78f/obsidian.d.ts#L1381

https://github.com/NatiAris/obsidian-random-todo/blob/37e29e3d0905a38bd51c04650e1bfbe4dcd77444/main.ts#L16
Please rename your classes/interfaces from the sample `My...` to your actual plugin name.

https://github.com/NatiAris/obsidian-random-todo/blob/37e29e3d0905a38bd51c04650e1bfbe4dcd77444/main.ts#L3
Avoid NodeJS APIs to make your plugin mobile compatible. `Math.random()` to replace `randomInt`.

https://github.com/NatiAris/obsidian-random-todo/blob/37e29e3d0905a38bd51c04650e1bfbe4dcd77444/main.ts#L98
Should remove anything you don't need from the sample code.

NatiAris:
- Removed `adapter.getBasePath()`
- Renamed `My...` stuff
- Switched to `Math.random()`
- Removed unused code

I don't see how to avoid reading files, I'm trying to match an arbitrary regex on arbitrary lines of arbitrary files.  
As far as I understand, the metadata cache keeps things like tags, task lists, and front matter, not the whole lines.  
I'm entirely unsure, though.

lishid:
Yeah that's what I meant by this:

> If you're ok to stick with standards

If you're going to regex the file then yeah you would have to read them.

I'm suggesting to consider using the checklist status that Obsidian caches as indications for your todos. That way you can skip reading the files unless that file contains un-checked checklist items. Then you can use the checklist cache's location information to locate the block.

lishid:
Of course it all depends on whether you want to stick to markdown checklists or be more flexible at the cost of extra compute.

NatiAris:
I think I'm more on the extra compute side, my marks often end up in the middle of a line or huge text block, having dedicated checklists every time seems to be not quite what I'd want.  
And I think I've seen an existing community plugin for that anyway, something about aggregating checklists from all files, etc.

lishid:
Sounds good.

lishid:
A few ideas to improve performance:
- Reuse your regexp pattern instead of making a new one for every file.
- Instead of parsing all files for all TODOs, randomly pick files, parse & choose a random todo, or if there is no TODO, continue to randomly pick files (excluding the ones you've already picked). This will decrease your amortized cost while maintain the same worst-case cost.
- Consider partially cache some of the results of your work, by storing the `file.stat.mtime` and only re-computing if the mtime has changed.

NatiAris:
All great suggestions, will do 👍 

What would be the conventional way to measure the performance, btw?  
My vault is not large enough (yet) for me to notice any hangups.

lishid:
Use the developer tools to take a performance snapshot, and measure from start to finish for time in ms.

NatiAris:
- Now saving the regexp object
- For "random file" shuffling the file list and only reading files until I find one with todos
- For "random item" I still need to traverse all files to retain the uniform-per-item distribution
- Now keeping a file -> (mtime, positions) mapping to avoid parsing files that haven't changed

lishid:
Looks great!

https://github.com/NatiAris/obsidian-random-todo/blob/c9cd72684041d652fb76dee9a3e026c3fe370453/main.ts#L47
mtime can actually go back in time (for example, when Sync reverts a file to an older version); suggest using `!==`

https://github.com/NatiAris/obsidian-random-todo/blob/c9cd72684041d652fb76dee9a3e026c3fe370453/main.ts#L72
Consider using your cache here. This also seems to be slightly different (it'll match two TODOs on the same line, whereas your other match will only count each line at most once) than your other algorithm.

PR #391, opened 2021-07-31
Add multi-line-formatting

nmady:

## Repo URL

Link to my plugin: https://github.com/nmady/obsidian-multi-line-formatting

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
Please give your code a quick auto-format to normalize spaces vs tabs.

https://github.com/nmady/obsidian-multi-line-formatting/blob/6bdf7bb400833f0416ab0365cc9ffabb7f3b68a8/src/main.ts#L73
Don't do this. `activeLeaf` may be null and `activeLeaf.view` is not guaranteed to be a MarkdownView.

If you need to access & verify the active markdown view, use `workspace.getActiveViewOfType(MarkdownView)`.

Even better, [here](https://github.com/nmady/obsidian-multi-line-formatting/blob/6bdf7bb400833f0416ab0365cc9ffabb7f3b68a8/src/main.ts#L106) you should use an `editorCallback` instead of a `callback` in your command, which will give you the editor that's currently active directly, while also hiding your command when no editor is available.

https://github.com/nmady/obsidian-multi-line-formatting/blob/6bdf7bb400833f0416ab0365cc9ffabb7f3b68a8/src/main.ts#L86
Keep in mind that this is the previously cached version of this file, and any edits made in the last few seconds may not have been saved and cached yet.

nmady:
Thank you so much for your thoughtful review.

I have autoformatted the code using Prettier and implemented the use of `editorCallback` as suggested: thank you!

I'm not sure what to use as an alternative to the `SectionCache`. Do you know of a better alternative for determining the type of section or for obtaining an up-to-date `MetadataCache`?

lishid:
I don't have a better alternative, sorry. I think the only other way is to use some kind of heuristics on the editor's source string like checking for empty lines, etc, which might be error prone for scenarios where the user's string doesn't render exactly you think it should based on the heuristics.

nmady:
The `MetadataCache` is only used to identify codeblocks, and specifically to avoid changing them. I believe that problems will only arise if the `MetadataCache` is out of date and the user is applying this plugin to a selection that includes lines that have recently changed status between 'code' and another type of section. I haven't encountered this problem myself, in testing. I have added a section to the README stating that codeblocks are a known limitation so that users will be aware of the potential problems. I think this plugin is useful, even with this minor limitation.

lishid:
Sounds good, I agree it probably isn't a big deal and will work most of the time.

https://github.com/nmady/obsidian-multi-line-formatting/blob/2152f36bc2fdca9942be85ca031775b9613f94f6/src/main.ts#L104
Small nit: The `async` here isn't necessary.

PR #392, opened 2021-08-02
Add Phone to Roam to Obsidian plugin

dgarrett:

## Repo URL

Link to my plugin: https://github.com/dgarrett/phone-to-roam-to-obsidian

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/dgarrett/phone-to-roam-to-obsidian/blob/5bcda395aac9a80d6cf992b5a01060344817caec/main.ts#L32
Maybe make this interval a tad bit longer? IMO once every minute should be ok.

Also, given this is fully automatic, I don't really see the need for a command to force-fetch - perhaps if there's a config option that disables auto-fetching then the command would make sense.

dgarrett:
Thanks for the review! I just pushed 1.0.1 with your recommendations.

lishid:
Actually, another suggestion I have is to consider naming the plugin "Phone to Obsidian" but keep the description to explicitly mention "Phone to roam", to avoid confusing users that they need to use Roam as an intermediary.

PR #393, opened 2021-08-02
Add WordNet

TfTHacker:

## Repo URL

Link to my plugin: https://github.com/TfTHacker/Obsidian-WordNet

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - [ ] `styles.css` _(optional)_ - not needed by my plugin


lishid:
Consider auto-formatting your code to normalize whitespaces and help readability.

https://github.com/TfTHacker/Obsidian-WordNet/blob/e1248a7e59d390130b5cab351bc8675e5c5b7068/src/main.ts#L44
If your command always runs, you can just use a regular `callback` instead of a `checkCallback`.

https://github.com/TfTHacker/Obsidian-WordNet/blob/e1248a7e59d390130b5cab351bc8675e5c5b7068/src/suggester.ts#L29
You can pass 0 as the timeout here, keep a reference to the notice, and hide it when your fetch is done.

https://github.com/TfTHacker/Obsidian-WordNet/blob/e1248a7e59d390130b5cab351bc8675e5c5b7068/src/suggester.ts#L30
Should consistently use `await` here instead of `.then`. Make sure to include the `catch` mechanics.

https://github.com/TfTHacker/Obsidian-WordNet/blob/e1248a7e59d390130b5cab351bc8675e5c5b7068/src/suggester.ts#L25
Consider putting `vault` into a variable so you aren't repeating `this.plugin.app.vault` (and maybe adapter as well).

https://github.com/TfTHacker/Obsidian-WordNet/blob/e1248a7e59d390130b5cab351bc8675e5c5b7068/src/suggester.ts#L80
Here you probably want to use `workspace.getActiveViewOfType(MarkdownView) and test for null return. `activeLeaf` can be null and `getState().mode` isn't guaranteed to only exist on markdown views.

https://github.com/TfTHacker/Obsidian-WordNet/blob/e1248a7e59d390130b5cab351bc8675e5c5b7068/src/suggester.ts#L107
This is not necessary, you can use the editor directly.

lishid:
Just a few more things (also given the changes already performed):

https://github.com/TfTHacker/Obsidian-WordNet/blob/ec22a683b3d18ee8d28bf997d1ed3b8f669168f0/src/suggester.ts#L19
This works too, and I wasn't very clear previously. I usually just store it as a variable at the beginning of a function. The advantage doing it that way is that you don't have a lingering reference that isn't really needed.

https://github.com/TfTHacker/Obsidian-WordNet/blob/ec22a683b3d18ee8d28bf997d1ed3b8f669168f0/src/suggester.ts#L36
Consider wrapping this line and `await response.json()` in a try-catch block.

https://github.com/TfTHacker/Obsidian-WordNet/blob/ec22a683b3d18ee8d28bf997d1ed3b8f669168f0/src/suggester.ts#L91
A better way here is to use `currentView.getMode()` instead of `.getState().mode` which causes unnecessary serialization of other state variables.

https://github.com/TfTHacker/Obsidian-WordNet/blob/ec22a683b3d18ee8d28bf997d1ed3b8f669168f0/src/suggester.ts#L110
Not sure how I missed last round but that's definitely a DOM injection vulnerability - I'd rewrite it using our convenience DOM creation utilities like this:
```ts
el.createEl('b', {text: item.item.Term});
el.createEl('br');
el.appendText(item.item.Definition);
```

https://github.com/TfTHacker/Obsidian-WordNet/blob/ec22a683b3d18ee8d28bf997d1ed3b8f669168f0/src/suggester.ts#L114
Same as before here, should use getActiveViewOfType(MarkdownView), then check for `getMode()` instead of `.getViewState()`

TfTHacker:
Thank  you for all these suggestions.  I have now implemented them and  performed testing. As far as I can see there are no  pending follow-up items.

lishid:
https://github.com/TfTHacker/Obsidian-WordNet/blob/ff7ad6fdc78118917d0fcc913f64b0addbe41ae7/src/suggester.ts#L46
Rather unconventional use of disabling your plugin. It's probably fine in theory, although I have not tested/intended nor seen any plugin attempt to this. Let me know if you run into issues doing this.

https://github.com/TfTHacker/Obsidian-WordNet/blob/ff7ad6fdc78118917d0fcc913f64b0addbe41ae7/src/suggester.ts#L97
I think you meant this to be `currentView.getMode() !== 'source'`?

https://github.com/TfTHacker/Obsidian-WordNet/blob/ff7ad6fdc78118917d0fcc913f64b0addbe41ae7/src/suggester.ts#L125
Redundant get of the same value from 2 lines above. Also variable naming convention not followed. 😉

---

Additionally, there's a couple of places where you can use `this.app` instead of `this.plugin.app` (since it's on the base class).

I'd also suggest getting into the habit of using `===` and `!==` instead of `==` and `!=` since they mean different things in JavaScript. The triple version is what you intend for most of the time normally.

lishid:
I think it's fair to have this merged since none of the feedback here is blocking.

TfTHacker:
Thank you for the feedback. Will incorporate these into the next update.

PR #396, opened 2021-08-03
Add "Apply Patterns" plugin to community plugins list.

publicus:

## Repo URL

Link to my plugin: https://github.com/publicus/obsidian-apply-patterns-plugin

## Release Checklist

  - I have tested this on Linux, which is the only system I have available.
  - [X] `main.js`
  - [X] `manifest.json`


lishid:
https://github.com/publicus/obsidian-apply-patterns-plugin/blob/4af8de9e0a7f76e3ecc6f817f6d9e7660aafa81d/src/main.ts#L7
Might need to rename the class 🙈 

https://github.com/publicus/obsidian-apply-patterns-plugin/blob/4af8de9e0a7f76e3ecc6f817f6d9e7660aafa81d/src/Commands/index.ts#L4
I don't really get the point of this class. IMO it would make more sense to move all the code here directly into the plugin's `onload` function.

https://github.com/publicus/obsidian-apply-patterns-plugin/blob/4af8de9e0a7f76e3ecc6f817f6d9e7660aafa81d/src/Commands/ApplyPattern.ts#L17
I don't think you need this test because `editorCallback` always happens in a MarkdownView.

https://github.com/publicus/obsidian-apply-patterns-plugin/blob/4af8de9e0a7f76e3ecc6f817f6d9e7660aafa81d/src/Commands/ApplyPattern.ts#L111
Consider storing all your changes and selection, and use `editor.transaction` instead, which would let the user undo in a single step rather than in multiple steps.

https://github.com/publicus/obsidian-apply-patterns-plugin/releases/tag/1.1.0
Your release seems rather large for the size of your code - consider stripping the sourcemaps so users can save some RAM and bandwidthh.

publicus:
@lishid, thank you for your review! I really appreciate your time and thoughtful feedback, and the opportunity to get know the API (and TypeScript) better.

I've incorporated your feedback in [this branch](https://github.com/publicus/obsidian-apply-patterns-plugin/compare/incorporate-feedback). I have a few questions below, if you have time and bandwidth:

> https://github.com/publicus/obsidian-apply-patterns-plugin/blob/4af8de9e0a7f76e3ecc6f817f6d9e7660aafa81d/src/main.ts#L7
> Might need to rename the class 

Lol, oops. Thank you : P I appreciated how the Tasks repo was set up, and so forked it to write this plugin (I originally had started writing what I thought would be a PR for Tasks, but then it grew to be big enough to be its own plugin). I missed changing this reference.

> https://github.com/publicus/obsidian-apply-patterns-plugin/blob/4af8de9e0a7f76e3ecc6f817f6d9e7660aafa81d/src/Commands/index.ts#L4
> I don't really get the point of this class. IMO it would make more sense to move all the code here directly into the plugin's `onload` function.

Looking back through, I agree. I've refactored such that the `Commands` file is removed, and the commands are defined in the plugin's `onload` function, following your suggestion.

> https://github.com/publicus/obsidian-apply-patterns-plugin/blob/4af8de9e0a7f76e3ecc6f817f6d9e7660aafa81d/src/Commands/ApplyPattern.ts#L17
> I don't think you need this test because `editorCallback` always happens in a MarkdownView.

Noted; thank you! I've removed the test, following your explanation.

> https://github.com/publicus/obsidian-apply-patterns-plugin/blob/4af8de9e0a7f76e3ecc6f817f6d9e7660aafa81d/src/Commands/ApplyPattern.ts#L111
> Consider storing all your changes and selection, and use `editor.transaction` instead, which would let the user undo in a single step rather than in multiple steps.

Thank you! I hadn't seen `editor.transaction`, and had been wondering how to best accomplish that. In this case, both commands currently just make a single change / undo step, but I've refactored following your advice to facilitate more complex collections of steps in the future.

> https://github.com/publicus/obsidian-apply-patterns-plugin/releases/tag/1.1.0
> Your release seems rather large for the size of your code - consider stripping the sourcemaps so users can save some RAM and bandwidthh.

Thank you for pointing this out! Setting `output.sourcemap` to `hidden` in `rollup.config.js` reduces the file size of `main.js` to about a quarter of its previous size (275KB vs. 1020KB)(!). Is that more in line with what you consider typical?

publicus:
I do have a question (and am happy to ask it elsewhere), if you have time and bandwidth: I just noticed that when the `FuzzySuggestModal` closes on selection of an item, the plugin makes the expected changes in the editor, but I then *can't type new text into the editor until I switch away and then back to the window, or open a new modal (such as the Command Palette).* Backspace, delete, enter, hotkeys (e.g., Ctrl + Z), and square brackets all work, but no other keys on the keyboard (e.g., [A-Za-z0-9] will produce new input, until I switch away from and back to the window or open a new modal.

Does this raise any immediate suspected causes in your mind? Is there some step to close out a FuzzySuggestModal that you've seen others forget or not understand they had to do? I'd be grateful for any thoughts you might have!

publicus:
(To follow-up from above, this behavior still occurs when I comment out [this line](https://github.com/publicus/obsidian-apply-patterns-plugin/blob/incorporate-feedback/src/PatternsModal.ts#L17), leading me to think that it's not about the plugin's logic, but rather something to do with how I'm handling the Modal.)

lishid:
> Thank you for pointing this out! Setting output.sourcemap to hidden in rollup.config.js reduces the file size of main.js to about a quarter of its previous size (275KB vs. 1020KB)(!). Is that more in line with what you consider typical?

Yeah that's better - Many plugins are tiny (~20kb) because they don't import external modules, but you are using dayjs and chrono-node so it's expected to take larger.

> (To follow-up from above, this behavior still occurs when I comment out this line, leading me to think that it's not about the plugin's logic, but rather something to do with how I'm handling the Modal.)

Have you tried not calling `this.close()`? I believe it's automatically handled when an item is chosen for the selection to be restored. Calling close twice may be disrupting the selection process.

publicus:
Thank you!

> Have you tried not calling `this.close()`? I believe it's automatically handled when an item is chosen for the selection to be restored. Calling close twice may be disrupting the selection process.

Removing `this.close()` fixed it! Thank you!! I'd stared at that for a while.

I've incremented the release with the feedback from above, [here](https://github.com/publicus/obsidian-apply-patterns-plugin/releases/tag/1.1.1).

Again, thank you for your review. I appreciate your feedback, and all of the work you're doing on such a neat product.

PR #400, opened 2021-08-05
Add Longform

kevboh:

## Repo URL

Link to my plugin: https://github.com/kevboh/longform

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/kevboh/longform/blob/e1aafd1e81b7b35def0e273417209e1979082f4f/src/main.ts#L46
Don't recommend keeping references to your custom views - you can get them when needed using `workspace.getLeavesOfType`. This is because there are cases where you're end up referring to an older detached copy of your view, which would then cause weird behaviors.

https://github.com/kevboh/longform/blob/e1aafd1e81b7b35def0e273417209e1979082f4f/src/main.ts#L67
You can test for file vs folder using `file instanceof TFile` or `file instanceof TFolder`.

https://github.com/kevboh/longform/blob/e1aafd1e81b7b35def0e273417209e1979082f4f/src/main.ts#L106
This can be replaced with `workspace.onLayoutReady(callback);`

https://github.com/kevboh/longform/blob/e1aafd1e81b7b35def0e273417209e1979082f4f/src/main.ts#L273
context is not required (it's used to set the `this` for the function call, which you're binding to something else anyway).

https://github.com/kevboh/longform/blob/e1aafd1e81b7b35def0e273417209e1979082f4f/src/model/index-file.ts#L1
Consider using Obsidian's `stringifyYaml` to avoid porting another copy of an YAML library.

kevboh:
@lishid thanks for the look-through! See [this release](https://github.com/kevboh/longform/commit/bc02f46e55acde824130ce3188cd3a19ef765f78) for those changes. Bumped up min app version to get `stringifyYaml`, happy to not ship extra deps.

Moonbase59:
@lishid’s checks & comments alone are worth a bookmark—thanks from here, too.

PR #401, opened 2021-08-05
Add Hide Sidebars When Narrow

mybuddymichael:

## Repo URL

Link to my plugin: https://github.com/mybuddymichael/obsidian-hide-sidebars-when-narrow

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


mybuddymichael:
I have tested this on macOS, but not Windows or Linux.

lishid:
https://github.com/mybuddymichael/obsidian-hide-sidebars-when-narrow/blob/166d3b44d4a45bb1aaa4cb3914550cfc55aea609/main.ts#L18
I don't think you need to wrap everything in the layout ready - at least you can load settings in your onload directly, with an await, as well as registering the settings tab.

https://github.com/mybuddymichael/obsidian-hide-sidebars-when-narrow/blob/166d3b44d4a45bb1aaa4cb3914550cfc55aea609/main.ts#L25
This timeout is probably not what you want to do  - I think you're looking for the `debounce` function that you can import from `obsidian`.

https://github.com/mybuddymichael/obsidian-hide-sidebars-when-narrow/blob/166d3b44d4a45bb1aaa4cb3914550cfc55aea609/main.ts#L32
I'd recommend using `innerWidth` since `outerWidth` will include the window border as well as things like the debugger window.

mybuddymichael:
Thanks for the review. I believe I've addressed all of these.

mybuddymichael:
I don't have access to a Windows or Linux machine with Obsidian. Do I need to find folks who have those who can help me test before I check that off?

PR #404, opened 2021-08-07
Add Obsidian Dev Tools Plugin (view console on Mobile)

KjellConnelly:
I am submitting a new plugin: `Obsidian Dev Tools Plugin` (currently only **1**: displays a console, **2**: log and eval to console, and **3** display available obsidian icons. Meant for plugin development, often filling the gap for mobile devices)

## Repo URL
https://github.com/KjellConnelly/obsidian-dev-tools

## Release Checklist
- [x] I have tested this on macOS and iOS
    - [x] `main.js`
    - [x] `manifest.json`
    - [x] `styles.css` (optional)
- [x] Github release name matches the exact version number specified in my manifest.json (**Note**: Use the exact version number, don't include a prefix v)
- [x] The `id` in my `manifest.json` matches the id in the `community-plugins.json` file.
- [x] `README` clearly describes the plugin's purpose and provides clear usage instructions.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1917,5 +1917,12 @@
         "author": "OwnJoke",
         "repo": "OwnJoke/obsidian-budget-wysiwyg",
         "branch": "main"
+    },
+    {
+    	"id": "obsidian-dev-tools",
+    	"name": "Obsidian Dev Tools Plugin",
+    	"description": "This plugin is initially used to view the console on mobile devices, though it may have other plugin tools added in the future.",
+    	"author": "Kjell Connelly",
+    	"repo": "KjellConnelly/obsidian-dev-tools"
     }
 ]


Vinzent03:
Sounds interesting, but you removed the pr template. At least a link to the repo would be great.

KjellConnelly:
@Vinzent03 Sorry about that. Didn't realize that was needed. Just updated my original post.

lishid:
Couple of things first:
- Please clean up & remove everything from the sample plugin that you aren't using, like settings, override functions, etc.
- Please auto-format your code to normalize whitespaces.

https://github.com/KjellConnelly/obsidian-dev-tools/blob/a921d6f1ae1edb4597440a8900baa51f181bff32/main.ts#L40
Please choose either a callback or a check callback - doesn't seem like you need a checkCallback here.

https://github.com/KjellConnelly/obsidian-dev-tools/blob/a921d6f1ae1edb4597440a8900baa51f181bff32/main.ts#L52
This imported script should not be ran in a function like this. Would it make more sense to be at the top of the file, right after the obsidian imports?

Actually, since these are so long, please move them each to their own files to avoid clogging up the main file.
`import 'filename';` should do the trick.

KjellConnelly:
Alright I cleaned up the code and added it to a separate class, removed unnecessary imports and functions. Also improved toggle capabilities and added it as a `boolean` under `saved settings so it would stay toggled on/off between app reloads.

KjellConnelly:
Just added a new feature while we wait - ability to view all icons. I found it difficult choosing icons in certain circumstances (ribbons and mobile command icons). I had no way of viewing them, and had to google a list of them. So I added them to `Settings` and allow users to click a button to copy the icon string to clipboard.

lishid:
Sorry for the late reply.

-  Your code still has mixed spaces & tabs. Please run it through an auto formatter so it's easier to read.

https://github.com/KjellConnelly/obsidian-dev-tools/blob/131d718691ecfcd940115b8d0a8692373c04d63f/src/plugin/Classes/Commands.ts#L15
You aren't using anything from the leaf, so I think you should just register a regular `callback` instead of a `checkCallback`. Honestly, I feel like `CommandsHelper` is pretty unnecessary.

https://github.com/KjellConnelly/obsidian-dev-tools/blob/131d718691ecfcd940115b8d0a8692373c04d63f/src/lib/ObsidianDevLibrary.ts#L29
Highly suggest avoiding the use of innerHTML as it introduces a chance for DOM injection attacks. I don't think this function is really used though? I've traced around and couldn't find a real use for it.

ericaxu:

PR #410, opened 2021-08-08
Add Emoji Shortcodes Plugin

phibr0:

## Repo URL

Link to my plugin: https://github.com/phibr0/obsidian-emoji-shortcodes

## Release Checklist

- [x] I have tested this on Windows, and Linux _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/phibr0/obsidian-emoji-shortcodes/blob/069d035be72d7f8313ad2f134b108c9b8f922cbe/src/main.ts#L6
I know it's a pain in the butt, but please don't use innerHTML since it's vulnerable to DOM injection attacks - instead, traverse through the DOM tree and scan for these in all the text nodes, and replace when needed.

https://github.com/phibr0/obsidian-emoji-shortcodes/blob/master/src/emojis.ts
Maybe add a comment to explain how this was generated/obtained so it's easier to get updated when more emojis are available.

phibr0:
> https://github.com/phibr0/obsidian-emoji-shortcodes/blob/069d035be72d7f8313ad2f134b108c9b8f922cbe/src/main.ts#L6
I know it's a pain in the butt, but please don't use innerHTML since it's vulnerable to DOM injection attacks - instead, traverse through the DOM tree and scan for these in all the text nodes, and replace when needed.

Okay done. I thought its fine, because I don't modify it with anything suspicious. I usually sanitize everything from external sources.

```ts
emojiReplace(e: keyof typeof emoji, el: HTMLElement | ChildNode){
	if (el.hasChildNodes()){
		el.childNodes.forEach((child: ChildNode) => this.emojiReplace(e, child));
	} else {
		el.textContent = el.textContent.replace(e, emoji[e] ?? e);
	}
}
```

I am now using this recursive function to loop over every child and if the element doesn't have any children the shortcodes in the textContent property will get replaced.

> https://github.com/phibr0/obsidian-emoji-shortcodes/blob/master/src/emojis.ts
Maybe add a comment to explain how this was generated/obtained so it's easier to get updated when more emojis are available.

Done!

phibr0:
I just added new functionality to immediately replace the shortcode with the unicode symbol, might want to review again: https://github.com/phibr0/obsidian-emoji-shortcodes/blob/master/src/immediateReplace.ts

lishid:
https://github.com/phibr0/obsidian-emoji-shortcodes/blob/4093992ac938ab38f88fe23e3847808bd67b7895/src/immediateReplace.ts#L7

This event is never un-registered when your plugin is disabled. Should probably do that...

phibr0:
> This event is never un-registered when your plugin is disabled. Should probably do that...

Fixed

PR #413, opened 2021-08-09
Add Readwise Official plugin

tadeoos:

## Repo URL

Link to my plugin: https://github.com/readwiseio/obsidian-readwise/

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1955,5 +1955,12 @@
         "description": "A plugin for Obsidian that helps you write and edit novels, screenplays, and other long projects.",
         "repo": "kevboh/longform",
         "branch": "main"
+    },
+    {
+        "id": "readwise-official",
+        "name": "Readwise Official",
+        "author": "Readwise",
+        "description": "Official Readwise <-> Obsidian integration.",
+        "repo": "readwiseio/obsidian-readwise"
     }
 ]


lishid:
- Consider auto-formatting to normalize semicolon usage (I'd recommend using them as opposed to stripping them).

https://github.com/readwiseio/obsidian-readwise/blob/29ffcdaaa123c979f42f9e9684eecd40bd8418a1/main.ts#L5
I don't think this is needed for anything.

https://github.com/readwiseio/obsidian-readwise/blob/29ffcdaaa123c979f42f9e9684eecd40bd8418a1/main.ts#L7
https://github.com/readwiseio/obsidian-readwise/blob/29ffcdaaa123c979f42f9e9684eecd40bd8418a1/main.ts#L238
I would suggest trying to get rid of this dependency and replacing it with a custom written one, that way the plugin will also be functional on mobile devices.

https://github.com/readwiseio/obsidian-readwise/blob/29ffcdaaa123c979f42f9e9684eecd40bd8418a1/main.ts#L10
I would suggest changing this to `process?.env?.READWISE_SERVER_URL` to make sure it doesn't error out in case `process` or `env` are undefined.

https://github.com/readwiseio/obsidian-readwise/blob/29ffcdaaa123c979f42f9e9684eecd40bd8418a1/main.ts#L12
I think `extends JSON` is probably not intended here?

https://github.com/readwiseio/obsidian-readwise/blob/29ffcdaaa123c979f42f9e9684eecd40bd8418a1/main.ts#L255
Recommend using `normalizePath` from `obsidian` to normalize paths to obsidian's vault/adapter standards - using `/` as path separators and trimming any leading/trailing `/` characters. This also normalizes unicode representations.

https://github.com/readwiseio/obsidian-readwise/blob/29ffcdaaa123c979f42f9e9684eecd40bd8418a1/main.ts#L284
This doesn't seem to do anything - you need a handle to to setInterval passed in here to clear it.

https://github.com/readwiseio/obsidian-readwise/blob/29ffcdaaa123c979f42f9e9684eecd40bd8418a1/main.ts#L192
There's a `.empty()` function that does this.

https://github.com/readwiseio/obsidian-readwise/blob/29ffcdaaa123c979f42f9e9684eecd40bd8418a1/main.ts#L396
I'd suggest using DOM manipulations instead of extracting and re-parsing the HTML. Something like this should be fine:
https://stackoverflow.com/a/20910214

lishid:
Oh and consider removing the inline sourcemap in the distribution build to reduce it from 1.3mb down to 300kb.

tadeoos:
@lishid thanks a lot! Looking into this now

tadeoos:
@lishid I updated our code. Didn't go with 2 of your suggestions:
- Changing to `process?.env?.READWISE_SERVER_URL` actually broke my dev setup... We're using this plugin: https://github.com/Andarist/rollup-plugin-dotenv which works by replacing the env vars, so the `process.env` will actually never be referenced in the output file
- the last one: I've decided to keep the reparsing of the HTML as this won't be much overhead and I prefer a brute simplicity of the current implementation.

Thanks for the feedback!

Regarding mobile: is there any guide on developing/testing mobile versions of plugins?

lishid:
> Changing to process?.env?.READWISE_SERVER_URL actually broke my dev setup... We're using this plugin: https://github.com/Andarist/rollup-plugin-dotenv which works by replacing the env vars, so the process.env will actually never be referenced in the output file

Oh ok so that's replaced in the build step and not outputted! That's perfectly fine, I'd recommend adding a comment to explain how it gets replaced at the build step.

> the last one: I've decided to keep the reparsing of the HTML as this won't be much overhead and I prefer a brute simplicity of the current implementation.

I agree it's simple and elegant. We normally scan and warn for `innerHTML` mostly due to DOM injection-related security concerns. I think in theory this should be fine as serializing from innerHTML should sanitize everything, but it's just a pattern we recommend against using.

The other concern here is breaking any registered DOM events that other plugins (or core) may have placed on the child elements directly, rather than using event delegation.

> Regarding mobile: is there any guide on developing/testing mobile versions of plugins?

I don't think there's anything official out there, but there is a utility in the desktop for emulating the mobile interface (but it does not test the mobile file system): type in console `app.emulateMobile(true)` (use false to revert), then using the element inspector you can restrict the screen size to a mobile device.

You can also test on Android with a USB/adb based debugger like Chrome. Unfortunately such is not possible on iOS due to Apple's app signing policy (Signed apps can't be debugged).

tadeoos:
Ah ok @lishid, didn't get the interaction/security concerns at first! I changed it now and added a comment for the `process.env...` line. Updated the tag and the files in a release package as well. It seems we may be good to go...? :)

lishid:
LGTM, will merge asap.

> the process.env variable will be replaced by it's target

I think you meant "its" 🤣 

> `strongEl.childNodes[0]`

I would even suggest using `strongEl.firstChild` for both the check and the move like this: https://stackoverflow.com/questions/20910147/how-to-move-all-html-element-children-to-another-parent-using-javascript/20910214#comment57130119_20910214

tadeoos:
will update! thanks a ton @lishid!

PR #418, opened 2021-08-13
Add Luhman plugin

Dyldog:

## Repo URL

Link to my plugin: https://github.com/Dyldog/luhman-obsidian-plugin

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1,5 +1,12 @@
 [
     {
+        "id": "luhman",
+        "name": "Luhman",
+        "author": "Dylan Elliott",
+        "description": "Commands for handling a zettelkasten with Luhman-style IDs as filenames",
+        "repo": "Dyldog/luhman-obsidian-plugin"
+    },
+    {
         "id": "nldates-obsidian",
         "name": "Natural Language Dates",
         "author": "Argentina Ortega Sainz",
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1,12 +1,5 @@
 [
     {
-        "id": "luhman",
-        "name": "Luhman",
-        "author": "Dylan Elliott",
-        "description": "Commands for handling a zettelkasten with Luhman-style IDs as filenames",
-        "repo": "Dyldog/luhman-obsidian-plugin"
-    },
-    {
         "id": "nldates-obsidian",
         "name": "Natural Language Dates",
         "author": "Argentina Ortega Sainz",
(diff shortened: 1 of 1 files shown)


Dyldog:
Hey, sorry, I just realised I hadn't added the plugin files to my release. Have done this now!

lishid:
- Should remove the `main.js` file from your repo and add it to your gitignore - the compiled file should stay in your releases :)
- I highly suggest you taking some time to learn & use the `async` and `await` syntax, instead of using `.then()`. That will make your code much cleaner and easier to write.
- Highly recommend auto-formatting your code to add semicolons to the end of statements. While JS allows both with and without, there are subtle cases where it will do things slightly differently, and it's always good to include them in your source code.

https://github.com/Dyldog/luhman-obsidian-plugin/blob/master/src/obsidian-ex.d.ts
I don't think this is used for anything.

https://github.com/Dyldog/luhman-obsidian-plugin/blob/3af37955640f22a698701c94bbc6c33ea8d6f183/src/main.ts#L38
Use `navigator.clipboard.writeText(string)` instead, which is supported on mobile as well.

https://github.com/Dyldog/luhman-obsidian-plugin/blob/3af37955640f22a698701c94bbc6c33ea8d6f183/src/main.ts#L125
Don't use the active leaf directly, instead use `workspace.getLeaf()` which is guaranteed to give you a non-pinned leaf to use.

https://github.com/Dyldog/luhman-obsidian-plugin/blob/3af37955640f22a698701c94bbc6c33ea8d6f183/src/main.ts#L126
Just test for null here and return instead of stringing everything else with `?`.

https://github.com/Dyldog/luhman-obsidian-plugin/blob/3af37955640f22a698701c94bbc6c33ea8d6f183/src/main.ts#L147
Same here.

https://github.com/Dyldog/luhman-obsidian-plugin/blob/3af37955640f22a698701c94bbc6c33ea8d6f183/src/main.ts#L173
There's a better way to rename & update links through Obsidian's API in `FileManager.renameFile`.

https://github.com/Dyldog/luhman-obsidian-plugin/blob/3af37955640f22a698701c94bbc6c33ea8d6f183/src/main.ts#L302
Just `getLeaf()` should do here.

https://github.com/Dyldog/luhman-obsidian-plugin/blob/3af37955640f22a698701c94bbc6c33ea8d6f183/src/main.ts#L377
The css class can be added as `{cls: "class"}` in the second param object.

https://github.com/Dyldog/luhman-obsidian-plugin/blob/3af37955640f22a698701c94bbc6c33ea8d6f183/src/main.ts#L400
Instead of getting it like this, just assign the element to a private class member of your modal when you create it.

Dyldog:
Hey @lishid, thanks for all the comments! :D I've followed the directions in all of them and the code is looking much better

PR #419, opened 2021-08-14
Added Obsidian Dotenv to community-plugins.json

metruzanca:
I am submitting a new plugin: `Obsidian Dotenv` which allows setting custom variables on `app.env` (via plugin settings) to facilitate configuring global constants usable by other plugins e.g. templater or dataview. This should also facilitate distributing templates and vaults with custom & easily editable configurations. 

## Repo URL

Link to my plugin: https://github.com/metruzanca/obsidian-dotenv

## Release Checklist

(Don't own a mac :/ should all work as the plugin is just javascript.)

  - [x] `main.js`
  - [x] `manifest.json`

---

This is my working MVP version. Tomorrow/this weekend I'll be fleshing out this plugin further and hopefully presenting it to the rest of the community

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1997,5 +1997,12 @@
         "author": "phibr0",
         "description": "Group multiple Commands into one Macro and set delays.",
         "repo": "phibr0/obsidian-macros"
+    },
+    {
+        "id": "obsidian-dotenv",
+        "name": "Dotenv",
+        "author": "metruzanca",
+        "description": "Easily manage custom variables accessible by other plugins e.g. templater (and it's templates)",
+        "repo": "metruzanca/obsidian-dotenv"
     }
 ]


lishid:
- Please auto-format your code to normalize semicolons - I'd highly suggest adding them instead of removing them since there are places where it makes semantic differences.

https://github.com/metruzanca/obsidian-dotenv/blob/45a9ec7d8d9d012b28db7a86e7c62dcbe8384d81/src/main.ts#L18
Might want to rename your class.

https://github.com/metruzanca/obsidian-dotenv/blob/45a9ec7d8d9d012b28db7a86e7c62dcbe8384d81/src/main.ts#L24
It does not exist on DataAdapter - only when on the desktop app when DataAdapter is an instance of FileSystemAdapter where it has a `getBasePath` function, but you shouldn't rely on that.

https://github.com/metruzanca/obsidian-dotenv/blob/45a9ec7d8d9d012b28db7a86e7c62dcbe8384d81/src/main.ts#L32
Might need some error checking in case `env[group]` doesn't exist.

https://github.com/metruzanca/obsidian-dotenv/blob/45a9ec7d8d9d012b28db7a86e7c62dcbe8384d81/src/main.ts#L65
I know this is useful for debugging, but I'd suggest avoiding this in the production build.

ericaxu:

PR #420, opened 2021-08-14
Add reminder plugin

uphy:

## Repo URL

Link to my plugin: https://github.com/uphy/obsidian-reminder

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/uphy/obsidian-reminder/blob/3b134488288b4f530141dda1c38b0f50c7fa8226/src/main.ts#L27
Don't keep a reference to views in your plugin - this could end up detached/removed and you'll reference a bad view.
Instead, when you need to, use `workspace.getLeavesOfType` to get all existing attached views of your type.

https://github.com/uphy/obsidian-reminder/blob/3b134488288b4f530141dda1c38b0f50c7fa8226/src/main.ts#L121
You'll need to wrap all of these with `this.registerEvent()` so that they are unhooked when your plugin unloads.

https://github.com/uphy/obsidian-reminder/blob/3b134488288b4f530141dda1c38b0f50c7fa8226/src/controller.ts#L25
Use await here instead of `.then()`?

https://github.com/uphy/obsidian-reminder/blob/3b134488288b4f530141dda1c38b0f50c7fa8226/src/controller.ts#L74
You can use `Vault.getMarkdownFiles()` instead.

https://github.com/uphy/obsidian-reminder/blob/3b134488288b4f530141dda1c38b0f50c7fa8226/src/controller.ts#L55
Remember to skip non-markdown files - otherwise you'll be attempting to read all the images/pdfs/etc.

https://github.com/uphy/obsidian-reminder/blob/3b134488288b4f530141dda1c38b0f50c7fa8226/src/controller.ts#L59
Use `cachedRead` here to perform better. Later on, you should use `read` when you write the contents of the file back to disk.

uphy:
@lishid Thank you for the comments!  I have fixed the issues.  Can you check them?
https://github.com/uphy/obsidian-reminder/pull/2

PR #430, opened 2021-08-21
Add daily-named-folder plugin

NemoAndrea:
Add new 'daily-named-folder' plugin

## Repo URL 

Link to my plugin: [https://github.com/NemoAndrea/obsidian-daily-named-folder](https://github.com/NemoAndrea/obsidian-daily-named-folder)

## Release Checklist

-  I have tested this on 
  - [x] Windows
  - [ ]  macOS
  - [ ]  Linux 
  - [x] `main.js`
  - [x] `manifest.json`


pjeby:
So, if I'm understanding the README correctly, it looks like the distinguishing feature for this plugin is the ability to have arbitrary extra text in the folder and note name after the date?  If so, this should probably be called out at the very top of the README, rather than buried further down, as it would help potential users understand why they need this plugin rather than a different one.

Creating daily notes as folders is actually well within Obsidian's built-in capabilities: all you have to do is add more slashes in the date format, and Obsidian will create the folders on the fly.  (There were some hiccups with that in the past, but that's no longer the case.)  The Periodic Notes and Calendar plugins are also quite compatible with daily folders, again as long as the filenames follow the same pattern for each day and the actual note file name contains a full date.

So, assuming I'm understanding the situation correctly, if all somebody wants to do is have daily folders, they don't actually need to use this plugin.  And if they want to use such folders in conjunction with Calendar, the core Daily Notes plugin, or the Periodic Notes plugin, it sounds like using this plugin might actually be counterproductive, since none of the others support varying filenames.  Clarifying this choice at the very start of the README would likely help potential users understand the tradeoffs of choosing varying vs fixed filenames, as distinct from the tradeoffs of folders vs. files.

NemoAndrea:
Those are good points. One of the problems with the 'native' folder nesting is that (as far as I can tell) the 'next/previous' folder navigation (with hotkeys) also breaks. I guess this might be something that is in the works to get fixed.

Once that is fixed, the naming (and renaming) is indeed the only difference from the core plugin. I should remove the option to 'skip naming', as that is then already covered by the core plugin. 

I will put a warning message at the top of the README in the next version.

pjeby:
Ah, you're right, I forgot that that issue is not fixed in the core Daily Notes plugin.  (To fix it, `iterateDailyNotes()` needs to use `this.getFormat().split("/").pop()` as the basis for parsing note file names (instead of `this.getFormat()`), so that it's only processing the filename part.)

Anyway, I got distracted by the fact that Periodic Notes came out around the same time as I was having that issue, and I switched to Calendar and history as my main way of navigating to different days, so I forgot all about the prev/next; I'm only using the "Open today's daily note" command.

lishid:
Did you get a chance to go through the last few things in this comment? https://github.com/obsidianmd/obsidian-releases/pull/430#issuecomment-907279525

NemoAndrea:
> Did you get a chance to go through the last few things in this comment? [#430 (comment)](https://github.com/obsidianmd/obsidian-releases/pull/430#issuecomment-907279525)

Yes, I have addressed your good suggestions in v`0.1.2`. I wonder if I should rename the plugin to `daily-named-folder`. Would be a bit of virtual paperwork, but maybe clearer?

lishid:
Yeah that might make sense - it wasn't obvious to me at first either how this plugin was different from Obsidian's core plugin.

NemoAndrea:
Okay, I think now everything should be in order.

lishid:
Ok cool, last thing:

https://github.com/NemoAndrea/obsidian-daily-named-folder/blob/f56e9aac2569829985b66a8eafa1a1ce3382d5a2/main.ts#L80
I think you shouldn't rely on a TypeError here for error checking - rather, you should test for whether `getAbstractFileByPath` returned an object and if it's `instanceof TFile` before attempting to read it. I'd also recommend using `vault.create()` rather than `vault.adapter.write()` [here](https://github.com/NemoAndrea/obsidian-daily-named-folder/blob/f56e9aac2569829985b66a8eafa1a1ce3382d5a2/main.ts#L61).

Then I think you should also read the template and write to the file once, instead of creating an empty file first (which causes other plugins to react to this empty file), then filling it with data.

https://github.com/NemoAndrea/obsidian-daily-named-folder/blob/f56e9aac2569829985b66a8eafa1a1ce3382d5a2/main.ts#L130

NemoAndrea:
Okay, I think your lasts suggestions are included in `0.1.4`, thanks for the tips again! 

Somehow, using app.vault.create also seems to have fixed the issue with `executeCommandById('file-explorer:reveal-active-file')` not working directly after file creation.

PR #437, opened 2021-08-23
add obsidian-card-view-mode

yo-goto:

## Repo URL

Link to my plugin: https://github.com/yo-goto/obsidian-card-view-mode

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


ericaxu:
Thank you for the pull request!

Could you go through the check list and check them as applicable? Thank you!

yo-goto:
I checked items👍

Quorafind:
@yo-goto seems like you didn't open issue, could i ask why?

yo-goto:
oh, sorry for that. I didn't noticed. I'll open issue now.

lishid:
https://github.com/yo-goto/obsidian-card-view-mode/blob/master/main.js
Recommend removing this file from the repo and adding it to .gitignore - it should stay in the releases only and not be committed with the source code.

https://github.com/yo-goto/obsidian-card-view-mode/blob/6e829fc19586da46b858bc898d2116dfeec4446c/src/main.ts#L26
Should use `workspace.onLayoutReady(callback)` instead, which does exactly this.

https://github.com/yo-goto/obsidian-card-view-mode/blob/6e829fc19586da46b858bc898d2116dfeec4446c/src/main.ts#L63
Remember to remove this element from DOM when your plugin unloads - you can keep a reference to it in your plugin's object when creating it to avoid having to give it a unique ID and query from DOM later on. You can also use `document.head.createEl('style')` as an easier way to create and append it to head directly.

https://github.com/yo-goto/obsidian-card-view-mode/blob/6e829fc19586da46b858bc898d2116dfeec4446c/src/main.ts#L99
You can just register this function in the layoutReady callback.

https://github.com/yo-goto/obsidian-card-view-mode/blob/6e829fc19586da46b858bc898d2116dfeec4446c/src/main.ts#L110
Hmmmm not sure if this will work properly - I don't think it'll observe any new panes that are created after the plugin initializes, right? Also, this observer isn't unloaded when your plugin unloads. (Also be aware that ResizeObserver is very new and may not be widely supported on mobile yet).

yo-goto:
hi, Licad.
I've fixed the points you pointed out and newly released!
https://github.com/yo-goto/obsidian-card-view-mode

What I wanted to do was to make the same note title styles in both preview & edit modes in the last point you indicated.
I didn't find the way to do that in edit mode cause edit mode structure and that of preview mode are different. So I tried using Observer to observe pane size change and refresh Embedded card title box size. I wanted to delete this unnecessary space and make the same size with card width. 



After all, I changed title style in edit mode with this release.

lishid:
https://github.com/yo-goto/obsidian-card-view-mode/blob/b0c1ab22ba273b7df207e4cef71bf52caa02b562/src/main.ts#L28

Don't need to check for `layoutReady` since `onLayoutReady` does that for you - the logic in there is literally just
```ts
if (layoutReady) { callback(); }
else { registerLayoutReadyCallback(callback); }
```

https://github.com/yo-goto/obsidian-card-view-mode/blob/b0c1ab22ba273b7df207e4cef71bf52caa02b562/src/main.ts#L33
This isn't needed either since `onLayoutReady` will only fire the callback once and dereference.

https://github.com/yo-goto/obsidian-card-view-mode/blob/b0c1ab22ba273b7df207e4cef71bf52caa02b562/src/main.ts#L59
You can keep a reference to this element in your plugin as a field/property, and reference to it later, instead of giving it an ID and trying to query it from the DOM every time.

> What I wanted to do was to make the same note title styles in both preview & edit modes in the last point you indicated.
I didn't find the way to do that in edit mode because the edit mode DOM structure and that of preview mode are different. So I tried using Observer to observe pane size change and refresh Embedded card title box size. I wanted to delete this unnecessary space and make the same size with card width.

Ah I see - normally you should be able to use CSS to do that. Maybe you can try something like `width: calc(100% - 10px)` as a calculated width instead of a hardcoded one? If you're using the size of the child elements then you'd be looking at applying a margin/padding.

yo-goto:
Thank you for the review. I've fixed my code.

PR #438, opened 2021-08-23
Add Obsidian Trello plugin

OfficerHalf:

This is a plugin to connect Trello cards to Obsidian notes. It uses the trello rest API along with a new view in Obsidian to retrieve data, display comments, add new comments. Plenty of future work ahead to expand what it can do.

Code review notes:

- It's a fairly large one. I used [rxjs](https://www.learnrxjs.io/) all over the place in it, but that's the only 3rd party library involved.
- Has a dependency on the [MetaEdit](https://github.com/chhoumann/MetaEdit) plugin, but it should fail gracefully.
- Uses rxjs ajax to make requests to the Trello API. Internally this uses standard XMLHttpRequest stuff, it just gets wrapped in some rxjs magic.
- Yes, there's an API key committed in the code. This is intentional - not sure why Trello calls it that, it's really an application ID.
- Things are fairly broken up and documented, styles should be isolated. I used built in colors/styling where possible.
- Trello has pretty strict rate limits not just for each token but also each app. I'll need to adjust these if people start seeing errors. I can investigate creating multiple apps if need be. Currently responses are cached per session with a timeout, but not stored if Obsidian is closed.

## Repo URL

Link to my plugin: https://github.com/OfficerHalf/obsidian-trello/

## Release Checklist

  - Works on mobile too, with a couple small styling issues. But it doesn't make much sense as a mobile plugin IMO anyway, so it's not a top priority.
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2185,5 +2185,13 @@
         "description": "Switch input method with fcitx-remote when Vim keymap is enabled.",
         "author": "yuanotes",
         "repo": "yuanotes/obsidian-vim-im-switch-plugin"
+    },
+    {
+        "id": "obsidian-trello",
+        "name": "Obsidian Trello",
+        "description": "Connect Trello cards to Obsidian notes.",
+        "author": "OfficerHalf",
+        "repo": "OfficerHalf/obsidian-trello",
+        "branch": "main"
     }
 ]


lishid:
> The MetaEdit Obsidian plugin (install from inside Obsidian)

I think it may be worth quickly explaining why this is required (for the YAML API) in your readme.

https://github.com/OfficerHalf/obsidian-trello/blob/2976a0a65d5dcad2b82ea3d199f194cde0d74dee/src/plugin.ts#L43
You might want to use `Object.assign({}, DEFAULT_DATA, savedData)` so that when you add new keys to your data, they'll keep the default value rather than being `undefined`.

https://github.com/OfficerHalf/obsidian-trello/blob/2976a0a65d5dcad2b82ea3d199f194cde0d74dee/src/plugin.ts#L69
Need to use `this.registerEvent()` so this will be properly unhooked when your plugin unloads.

https://github.com/OfficerHalf/obsidian-trello/blob/2976a0a65d5dcad2b82ea3d199f194cde0d74dee/src/settings/board-select-modal.ts#L89
Tip: Use the shorthand `element.addClass()` and `element.removeClass()`. There's also `element.toggleClass('class', state)`.

OfficerHalf:
@lishid Done, done, and done. Updated the release with those changes and this PR to resolve the merge conflict. 🙂

PR #439, opened 2021-08-23
Add Hover External Link plugin

jamiebrynes7:

This plugin adds tooltips on hover for external links. These tooltips display the destination URL.

## Repo URL

Link to my plugin: https://github.com/jamiebrynes7/obsidian-hover-external-link

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2018,5 +2018,12 @@
         "description": "Automatically number or re-number headings in an Obsidian document.",
         "author": "Kevin Albrecht",
         "repo": "onlyafly/number-headings-obsidian"
+    },
+    {
+        "id": "hover-external-link",
+        "name": "Hover External Link",
+        "description": "Hover on external links to see the destination URL.",
+        "author": "Jamie Brynes",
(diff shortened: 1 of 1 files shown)


pjeby:
Rather than using Tippy, it might be a good idea to simply set the `aria-label` property of the link to its target URL.  Obsidian already has built-in support for displaying tooltips using that property, so that would eliminate the dependency on tippy, use less resources at runtime (i.e. no creating an object for every link for every preview including hover previews), and match the "native" theming/look-and-feel of Obsidian's existing tooltips.

lishid:
Actually, come to think of it - this would be a great addition to core! Thank you so much for the inspiration and hope you don't mind me stealing your ideas 🤣

jamiebrynes7:
> Actually, come to think of it - this would be a great addition to core! Thank you so much for the inspiration and hope you don't mind me stealing your ideas 🤣 

Please do! I'm more than happy for this to make its way into core 😄

PR #440, opened 2021-08-25
Add Code Block Enhancer Plugin

nyable:
This plugin provides copy button,linenumber,language name for the markdown code block in preview.
## Repo URL
Link to my plugin: [https://github.com/nyable/obsidian-code-block-enhancer](https://github.com/nyable/obsidian-code-block-enhancer)

## Release Checklist
 

- [x]  My GitHub release contains all required files
  - [x]  main.js
  - [x] manifest.json
  - [x] styles.css (optional)

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2054,5 +2054,12 @@
         "description": "This Plugin enables the use of Markdown Emoji Shortcodes :smile:, just like in Slack or Discord",
         "author": "phibr0",
         "repo": "phibr0/obsidian-emoji-shortcodes"
+    },
+    {
+        "id": "obsidian-code-block-enhancer",
+        "name": "Code Block Enhancer",
+        "description": "Enhance obsidian markdown code block,provides copy button,linenumber,language name tip and so on.",
+        "author": "hafuhafu",
+        "repo": "nyable/obsidian-code-block-enhancer"
     }
 ]


lishid:
https://github.com/nyable/obsidian-code-block-enhancer/blob/64595ac4dbb7d4701092381c34b64ed1d3e98fa1/src/main.ts#L23

This doesn't need to be awaited (even though it doesn't hurt anything)

https://github.com/nyable/obsidian-code-block-enhancer/blob/64595ac4dbb7d4701092381c34b64ed1d3e98fa1/src/main.ts#L25

This doesn't need to be awaited either - and this function doesn't need to be async.

https://github.com/nyable/obsidian-code-block-enhancer/blob/64595ac4dbb7d4701092381c34b64ed1d3e98fa1/src/core.ts#L104
Really shouldn't use this like that - The pane you're hovering may not be the same pane as the active one (which could also be null, or a different type of pane that isn't a `MarkdownView`). It doesn't seem like you really need it though since if the user was able to right-click on the element, it must be in preview mode.

https://github.com/nyable/obsidian-code-block-enhancer/blob/64595ac4dbb7d4701092381c34b64ed1d3e98fa1/src/core.ts#L114
I recommend calculating these when the user press the button, instead of before the menu shows up. That will save some computation as well as making sure that if the user scrolled the page after the menu showed it, it will still work correctly.

nyable:
Thanks for your code review.
Some related code has been cleaned up,and fixed the problem caused by scrolling the page after opening the menu.

> I recommend calculating these when the user press the button, instead of before the menu shows up.

Because different options need to be provided according to the height and position of the code block to reduce choices, I calculated in advance.

lishid:
> Because different options need to be provided according to the height and position of the code block to reduce choices, I calculated in advance.

Oh ok, that's fine yeah.

> https://github.com/nyable/obsidian-code-block-enhancer/blob/64595ac4dbb7d4701092381c34b64ed1d3e98fa1/src/main.ts#L25
>
> This doesn't need to be awaited either - and this function doesn't need to be async.

I think you missed this one.

PR #442, opened 2021-08-25
Add Annotator Plugin

elias-sundqvist:

## Repo URL

Link to my plugin: https://github.com/elias-sundqvist/obsidian-annotator

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2054,5 +2054,12 @@
         "description": "This Plugin enables the use of Markdown Emoji Shortcodes :smile:, just like in Slack or Discord",
         "author": "phibr0",
         "repo": "phibr0/obsidian-emoji-shortcodes"
+    },
+    {
+        "id": "obsidian-annotator",
+        "name": "Annotator",
+        "description": "A plugin for reading and annotating PDFs and EPUBs in obsidian.",
+        "author": "Elias Sundqvist",
+        "repo": "elias-sundqvist/obsidian-annotator"
     }
 ]


6-AND-9:
> I think epub support and maybe even the darkreader stuff isn't necessary for everyone. 

Believe me, dark mode is absolutely critical. 

As for EPUB support, it is a great addition, as I often read EPUB books and mark them up all over the place with various software. If I can integrate everything in Obsidian that'd be awesome. But as Elias said, its size is negligible anyway.

lishid:
We actually do ship PDF.js with Obsidian, it just isn't the most up-to-date version and isn't automatically loaded into the app unless requested - perhaps we can expose an API to load PDF.js on demand? I've seen another plugin use it so that might be helpful.

elias-sundqvist:
Which parts of pdf.js do you ship? 
I bundle both the web.js javascript library as well as the web viewer. 
Would I be able to remove both parts or just the former?
image

lishid:
Ah I see! Currently we only ship `pdf.js` and `pdf.worker.js` (in their minified form too) but we can look into adding the viewer. IIRC mozilla encourage developers to make UI tweaks to the viewer so as to avoid it looking like the one in Firefox? Not sure how much work it would be.

lishid:
PS. Minifying the source codes actually makes a big difference - pdf.js comes un-minified by default.

elias-sundqvist:
I've managed to reduce the plugin to around 8MB now. I'm now out of tricks. 
* The files are minimized before they are zipped. 
* The zip file is now stored as unicode in a multiline comment instead of a base64 string variable. This allows for higher information density. 
   * You said that source code, in particular, takes a lot of ram @lishid. Is this true for comments as well?
* The rollup output is also minimized. 
I could possibly use the built-in `pdf.js` and `pdf.js.worker` files that are bundled. Not sure how much that would help though, and it's kind of nice to have everything self-contained in the plugin.

lishid:
I think that's fine! Some first-pass minimization is reasonable but don't waste too much time trying to squeeze the last few bits.

lishid:
One thing to keep in mind though, we've gotten reports that another plugin which includes its own PDF.js causes PDFs to stop working in Obsidian mobile - once we ship the API for PDF.js it might be a good idea to hook it up if available.

PR #443, opened 2021-08-26
Add customJS plugin

SamLewis0602:

## Repo URL

Link to my plugin: https://github.com/samlewis0602/obsidian-custom-js

## Release Checklist

- [x] I have tested this on macOS, iOS, and windows
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2063,7 +2063,7 @@
         "repo": "uphy/obsidian-reminder"
     },
     {
-        "id": "obsidian-custom-js",
+        "id": "customjs",
         "name": "CustomJS",
         "author": "Sam Lewis",
         "description": "Reuse custom javascript across desktop and mobile.",
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2061,5 +2061,12 @@
         "author": "uphy",
         "description": "Manage markdown TODOs with reminder.",
         "repo": "uphy/obsidian-reminder"
+    },
+    {
+        "id": "obsidian-custom-js",
+        "name": "CustomJS",
+        "author": "Sam Lewis",
+        "description": "Reuse custom javascript across desktop and mobile.",
+        "repo": "samlewis0602/obsidian-custom-js"
     }
 ]


pjeby:
Rather than adding the constructors to your plugin object, may I suggest something like:

```js
 async loadFunctions() {
    window.customJS = {}
    // ...
        window.customJS[cls.constructor.name] = cls
}

onunload() {
    delete window.customJS;
}
```

With this approach, `customJS` will be a global variable accessible from everywhere in Obsidian, eliminating the need for people to use `app.plugins.plugins.customjs`.  They would also not be accessing your plugin object and all its other data and methods, and it wouldn't be possible to accidentally break the plugin by a naming conflict.  (e.g. if somebody named a class `saveSettings`, it would overwrite your saveSettings method!)

Also, it looks like changing the content of any .js files currently requires Obsidian to be restarted or for the plugin to be disabled and re-enabled for the changes to take effect.  You might want to consider listening to the vault `modify` event for .js files (i.e., checking if the file.extension is `js`) and then rerunning loadFunctions if the file is on your active list.  That way people can edit their existing JS to fix a mistake or add something new without needing to restart, or to fake-edit the file list to trick the plugin into reloading.

SamLewis0602:
Great tips, thank you! I just implemented them both in release `1.0.3`. I have a couple questions for you:
1. I added the `vault.on('modify', ...)` call in `onload`. Do I need to do anything in `onunload` to clean up that listener?
2. Though my plugin is now reloading the tracked files on change, it seems that Dataview caches the result of a block, so I have to add a space or something to get Dataview to rerun the block and see the updated result of my custom code. I know this is likely an optimization from Dataview, but curious if you're aware of anything I could use to make that more seamless for users.

pjeby:
1. You need to do it as `this.registerEvent(this.app.vault.on('modify', this.reloadIfNeeded, this))`, and then it will be removed at unload automatically.
2. Not really, no; you might ask the Dataview plugin author if they can provide an API to invalidate their JS cache (or whatever it is that needs to be invalidated, refreshed, etc.)

SamLewis0602:
Perfect, release 1.0.4 is using `registerEvent`. Appreciate your help, this is my first plugin so still learning the ins and outs of the API. Let me know if there's anything else that needs to be done to get it on the community store!

lishid:
https://github.com/SamLewis0602/obsidian-custom-js/blob/95d8212901c1488fc71b8de18ff0cba53f425c9c/main.ts#L11

Please rename these sample class/interface names to your plugin's name.

https://github.com/SamLewis0602/obsidian-custom-js/blob/95d8212901c1488fc71b8de18ff0cba53f425c9c/main.ts#L17
Don't need to await this.

https://github.com/SamLewis0602/obsidian-custom-js/blob/95d8212901c1488fc71b8de18ff0cba53f425c9c/main.ts#L60
You can assign this first before starting to load each file.

pjeby:
> Don't need to await this.

Awaiting there avoids re-entry to the load routine, if a modification occurs during loading.

Though to be fair there isn't any re-entry protection on subsequent reloads, and that's a more likely scenario than re-entry during plugin start.  It would probably be best for rename to trigger a debounced or queued version of the reload function if the goal is to prevent such re-entry.

So await isn't really helping *much*, but it's also not hurting anything.

> You can assign this first before starting to load each file.

Not if the goal is to load or update atomically, such that the exposed `customJS` is always a consistent snapshot.

lishid:
> Awaiting there avoids re-entry to the load routine, if a modification occurs during loading.

The function isn't even awaiting its child tasks - they're all ran in parallel and not awaited for. It's missing a `Promise.all()` (or better, use `for(in)`)?

> Not if the goal is to load or update atomically, such that the exposed customJS is always a consistent snapshot.

Right.

pjeby:
> The function isn't even awaiting its child tasks - they're all ran in parallel and not awaited for. It's missing a Promise.all() (or better, use for(in))?

Ah, I missed that.  It should be a for loop, not a `forEach`, or else a `Promise.all(.map)`.

SamLewis0602:
> Please rename these sample class/interface names to your plugin's name.

Done, not sure how I missed that one 🤦‍♂️

> > You can assign this first before starting to load each file.

> Not if the goal is to load or update atomically, such that the exposed customJS is always a consistent snapshot.

pjeby's assumption is correct here. My thinking was that this prevents intermediate states where an expected plugin is missing from the `customJS` object. Which leads us to...

> The function isn't even awaiting its child tasks - they're all ran in parallel and not awaited for. It's missing a Promise.all() (or better, use for(in))?

Good catch, I switched this to a `for..of` so that it actually waits to assign to `window.customJS` as I originally intended.

Just released 1.0.5 with the changes.

lishid:
Ok fair enough. One small thing:

https://github.com/SamLewis0602/obsidian-custom-js/blob/dbd46c27284093297469fed96d357044aca75f52/main.ts#L47
Might want to use `.endsWith` instead of `.includes`.

Also, I don't seem to have permission to modify your PR to merge the conflict, so you'll have to do it yourself before I can have it merged.

SamLewis0602:
Oops sorry about that, I just gave edit perms.

1.0.6 released that uses `.endsWith` instead of `.includes`

PR #445, opened 2021-08-26
Add Banners plugin

noatpad:
This plugin adds the ability to add banners to your notes, through either an image URL or a local image file.

## Repo URL
Link to my plugin: https://github.com/noatpad/obsidian-banners

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2068,5 +2068,12 @@
         "author": "phibr0",
         "description": "This Plugin provides a go to Line Command",
         "repo": "phibr0/obsidian-go-to-line"
+    },
+    {
+        "id": "obsidian-banners",
+        "name": "Banners",
+        "author": "Danny Hernandez",
+        "description": "Add banner images to your notes!",
+        "repo": "noatpad/obsidian-banners"
     }
 ]


lishid:
https://github.com/noatpad/obsidian-banners/blob/4bb9f15c7eac0c52aa641100fc8ae25ffe3dc2dd/src/main.ts#L49

This isn't a great idea for performance as it will cause a lot of wasted effort unloading and re-loading every pane on each startup.

https://github.com/noatpad/obsidian-banners/blob/4bb9f15c7eac0c52aa641100fc8ae25ffe3dc2dd/src/main.ts#L140
You can use `this.registerEvent(vault.on(....))` on all Obsidian events to have it automatically unhook when your plugin unloads.

https://github.com/noatpad/obsidian-banners/blob/4bb9f15c7eac0c52aa641100fc8ae25ffe3dc2dd/src/main.ts#L103
Make sure you always check if it's a markdown file using `file.extension === 'md'`

https://github.com/noatpad/obsidian-banners/blob/4bb9f15c7eac0c52aa641100fc8ae25ffe3dc2dd/src/main.ts#L175
That's a bit weird? Would this call `Elephant Tail` to be a duplicate of `Elephant`?

https://github.com/noatpad/obsidian-banners/releases/tag/1.0.0
Your release is quite big for what I think should have been for your plugin - maybe it's because of the `common-tags` and `lodash` dependencies? Are those necessary? 

https://github.com/noatpad/obsidian-banners/blob/4bb9f15c7eac0c52aa641100fc8ae25ffe3dc2dd/src/MetaManager.ts#L91
I think this will fail if there's a mistyped frontmatter like
```
----
```

https://github.com/noatpad/obsidian-banners/blob/4bb9f15c7eac0c52aa641100fc8ae25ffe3dc2dd/src/BannersProcessor.ts#L86
I see that a lot of your code is done to manage banners that you add to the top of the view's container, which is normally out of your control - why not add the banner to the first element by testing for `element.querySelector('pre.frontmatter')`? That way the first section is always re-rendered when the frontmatter changes and swapping files will automatically cause a re-render. This will free you up from having to manually manage DOM elements outside and having to remember to clean things up.

https://github.com/noatpad/obsidian-banners/blob/4bb9f15c7eac0c52aa641100fc8ae25ffe3dc2dd/src/BannersProcessor.ts#L201
These could be managed via global CSS variables rather than toggling classes onto each element.

https://github.com/noatpad/obsidian-banners/blob/4bb9f15c7eac0c52aa641100fc8ae25ffe3dc2dd/src/BannersProcessor.ts#L215
I have a feeling what you're trying to accomplish here with positioning can be done with CSS, without having to resort to manual JS computations. Not 100% sure if that's possible but it would make things much simpler and more performance if that's possible.

noatpad:
Heyo, I took a bit to work through these suggestions, so thanks for taking a look at it. Lemme address a few of these:

> https://github.com/noatpad/obsidian-banners/blob/4bb9f15c7eac0c52aa641100fc8ae25ffe3dc2dd/src/BannersProcessor.ts#L86
I see that a lot of your code is done to manage banners that you add to the top of the view's container, which is normally out of your control - why not add the banner to the first element by testing for element.querySelector('pre.frontmatter')? That way the first section is always re-rendered when the frontmatter changes and swapping files will automatically cause a re-render. This will free you up from having to manually manage DOM elements outside and having to remember to clean things up.

So I thought at first this wouldn't be doable, but after playing around with it some more, I've managed to get it to look the way I want with `pre.frontmatter` as the wrapper. And yeah, that's helped a _ton_ 'cause I can not worry about all the checks for when a file changes and such. I could take off some unnecessary checks and logic 'cause of that, so merci.

> https://github.com/noatpad/obsidian-banners/blob/4bb9f15c7eac0c52aa641100fc8ae25ffe3dc2dd/src/main.ts#L49
This isn't a great idea for performance as it will cause a lot of wasted effort unloading and re-loading every pane on each startup.

Yeah, it's also caused an issue shortly after release too. So I've opted for a different way to "refresh" the Markdown views with `MarkdownView.previewMode.rerender()`.

> https://github.com/noatpad/obsidian-banners/blob/4bb9f15c7eac0c52aa641100fc8ae25ffe3dc2dd/src/MetaManager.ts#L91
I think this will fail if there's a mistyped frontmatter like `----`

For this one, I used `MetadataCache.getFileCache()` to get the YAML line numbers, so now that check's been removed.

> https://github.com/noatpad/obsidian-banners/blob/4bb9f15c7eac0c52aa641100fc8ae25ffe3dc2dd/src/BannersProcessor.ts#L201
These could be managed via global CSS variables rather than toggling classes onto each element.

I wasn't quite sure what you mean by this, since I need styling from a common class. And I think I should mention one of the things I have on the todo list is to give notes the ability to have note-specific style settings that would override the global settings.

> https://github.com/noatpad/obsidian-banners/blob/4bb9f15c7eac0c52aa641100fc8ae25ffe3dc2dd/src/BannersProcessor.ts#L215
I have a feeling what you're trying to accomplish here with positioning can be done with CSS, without having to resort to manual JS computations. Not 100% sure if that's possible but it would make things much simpler and more performance if that's possible.

I also initially thought that JS was my only go-to, but after playing around with it some more, I found a method that works. So now there's no more extra work when resizing panes.

> https://github.com/noatpad/obsidian-banners/releases/tag/1.0.0
Your release is quite big for what I think should have been for your plugin - maybe it's because of the `common-tags` and `lodash` dependencies? Are those necessary?

Those libraries are more so for developer convenience since they help in a few areas, & I used tree-shaking to minimize that footprint. But I looked into it, & I forgot to remove the sourcemap from the initial release, so that's why it was bumped so high, haha. Now it's down to about ~60KB, & you can see it as a pre-release here: https://github.com/noatpad/obsidian-banners/releases/tag/1.1.0

---

The rest of the points can be dismissed since the related code was removed as part of the rewrite. Thanks again for the pointers and reviewing this!

lishid:
> I wasn't quite sure what you mean by this, since I need styling from a common class. And I think I should mention one of the things I have on the todo list is to give notes the ability to have note-specific style settings that would override the global settings.

Ah yes you're right - that wouldn't be possible with a global prop.

> Those libraries are more so for developer convenience since they help in a few areas, & I used tree-shaking to minimize that footprint. But I looked into it, & I forgot to remove the sourcemap from the initial release, so that's why it was bumped so high, haha. Now it's down to about ~60KB, & you can see it as a pre-release here: https://github.com/noatpad/obsidian-banners/releases/tag/1.1.0

Sounds good. I usually avoid using "convenience" libraries when those functions can be copy & pasted right into the source code, to avoid things like the left-pad incident. If you can shake them clean in your release build that's fine too 👍

PR #446, opened 2021-08-26
Obsidian42 - Jump-to-Date

TfTHacker:
Link to my plugin: https://github.com/TfTHacker/obsidian42-jump-to-date

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1932,7 +1932,7 @@
         "author": "TfTHacker",
         "repo": "TfTHacker/Obsidian-WordNet",
         "branch": "main"
-    }, 
+    },
     {
         "id": "multi-line-formatting",
         "name": "Multi-line Formatting",
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1932,14 +1932,7 @@
         "author": "TfTHacker",
         "repo": "TfTHacker/Obsidian-WordNet",
         "branch": "main"
-    },
-    {
-        "id": "obsidian-jump-to-date-plugin",
-        "name": "Obsidian42 - Jump-to-Date",
-        "description": "Popup calendar for quickly navigating dates.",
-        "author": "TfTHacker",
-        "repo": "TfTHacker/obsidian42-jump-to-date"
-    },    
+    }, 
     {
         "id": "multi-line-formatting",
         "name": "Multi-line Formatting",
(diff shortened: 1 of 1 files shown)


TfTHacker:
Once again, thank you for the time you put into providing feedback. Let me address the issues:

- You should use moment from obsidian to avoid importing another copy into your plugin.
  - I removed this import. I am not sure though how to get type checking without including moment in the import.

- Two things here - first, you can use workspace.onLayoutReady(callback) to run code after all plugins load and the workspace is ready, but second, I don't think this is necessary, since you're only triggering the import when the callback is fired, and users shouldn't have been able to run commands before all plugins finish loading.
  - I changed to onLayoutReady. this feature should only be available if this auxillary plugin is install, so for this reason I am doing a test if its installed, either to display it or not.


- Here, the better way should be using workspace.getLeaf() if not requesting a new pane, or workspace.splitActiveLeaf(direction). Avoid using workspace.activeLeaf as much as possible since it can be null.
  - Thank you - changed this throughout my code

lishid:
> I removed this import. I am not sure though how to get type checking without including moment in the import.

I think the practice I've seen is to use
```ts
import { moment } from 'obsidian';
import type { Moment } from 'moment'; // If you need to store Moment objects for typing
```

> I changed to onLayoutReady. this feature should only be available if this auxillary plugin is install, so for this reason I am doing a test if its installed, either to display it or not.

I see, yeah that works!

https://github.com/TfTHacker/obsidian42-jump-to-date/blob/cc1d7dc803de7ccd41d334c2b635415ae0a8caaf/src/main.ts#L80
Should replace this with `workspace.splitActiveLeaf(direction)`, since `getLeaf` may split a new leaf for you, and you will accidentally create two new panes.

https://github.com/TfTHacker/obsidian42-jump-to-date/blob/cc1d7dc803de7ccd41d334c2b635415ae0a8caaf/src/main.ts#L87
These ignore shouldn't be necessary I think.

TfTHacker:
Thanks again for the feedback. Made some changes based on this feedback and added an eslint verification (so some TS cleanup resulting from that).

The only issue is TS ignore. I am making some calls in a number of places where the obsidian.d.ts file doesn't have definitions. The warnings then appear in vscode, which distracts me :-)

I am not sure if there is another way to solve it or if it is one of the side effects of using TS,

lishid:
> The only issue is TS ignore. I am making some calls in a number of places where the obsidian.d.ts file doesn't have definitions. The warnings then appear in vscode, which distracts me :-)

I'm looking at the code there and I am not seeing anything we forgot to expose in the typescript definitions. Can you show me what the warnings are?

TfTHacker:
As always, thanks for all your help.

So here are  some  examples: 

# plugins

- https://github.com/TfTHacker/obsidian42-jump-to-date/blob/424ee63a6a2df061d9deb912dff8e47d21792bc2/src/main.ts#L41
- https://github.com/TfTHacker/obsidian42-jump-to-date/blob/424ee63a6a2df061d9deb912dff8e47d21792bc2/src/ui/datenlpModal.ts#L26

# view.file
- https://github.com/TfTHacker/obsidian42-jump-to-date/blob/424ee63a6a2df061d9deb912dff8e47d21792bc2/src/ui/calendarPicker.ts#L34

The other cases are from the 3rd party component  I am using.

PR #447, opened 2021-08-27
Add obsidian-icon-folder

FlorianWoelki:

## Repo URL

Link to my plugin: https://github.com/FlorianWoelki/obsidian-icon-folder

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2125,5 +2125,12 @@
         "author": "phibr0",
         "description": "This Plugin lets you search relative to your Cursor Position.",
         "repo": "phibr0/obsidian-relative-find"
+    },
+    {
+        "id": "obsidian-icon-folder",
+        "name": "Obsidian Icon Folder",
+        "author": "FlorianWoelki",
+        "description": "This plugin allows you to add icons to your folders.",
+        "repo": "FlorianWoelki/obsidian-icon-folder"
     }
 ]


pjeby:
You might want to add vault event handlers for rename and delete, so that you can avoid orphaned icons in your data.json.  Currently if a folder is moved, renamed, or deleted, your icon database isn't being updated.

FlorianWoelki:
> You might want to add vault event handlers for rename and delete, so that you can avoid orphaned icons in your data.json. Currently if a folder is moved, renamed, or deleted, your icon database isn't being updated.

Thank you for the suggestion! I will add this to the plugin asap.

FlorianWoelki:
> https://github.com/FlorianWoelki/obsidian-icon-folder/blob/83ff2274557e51325fa9c0c777fbb227bed2dd69/src/main.ts#L25
> 
> I think you don't need to hide the menu manually - it should be automatic.
> 
> https://github.com/FlorianWoelki/obsidian-icon-folder/blob/83ff2274557e51325fa9c0c777fbb227bed2dd69/src/util.ts#L12
> 
> Maybe only query the changed elements instead of doing it from global? MutationObserver can fire very rapidly and running queries many times can be a performance concern. Also, perhaps keep the results in a variable instead of querying again if the test passes?
> 
> https://github.com/FlorianWoelki/obsidian-icon-folder/blob/83ff2274557e51325fa9c0c777fbb227bed2dd69/src/util.ts#L50
> Hmmm... is there not a better way to import these icons instead of having it rendered with react and then setting it to innerHTML? I took a quick look and it seems that someone's gotta do the legwork to write a script to import those icons as SVG strings into the plugin.

Thank you for your review!

I've tried to optimize the use of the MutationObserver functionality. Btw. it would be pretty awesome if the plugin api has some kind of lifecycle hooks where for example you can catch whenever the editor is fully loaded.

Yeah, I did not find any better and easier way for me as well. But I will do the necessary work.

pjeby:
Hm.  Now that I'm looking at it, it looks like the plugin isn't considering the possibility of explorer views being opened and closed.  Not only can somebody do that manually, but it would also happen upon changing workspaces, I think?  It definitely would happen if the explorer pane was at different places in the workspace, or if you switched between workspaces that had explorer views and ones that didn't.  It's just using MutationObserver to detect the startup of the first file explorer, or more precisely, for the first rendering of each folder.  So they'll hang around until that specific folder is rendered, creating runtime overhead.

You could possibly get rid of MutationObserver entirely if you just had the plugin operate on the `.fileItems[folderPath].el` of all File Explorer views.  The only hard part of that (besides being undocumented) is that you wouldn't know when explorer views were being opened and thus when to perform the operation.  But you'd only need to do it once per opened explorer view, since the fileItems and their DOM elements aren't re-created, I don't think.  Worst case, I suppose you could keep a WeakMap of explorer views with a flag as to whether you'd updated them, and then on the `layout-changed` event look for any explorer views that aren't in the weakmap and update them.  That would drop runtime overhead to as close to zero as you can get it, and you certainly wouldn`t be using an observer for each folder, let alone having them hang around until a folder's actually rendered,

pjeby:
Background I left out of the previous comment: the file explorer view lazily renders DOM elements for performance reasons, but it does create an element for every entry in the vault when the view loads, and it stores them in a map of `.fileItems`.  These `.fileItems` have `.file` and `.el` properties (among others) that correspond to the file and the DOM element that will render that file.  So for purposes of this plugin, you need only manipulate the `.el` of those fileItems that match the paths you're planning to manipulate.

The view objects also have a `.ready` property that should be true if `fileItems` is fully populated, so you can skip any views that aren't ready yet.

FlorianWoelki:
> Background I left out of the previous comment: the file explorer view lazily renders DOM elements for performance reasons, but it does create an element for every entry in the vault when the view loads, and it stores them in a map of `.fileItems`. These `.fileItems` have `.file` and `.el` properties (among others) that correspond to the file and the DOM element that will render that file. So for purposes of this plugin, you need only manipulate the `.el` of those fileItems that match the paths you're planning to manipulate.
> 
> The view objects also have a `.ready` property that should be true if `fileItems` is fully populated, so you can skip any views that aren't ready yet.

I really appreciate your suggestion. I cannot really find the `.fileItems` in the types file. Can you please tell me where it actually is or how I can access this property?

lishid:
> I really appreciate your suggestion. I cannot really find the .fileItems in the types file. Can you please tell me where it actually is or how I can access this property?

These are private fields that are not exposed through the API, which is why you aren't seeing it in the typings file. You may choose to use them using `// @ts-ignore` but it's more likely to break if we choose to change the internal implementations.

FlorianWoelki:
> > I really appreciate your suggestion. I cannot really find the .fileItems in the types file. Can you please tell me where it actually is or how I can access this property?
> 
> These are private fields that are not exposed through the API, which is why you aren't seeing it in the typings file. You may choose to use them using `// @ts-ignore` but it's more likely to break if we choose to change the internal implementations.

I still didn't really get in which object I can find this property. Can you please tell me where this property belongs to?

In addition, I've made some changes to the project where react is completely removed. There is now a small script that generates the icons with their corresponding svg string in a simple file.

lishid:
> I still didn't really get in which object I can find this property. Can you please tell me to what object this property belongs?

I think he meant the file explorer pane's view which you can get from `workspace.getLeavesOfType('file-explorer')` (returns an array of possible file explorer panes, and you can use `leaf.view` to access the view)

> In addition, I've made some changes to the project where react is completely removed. There is now a small script that generates the icons with their corresponding svg string in a simple file.

Cool, can you try to build the final file and upload to the release? Would be nice to compare the difference.

pjeby:
Yep.  The Typescript declarations you'd need (if you don't use `any` or `ts-ignore`) would look something like:

```ts
declare module "obsidian" {
    interface Workspace {
        getLeavesOfType(viewType: "file-explorer"): ExplorerLeaf[]
    }
}

interface ExplorerLeaf extends WorkspaceLeaf {
    view: ExplorerView
}

interface ExplorerView extends View {
    fileItems: Record<string,FileItem>  // keyed by path
    ready: boolean   // true if fileItems is populated
}

interface FileItem {
    el: HTMLDivElement  // The DOM element that will (eventually) be shown for this file/folder
}
```

pjeby:
If you go this route, you will probably want to listen to `layout-ready` and `layout-changed` (as well as triggering on the initial `onLayoutReady()`), scan the list of explorer views and see if you have them in your weakmap.  If they are, skip them, but if not, check if they're ready, and if so, patch the dom elements and add the view to your weakmap.

I think that algorithm will actually always end up with the view being ready, *except* when Obsidian first starts, and even then, only if its layout-ready handler fires after yours.  So it might actually be safe to just ignore the ready flag altogether.

FlorianWoelki:
> Cool, can you try to build the final file and upload to the release? Would be nice to compare the difference.

I've updated the [release files](https://github.com/FlorianWoelki/obsidian-icon-folder/releases/tag/1.0.0). I hope it is ok that I did not tag any new version :D 



> If you go this route, you will probably want to listen to `layout-ready` and `layout-changed` (as well as triggering on the initial `onLayoutReady()`), scan the list of explorer views and see if you have them in your weakmap. If they are, skip them, but if not, check if they're ready, and if so, patch the dom elements and add the view to your weakmap.
> 
> I think that algorithm will actually always end up with the view being ready, _except_ when Obsidian first starts, and even then, only if its layout-ready handler fires after yours. So it might actually be safe to just ignore the ready flag altogether.

Thank you for your typing definition. Interestingly enough, I do not have any explorer views or the array is empty for me.

pjeby:
The array will definitely be empty if you do it before layoutReady, or if there are no Explorer leaves open.  You have to do it from onLayoutReady() or a later event, otherwise there will be no leaves or views of any kind, let alone explorer ones.

FlorianWoelki:
> The array will definitely be empty if you do it before layoutReady, or if there are no Explorer leaves open. You have to do it from onLayoutReady() or a later event, otherwise there will be no leaves or views of any kind, let alone explorer ones.

Thank you! As you said, I've ignored the `ready` boolean completely. Can you quickly check if this code makes sense:

```ts
private registeredViews = new WeakMap();
//...
const addIconsToDOM = () => {
  const fileExplorers = this.app.workspace.getLeavesOfType('file-explorer');
  fileExplorers.forEach((fileExplorer) => {
    if (this.registeredViews.has(fileExplorer)) {
      return;
    }

    this.registeredViews.set(fileExplorer, true);
    data.forEach(([key, value]) => {
      const fileItem = fileExplorer.view.fileItems[key];
      if (fileItem) {
        const titleEl = fileItem.titleEl;
        const titleInnerEl = fileItem.titleInnerEl;

        const iconNode = titleEl.createDiv();
        iconNode.classList.add('obsidian-icon-folder-icon');
        iconNode.innerHTML = getIcon(value.substring(2));

        titleEl.insertBefore(iconNode, titleInnerEl);
      }
    });
  });
};

  addIconsToDOM();
});

this.registerEvent(
  this.app.workspace.on('layout-change', () => {
    addIconsToDOM();
  }),
);
```

pjeby:
Apart from the DOM manipulation, which I'm assuming you've tested, the above looks good to me.  You might want to test that it works when you close and re-open an explorer view, or switch between workspaces, or restart Obsidian.  If it does, then I think everything's covered.

FlorianWoelki:
> Apart from the DOM manipulation, which I'm assuming you've tested, the above looks good to me. You might want to test that it works when you close and re-open an explorer view, or switch between workspaces, or restart Obsidian. If it does, then I think everything's covered.

Yeah I've tested it. Thank you for your help, really appreciate it! :)

PR #448, opened 2021-08-27
Add zk-new-pane (Open Zettelkasten in New Pane)

pjeby:
...but this is such a trivial feature that you might want to just add the same functionality to the core plugin for a future release, and it might actually be less work than reviewing this plugin.  :wink:

(Especially since it can just be done by adding a new-pane flag to `onCreateNote()` and passing it through to `getLeaf()`, then adding a second command with an extra bind parameter of true, and making the button handler set the flag based on `Keymap.isModifier(event, "Mod")`.)

So yeah, if you want to implement the functionality in the next insider release, feel free to just close this PR.

## Repo URL
Link to my plugin: [zk-new-pane](/pjeby/zk-new-pane)

## Release Checklist
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2132,5 +2132,12 @@
         "description": "When you click an image, it will be displayed in a popup layer and you can view, drag, zoom, rotate the image.",
         "author": "sissilab",
         "repo": "sissilab/obsidian-image-toolkit"
+    },
+    {
+        "id": "zk-new-pane",
+        "name": "Open Zettelkasten in New Pane",
+        "description": "Adds a command to open a new Zettel in a new pane (or Ctrl/Cmd-click the ribbon button)",
+        "author": "pjeby",
+        "repo": "pjeby/zk-new-pane"
     }
 ]


lishid:
Will be added in next release.

lishid:
(Will follow convention for `Mod` as well as middle click - which will also apply for command palette).

pjeby:
Sorry, didn't quite follow the mod vs. middle click vs. command pallette thing.  I know I forgot about middle click (and now that I've remembered, I probably have other plugins that need some fixing on that front), but was there something else I missed?  Thanks.

lishid:
Oh no, I meant that we'll also add the mod+click/mod+enter and middle click triggers for the command palette, that way it's consistent and also don't require adding a second "open in new pane" command for each one that opens a file.

pjeby:
But how can it be mapped to a hotkey then?  (That was the main point of the plugin, at least for me.)

lishid:
Ahhh you're right! I did not think of that... Let me give it a bit more thought.

PR #449, opened 2021-08-27
Add Plugin Dynamic Table of Contents

Aidurber:
## Repo URL
Link to my plugin: https://github.com/aidurber/obsidian-plugin-dynamic-toc

## Release Checklist

  - [X] `main.js`
  - [X] `manifest.json`


lishid:
Don't really recommend this - you will be creating multiple objects overwriting each other when multiple panes are open, and they will stop working properly.

https://github.com/Aidurber/obsidian-plugin-dynamic-toc/blob/f94f765b45af9af4c81bc51131c8f4c80078fb72/src/main.ts#L55
This event is not properly unregistered if you create more than one renderer.

https://github.com/Aidurber/obsidian-plugin-dynamic-toc/blob/f94f765b45af9af4c81bc51131c8f4c80078fb72/src/HeadingExtractor.ts#L7
This just seems like helper functions... perhaps you don't have to create a whole class/interface combo and just export `clean` and `extract` as regular functions?

https://github.com/Aidurber/obsidian-plugin-dynamic-toc/blob/f94f765b45af9af4c81bc51131c8f4c80078fb72/src/DynamicInjectionPostProcessor.ts#L17
This doesn't seem conventional, I'm not familiar with the performance characteristics of it compared to something like querySelector.

https://github.com/Aidurber/obsidian-plugin-dynamic-toc/blob/f94f765b45af9af4c81bc51131c8f4c80078fb72/src/DynamicInjectionPostProcessor.ts#L28
The block being rendered here may not be the active file - the current file path is passed through the context object as `sourcePath`



---

Either way - it seems like your plugin is designed for only a single (preview) pane open, and will likely not work well/break when multiple panes are used.

Aidurber:
@lishid Thanks again for the review, this all makes sense. 

Regarding the document.evaluate vs querySelector. It's not possible to use querySelector to find a node by child tex, using document.evaluate with xpath it's more performant than me recursively walking the DOM (I'd need to benchmark that but I'm pretty confident). The alternative is I come up with another convention to allow compatibility with other renderers, some users on Reddit wanted to inject custom nodes such as `[[__TOC__]]` so it works with GitLab. 

I do have one question which will help me fix the rendering/multiple pane issues: 
- What's a recommended way to re-trigger the code block processor when settings have changed?

I could keep a hash/array of sourcePaths and a render function which I trigger when settings have saved. Are there any patterns/Obsidian APIs that you're aware of which will help me here?

lishid:
Yeah so, in theory, the block renderers are designed to be self-contained - it's not a good idea to keep them referenced from your plugin because you don't know when they're removed, causing memory leaks.

What you'll want to do, is to create a child component, and call `context.addChild(myComponent)` when rendering the block. The child component can have an `onload` and `onunload` function which will be called when the component is loaded and removed. This component should be self contained, but otherwise it can register events like `changed` using `this.registerEvent(vault.on('changed', callback))`.

Currently there is no guaranteed way to completely re-render a block, but if you're rendering the block and reacting to events to update the contents, it should be doable from within that component. Just make sure you don't keep references to it globally.

Aidurber:
Interesting, thanks for sharing. I'll look into it.

Aidurber:
@lishid I've taken your advice: 
- CodeBlock and Markdown Processors are now components and correctly register their events with the `this.registerEvent` and `onload` lifecycle. 
- Extracted `HeadingExtractor` into pure functions
- I dropped the XPath approach for selectors. I had more control over casing and it's much more readable/less arcane. 

In extracting the codeblock processor and markdown post processors into components that fixed the issue with multiple panes. Thanks again for your time on this.

lishid:
https://github.com/Aidurber/obsidian-plugin-dynamic-toc/blob/bb55ea2c830c85f70a06e2a9905af3e134bae628/src/main.ts#L25
This isn't a good move - you'll be registering a life-cycle managed child for each and every markdown block of each file. You should perform any matching/searching of the text first, and then only if you find a match, create and add the child component.

Aidurber:
I've incorporated those changes

PR #453, opened 2021-08-28
Add obsidian-file-link plugin

marcjulianschwarz:

## Repo URL

Link to my plugin: https://github.com/marcjulianschwarz/obsidian-file-link

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2132,5 +2132,12 @@
         "description": "When you click an image, it will be displayed in a popup layer and you can view, drag, zoom, rotate the image.",
         "author": "sissilab",
         "repo": "sissilab/obsidian-image-toolkit"
+    },
+    {
+        "id": "obsidian-file-link",
+        "name": "Obsidian File Link",
+        "description": "This plugin lets you create simple file links in your notes.",
+        "author": "Marc Julian Schwarz",
+        "repo": "marcjulianschwarz/obsidian-file-link"
     }
 ]


pjeby:
It might be good to include some info in your README about the differences between this and using drag-drop from Finder or Explorer to create file links (by holding Cmd or Ctrl respectively when dropping).

It looks like the only differences are that this plugin 1) has a different interface, 2) allows including the extension, and 3) allows prefix characters when dropping multiple files.  But if there are any other differences that would be good to note as well.

marcjulianschwarz:
Thank you for the information, I actually didn't know you could drag and drop files into Obsidian. I guess the 3 points mentioned by you are the only differences then. Therefore this plugin definitely needs more work (features). I thought this functionality didn't already exist. 
I don't know if it makes sense to put it on the community list.

pjeby:
Plugins don't have to have a lot of features, but it's important to call out the distinguishing features so that users can tell right away if it's something they need.  If you're still planning to use your plugin instead of dragging and dropping, that probably means there are other people who might want those same features.  And it'll be easier for them to tell this is the plugin for them if those features are called out up front.

marcjulianschwarz:
Just updated the README file. I hope it is now easier to see the advantages of this plugin. 
I changed the main.ts file too, added one more setting. I should probably delete the release now and add a new one with the updated file?

marcjulianschwarz:
By the way, right now this plugin only works on macOS since I wasn't able to test it on windows or linux. I should probably do that, before releasing.

I am sorry for the problems. This is my first time contributing to a project on GitHub. I think it would make sense to close this pull request and I will open a new one when the plugin is tested on all platforms and ready to be released. Now I know what to look for.

pjeby:
The Readme change looks good.  No need to delete the old release, just add a new one.

No need to apologize, either -- reviews are just part of the process for adding a plugin.  You can close this if you want, but it'd probably be simpler to leave it open.  Cross-platform testing isn't an absolute requirement; you can just note in your README that it's only been tested on macOS, though you can certainly ask on the Obsidian Discord for testers to try it on other platforms.

marcjulianschwarz:
Thanks. Just created the new release with all changes. Updated the README with a warning that this plugin has only been tested on macOS. 
Will work on the support for other operating systems for a future release.

marcjulianschwarz:
This branch now conflicts with the master branch. Do I have to change something?

pjeby:
There should be a button you can click to do it in the browser.  Or you can pull and merge the branch on the command line, then push the result back.

lishid:
Don't worry about the PR conflict, I can resolve it when we're ready to merge after the review.

lishid:
Ok, to me it seems like the biggest surprise about this plugin is that it's linking files **outside** of the vault with `file:///` links.  Perhaps the branding/naming of this plugin should be focused on "external" file linking?

https://github.com/marcjulianschwarz/obsidian-file-link/blob/master/main.js
Please remove this file from the repo and add it to your .gitignore - compiled release files shouldn't get mixed with source code in your commits and should go to the releases files instead.

https://github.com/marcjulianschwarz/obsidian-file-link/blob/dba36be156f81b15d5bfbb031c14d42e7a1f96cb/main.ts#L68
You can more easily create these DOM elements manually using something like `contentEl.createEl('h3')`. That way you'll have references to the elements and won't need to query them out of the DOM again. (Also eliminates the use of innerHTML, which is discouraged).

https://github.com/marcjulianschwarz/obsidian-file-link/blob/dba36be156f81b15d5bfbb031c14d42e7a1f96cb/main.ts#L30
Seems that you should just use an editorCallback instead?

marcjulianschwarz:
I agree, the naming might be a bit confusing. Including the word "external" would maybe make it more clear (obsidian-external-file-link or obsidian-better-external-file-link to make it more precise).  

Removed the main.js file.

Implemented the DOM elements and used the editorCallback now. I am quite new to plugin development, is there a way to better learn things like editorCallback. Didn't even know this exists.

lishid:
> Implemented the DOM elements and used the editorCallback now. I am quite new to plugin development, is there a way to better learn things like editorCallback? Didn't even know this exists.

Not yet - we're still quite behind on plugin documentation in terms of teaching and learning - they exist in the typings file, sometimes with comments explaining what they do, but there isn't yet a guide.

lishid:
https://github.com/marcjulianschwarz/obsidian-file-link/blob/cbe8a74268aa4413cde12a058c38e58635cc9e55/main.ts#L30
I think you only need `editorCallback` not `editorCheckCallback`.

marcjulianschwarz:
Just changed it to `editorCallback`.

marcjulianschwarz:
To release a new version I just have to create a new release on GitHub right? And update manifest and versions files.

marcjulianschwarz:
And one more question. Is there a way to change the name and description shown in the community plugins tab? Thought it would use name and description from the manifest.json file but apparently it doesn't.

Do I have to create a new pull request with updated attributes?

pjeby:
Yes to both questions.

marcjulianschwarz:
Thank you!

PR #455, opened 2021-08-30
Resubmission: Block Reference Counter

shabegom:

Previous Submission: https://github.com/obsidianmd/obsidian-releases/pull/317

All feedback from the previous submission has been addressed in addition to further performance improvements to indexing. Indexing of new references only occurs if:
- it has not indexed recently
- the cache for the given file has changed
- the user is not currently typing into a note

Settings have been added for display of the references and what format the table of references should be. The Search Results table works by:
1. Creating a new search results leaf in the right sidebar
2. Autogenerating the search query for that search
3. Detaching the leaf from the sidebar and re-attaching it to the element within the note
4. Hiding some search specific UI
This is really similar to how the old "Backlinks in Note" plugin was working.

## Repo URL

Link to my plugin: https://github.com/shabegom/obsidian-reference-count

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


shabegom:
I've had one report that the indexing is causing a slowdown. I've updated the cache checking code to only run an index if links, headings, blocks, or embeds have changed to reduce the number of indexes even further. The other thing I could try is increasing the debounce times.

lishid:
https://github.com/shabegom/obsidian-reference-count/blob/3edd294613f61523961a83bbde0e0c22f512432b/src/main.ts#L18
I suspect this lodash import is blowing up your plugin's size (it really shouldn't be anywhere close to 1mb) - maybe just implement isEqual in your own code?

https://github.com/shabegom/obsidian-reference-count/blob/3edd294613f61523961a83bbde0e0c22f512432b/src/main.ts#L68
Move this 2 lines up in case your functions error out.

https://github.com/shabegom/obsidian-reference-count/blob/3edd294613f61523961a83bbde0e0c22f512432b/src/main.ts#L167
This doesn't need to be an async function.

https://github.com/shabegom/obsidian-reference-count/blob/3edd294613f61523961a83bbde0e0c22f512432b/src/main.ts#L186
You can use `workspace.onLayoutReady(callback)` instead.

https://github.com/shabegom/obsidian-reference-count/blob/3edd294613f61523961a83bbde0e0c22f512432b/src/main.ts#L217
This should already give you the view, not the leaf.

https://github.com/shabegom/obsidian-reference-count/blob/3edd294613f61523961a83bbde0e0c22f512432b/src/main.ts#L569
You can just use `parentElement.createEl` to automatically append the child to the parent.

https://github.com/shabegom/obsidian-reference-count/blob/3edd294613f61523961a83bbde0e0c22f512432b/src/main.ts#L612
Just use `debounce` from `obsidian`? (We actually have a slightly more performant version of debounce)

https://github.com/shabegom/obsidian-reference-count/blob/3edd294613f61523961a83bbde0e0c22f512432b/src/indexer.ts#L224
This still isn't the right way. See the original comment from last review:
> You should use Obsidian's parseLinktext function (https://github.com/obsidianmd/obsidian-api/blob/48bd2b91f66aa7d3768629be48c52e5351ec6696/obsidian.d.ts#L2026) which is what we use to split the link. This will give you a linkpath which is the path that will resolve to a note, and a subpath which starts with #.

parseLinktext splits the path at the `#` so if you had a link like `path/filename#hash` it will be split into `path/filename` and `#hash`. You should then check if the hash part `startsWith('#^')` to determine if it's a block ID.

shabegom:
Dearest @lishid,

Thank you for the feedback! I have addressed the notes and updated to 0.1.5.

Without lodash, we're back to a reasonable 136kb!

The one place I still had trouble was using `parseLinkText` as it includes the # and my code was expecting no #. Nothing a quick replace can't fix.

I also learned how long a regex.replace takes when trying to optimize the indexing. If there is some better way to grab a heading with special characters like `Heading | With Line` and make it match the link: `Heading With Line` without doing a `.replace()` please educate me!

Please let me know if there is anything else I can improve.

All the best,

shabegom

lishid:
- I suggest auto-formatting your code to add semicolons - while they are optional, there are cases where they actually do make a subtle difference that would make the code do different things when omitted.

https://github.com/shabegom/obsidian-reference-count/blob/a89712d91c3deb6645f10ac08b2b3f4296c67ae3/src/main.ts#L38
This isn't necessary anymore since you have the onLayoutReady [here](https://github.com/shabegom/obsidian-reference-count/blob/a89712d91c3deb6645f10ac08b2b3f4296c67ae3/src/main.ts#L183) right?



> I also learned how long a regex.replace takes when trying to optimize the indexing. If there is some better way to grab a heading with special characters like `Heading | With Line` and make it match the link: `Heading With Line` without doing a `.replace()` please educate me!

We haven't exposed the function to normalize headings for matching, but it's basically a replace for special char with space, then replace for consecutive whitespace chars `/\s+/g` with a single space.

shabegom:
I'm pretty anti semi-colon, but for you I would do anything.

Removed the extra unload!

PR #459, opened 2021-08-31
Add random structural diary plugin

ShockThunder:
## Repo URL
Link to my plugin: https://github.com/ShockThunder/RandomStructuralDiary

Release Checklist
 - [x] I have tested this on Windows, macOS, and Linux (if applicable)
   - [x] main.js
   - [x] manifest.json
   - [ ] styles.css (optional)

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2132,5 +2132,12 @@
         "description": "When you click an image, it will be displayed in a popup layer and you can view, drag, zoom, rotate the image.",
         "author": "sissilab",
         "repo": "sissilab/obsidian-image-toolkit"
+    },
+    {
+        "id": "random-structural-diary-plugin",
+        "name": "Random Structural Diary",
+        "description": "This is a plugin for picking random questions from prepared question list. It allows you answer on different questions each time.",
+        "author": "Timur Sidoriuk",
+        "repo": "ShockThunder/RandomStructuralDiary"
     }
 ]


lishid:
https://github.com/ShockThunder/RandomStructuralDiary/blob/b4fc1f319474a8a75af11b2b5af7b2a2b4993ddc/main.ts#L17

Please rename your classes/interfaces from the sample names to your plugin's actual name.

https://github.com/ShockThunder/RandomStructuralDiary/blob/b4fc1f319474a8a75af11b2b5af7b2a2b4993ddc/main.ts#L31

Don't do this - instead, use `vault.getAbstractFileByPath(path)` and then check for `file instanceof TFile`. Then you can use `vault.cachedRead` rather than `adapter.read()`. You should also use `async` and `await` instead of `.then` when possible.

Here I don't think you need a `checkCallback` either - you can just convert it into a regular `callback` command.

https://github.com/ShockThunder/RandomStructuralDiary/blob/b4fc1f319474a8a75af11b2b5af7b2a2b4993ddc/main.ts#L84
Make sure you test for null here and check that it's actually a markdown file. You are also completely overwriting this file's content, which may cause data-loss if the user was working on something.

ShockThunder:
Thank you for your review.
I updated the code.
I decided not to create a default file and use one in the plugin folder.

Unfortunately, i couldn't get file in plugin folder by `vault.getAbstractFileByPath(path)`, so, i use the adapter
https://github.com/ShockThunder/RandomStructuralDiary/blob/a0ca0f1c6bc06bd8f105d43db691ba1cc45c8044/main.ts#L37

I changed `checkCallback` to async `callback` and remove `then` in promises.

Also, I save user content and place my generated text after this.

lishid:
> Unfortunately, i couldn't get file in plugin folder by vault.getAbstractFileByPath(path), so, i use the adapter
https://github.com/ShockThunder/RandomStructuralDiary/blob/a0ca0f1c6bc06bd8f105d43db691ba1cc45c8044/main.ts#L37

Sure, files in the config folder (including the plugin folder) doesn't exist in the vault structure, so using the adapter is correct.

PR #460, opened 2021-08-31
Add ghost-fade-focus plugin.

skipadu:

## Repo URL

Link to my plugin: https://github.com/skipadu/obsidian-ghost-fade-focus

## Release Checklist

- [x] I have tested this on  **macOS**  _Windows & Linux not tested_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2132,5 +2132,13 @@
         "description": "When you click an image, it will be displayed in a popup layer and you can view, drag, zoom, rotate the image.",
         "author": "sissilab",
         "repo": "sissilab/obsidian-image-toolkit"
+    },
+    {
+        "id": "ghost-fade-focus",
+        "name": "Ghost Fade Focus",
+        "author": "Sami Korpela",
+        "description": "Focused on the current line, others faded like a ghost!",
+        "repo": "skipadu/obsidian-ghost-fade-focus",
+        "branch": "main"
     }
 ]


lishid:
https://github.com/skipadu/obsidian-ghost-fade-focus/blob/5229cec96ad83a5f22a96579846acd83306504a9/src/main.ts#L33
Please unhook this event when your plugin unloads in the `onunload()` function. You can use the `iterateCodeMirrors` to go through the currently available editors.

lishid:
You should also cleanup everything (class names) when unloading.

skipadu:
I released new version, [`1.0.1`](https://github.com/skipadu/obsidian-ghost-fade-focus/releases/tag/1.0.1), where these issues mentioned should be solved.
Is it okay now?

lishid:
https://github.com/skipadu/obsidian-ghost-fade-focus/blob/1a82808fcf0d7d9fa89a6ebbac86fa17d8ff8bc3/src/main.ts#L26

I think this should be a regular `callback` instead of an `editorCallback`.

https://github.com/skipadu/obsidian-ghost-fade-focus/blob/1a82808fcf0d7d9fa89a6ebbac86fa17d8ff8bc3/src/main.ts#L95

I don't think passing this parameter in is necessary since you can query it in this function from `cm`.

---
Both are small issues, so I'll have it merged now and you can fix it later.

skipadu:
> Both are small issues, so I'll have it merged now and you can fix it later.

Thanks for the findings! I will try to address these issues on next release version.

PR #463, opened 2021-09-01
Add jupyter plugin.

tillahoffmann:

## Repo URL

Link to my plugin: https://github.com/tillahoffmann/obsidian-jupyter

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2185,5 +2185,13 @@
         "description": "Switch input method with fcitx-remote when Vim keymap is enabled.",
         "author": "yuanotes",
         "repo": "yuanotes/obsidian-vim-im-switch-plugin"
+    },
+    {
+        "id": "obsidian-jupyter",
+        "name": "Jupyter",
+        "author": "tillahoffmann",
+        "description": "Run python code in Obsidian using jupyter.",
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/tillahoffmann/obsidian-jupyter/blob/42674745daf57a0ff5586a79bfd57978369c09fa/main.ts#L90
Discouraged use of `var`. There's a couple more in the code.

https://github.com/tillahoffmann/obsidian-jupyter/blob/42674745daf57a0ff5586a79bfd57978369c09fa/main.ts#L92
These can be set like this:
```ts
el.createEl('button', {cls: 'css-classes', text: 'inner text'});
```
https://github.com/tillahoffmann/obsidian-jupyter/blob/42674745daf57a0ff5586a79bfd57978369c09fa/main.ts#L99
Recommend using `async` and `await` here.

https://github.com/tillahoffmann/obsidian-jupyter/blob/42674745daf57a0ff5586a79bfd57978369c09fa/main.ts#L130
Recommend checking for the type using `instanceof` rather than casting, because it might be possible to have a different adapter here in the future.

PR #464, opened 2021-09-02
adding Obsidian HackerNews

arpitbbhayani:
Periodically fetches and displays top stories from HackerNews.

## Repo URL

Link to my plugin: https://github.com/arpitbbhayani/obsidian-hackernews

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2185,5 +2185,13 @@
         "description": "Switch input method with fcitx-remote when Vim keymap is enabled.",
         "author": "yuanotes",
         "repo": "yuanotes/obsidian-vim-im-switch-plugin"
+    },
+    {
+        "id": "obsidian-hackernews",
+        "name": "HackerNews",
+        "author": "arpitbbhayani",
+        "description": "Periodically fetches and displays top stories from HackerNews.",
+        "repo": "arpitbbhayani/obsidian-hackernews",
+        "branch": "master"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2187,14 +2187,6 @@
         "repo": "yuanotes/obsidian-vim-im-switch-plugin"
     },
     {
-        "id": "obsidian-hackernews",
-        "name": "HackerNews",
-        "author": "arpitbbhayani",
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/arpitbbhayani/obsidian-hackernews/blob/d03d91fe86fa59120d43483b2b30cfdb00116806/src/main.ts#L20
This `Promise.all` doesn't seem necessary.

https://github.com/arpitbbhayani/obsidian-hackernews/blob/d03d91fe86fa59120d43483b2b30cfdb00116806/src/main.ts#L53
Usually the procedure here is to `for(let leaf of getLeavesOfType()) { leaf.detach() }`.

https://github.com/arpitbbhayani/obsidian-hackernews/blob/d03d91fe86fa59120d43483b2b30cfdb00116806/src/apiManager.ts#L3
Should import `moment` from `obsidian` so you don't package another copy in your plugin.

https://github.com/arpitbbhayani/obsidian-hackernews/blob/d03d91fe86fa59120d43483b2b30cfdb00116806/src/apiManager.ts#L2
Don't use `path` since it's only available on Electron. Instead, just join with `/` and then use `normalizePath` from `obsidian`.

https://github.com/arpitbbhayani/obsidian-hackernews/blob/d03d91fe86fa59120d43483b2b30cfdb00116806/src/apiManager.ts#L38

Could probably put vault in a variable here to avoid the duplicated chain.

arpitbbhayani:
Thank you @lishid for reviewing. I have released v 1.0.4 incorporating all the mentioned changes

 - removed unnecessary `Promise.all`
 - detaching all the leaves instead of just one
 - using Obsidian's `moment` instead of importing the Moment package
 - Have joined the directory and filename instead of using `path`
 - have put `vault` in a separate variable to avoid duplicated chains.

I have also tested the code through various flows and everything is working fine.

lishid:
https://github.com/arpitbbhayani/obsidian-hackernews/blob/2c242de3c92cf0f70b7ada93ad275115242d7672/src/apiManager.ts#L2
While this works, you should probably just add `moment` to the import list above.

PR #467, opened 2021-09-05
Add Stille

michaellee:

## Repo URL

Link to my plugin: https://github.com/michaellee/stille

## Release Checklist

- [ ] I have tested this on Windows, macOS, and Linux _(if applicable)_ [I have only tested on macOS, as I don't have access to Linux or Windows]
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/michaellee/stille/blob/76a27cd33317c9d33554c55ebf1fbf8afb4fa19d/main.ts#L48
Should remove this.

https://github.com/michaellee/stille/blob/76a27cd33317c9d33554c55ebf1fbf8afb4fa19d/main.ts#L58
Don't save settings when unloading - only do it when the user change those settings.

https://github.com/michaellee/stille/blob/76a27cd33317c9d33554c55ebf1fbf8afb4fa19d/main.ts#L106
`addStyleToView` already calls `updateStyles`.

https://github.com/michaellee/stille/blob/76a27cd33317c9d33554c55ebf1fbf8afb4fa19d/main.ts#L91
Should also set `this.styleElement` to null.

https://github.com/michaellee/stille/blob/76a27cd33317c9d33554c55ebf1fbf8afb4fa19d/main.ts#L4
This doesn't seem used - you're using the other identically named class property on the plugin.

https://github.com/michaellee/stille/blob/76a27cd33317c9d33554c55ebf1fbf8afb4fa19d/main.ts#L133
Maybe want to use a slider here instead?

https://github.com/michaellee/stille/blob/76a27cd33317c9d33554c55ebf1fbf8afb4fa19d/main.ts#L99
Missing setting toggle to change this value? Maybe it's intended?

michaellee:
Hey @lishid thank you so much for reviewing my code! I've made some adjustments based on your feedback and tagged version 1.0.1 on the repo with the compiled binary files https://github.com/michaellee/stille/releases/tag/1.0.1

RE: Slider — do you have any examples of a slider? I actually prefer text inputs for floating point values but am willing to test out the UX change.

RE — Missing setting toggle, this was a feature I was on the fence about with version 1.0. Since it isn't currently implemented, I decided to remove it altogether and will revisit if Obsidian users ask for it.

Thank you so much again for your help 👍 

Please let me know if there's anything else I can do to prepare my plugin for it to be included as a community plugin.

lishid:
> RE: Slider — do you have any examples of a slider? I actually prefer text inputs for floating point values but am willing to test out the UX change.

Use `addSlider` instead of `addText`. Sliders have these functions you can use: https://github.com/obsidianmd/obsidian-api/blob/bde556afa033e909ebfb9fcee8f5ef288276f78f/obsidian.d.ts#L2610

You'll most likely want to set a min/max and step, then set the current value and set dynamic tooltip (a tooltip pops up when the user drags it).

> RE — Missing setting toggle, this was a feature I was on the fence about with version 1.0. Since it isn't currently implemented, I decided to remove it altogether and will revisit if Obsidian users ask for it.

Sounds good. Let me know if you want me to merge now or when you make the slider change, I don't mind either way.

michaellee:
@lishid thanks for the reference!

I gave it a shot and I don't think it is the correct UI/UX for what I'd like to achieve. For this reason, I'm going to stick with the text input.

If you could merge now, that would be great! Thanks again for all your help!

PR #471, opened 2021-09-06
Add "Podcast Note" plugin to community list

marcjulianschwarz:

## Repo URL

Link to my plugin: https://github.com/marcjulianschwarz/obsidian-podcast-note

## Release Checklist

- [x] I have tested this on macOS (but it should work on windows and linux too, will test for future release) _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/marcjulianschwarz/obsidian-podcast-note/blob/cf696a62747ae2f901adb76475f70b5560dda70a/main.ts#L104
Suggest making use of `async` `await` here (and try-catch), instead of .then/.catch.

https://github.com/marcjulianschwarz/obsidian-podcast-note/blob/cf696a62747ae2f901adb76475f70b5560dda70a/main.ts#L146
Don't do this yourself (this also doesn't work on mobile since NodeJS does not exist there). Use either `XMLHttpRequest`, `fetch`, or Obsidian's provided `request` API which also ignores CORS restrictions.

https://github.com/marcjulianschwarz/obsidian-podcast-note/blob/cf696a62747ae2f901adb76475f70b5560dda70a/main.ts#L226
`mdView` may be null - make sure you check for this.

https://github.com/marcjulianschwarz/obsidian-podcast-note/blob/cf696a62747ae2f901adb76475f70b5560dda70a/main.ts#L231
I think best to add a `/` between folder and filename?

https://github.com/marcjulianschwarz/obsidian-podcast-note/blob/cf696a62747ae2f901adb76475f70b5560dda70a/main.ts#L230
Duplicate replace for `:`, also consider using a regex instead? Perhaps something like `.replace(/[\/\\:]/g, '')`?

lishid:
https://github.com/marcjulianschwarz/obsidian-podcast-note/blob/cf696a62747ae2f901adb76475f70b5560dda70a/main.ts#L220
Discouraged use of `var`

https://github.com/marcjulianschwarz/obsidian-podcast-note/blob/cf696a62747ae2f901adb76475f70b5560dda70a/main.ts#L174
You can use `import { moment } from 'obsidian'` instead of doing this yourself.

marcjulianschwarz:
Thanks for all suggestions and improvements.
Just implemented and pushed everything you suggested.

lishid:
> https://github.com/marcjulianschwarz/obsidian-podcast-note/blob/cf696a62747ae2f901adb76475f70b5560dda70a/main.ts#L226
mdView may be null - make sure you check for this.

You added a check in `getEditor` but your other functions that call `getEditor` doesn't check for undefined returns 😅

lishid:
Not a huge deal, so I have it merged. Still recommend fixing it though 👍

marcjulianschwarz:
Ups... forgot to do that. Will fix it. Thanks for merging.

marcjulianschwarz:
If mdView is null, does this mean the editor is not in focus or open?

lishid:
There's a couple of possibilities - the user could have a non-md view open (like Calendar, image, video, or pdf), or they could have closed all panes.

PR #479, opened 2021-09-14
Shortening obsidian-pocket name

nybbles:
I would like to change the human-readable name of the `obsidian-pocket` plugin, based on user feedback from https://github.com/nybbles/obsidian-pocket/issues/49.


ericaxu:
Sure, thanks for the PR!

PR #480, opened 2021-09-15
Add Workspaces Plus Plugin

nothingislost:

This plugin expands the functionality of the workspaces feature:
        - Status bar indicator shows the currently active workspace name
        - Workspace picker pop up when clicking on the status bar indicator
        - Workspace picker and modal with ability to create, delete, rename, and switch workspaces
        - Plugin option: auto-save workspace on switch

## Repo URL

Link to my plugin: https://github.com/nothingislost/obsidian-workspaces-plus

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/nothingislost/obsidian-workspaces-plus/blob/df8a70145fd186066a41e39d101bc3d48c59bb87/src/main.ts#L22
Variable names should be `camelCase`.

https://github.com/nothingislost/obsidian-workspaces-plus/blob/df8a70145fd186066a41e39d101bc3d48c59bb87/src/main.ts#L69
This is a very expensive operation as it asks every pane to serialize its state to a js object. Don't do this frequently, or make sure you have the proper debounces. Also, the usage of `var` is discouraged.

nothingislost:
@lishid thanks for the feedback. Agreed on the workspace modification indicator. I've removed it completely as it wasn't providing much value and wasn't very accurate.

The vars and variable names should be resolved as of https://github.com/nothingislost/obsidian-workspaces-plus/commit/4e6339f742dda3743a0b9f75a287a67efd74bf8b

nothingislost:
https://github.com/nothingislost/obsidian-workspaces-plus/tree/0.2.0 should be a bit easier to read through now. refactored things into smaller functions and removed the ts-ignores and other bad practices.

lishid:
https://github.com/nothingislost/obsidian-workspaces-plus/blob/89233bc5342d7cb0c4c65598fce01fd94a885915/src/main.ts#L12
Here you should use the recommended way to load settings: `Object.assign({}, DEFAULT_CONFIG, await this.loadData());`

Two issues with the way you're currently doing it: 1. loadData gives you a plain object, whereas you're initializing a class object instance, which has different types. 2. loading a previously saved object will not have properties you add later on to your plugin, so you will end up with undefined as the default for any new properties you add, rather than the default one.

https://github.com/nothingislost/obsidian-workspaces-plus/blob/89233bc5342d7cb0c4c65598fce01fd94a885915/src/main.ts#L15
I don't understand how saveOnSwitch can both be truthy and triple-equals to undefined.

nothingislost:
@lishid the settings logic has been addressed here https://github.com/nothingislost/obsidian-workspaces-plus/commit/ea685e5120bf9d236374aa4a10c311daa05b98ab

The saveOnSwitch concern looks to be a misread. Those are two different variables saveOnSwitch and saveOnSave. I was migrating a setting over to a new name and trying to retain the user's previous setting since the two features are functionally similar. I've removed this logic completely since it is no longer needed.

lishid:
> The saveOnSwitch concern looks to be a misread. Those are two different variables saveOnSwitch and saveOnSave. I was migrating a setting over to a new name and trying to retain the user's previous setting since the two features are functionally similar. I've removed this logic completely since it is no longer needed.

Ohhh yeah haha my bad!

PR #485, opened 2021-09-18
Add obsidian-audio-speed-plugin

kaizau:

## Repo URL

Link to my plugin: https://github.com/kaizau/obsidian-audio-speed-plugin

## Release Checklist

- [ ] I have tested this on
  - [ ] Windows
  - [x] macOS
  - [ ] Linux _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
  - [x] ~~`styles.css`~~ _(optional)_

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2354,5 +2354,13 @@
         "description": "Move completed tasks to an archive with a date tree",
         "author": "ivan-lednev",
         "repo": "ivan-lednev/obsidian-task-archiver"
+    },
+    {
+        "id": "obsidian-audio-speed-plugin",
+        "name": "Audio Speed Plugin",
+        "description": "Change the playback rate of audio files during markdown preview.",
+        "author": "kaizau",
+        "repo": "kaizau/obsidian-audio-speed-plugin",
+        "branch": "main"
     }
 ]


lishid:
https://github.com/kaizau/obsidian-audio-speed-plugin/blob/4e8fbbd50444b7d077d7cf1f58b9fdc0114c92c7/main.ts#L30
Perhaps just try to find `audio` elements in the embed instead of creating a new one?

https://github.com/kaizau/obsidian-audio-speed-plugin/blob/4e8fbbd50444b7d077d7cf1f58b9fdc0114c92c7/main.ts#L79
Consider using a slider instead of a text input.

kaizau:
@lishid Thanks for the feedback. Looking for the `<audio>` was actually my very initial attempt. 

At the point in time when the markdown post processor runs, no `<audio>` elements are in the DOM. Instead, `<span class="internal-embed" src="...">filename.mp3</span>` was the closest thing I could find.

Is there a different, later stage "hook" into the post processing pipeline? Or some kind of `ready` event to listen for?

lishid:
Ah you're right, embeds are processed after the regular post processors run. That's fine then - you can do what you're doing now, or maybe look for the `internal-embed` class and add a timeout to check again soon.

kaizau:
@lishid Updated: https://github.com/kaizau/obsidian-audio-speed-plugin/blob/main/main.ts#L27

- Now using 500ms timeout instead of recreating the audio elements.
- Adds a slider for setting playback speed.

PR #486, opened 2021-09-18
Add initial version of ReadItLater

DominikPieper:

## Repo URL

Link to my plugin: https://github.com/DominikPieper/obsidian-ReadItLater

## Release Checklist

- [ x ] I have tested this on Windows, macOS, and Linux _(if applicable)_
- [ x ] My GitHub release contains all required files
  - [ x ] `main.js`
  - [ x ] `manifest.json`
  - [ x ] `styles.css` _(optional)_
- [ x ] GitHub release name matches the exact version number specified in my manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
- [ x ] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [ x ] README clearly describes the plugins purpose and provides clear usage instructions.
- [ x ] I have added a license in the LICENSE file.


lishid:
https://github.com/DominikPieper/obsidian-ReadItLater/blob/422f3ae8bee47ad905fc861041d9e00775c14b21/src/main.ts#L11
Should normalize the mkdir as well. Honestly though, I don't think it's a good idea to do this on plugin enable because you'll be attempting to do this every time the user open the app/vault.

https://github.com/DominikPieper/obsidian-ReadItLater/blob/422f3ae8bee47ad905fc861041d9e00775c14b21/src/converter.ts#L3
Don't import moment from moment, import it from obsidian so you don't end up bunding a whole copy of moment.js into your plugin.

https://github.com/DominikPieper/obsidian-ReadItLater/blob/422f3ae8bee47ad905fc861041d9e00775c14b21/src/converter.ts#L7
I think you can probably move all this code into your plugin, so you don't end up creating new objects (of ContentConverter) every time the command is executed.

DominikPieper:
@lishid thanks for the review. I refactored my code and for the last point, I'm not sure where's a better place? Or maybe just a Notice for the user?

lishid:
> @lishid thanks for the review. I refactored my code and for the last point, I'm not sure where's a better place? Or maybe just a Notice for the user?

Yeah I think you can either create the directory when writing the file (if it doesn't exist), or notify the user that the inbox directory doesn't exist and have them create it.

DominikPieper:
Okay, I removed the check from `onLoad` and added a notice for the user

PR #487, opened 2021-09-19
Addition of plugin Obsidian42 - Text Transporter

TfTHacker:

## Repo URL

Link to my plugin: https://github.com/TfTHacker/obsidian-spaced-repetition

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2354,5 +2354,12 @@
         "description": "Move completed tasks to an archive with a date tree",
         "author": "ivan-lednev",
         "repo": "ivan-lednev/obsidian-task-archiver"
+    },
+    {
+        "id": "obsidian42-text-transporter",
+        "name": "Obsidian42 - Text Transporter",
+        "description": "Advanced text tools for working with text in your vault",
+        "author": "TfTHacker",
+        "repo": "https://github.com/TfTHacker/obsidian-text-transporter",
     }
 ]


TfTHacker:
Sorry, looks like I have a trailing comma in my submission.

Also this plugin is doing crazy text changes, so will not be easy to follow :-)

nothingislost:
Just a heads up, I believe you have the wrong link in your PR description. It should be https://github.com/TfTHacker/obsidian42-text-transporter

TfTHacker:
> Just a heads up, I believe you have the wrong link in your PR description. It should be https://github.com/TfTHacker/obsidian42-text-transporter

Yes you are right. thank you. I have now corrected this.

lishid:
https://github.com/TfTHacker/obsidian42-text-transporter/blob/5fcef58d775d1d277fd8a33e4c0175490fb588d3/src/main.ts#L2
Class/interface names should be PascalCase, whereas function names should be camelCase. There's a lot of classes to fix though... 

https://github.com/TfTHacker/obsidian42-text-transporter/blob/5fcef58d775d1d277fd8a33e4c0175490fb588d3/src/utils/fileSystem.ts#L27
Don't suggest using `list` directly, since that will report what the OS sees, which is slightly different than what Obsidian sees (as noted by your tests for dotfiles). Instead, you can use `TFolder.children` and recursively scan the folder tree. There's a helper function called `Vault.recurseChildren` that you can use like this:

```ts
Vault.recurseChildren(vault.getRoot(), (file: TAbstractFile) => {
  if (file instanceof TFile) {
    // Maybe also check for (file.extension === 'md')
    ...
  }
  else if (file instanceof TFolder) {
    ...
  }
});
```

https://github.com/TfTHacker/obsidian42-text-transporter/blob/5fcef58d775d1d277fd8a33e4c0175490fb588d3/src/utils/transporterFunctions.ts#L10
activeLeaf can be null, which will crash the function. Since you're using MarkdownView anyway, might as well just use `getActiveViewOfType(MarkdownView)`. That way, you won't need to `// @ts-ignore` in the next few lines.

Actually, `getContextObjects` is a very particularly interesting (and problematic) function here. It's using `this` but `this` doesn't refer to anything here, so it's falling back to `window`. Which means you're using `window.app`, our debug app reference. Don't use it as it's not guaranteed to always exist. Instead, you should use the app reference from your plugin, which means you need to pass the `plugin` reference as a parameter.

What I don't like is that you're calling this from various "helper" functions, and some functionality of your plugin will end up calling this function multiple times just to access a single `Editor` reference, so you end up doing a lot of wasted or duplicated work. What I would suggest is to gather all this at the beginning of each command, and pass it to your helper functions. I also think your helper functions are a bit all over the place; some of them are root-level functions that take a `plugin` as parameter, some of them exists on an object like FileSystem

https://github.com/TfTHacker/obsidian42-text-transporter/blob/5fcef58d775d1d277fd8a33e4c0175490fb588d3/src/utils/transporterFunctions.ts#L29
You can do all of this with a single `.replace(/\[\]#\|:/g, '')` regex replace.

https://github.com/TfTHacker/obsidian42-text-transporter/blob/5fcef58d775d1d277fd8a33e4c0175490fb588d3/src/utils/transporterFunctions.ts#L120
Missing await.

https://github.com/TfTHacker/obsidian42-text-transporter/blob/5fcef58d775d1d277fd8a33e4c0175490fb588d3/src/utils/transporterFunctions.ts#L129
Damn, do you have to do this on all your functions? Most IDEs provide visuals without requiring you to manually write it, and it'd make a pain for function refactor/renames.

> copyOrPushLineOrSelectionToNewLocationUsingCurrentCursorLocationAndBoomark

Also I think your function names are too verbose - function names this long is usually a sign that your function is way too specific in what it does, or you're trying to explain too much what the function does, or the function does too much and needs to be split up. Remember you can write comments for what the function does in a standard way using `JSDoc` (or TSDoc) which looks like this: https://github.com/obsidianmd/obsidian-api/blob/5af3ef982328a2dd0a2f5242f3c74e2f45e07896/obsidian.d.ts#L234

lishid:
This is quite a large plugin, and all the usage of obsidian's API is pretty much all over the place, so I probably missed a lot of things when skimming. I would consider possibly putting some thoughts into re-structuring the helper functions and utils so they're easier to read/understand.

lishid:
> So first Licat, thank you for the time you put into this. It is a large plugin, doing a lot of things. I made the code corrections and simplifications as suggested. I did leave some of the long names, but those are just a few commands. So I left at least that since they help me know what the function is doing.

Yeah that's fine. I just don't see long function names very often 😅 

> By the way, probably obvious, I am not a TypeScript developer :-)

Well now you are! Go ahead and be proud 🥂 

> So would appreciate your further observations.

I've likely missed things since this is pretty big, but I think overall if it all works fine in your testing we can go ahead and merge.

PR #490, opened 2021-09-21
add marginnote-companion

aidenlx:

## Repo URL

Link to my plugin: <https://github.com/aidenlx/marginnote-companion>

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/aidenlx/marginnote-companion/blob/8f21e16a2d2b2f371c8f568bce665c688c960215/src/mn-main.ts#L41

This is problematic because you will keep all codemirror instances in memory until your plugin is unloaded, which can cause memory leaks. Instead, you can use `workspace.iterateCodeMirrors(callback)` in your `onunload` function to un-register these event handlers, which only goes through currently active CodeMirror instances.

https://github.com/aidenlx/marginnote-companion/blob/8f21e16a2d2b2f371c8f568bce665c688c960215/src/receivers/input-handler.ts#L158
Remember to call this when your plugin unloads, just in case it wasn't stopped.

PR #494, opened 2021-09-24
Added new community plugin - Quick Latex for Obsidian

joeyuping:

## Repo URL

Link to my plugin: https://github.com/joeyuping/quick_latex_obsidian

## Release Checklist

- I have tested this on:
  - [X]  Windows
  - [x] macOS
  - [x] Linux
  - [X] `main.js`
  - [X] `manifest.json`


lishid:
https://github.com/joeyuping/quick_latex_obsidian/blob/d4bf584c781e0b7413d376622df02092567fa8b3/main.ts#L14
Don't keep this list as it will cause a memory leak. Instead, you can use `workspace.iterateCodeMirrors` to go through all currently active CodeMirror instances in your plugin onunload.

https://github.com/joeyuping/quick_latex_obsidian/blob/d4bf584c781e0b7413d376622df02092567fa8b3/main.ts#L37
I feel like these should be editor commands rather than a direct CodeMirror hook.

https://github.com/joeyuping/quick_latex_obsidian/blob/d4bf584c781e0b7413d376622df02092567fa8b3/main.ts#L88
https://github.com/joeyuping/quick_latex_obsidian/blob/d4bf584c781e0b7413d376622df02092567fa8b3/main.ts#L128
activeLeaf may be null causing a crash here - use `workspace.getActiveViewOfType(MarkdownView)` and check for null before continuing.

joeyuping:
@lishid Thank you so much for the kind feedbacks!

I have incorporated your suggestions, and also added a settings page.  New files have been published in the same repository as 1.3.0.

PR #495, opened 2021-09-25
Add Move Folder plugin

piotryordanov:

## Repo URL

Link to my plugin:

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2389,5 +2389,12 @@
         "description": "Quickly switch and manage workspaces",
         "author": "NothingIsLost",
         "repo": "nothingislost/obsidian-workspaces-plus"
+    },
+    {
+        "id": "move-folder-plugin",
+	    "name": "Move Folder",
+	    "description": "This plugin allows you to move a folder into another location",
+	    "author": "Piotr Yordanov",
+        "repo": "piotryordanov/obsidian-move-folder"
     }
 ]


pjeby:
You did not include a link to your repository in your submission.

You have also not actually published a Github *release* of your plugin, or even tagged it for release.  (The repo page shows zero tags and zero releases.)

Next, your README file does not clearly describe the purpose or provide clear usage instructions.  (A video is not usage instructions, and I had to watch it several times to get even a bare idea of what was going on; the README doesn't even say that it adds a *command* to move folders, or that it operates on the folder containing the active file, just that it "allows" you to move "a folder".)  And even after reading some of the code, I'm still not sure what it means to "go over parent directories beyond the current directory".

The plugin is called "Move Folder", but in actuality it *copies* the folder, and optionally deletes it.  This is an important distinction because copying and deletion *will not update Obsidian links*, so a user who takes the plugin name at face value may end up breaking links in their repository due to the lack of movement.  (It will also disrupt the operation of other plugins that track the location of notes, since they will see new notes and deleted notes, but not any *moved* notes.)

Finally, your plugin itself is using `fs` and `path` operations instead of Vault methods: this means it will not work on Obsidian mobile, and even if it were doing a rename instead of copy, Obsidian would see this as the external creation and deletion of files, again disrupting the operation of plugins that watch for rename operations inside Obsidian.

If the intention is to be a copy-and-maybe-delete operation, then instead of working with `fs`  and `path`, you should implement this plugin's functionality *entirely* using the Obsidian Vault's `createFolder()`, `getAbstractFileByPath()`, `copy()`, and `trash()` methods, which would make it compatible with Obsidian mobile.

However, you should probably decide whether your plugin is really intended to move folders, or copy them, and make sure the name matches what it actually does.  Copying and moving are entirely different operations in Obsidian, and you should not claim your plugin "moves" notes if it actually copies them, as users can and will experience corruption of links or loss of other plugin functionality if they assume this works the same as normal folder moves inside Obsidian.

If the intent is to provide an actual "move folder" operation, you should use the FileManager `renameFile()` method, passing in the folder object to be moved and its desired new path in the vault.  For example, `app.fileManager.renameFile(activeFile.parent, destinationFolder.path + "/" + activeFile.parent.name)` would move the current note's enclosing folder to just inside `destinationFolder`, *and* update any Obsidian links to the file(s) being moved, as well as notifying other plugins that the move happened, so they can update anything they need to as well.

For the convenience of future reviewers, please edit the issue description to include a link to your plugin repository.  You will also need to make a Github release with the proper files, in order for anyone to be able to install or test your plugin.

piotryordanov:
Thank you so much for taking the time to give me such an in depth review. 

I will make the necessary changes then make a pull request.

pjeby:
You do not need to make a new pull request, just update the description of this one to link to your plugin, and add a comment when it is ready for the next review.

ericaxu:

PR #496, opened 2021-09-26
Add obsidian-plotly plugin.

Dmitriy-Shulha:
I am submitting new Community Plugin.

## Repo URL

Link to my plugin:
https://github.com/Dmitriy-Shulha/obsidian-plotly

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


kzhovn:
Just a suggestion, speaking as a potential user rather than a code reviewer: functions like what your plugin adds to the right click menu are usually registered as a command instead, which means it shows up in the command palette and can be assigned a hotkey.

Dmitriy-Shulha:
Thank you for a suggestion. 
Any ideas how can I use same function for both right click menu and command? Struggle to get editor instance from addCommand.
Also, is there any way to remove my button from right click menu on plugin unload?

Dmitriy-Shulha:
Actually, regarding second question, I figured it out - it looks like Obsidian automatically removes buttons introduced by `this.registerEvent()`. It just my local build had some errors and plugin was not unloading properly, what resulted in button duplicates on each plugin off/on toggle.
Still searching for an answer to first question though.

kzhovn:
> Any ideas how can I use same function for both right click menu and command? 

Factor out your onClick logic into something like `makePlot(editor: Editor)`, and then you can just add the command with something like

```
this.addCommand({
	id: 'make-plotly',
	name: 'Plotly',
	editorCallback: (editor: Editor) => { //editorCallback only shows the command if and editor is active
		this.makePlot(editor);
	}
});
```

You might want to take a look at Marcus' [docs for commands](https://marcus.se.net/obsidian-plugin-docs/guides/commands).

Dmitriy-Shulha:
Thank you very much!
I will release this improvement in 0.0.5

lishid:
> Any ideas how can I use same function for both right click menu and command? Struggle to get editor instance from addCommand.
Also, is there any way to remove my button from right click menu on plugin unload?

Actually, I don't recommend adding your plugin function to the right click menu at all - commands are much more suitable. Right click menu is called a "context menu" because it reacts to the context. Right clicking on different things gives you actions depending on the context, and the menu space is fairly precious (for those having a dozen of plugins enabled, the right click menu might get bloated). Commands on the other hand, you can have hundreds and there are tools like the command palette and hotkeys to facilitate triggering commands.

https://github.com/Dmitriy-Shulha/obsidian-plotly/blob/6d597a50882641afc986f3f3f56cf6df32f01203/src/main.ts#L20
Please remove any code from the sample plugin that you don't use.

https://github.com/Dmitriy-Shulha/obsidian-plotly/blob/6d597a50882641afc986f3f3f56cf6df32f01203/src/preprocessor.ts#L7
Pretty sure parseYaml is synchronous, so you don't need the await. Your function also don't need to be async either.

https://github.com/Dmitriy-Shulha/obsidian-plotly/blob/6d597a50882641afc986f3f3f56cf6df32f01203/src/ui/modal.ts
Doesn't seem used.

https://github.com/Dmitriy-Shulha/obsidian-plotly/blob/6d597a50882641afc986f3f3f56cf6df32f01203/src/ui/settingsTab.ts
Doesn't seem used either.

https://github.com/Dmitriy-Shulha/obsidian-plotly/blob/6d597a50882641afc986f3f3f56cf6df32f01203/src/show-hint.js
What's this used for?

Dmitriy-Shulha:
I wanted to put Plotly logo somewhere, and context menu item was best option for that :)
However, I agree with your arguments and remove this item.

Show Hint is an addon for CodeMirror I intend to use for future code completion feature. 
For now it is not in use, so I remove it with all other unused and unnecessary code.

All this changes are already in 0.0.5 release.

Dmitriy-Shulha:
Vulnerability fixed.

PR #499, opened 2021-09-27
Adds obsidian-itinerary plugin allowing users to render calendars from event information found in notes

coddingtonbear:
This plugin adds support for rendering calendars of various types from event information you can store in-line in your notes.

## Repo URL

Link to my plugin: https://github.com/coddingtonbear/obsidian-itinerary

## Release Checklist

  - [x] Linux
  - [x] Windows
  - [ ] MacOS (but I can't imagine why this wouldn't work) 
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/coddingtonbear/obsidian-itinerary/blob/f8851f8fdb02b62d6da1695b092eb34be3c94102/src/main.ts#L30
Instead of `any` you should keep the TAbstractFile, and check for `file instanceof TFile` before proceeding, since this could be a folder.

https://github.com/coddingtonbear/obsidian-itinerary/blob/f8851f8fdb02b62d6da1695b092eb34be3c94102/src/main.ts#L37
Prefer using async-await instead of .then.

https://github.com/coddingtonbear/obsidian-itinerary/blob/f8851f8fdb02b62d6da1695b092eb34be3c94102/src/main.ts#L85
I recommend using `vault` instead of `adapter` here - `let file = vault.getAbstractFileByPath(path)`, `file instanceof TFile`, and `vault.cachedRead(file)`. Using these three would give you better performance, as well as better isolation for when we swap vault/adapter implementations for example for mobile.

https://github.com/coddingtonbear/obsidian-itinerary/blob/f8851f8fdb02b62d6da1695b092eb34be3c94102/src/render.ts#L10
Not sure if it's possible, but we do have Moment.JS built into Obsidian you can import directly from `obsidian`, which could reduce your plugin's bundle size and eliminate a lot of duplicated functionality for time manipulations.

coddingtonbear:
Awesome, thanks for the feedback @lishid!

For your first three points -- I've applied all of the changes you'e suggested.  I'm a little embarassed that I hadn't even considered that `onFileChange` could be marked async.  All fixed now, though, I believe.

On the last point re: Luxon.  I actually need Luxon specifically for its timezone support -- moment doesn't provide timezone support directly, and unfortunately timezones are an unfortunately important thing for an itinerary.  I have to say, though, that I hadn't until moments ago actually documented any of that timezone support yet, and hadn't fully implemented timezone support on the event parsing side.  I've just now added an implementation of that timezone support via  (https://github.com/coddingtonbear/obsidian-itinerary/commit/31339d0b5188bfc26e02b98da80b3c2b7638d510) that'll hopefully justify for you my using it instead of the `moment` that's available via Obsidian.

Side note, though -- I'm a little surprised you'd want to encourage folks to use undocumented internal dependencies; won't that limit the Obsidian team's flexibility with regard to being able to change those internals in the future?

lishid:
> I actually need Luxon specifically for its timezone support

Ah you're right, too bad then!

> Side note, though -- I'm a little surprised you'd want to encourage folks to use undocumented internal dependencies; won't that limit the Obsidian team's flexibility with regard to being able to change those internals in the future?

You mean moment.js? We actually expose it through out API. In the worst case we can always mark it as deprecated and provide a new API for migration in the future.

PR #502, opened 2021-09-27
Add URI Commands plugin

kzhovn:

## Repo URL

Link to my plugin: https://github.com/kzhovn/uri-commands-obsidian

## Release Checklist

- [x] I have tested this on ~~Windows, macOS, and~~ Linux _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


pjeby:
Very nice!  I'm planning to install this so I can replace my QuickAdd macro for looking up books in Calibre.  (`calibre://search/_?q={{selection}}`.)

May I suggest perhaps using `{{basename}}` as the insert for the basename, since it doesn't contain the file extension or the path?  There isn't anything in the README that otherwise indicates the name doesn't include the path or extension.

Actually, it might also be good to indicate somewhere that the placeholders are escaped, so that people should not expect to e.g. insert slashes into a path from the selection.

The only (very very minor) code issue I found was that in main.ts, you have a lot of async functions/methods that have no reason to be async, as they do not contain any await operations:

* `addURICommand()` doesn't await anything, so doesn't need to be async
* The non-placeholder `callback:` doesn't await anything, so doesn't need to be async
* `addCommands()` only calls `addURICommand()`, and since the latter doesn't need to be async, nor does the foreach callback need to be async or await.  (And if they did, it wouldn't work right because the forEach would start them all in parallel and then proceed without waiting for any of them to finish.)
* `addFeatherIcons()` doesn't await anything, so doesn't need to be async

A function only needs to be async if it awaits something or needs to return a promise.  This requirement does *not* cascade down to enclosed functions, unless they too need to wait for something to finish.  Your editorCallback only has to be async because it may need to wait for a vault read, but if it didn't then it too would not need to be async.  But even though it's async, it doesn't mean the function that defines the editorCallback needs to be async -- it's just defining the function, not calling it.

Likewise, just because a function is async, doesn't mean it can't call non-async functions: it's okay for `onload()` to call the other methods (aside from `loadSettings()`) without using `await`, so long as they're either synchronous or it doesn't need them to finish before continuing.

kzhovn:
I'm glad you like it! Do you mind if I add your usecase to the examples in the README?


I find {{basename}} unintuitive and expect most people who haven't encountered it in the API would too. Do you think {{fileTitle}} would be better than {{fileName}}? 

Updated the README, hope it's less ambiguous now.

Thanks for the async/await explanation, I'm unfortunately not very comfortable with them but that clarified a couple things for me! Updated.

pjeby:
You still have an `(async command =>` in `addCommands()`, by the way.  This will make it run *later*, after the enclosing function returns, which is probably not what you want.

(It's important to understand that async functions do not run when they are called!  They run *later*.  Calling an async function puts it on a to-do list that will be done later.  `await` stops the current function running until the promise is resolved, at which point that function is put on the to-do list to resume running.  This means that `foreach`, `map`, etc. don't work right with async functions or promises: foreach will just put a bunch of async functions on a to-do list, and then blindly continue without waiting for any of them to start, let alone finish.)

The simplest rules for async/await are:

* If you are calling something that returns a Promise, and you don't want to use `then()` callbacks, you need to use `await` to either get the actual return value, or to wait for that called function/method to *finish* (succeed or throw an error).
* If the thing you're calling returns a Promise, and you don't *need* to know when it finishes, get a result, or trap errors, you don't need to `await` the promise!
* If you use `await` in a function, you need to make the *immediately* enclosing function `async`, which will make *it* return a Promise
* Go back to step 1 if needed, to make any function that calls your function do an await, if needed

So there is no need to make everything async/await, and if you do you will introduce all kinds of subtle errors into things.  Only use them with things that have to *wait* for something else to happen to finish.

Regarding fileTitle, that might help, but the important thing is just to document things people can't know without reading your mind (or the code).  :grin:  And yes, feel free to share the Calibre example since I know other people have talked about using Calibre URLs with Obsidian.  The given pattern above is for "search the current library"; I was using a macro to turn a book title into a link to read the book, so I can include it in notes I make while reading.

lishid:
https://github.com/kzhovn/uri-commands-obsidian/blob/fd2fa6af0860bce8ea49f54f2a6295cc5eddc24c/src/main.ts#L61
Here you should add the second parameter (MarkdownView) and use `view.file` instead of getting the active file from workspace.

https://github.com/kzhovn/uri-commands-obsidian/blob/fd2fa6af0860bce8ea49f54f2a6295cc5eddc24c/src/main.ts#L116
Yes this is available everywhere, so you won't need the check callback.

pjeby:
> Here you should add the second parameter (MarkdownView) and use view.file instead of getting the active file from workspace.

Actually, now that you've pointed it out, it looks to me like the editorCallback version should only be used if the URI has a `{{selection}}` or `{{line}}` pattern included, because those are the only insertions that require an `Editor` object.  Right now, if a command contains any insertion, you can only use it in edit mode on a markdown file, even if it's only inserting the file name or some metadata.

Overall, it might be simplest to just always use a plain callback for the command, and just have the selection/line inserts grab the editor via `getActiveViewOfType(MarkdownView)?.editor` if needed, or throw up a notice saying you can only use `selection` or `line` on markdown files.

kzhovn:
@pjeby I just caught that bug (I spend very little time in preview mode, it didn't occur to me to check earlier) and did a large refactor which ended up with the {{selection}} and {{line}} placeholders in an editorCallback and the rest in the regular callback. I tried a checkCallback for some which are only valid in some contexts (fileText in an md file, for instance), but that ended up being too complicated to implement.

@lishid Fixed.

PR #504, opened 2021-09-29
Add Map of Content plugin

Robin-Haupt-1:

## Repo URL
https://github.com/Robin-Haupt-1/Obsidian-Map-of-Content 

Link to my plugin:

## Release Checklist

- [X ] I have tested this on Windows, macOS, and Linux _(if applicable)_
- [ x] My GitHub release contains all required files
  - [x ] `main.js`
  - [ x] `manifest.json`
- [ x] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [ x] README clearly describes the plugins purpose and provides clear usage instructions.
- [ X] I have added a license in the LICENSE file.


lishid:
Please auto-format your code to normalize whitespaces.

https://github.com/Robin-Haupt-1/Obsidian-Map-of-Content/blob/2d9a294b01370c0927a93928c1b9d30b489d84c2/src/main.ts#L65
You can use the shortcut `detachLeavesOfType`.

https://github.com/Robin-Haupt-1/Obsidian-Map-of-Content/blob/2d9a294b01370c0927a93928c1b9d30b489d84c2/src/main.ts#L14
Don't keep a reference to your view, you might keep a reference to an outdated/detached view. Instead, use `workspace.getLeavesOfType(MY_VIEW_TYPE)` to get live references to your views.

https://github.com/Robin-Haupt-1/Obsidian-Map-of-Content/blob/2d9a294b01370c0927a93928c1b9d30b489d84c2/src/db.ts#L43
Since your entire update is completely synchronous, it will freeze the app until it's done, which means you'll always see this message and the next "complete" message at the exact same time, once the update is completed.

To allow the UI thread to update and not completely freeze, you can make your function `async` add various `await new Promise(r=>setTimeout(r, 0))` to make it "break" at those locations and let the UI do some work.

Robin-Haupt-1:
Thanks for the code review!

I have normalized the whitespace.

I have changed the code to use `detachLeavesOfType`

I've switched to getting live references to my view. I do that in a function that executes callback functions on the found view. I only use the static reference to the view once to initialize (`view.init`) it at plugin startup, because for some reason getting the fresh reference for that didn't work (it didn't see the view). 

I have included many `await new Promise(r=>setTimeout(r, 0))` in my now async update function and it works like expected.

lishid:
> I've switched to getting live references to my view. I do that in a function that executes callback functions on the found view. I only use the static reference to the view once to initialize (view.init) it at plugin startup, because for some reason getting the fresh reference for that didn't work (it didn't see the view).

Ok I see what you're trying to do here. This isn't the correct way. What you should do here is to pass the reference of your plugin to the constructor of your view, which stores a reference on the view:

```ts
this.registerView(MOC_VIEW_TYPE, (leaf) => new MOCView(leaf, this));
```

Then in your view's `onload()` function, do the init of registering events.

That way you can get rid of the view reference from your plugin/

Robin-Haupt-1:
Okay, that works. 

Out of curiosity, why exactly is it so important not to statically reference the view? 
After all i'm not aware it's even possible to manually open a second one without reloading the plugin. So isn't the worst that could happen that the plugin keeps writing to an detached view?

While trying to figure out how to implement your request i researched other plugin's code and many popular ones like the [Calendar plugin](https://github.com/liamcain/obsidian-calendar-plugin/blob/master/src/main.ts#L23) or the [Checklist plugin](https://github.com/delashum/obsidian-checklist-plugin/blob/master/src/main.ts#L9) also keep static references to their views.

lishid:
> Out of curiosity, why exactly is it so important not to statically reference the view?
After all i'm not aware it's even possible to manually open a second one without reloading the plugin. So isn't the worst that could happen that the plugin keeps writing to an detached view?
> While trying to figure out how to implement your request i researched other plugin's code and many popular ones like the Calendar plugin or the Checklist plugin also keep static references to their views.

Some of the earlier plugins haven't been reviewed with the latest recommended practices; I would suggest them the same.

The biggest factor here is the semantics and architecture. View creator functions aren't meant to have any side effects when constructing the view. Plugins also aren't supposed to "keep track" of active views, because now you have to different "source of truths" while one of them is the real one (workspace).

Assigning it to your plugin's field makes your code brittle - if we do end up changing the internals of how Obsidian construct new views, it might break your plugin functionality. For example, if there was a case that views can be constructed without being used (it may be discarded, or reloaded, etc), then you'll have a reference to a new but unused view while the real one that's still currently being used will no longer get updates (or whatever it is that you use the plugin's reference to the view for).

Most of the time, you can perform all your tasks in the view itself with a reference to the plugin.

Robin-Haupt-1:
Lishid, can you tell me where i can find the list of icon names i could use for my view?

lishid:
There was a list shared on discord somewhere... or you can add your own svg path to the icon database using the `addIcon` function.

PR #508, opened 2021-09-30
add linked data vocabularies

kometenstaub:

## Repo URL

Link to my plugin: https://github.com/kometenstaub/obsidian-linked-data-vocabularies

## Release Checklist

- [x] I have tested this on Linux (and Windows, thanks @SkepticMystic)
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2412,5 +2412,12 @@
         "author": "TfTHacker",
         "repo": "TfTHacker/obsidian42-text-transporter",
         "branch": "main"
+    },
+    {
+        "id": "linked-data-vocabularies",
+        "name": "Linked Data Vocabularies (SKOS)",
+        "description": "Use structured linked data as metadata in your notes.",
+        "author": "kometenstaub",
+        "repo": "kometenstaub/obsidian-linked-data-vocabularies"
     }
 ]


lishid:
https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/f3e71f0ae576c167fe5dba9c03ffeb12bd3ae2a4/src/main.ts#L57

This is not necessary - editorCheckCallback` pass you a `MarkdownView` rather than a `View`, and you don't need to check for it. You should use `editorCallback` instead.

https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/f3e71f0ae576c167fe5dba9c03ffeb12bd3ae2a4/src/methods.ts#L38
That's a bad idea. For one, the URL may indeed only work through http, but it could also already be `https` and now it will be `httpss`. It could also have http somewhere inside the URL like `www.my-site.com/what-is-http.htm` and now you converted it to `what-is-https`.

https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/f3e71f0ae576c167fe5dba9c03ffeb12bd3ae2a4/src/methods.ts#L62
Remember to encodeURIComponent on all of these.

https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/f3e71f0ae576c167fe5dba9c03ffeb12bd3ae2a4/src/suggester.ts#L47
Use `parent.createEl()` rather than `child = createEl() ... parent.appendChild(child)`.

There's also shortcuts for `createSpan` and `createDiv`.

You can also set the css class and inner text directly using `createDiv({cls: 'css-class', text: 'inner text'})`.

kometenstaub:
> https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/f3e71f0ae576c167fe5dba9c03ffeb12bd3ae2a4/src/main.ts#L57
> This is not necessary - editorCheckCallbackpass you aMarkdownViewrather than aView, and you don't need to check for it. You should use editorCallback` instead.

Done. https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/0c196e797137f218270a116abc99a1a468f5fbb4/src/main.ts#L52

> https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/f3e71f0ae576c167fe5dba9c03ffeb12bd3ae2a4/src/methods.ts#L38
> That's a bad idea. For one, the URL may indeed only work through http, but it could also already be https and now it will be httpss. It could also have http somewhere inside the URL like www.my-site.com/what-is-http.htm and now you converted it to what-is-https.

Not really in my case I think, because they are all of this type (http://id.loc.gov/authorities/subjects/sh85055634), which only `sh<number>` being different each time, but I've added another check. https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/0c196e797137f218270a116abc99a1a468f5fbb4/src/methods.ts#L39
I had it as `http` before, but on Android it errored and told me that http is not supported. As you can verify [here](https://id.loc.gov/authorities/subjects/sh85055634.json), `https` is supported.

> https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/f3e71f0ae576c167fe5dba9c03ffeb12bd3ae2a4/src/methods.ts#L62
> Remember to encodeURIComponent on all of these.

They are the parameters for the query (see [here](https://id.loc.gov/techcenter/searching.html), you need to scroll down to Suggest2), counter can only be an integer and the searchtype is hardcoded for either `keyword` or `leftanchored`, so they don't need to be encoded either.

> 
> https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/f3e71f0ae576c167fe5dba9c03ffeb12bd3ae2a4/src/suggester.ts#L47
> Use parent.createEl() rather than child = createEl() ... parent.appendChild(child).
> There's also shortcuts for createSpan and createDiv.
> You can also set the css class and inner text directly using createDiv({cls: 'css-class', text: 'inner text'}).

Thank you for these pointers. I exchanged `createEl` for `createDiv` where possible now, I still haven't figured out how to properly use `parent.createEl` though. Do I need to change that for this PR or can I take a look at that later?

kometenstaub:
Everything should be done now.

lishid:
> Not really in my case I think, because they are all of this type (http://id.loc.gov/authorities/subjects/sh85055634), with only sh<number> being different each time, but I've added another check. https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/0c196e797137f218270a116abc99a1a468f5fbb4/src/methods.ts#L39
I had it as http before, but on Android it errored and told me that http is not supported. As you can verify here, https is supported.

Weird, it seems that the http page just redirects to the https page, but I guess it's possible that the android side doesn't like plain http. I recommend doing it like this:
```ts
url = url.replace(/^https?:\/\//, 'https://');
```
That way, it'll only replace if the string is at the beginning, and not do anything if it's already https.

https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/daa18a384a1cb731f181691112f520fcd495fa6d/src/methods.ts#L38
Also this will be empty if the url isn't `http://`, and your request won't work.

lishid:
> They are the parameters for the query (see here, you need to scroll down to Suggest2), counter can only be an integer and the searchtype is hardcoded for either keyword or leftanchored, so they don't need to be encoded either.

I understand that you're fairly confident about the data types for these, but it's still a good practice to systematically encode every parameter for a couple of reasons:
- In case the caller function doesn't check/verify the parameter type and sends a bad parameter
- Doing so avoid URL parameter injection (i.e. accidentally passing in `&id=2`)

For example, you don't actually check that `this.plugin.settings.elementCounter` is a number, so if someone modifies the JSON file of your plugin config, any kind of parameter could be put in here.

(Also remember to use `String(variable)` to convert non strings to strings before uri encoding them).

> I still haven't figured out how to properly use parent.createEl though.

You're doing it correctly here: https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/d5bfe28e298fb2e5781c864ba6e968850c4bb312/src/suggester.ts#L93

kometenstaub:
> `url = url.replace(/^https?:\/\//, 'https://');`

I've done the http -> https the way you recommended. <https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/5d8211ce889c9b124964a40b5c990d66c061e1cc/src/methods.ts#L38>

I have encoded everything:
https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/1bd6193b346daabcee2c90817be857035bb4ba99/src/methods.ts#L69

I'm checking whether it is an integer in the settings (https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/1bd6193b346daabcee2c90817be857035bb4ba99/src/settings.ts#L35), but I've added another check: https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/1bd6193b346daabcee2c90817be857035bb4ba99/src/methods.ts#L60

> You're doing it correctly here: https://github.com/kometenstaub/obsidian-linked-data-vocabularies/blob/d5bfe28e298fb2e5781c864ba6e968850c4bb312/src/suggester.ts#L93

My issue wasn't that I didn't know how to create it like that, but how to interact with it then.

lishid:
> My issue wasn't that I didn't know how to create it like that, but how to interact with it then.

Ah! The rest is actually exactly the same. `parent.createEl()` will return you the HTMLElement that you can then keep as a variable and do what you want with it.

kometenstaub:
> > My issue wasn't that I didn't know how to create it like that, but how to interact with it then.
> 
> Ah! The rest is actually exactly the same. `parent.createEl()` will return you the HTMLElement that you can then keep as a variable and do what you want with it.

Ah, that was the piece I was missing. Thank you so much! :heart:

PR #509, opened 2021-10-01
Add update-time-on-edit-obsidian plugin

beaussan:

This is my first obsidian plugin, I tested it for 5 month in my local obsidian and waited to before releasing. Feel free to give me any feedback !

## Repo URL

Link to my plugin: https://github.com/beaussan/update-time-on-edit-obsidian

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2412,5 +2412,13 @@
         "author": "TfTHacker",
         "repo": "TfTHacker/obsidian42-text-transporter",
         "branch": "main"
+    },
+    {
+        "id": "update-time-on-edit",
+        "name": "Update frontmatter time on edit",
+        "description": "Keep front matter in sync with the last edit time",
+        "author": "beaussan",
+        "repo": "beaussan/update-time-on-edit-obsidian",
+        "branch": "main"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2414,6 +2414,58 @@
         "branch": "main"
     },
     {
+        "id": "marginnote-companion",
+        "name": "MarginNote Companion",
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/beaussan/update-time-on-edit-obsidian/blob/5fb46dc8ebb8dbb49c1028ca0e20f1d806d79da6/main.ts#L10
Recommend looking into using `moment` from `obsidian` since we already bundle it for internal use and also various plugins. That'll help keep your plugin bundle smaller.

https://github.com/beaussan/update-time-on-edit-obsidian/blob/5fb46dc8ebb8dbb49c1028ca0e20f1d806d79da6/main.ts#L126
Don't do this. You already have a reference to file from earlier, you can reuse it.

beaussan:
> Recommend looking into using moment from obsidian since we already bundle it for internal use and also various plugins. That'll help keep your plugin bundle smaller.

I saw that moment was already used in Obsidian, but I went with date-fns because it can parse a lot of things by default, since the first version of the package is very opinionated. In a second version I plan to migrate to `moment` and have a dynamic (via config) date format that will be use for parse + write

> Don't do this. You already have a reference to file from earlier, you can reuse it.

After more digging, I found out I can directly call `app.vault.modify` with the path, good call ! I didn't found this api last time while I fiddle with the plugin. Will change accordingly !

lishid:
> I saw that moment was already used in Obsidian, but I went with date-fns because it can parse a lot of things by default, since the first version of the package is very opinionated. In a second version I plan to migrate to moment and have a dynamic (via config) date format that will be use for parse + write. The footprint of date-fns with the tree shake is small hence the decision I made to go fast in the beginning. If this is a blocker, I can reconsider

No problem!

PR #518, opened 2021-10-05
add a new community plugin - Cryptsidian

triumphantomato:

## https://github.com/triumphantomato/cryptsidian/

Link to my plugin: https://github.com/triumphantomato/cryptsidian/

## Release Checklist

Don't have a Windows machine for testing, everything else is done.

- [ ] I have tested this on: 
  - [ ] Windows
  - [x] macOS
  - [x] Linux
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/triumphantomato/cryptsidian/blob/f661b6dbeda2d17ae203b79ba13959bd74da784c/cryptsidian.mjs#L39

If you're using NodeJS libraries, then your plugin will not work on any other platform than the electron app (i.e. mobile, possible web in the future). There's a browser API that does the same thing called SubtleCrypto, which is what Obsidian uses for end-to-end encryption on desktop and mobile.

https://github.com/triumphantomato/cryptsidian/blob/f661b6dbeda2d17ae203b79ba13959bd74da784c/cryptsidian.mjs#L47
This is a really really bad idea. You should save your initial value/salt as the first block of each file to preserve the strength of the encryption used.

https://github.com/triumphantomato/cryptsidian/blob/f661b6dbeda2d17ae203b79ba13959bd74da784c/main.ts#L89
Don't advise using the base path (and/or using node.js's fs) directly; use the vault interface as much as possible, and fallback to using the adapter. Avoid going to the native FS directly.


https://github.com/triumphantomato/cryptsidian/blob/f661b6dbeda2d17ae203b79ba13959bd74da784c/cryptsidian.mjs#L401
Here you're processing ALL files including attachments and unsupported extensions - might not be a great idea, especially since you're only closing down markdown panes and not other kind of file panes: https://github.com/triumphantomato/cryptsidian/blob/f661b6dbeda2d17ae203b79ba13959bd74da784c/main.ts#L197

triumphantomato:
Thanks for your comments, @lishid. 

> https://github.com/triumphantomato/cryptsidian/blob/f661b6dbeda2d17ae203b79ba13959bd74da784c/cryptsidian.mjs#L39
> 
> If you're using NodeJS libraries, then your plugin will not work on any other platform than the electron app (i.e. mobile, possible web in the future). There's a browser API that does the same thing called SubtleCrypto, which is what Obsidian uses for end-to-end encryption on desktop and mobile.

Thanks, I can look into SubtleCrypto for a future revision. This version of Cryptsidian is intended for use only on desktop, and is declared `isDesktopOnly: true` in the [manifest.json#L9](https://github.com/triumphantomato/cryptsidian/blob/main/manifest.json#L9). 

> 
> https://github.com/triumphantomato/cryptsidian/blob/f661b6dbeda2d17ae203b79ba13959bd74da784c/cryptsidian.mjs#L47 This is a really really bad idea. You should save your initial value/salt as the first block of each file to preserve the strength of the encryption used.
 
The `SALT` value is not used in encryption. It's used in a PBKDF routine with `scrypt` later: [cryptsidian.mjs#L201](https://github.com/triumphantomato/cryptsidian/blob/main/cryptsidian.mjs#L201)

The initialization vector (IV) is unique per-file ([cryptsidian.mjs#L248](https://github.com/triumphantomato/cryptsidian/blob/main/cryptsidian.mjs#L248)), and is pre-pended to the file as suggested ([cryptsidian.mjs#L260](https://github.com/triumphantomato/cryptsidian/blob/main/cryptsidian.mjs#L260)).

> https://github.com/triumphantomato/cryptsidian/blob/f661b6dbeda2d17ae203b79ba13959bd74da784c/main.ts#L89 Don't advise using the base path (and/or using node.js's fs) directly; use the vault interface as much as possible, and fallback to using the adapter. Avoid going to the native FS directly.

I'm new to Obsidian and not sure how to accomplish this. `this.app.vault.adapter.getBasePath();` was suggested in the Discord. I'm using it to get the directory path I need for `getVaultFiles()` [(cryptsidian.mjs#L401)](https://github.com/triumphantomato/cryptsidian/blob/main/cryptsidian.mjs#L401) to work.

Could you help me understand the scope of changes suggested here and how to implement if needed?
 
>https://github.com/triumphantomato/cryptsidian/blob/f661b6dbeda2d17ae203b79ba13959bd74da784c/cryptsidian.mjs#L401 Here you're processing ALL files including attachments and unsupported extensions - might not be a great idea, especially since you're only closing down markdown panes and not other kind of file panes: https://github.com/triumphantomato/cryptsidian/blob/f661b6dbeda2d17ae203b79ba13959bd74da784c/main.ts#L197

Thanks yes - the intent is to encrypt all files, regardless of file type, within the Vault (including attachments). 

Could you suggest how to close all open panes within Obsidian? The Discord was not sure how to do this and I couldn't find documentation in the API.

triumphantomato:
hey @lishid thanks for your help here - i've refactored to `iterateAllLeaves` and `leaf.detach` for leaves that pass `instanceof FileView` (see [0155e9d](https://github.com/triumphantomato/cryptsidian/commit/0155e9d30c6dea8de8d3b9ae2b8dd17ad8ab832d) - _please note, i removed an unused import of `util` in a later commit_).

regarding using the Obsidian vault API for file access, all the reasons you list are compelling and i'd love to benefit from them. i did a bit of digging and had two thoughts - 

first, there doesn't seem to be an easy way to get a file descriptor from Obsidian, say after it's cleaned up its memory caches for files. `vault.getFiles()` returns a set of objects that includes the file _paths_, but this leaves me in the same position of using `fs` to secure an `fd` and read/write to disk

the alternative is to rewrite the plugin to rely on the Vault API exclusively for file access. this has the concurrency and file cache advantages you mentioned above, and makes a future port to mobile simpler - but, it has a high cost since i'll need to rewrite all the file handling logic to use `TFile`'s (and possibly `modifyBinary()`?) and rewrite the tests to ensure the logic is operating properly

Second, I'm wondering if we already get the concurrency safety benefit by requiring Obsidian to close all open leaves of instance `FileView` before doing any crypto? This leaves the mobile port and file caching still on the table.

If we think this is safe for general use from the concurrency angle with the current refactor in [0155e9d](https://github.com/triumphantomato/cryptsidian/commit/0155e9d30c6dea8de8d3b9ae2b8dd17ad8ab832d), then I'm inclined to leave a major rewrite of the plugin for future work.

What do you think?

lishid:
> first, there doesn't seem to be an easy way to get a file descriptor from Obsidian, say after it's cleaned up its memory caches for files. vault.getFiles() returns a set of objects that includes the file paths, but this leaves me in the same position of using fs to secure an fd and read/write to disk

Agreed - we don't expose anything like that because not all platforms uses file descriptors for their file system API. Remember that Obsidian works across desktop and mobile apps, and possibly as a webpage app in the future (with a virtual/emulated filesystem), and our API has to accommodate all of these.

> but, it has a high cost since i'll need to rewrite all the file handling logic to use TFile's (and possibly modifyBinary()?) and rewrite the tests to ensure the logic is operating properly

Yeah that's unfortunate.

>Second, I'm wondering if we already get the concurrency safety benefit by requiring Obsidian to close all open leaves of instance FileView before doing any crypto? This leaves the mobile port and file caching still on the table.

> If we think this is safe for general use from the concurrency angle with the current refactor in 0155e9d, then I'm inclined to leave a major rewrite of the plugin for future work.

You might want to wait a little bit after closing out the leaves in case the user made changes that has to be saved - to do that, you should call `workspaceLeaf.setViewState({type: 'empty'})` and await the resulting promise. You can then safely detach the leaf after the await and start processing your files.

I think if you want to have your plugin work across other platforms this is the way to go. It is definitely quite a challenge to rewrite though, good luck if you decide to go that way!

> Password is required to posess sufficient entropy, but you can change the amount of entropy required in the hasEnoughEntropy function.

IMO from a usability's perspective this isn't really needed - if a user is educated enough to look for a full vault encryption solution, they are likely to know the implications of weak passwords. If they don't, then maybe the plugin isn't really for them 🤣.

lishid:
Oh and just curious - what are the benefit of using encryption like this as a plugin rather than using something like Cryptonator or VeraCrypt? I think it would be useful to explain this in the README too.

triumphantomato:
> You might want to wait a little bit after closing out the leaves in case the user made changes that has to be saved - to do that, you should call workspaceLeaf.setViewState({type: 'empty'}) and await the resulting promise. You can then safely detach the leaf after the await and start processing your files.

Done! [3c18712](https://github.com/triumphantomato/cryptsidian/commit/3c18712a5beed03a0ddb7473a955f0f4ce005dd5)

> I think if you want to have your plugin work across other platforms this is the way to go. It is definitely quite a challenge to rewrite though, good luck if you decide to go that way!

Thanks, something i'll have to noodle on for later :)

> Cryptonator or VeraCrypt? 

I'm not familiar with Cryptonator, but google tells me it's a wallet solution. VeraCrypt is an interesting piece of software, but probably overkill for someone who wants to just protect their Obsidian vault directory. The [beginner's tutorial](https://www.veracrypt.fr/en/Beginner%27s%20Tutorial.html) is pretty complicated to setup a simple volume, so having something within Obsidian that's dead simple ("enter a password") feels compelling.

You can imagine a situation where one device is shared between multiple people (say a family can only afford one computer, or a computer lab for students), each of whom is using Obsidian, but wants to keep their notes private from each other. Vault encryption as a plugin is helpful here! Will update the README.

Please let me know if this plugin can now be accepted!

lishid:
> I'm not familiar with Cryptonator, but google tells me it's a wallet solution. VeraCrypt is an interesting piece of software, but probably overkill for someone who wants to just protect their Obsidian vault directory. The beginner's tutorial is pretty complicated to setup a simple volume, so having something within Obsidian that's dead simple ("enter a password") feels compelling.

Oops I meant [Cryptomator](https://cryptomator.org/)!

I think having a drive like encryption system has the advantage of not requiring an encrypting/decrypting step (it's done on the fly via the emulated file system), and is thus less likely to break (for example, failing to decrypt halfway and leaving some files encrypted and some decrypted).

triumphantomato:
🎉

PR #519, opened 2021-10-05
Attempt #2 for Enlightenment

ryanjamurphy:
# New plugin: Enlightenment

## Repo URL

Link to my plugin: https://github.com/ryanjamurphy/enlightenment-obsidian

## Release Checklist

- [x] I have tested this on macOS and MS-DOS
  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css`


liamcain:
Found a few small things!

---

https://github.com/ryanjamurphy/enlightenment-obsidian/blob/40ddc4347ff983001f3760cff2e998bb1ccbfd56/main.ts#L111 I recommend you create a Typescript `type` called `EnlightmentType`. You can define it as `type EnlightenmentType = "full" | "active-pane"`. Then you can change this function to accept a param `type: EnlightenmentType`. The benefit to this is that when you compile the code, typescript will make sure you only ever pass in a valid enlightenment type.

https://github.com/ryanjamurphy/enlightenment-obsidian/blob/40ddc4347ff983001f3760cff2e998bb1ccbfd56/main.ts#L45 since CSS variables are global, I recommend making the name here a bit more unique just to avoid the chance that another plugin has a css variable with the same name. Just `--enlightenment-background-transparency` is fine.

https://github.com/ryanjamurphy/enlightenment-obsidian/blob/40ddc4347ff983001f3760cff2e998bb1ccbfd56/main.ts#L105 you could put `removeStyle` in the `addStyle` function. That way you don't have to call it in each of these logic branches here and it removes the risk of having 2 modes active at once somehow.

https://github.com/ryanjamurphy/enlightenment-obsidian/blob/40ddc4347ff983001f3760cff2e998bb1ccbfd56/main.ts#L176 you don't really need the `await` here. It's not doing anything for you since you aren't waiting for anything.

https://github.com/ryanjamurphy/enlightenment-obsidian/blob/40ddc4347ff983001f3760cff2e998bb1ccbfd56/main.ts#L49 typo!!!!! ❗ 

https://github.com/ryanjamurphy/enlightenment-obsidian/blob/40ddc4347ff983001f3760cff2e998bb1ccbfd56/main.ts#L68 I don't think you actually need to worry about if there's an activeLeaf or not. Your plugin seems safe to activate if there is no active leaf and then you activate one after. I would recommend changing this from `checkCallback` to just `callback`

lishid:
https://github.com/ryanjamurphy/enlightenment-obsidian/blob/40ddc4347ff983001f3760cff2e998bb1ccbfd56/main.ts#L4
I think this just need your path here - but you need to resize the icon to `0 0 100 100`.

https://github.com/ryanjamurphy/enlightenment-obsidian/blob/40ddc4347ff983001f3760cff2e998bb1ccbfd56/main.ts#L65
I think you don't need to check for activeLeaf here since your style is applied globally anyway.

https://github.com/ryanjamurphy/enlightenment-obsidian/blob/40ddc4347ff983001f3760cff2e998bb1ccbfd56/main.ts#L96
There's a shortcut here for `.hasClass('class')`.

https://github.com/ryanjamurphy/enlightenment-obsidian/blob/40ddc4347ff983001f3760cff2e998bb1ccbfd56/main.ts#L114
Same here for `.addClass('class')`, and `.removeClass('class')`.

ericaxu:

liamcain:
Looks like you need to cut a release for 1.2 still (there isn't a Github Release yet). My might want to look at the sample plugin again since it has a script that automatically bumps the version for you. There's also github hooks that can automatically create a release after you push a new version (https://github.com/liamcain/obsidian-periodic-notes/blob/main/.github/workflows/publish.yml>)

> I tried this, but the icon became tiny

That's because you'll need to modify the SVG itself. All the points are currently relative to the 512 x 512 viewbox, so just changing the viewbox dimensions isn't enough. You would need to open the svg up in a vector image program and reexport it to 100 x 100

https://github.com/ryanjamurphy/enlightenment-obsidian/blob/73804916e3ff1d773d327f629863339a3dda84ad/main.ts#L154

Super nit-picky, but maybe:

> Set the transparency for items in the background. The default is 0.7. 

Just seeing `.7.` made my brain struggle to parse it for a second.

https://github.com/ryanjamurphy/enlightenment-obsidian/blob/73804916e3ff1d773d327f629863339a3dda84ad/main.ts#L163 I recommend removing the console.log statements. Or change the to `console.debug` and add info about what plugin is printing things. Otherwise it can get messy for other plugin developers.

liamcain (approved):

PR #520, opened 2021-10-06
Add Another Quick Switcher plugin

tadashi-aikawa:

## Repo URL

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher

## Release Checklist

  - Windows only
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/blob/ba35fe444911c4a2ff0bbac68b7841528cb95b30/src/app-helper.ts#L2
Don't use any NodeJS packages if you intend for your plugin to work on mobile since NodeJS isn't available there.

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/blob/ba35fe444911c4a2ff0bbac68b7841528cb95b30/src/app-helper.ts#L42

activeLeaf may be null, and might not be a MarkdownView.

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/blob/ba35fe444911c4a2ff0bbac68b7841528cb95b30/src/app-helper.ts#L46
Missing null check on the function return.

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/blob/ba35fe444911c4a2ff0bbac68b7841528cb95b30/src/app-helper.ts#L65
Here you should just use `MarkdownView.file` instead of getting the active file (which gets you the active view anyway).

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/blob/ba35fe444911c4a2ff0bbac68b7841528cb95b30/src/commands.ts#L19
You don't need a checkCallback here if you don't have any condition to check here.

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/blob/ba35fe444911c4a2ff0bbac68b7841528cb95b30/src/ui/AnotherQuickSwitcherModal.ts#L119
A better (but still undocumented way) is to use `this.chooser.useSelectedItem(KeyEvent)`.

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/blob/ba35fe444911c4a2ff0bbac68b7841528cb95b30/src/ui/AnotherQuickSwitcherModal.ts#L141
You're querying for the active file many times here. Store the value in a variable so you aren't doing that thousands of times.

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/blob/ba35fe444911c4a2ff0bbac68b7841528cb95b30/src/ui/AnotherQuickSwitcherModal.ts#L238
Don't use raw HTML because this leads to DOM injection attacks. Always build your DOM from elements. Obsidian provides some shortcuts to create elements quickly, like `parentEl.createEl/createDiv({cls: 'css-class', text: 'innerText'})`.

tadashi-aikawa:
@lishid 
Thank you for your kind reviews! I have fixed all as follows.

> Don't use any NodeJS packages if you intend for your plugin to work on mobile since NodeJS isn't available there.

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/commit/cc3408028d7d7361af0180e619baeb3ec7de5aca

> activeLeaf may be null, and might not be a MarkdownView.

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/commit/f1e25b3a8dee404cbb6b2da56fd6ac9acb6a06fb

> Missing null check on the function return.

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/commit/fa359b394c29e5881ad415e305099e7ae866cb82

> Here you should just use MarkdownView.file instead of getting the active file (which gets you the active view anyway).

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/commit/b1c288c9992ba92a0ade88e74afb3281593d4e6a

> You don't need a checkCallback here if you don't have any condition to check here.

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/commit/3969a9e3b556bf4a17f18398c44ca6f21ba359d8

> A better (but still undocumented way) is to use this.chooser.useSelectedItem(KeyEvent).

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/commit/c14523495e236846a6042fb260da2dec63a338f1

> You're querying for the active file many times here. Store the value in a variable so you aren't doing that thousands of times.

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/commit/e66315cd29b517a7bfb6e5cc439394e0fa30d21f

> Don't use raw HTML because this leads to DOM injection attacks. Always build your DOM from elements. Obsidian provides some shortcuts to create elements quickly, like parentEl.createEl/createDiv({cls: 'css-class', text: 'innerText'}).

https://github.com/tadashi-aikawa/obsidian-another-quick-switcher/commit/a1ee854dfc7ea43a14d45ef02e0e901f7740d601

Best regards.

PR #522, opened 2021-10-07
Added Snippet Commands plugin

deathau:
Adds all your CSS snippets as commands, accessible from the command palette and even assignable to hotkeys!


liamcain:
> Snippet Commands: Toggle

🤦 d'oh. Yeah. great point, bad suggestion by me. pls ignore.

PR #525, opened 2021-10-07
Add Obsidian Link Converter

ozntel:

## Repo URL

Link to my plugin:
https://github.com/ozntel/obsidian-link-converter

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css


liamcain (approved):
Very handy! Code LGTM

---

### Small suggestions

https://github.com/ozntel/obsidian-link-converter/blob/77407a0e76840352c220f0668994a14c66387f97/src/converter.ts#L248 you might want to return an empty string here just so the return type is `string` instead of `string?`

https://github.com/ozntel/obsidian-link-converter/blob/77407a0e76840352c220f0668994a14c66387f97/src/main.ts#L41 Since converting all the links in a vault might be slow/expensive, maybe add a notice when the conversion is done?

ozntel:
Hi Liam, thanks for your review. I just pushed new changes to the plugin repo:

https://github.com/ozntel/obsidian-link-converter/commit/4bc80abac4c2d8d95bcdeb17fbcc19532767118c

ozntel:
> I would suggest hooking this event regardless, and check for your settings flag and return if it's not enabled - that way users won't have to reload the plugin when changing the option. Also, you can use this.registerEvent( workspace.on(...)) to automatically have the event unhooked when your plugin unloads, rather than manually unloading it.

If it is ok, I'd prefer leaving since the plugin settings already handles the option change (no need for reload):
https://github.com/ozntel/obsidian-link-converter/blob/e11eb558f7a3a310db73b89d15386c903680fc65/src/settings.ts#L40
All event handlers refer to the same function and works good with loading and unloading.

> Please use vault.read(file) rather than vault.adapter.read(file.path) whenever possible.
> Make sure you check that the file here is a markdown file, because it could be a different extension like PDF.

Both Done
https://github.com/ozntel/obsidian-link-converter/commit/6f8444e159b084435cbfd75c32dc20b2ac2a8d9e

> Same here. You can use vault.write(file, text). This helps with caching and reduces the chance of race conditions because direct
adapter writes are considered external.

I tried but I have an error that `app.vault.write is not a function`

> An even better way here is to display a notice when you start, set its timeout to 0 (always shown), and continuously update the notice's content with a counter like (x/y files processed). Then at the end, you can use Notice.hide() to remove it.

Done
https://github.com/ozntel/obsidian-link-converter/commit/15be898d2d0c9362e0156e860cc51c211becfeef

PR #526, opened 2021-10-07
Update community-plugins.json

jmilldotdev:

## Repo URL

Link to my plugin: https://github.com/jmilldotdev/obsidian-wikipedia

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


phibr0:
https://github.com/jmilldotdev/obsidian-wikipedia/blob/47ffbafb51b9c8948b70d8fccb2ae9003a956ca8/main.ts#L125
Maybe use `await` only

https://github.com/jmilldotdev/obsidian-wikipedia/blob/47ffbafb51b9c8948b70d8fccb2ae9003a956ca8/main.ts#L181
Instead of using regular callbacks here, you should use `editorCallback` here. Your command won't show if there is no active editor and you don't need to get the editor manually, since it gets passed as a parameter.

https://github.com/jmilldotdev/obsidian-wikipedia/blob/47ffbafb51b9c8948b70d8fccb2ae9003a956ca8/main.ts#L271
Not necessarily needed but you could use a dropdown with the available languages here.

-> Remove all `console.log()`s or change them to `console.debug()` to avoid spamming the console so much

lishid:
https://github.com/jmilldotdev/obsidian-wikipedia/blob/47ffbafb51b9c8948b70d8fccb2ae9003a956ca8/main.ts#L39
Please rename your plugin, settings, modals, and other default class names to your actual plugin's name :)

https://github.com/jmilldotdev/obsidian-wikipedia/blob/47ffbafb51b9c8948b70d8fccb2ae9003a956ca8/main.ts#L125
Consider using Obsidian's `request` function that does not run into CORS issues that `fetch` has.

https://github.com/jmilldotdev/obsidian-wikipedia/blob/47ffbafb51b9c8948b70d8fccb2ae9003a956ca8/main.ts#L167
This may return null.

https://github.com/jmilldotdev/obsidian-wikipedia/blob/47ffbafb51b9c8948b70d8fccb2ae9003a956ca8/main.ts#L162
Missing null check here.

PR #527, opened 2021-10-08
Add Vale plugin

marcusolsson:

## Repo URL

Link to my plugin: https://github.com/marcusolsson/obsidian-vale

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
Very excited about this plugin!! I took a look at the code and just had a few small comments, but it looks great. Nicely done.

## Code Review

https://github.com/marcusolsson/obsidian-vale/blob/b8737160c82aab23098c67438b80b602d239cd9d/src/main.ts#L281 Oooh, this is a cool pattern! I wonder if this could work with Typescript decorators 🤔 

### Nits

https://github.com/marcusolsson/obsidian-vale/blob/b8737160c82aab23098c67438b80b602d239cd9d/src/main.ts#L18 `the the` 👀 

https://github.com/marcusolsson/obsidian-vale/blob/b8737160c82aab23098c67438b80b602d239cd9d/src/main.ts#L76 I'm (admittedly a bit rusty) but you shouldn't have to call this yourself. `onClose` will get called automatically.

https://github.com/marcusolsson/obsidian-vale/blob/b8737160c82aab23098c67438b80b602d239cd9d/src/main.ts#L120 instead of nesting `Object.assign` you can just add an extra argument to the outer one:

```ts
    this.settings = Object.assign(
      {}, // only this object will get mutated
      DEFAULT_SETTINGS, // left intact
      await this.loadData()
    );
```

https://github.com/marcusolsson/obsidian-vale/blob/b8737160c82aab23098c67438b80b602d239cd9d/src/main.ts#L177 not quite sure this is the right error message since the only way to get to this code path is if the adapter isn't a FileSystemAdapter (aka, it's on mobile).

https://github.com/marcusolsson/obsidian-vale/blob/b8737160c82aab23098c67438b80b602d239cd9d/src/hooks.ts#L37 why not use the [normalizePath](https://github.com/obsidianmd/obsidian-api/blob/763a243b4ec295c9c460560e9b227c8f18d8199b/obsidian.d.ts#L2122) from the api?

https://github.com/marcusolsson/obsidian-vale/blob/b8737160c82aab23098c67438b80b602d239cd9d/src/vale/ValeConfigManager.ts#L15 the local variable `path` is shadowing the `path` import. You might want to rename one of them to avoid confusion.

marcusolsson:
Thanks for taking a look!

> I wonder if this could work with Typescript decorators

Not sure. I haven't found a reason to learn about decorators yet, but I'll do some reading!

> `onClose` will get called automatically

That would make sense. I think I saw this in an existing plugin and didn't think to much of it. I'll remove it and test.

> instead of nesting `Object.assign` you can just add an extra argument to the outer one

I learned something new!

> not quite sure this is the right error message

You're right. I think I didn't know what it meant for the adapter to not be a FileSystemAdapter. I'll update it accordingly.

> why not use the normalizePath from the api

For some reason I thought its purpose was to deal with delimiters across OSs. I'll do some testing!

I'll let you know once I've made the suggested changes!

marcusolsson:
I've updated the plugin with your suggestions. Regarding `normalizePath` however, it doesn't appear to do the same thing. 

```
console.log(normalizePath("foo"));
console.log(normalizePath("folder/note"));
console.log(normalizePath("C:\\folder1\\Untitled 2"));
console.log(normalizePath("/Users/foo/Desktop"));
```

Prints

```
foo
folder/note
C:/folder1/Untitled 2
Users/foo/Desktop
```

Notice the missing forward slash at the start of the last path.

liamcain:
> Regarding normalizePath however, it doesn't appear to do the same thing

Aah, good catch. The API `normalizePath` function is for normalizing paths _within_ the vault. So it doesn't handle absolute paths.

lishid:
https://github.com/marcusolsson/obsidian-vale/blob/f12d55fc24b4b3fbdd011cfdb8bec86d50ce62ab/src/main.ts#L59
Suggest avoiding assigning created views to your plugin. The view creator function should be side-effect free, as it may be called to create multiple views, some of which are possibly trashed/unused (in theory at least, in practice Obsidian doesn't do this yet).

You should always use `workspace.getLeavesOfType(MY_VIEW_TYPE)` to get references to views of your plugin that's currently live, rather than keeping a reference to the last created view. (Use `leaf.view instanceof MyViewType` after to double check)

https://github.com/marcusolsson/obsidian-vale/blob/f12d55fc24b4b3fbdd011cfdb8bec86d50ce62ab/src/main.ts#L77
There's a shortcut for `detachLeavesOfType`.

https://github.com/marcusolsson/obsidian-vale/blob/f12d55fc24b4b3fbdd011cfdb8bec86d50ce62ab/src/main.ts#L167
`getFullPath` expects a vault internal normalized path here.

marcusolsson:
Thanks for taking a look! I hope I've addressed all your comments here:

https://github.com/marcusolsson/obsidian-vale/commit/2fdaeb860440e568e993026604e6c8b246943094

PR #528, opened 2021-10-08
Add JavaScript Init plugin.

ryanpcmcquen:

## Repo URL

Link to my plugin:
https://github.com/ryanpcmcquen/obsidian-javascript-init/

## Release Checklist

- [x] I have tested this on Windows.
- [x] I have tested this on macOS.
- [ ] I have tested this on Linux.
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-css-themes.json
+++ b/community-css-themes.json
@@ -619,5 +619,12 @@
         "screenshot": "Faded.png",
         "modes": ["dark"],
         "branch": "main"
+    },
+    {
+        "name": "Creature",
+        "author": "Marcus Olsson",
+        "repo": "marcusolsson/obsidian-creature-theme",
+        "screenshot": "screenshot.png",
+        "branch": "main"
     }
 ]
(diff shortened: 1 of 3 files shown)


phibr0:
https://github.com/ryanpcmcquen/obsidian-javascript-init/blob/58fef8cadb6e5e5ee385b89ae58fdb0cee8be62e/main.ts#L86
This should be fine, but you could also just use the  `setClass()` function on the Setting component and add these styles to the `styles.css`.

https://github.com/ryanpcmcquen/obsidian-javascript-init/blob/58fef8cadb6e5e5ee385b89ae58fdb0cee8be62e/main.ts#L13
Why don't you pass the code saved in your settings Object into this function? If it wasn't set by the user it still just is an empty string, since thats the default value for your settings.

lishid:
https://github.com/ryanpcmcquen/obsidian-javascript-init/blob/c213d858b3a94b7b294438cba5f4650153b795e0/main.ts#L14
Should keep a reference to this and remove it from DOM when your plugin unloads.

PR #531, opened 2021-10-10
add key promoter

joethei:

## Repo URL

Link to my plugin: https://github.com/joethei/obsidian-key-promoter

## Release Checklist

  - [X] `main.js`
  - [X] `manifest.json`


liamcain:
Nice! I like this idea a lot. You might want to add a screenshot to your README file to help others understand what it does though.

---

https://github.com/joethei/obsidian-key-promoter/blob/8e0c24f94035ee668234f5f04ea483e6ea3440a2/main.ts#L31 You can set the type here to [`Command`](https://github.com/obsidianmd/obsidian-api/blob/763a243b4ec295c9c460560e9b227c8f18d8199b/obsidian.d.ts#L416) instead of `any`

https://github.com/joethei/obsidian-key-promoter/blob/8e0c24f94035ee668234f5f04ea483e6ea3440a2/main.ts#L30 Small nit: you can just use `Object.values` here instead since you're never using the keys.

lishid:
https://github.com/joethei/obsidian-key-promoter/blob/2fcbd8322660b0d32757df99110f1e9ed24a5c6b/main.ts#L45
Here you might run into issues for example when having a command named `copy` and another named `copy text under cursor`, you'll show both when `copy` is triggered.

PR #532, opened 2021-10-11
Add Meld Calc plugin

meld-cp:

## Repo URL

Link to my plugin: https://github.com/meld-cp/obsidian-calc

## Release Checklist

- [x] I have tested this on Windows (don't have access to macOS or Linux)
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
Oh wow, Fcal is super cool. I didn't know this existed but I like the idea of Obsidian + Soulver.

---

Code LGTM just saw a few small things I figured I would point out.

### Small code nitpicks

https://github.com/meld-cp/obsidian-calc/blob/1b517adc8b89de9da6650ad9c78d05eb05e8f1a0/src/main.ts#L55 This should be triple equals `===`

https://github.com/meld-cp/obsidian-calc/blob/1b517adc8b89de9da6650ad9c78d05eb05e8f1a0/src/main.ts#L41 triple equals `===`

https://github.com/meld-cp/obsidian-calc/blob/1b517adc8b89de9da6650ad9c78d05eb05e8f1a0/src/main.ts#L7 instead of making this a static variable, why not make this an instance variable on the plugin instead. There will only ever be a single instead of the plugin and it will be easier to call from within the plugin instance.

https://github.com/meld-cp/obsidian-calc/blob/1b517adc8b89de9da6650ad9c78d05eb05e8f1a0/src/main.ts#L85 might want to clean this up, or optionally change this to `console.debug` so it doesn't pollute the console as much.

meld-cp:
Thanks @lishid I'll look into it.   Has checkCallback been deprecated?

PR #560, opened 2021-10-20
Add Tweet to Markdown plugin

kbravh:

## Repo URL

Link to my plugin: https://github.com/kbravh/obsidian-tweet-to-markdown

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/kbravh/obsidian-tweet-to-markdown/blob/8894b1ba4b4b7ebcf49bef83712387179b81cdf5/src/TweetCompleteModal.ts#L55
This catch is eating up the error and not stopping execution - probably want to re-throw here. Also recommend using the async-await code style rather than `.then/.catch` here (when using await, you can use a simple try-catch block which translates to what you're doing with catch).

PR #570, opened 2021-10-23
Added Habitica Sync Plugin

SuperChamp234:

## Repo URL

Link to my plugin: https://github.com/SuperChamp234/habitica-sync

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


SuperChamp234:
> Not sure why they're still recommending using then/catch, but the latest js/typescript syntax can make things a lot simpler and more like regular code.

Ah I see, it is def more simpler than theirs! Updated my program!

> Wait, why aren't these variables on your App class, and instead it's in a global variable? That seems like pretty bad practice here...

Oops! Might have missed that while going through the program, fixed it. Thanks for pointing it out!

PR #576, opened 2021-10-24
Added Image Caption plugin

bicarlsen:

## Repo URL

Link to my plugin: https://github.com/bicarlsen/obsidian_image_caption

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/bicarlsen/obsidian_image_caption/blob/1e45e7ea5c641a767f6ec434e7bf8771f18a6d14/src/main.ts#L37

Remember to also remove your stylesheet when your plugin unloads.

> Do you know how to remove an individual element from a MutationObservers watch list?
Do you think having one observer per element would be a better implementation?

I think that's exactly what most people do to unhook MutationObservers. MutationObservers also tie their lifecycle (for garbage collection) to the elements you observe, so by creating individual observers for each DOM element, they can be garbage collected along side the DOM element when that element is removed and ready to be collected.

bicarlsen:
@lishid Thank you for the input. I've updated again, unloading styles and MutationObservers, and removing captions. I also now have one MutationObserver per element, which is removed once the caption is added.

lishid:
https://github.com/bicarlsen/obsidian_image_caption/blob/a1c003d77a973ade976a6370733922344c32eb27/src/main.ts#L26
Does the building process not give you a bunch of errors? You aren't declaring the types for most of your class properties like `caption_tag`, etc. (Also the typical convention is to use camelCase for class members).

bicarlsen:
Thanks @lishid . No, I don't get any build errors. I was just using type inference where I didn't explicitly declare the types as the plugin is so simple. As bit of personal preference. I moved those class wide variables to be static class variables, and explicitly state their type now, though.

I know the camel case convention, but have been programming in Python for the past few years and have grown 
aesthetically more fond of snake case over camel case for readability, so decided to stick with it here. Just personal preference again.

Thank you again for all your inpu through this process.

lishid:
> Thanks @lishid . No, I don't get any build errors. I was just using type inference where I didn't explicitly declare the types as the plugin is so simple. As bit of personal preference. I moved those class wide variables to be static class variables, and explicitly state their type now, though.

I see - I double checked and it turns out ESBuild skips most of the type checks to improve the build speed, but with a proper TypeScript compiler it will throw those errors. (Also most IDEs will show errors because class properties aren't typically inferred as opposed to inline variables).

> I know the camel case convention, but have been programming in Python for the past few years and have grown
aesthetically more fond of snake case over camel case for readability, so decided to stick with it here. Just personal preference again.

I understand. I still urge people to adopt the formatting policy of "best match" for the language/ecosystem just because you will be using code and API from others in the ecosystem (like the DOM and browser APIs) which will be using a specific formatting. Using a personal preference that is different from the majority of the ecosystem will end up with different styles throughout the codebase of your plugin which is confusing and not very aesthetically pleasing 😉. Regardless - just a suggestion from my side, it's all up to you in the end.

PR #578, opened 2021-10-25
Added Asciinema Player plugin

nekomeowww:

## Repo URL

Link to my plugin: https://github.com/nekomeowww/obsidian-asciinema-player

## Release Checklist

- [x] I have tested this on Windows, macOS, iPadOS, and Linux _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2804,5 +2804,13 @@
         "author": "triumphantomato",
         "repo": "triumphantomato/cryptsidian",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-asciinema-player",
+        "name": "Asciinema Player",
+        "description": "This is a Obsidian plugin which supports embedding asciicast files into Markdown files",
+        "author": "Ayaka Neko",
+        "repo": "nekomeowww/obsidian-asciinema-player",
+        "branch": "main"
     }
 ]


nekomeowww:
> https://github.com/nekomeowww/obsidian-asciinema-player/blob/02dbb044f75f26e2064b586978e230ea31dcd3ab/src/main.ts#L16 You should use vault.modify instead of vault.adapter.write whenever possible - the vault interface offers an additional file caching layer that would otherwise be invalidated if you use adapter directly. You'll probably want to look into `vault.getAbstractFileByPath()`, `if (file instanceof TFile)` and `vault.modify()`.
> 
> https://github.com/nekomeowww/obsidian-asciinema-player/blob/02dbb044f75f26e2064b586978e230ea31dcd3ab/src/main.ts#L89 Your plugin should take care of cleaning up itself and removing any additional scripts/styles loaded when it's disabled. Override the `onunload()` function to perform the cleanup. (Tip: You can keep a reference to each script/style tag you create in your plugin's class to use in your unload function)
> 
> https://github.com/nekomeowww/obsidian-asciinema-player/blob/02dbb044f75f26e2064b586978e230ea31dcd3ab/src/main.ts#L124 This is rather inefficient - use `vault.getAbstractFileByPath` instead.
> 
> https://github.com/nekomeowww/obsidian-asciinema-player/blob/02dbb044f75f26e2064b586978e230ea31dcd3ab/src/main.ts#L143 Avoid the use of HTML directly, especially with user controlled content, which leads to DOM injection vulnerabilities. Instead, build the DOM directly by creating the elements and setting their properties. Obsidian has a set of helper functions to create DOM elements easier using something like `parentEl.createEl/createDiv/createSpan({cls: 'css-class', text: 'innerText'})`. Here you'll want to create the DOM element and use `img.replaceWith(newElement)`.
> 
> https://github.com/nekomeowww/obsidian-asciinema-player/blob/02dbb044f75f26e2064b586978e230ea31dcd3ab/src/main.ts#L203 This might make it a bit difficult to update your plugin in the future (or at least, update the dependency). It might be better to try to bundle the dependency inside your plugin, and/or put some of its code as a base64 string.

Actually, function `getAbstractFileByPath()` and other related file modification functions has no detailed description, such as:
https://github.com/obsidianmd/obsidian-api/blob/763a243b4ec295c9c460560e9b227c8f18d8199b/obsidian.d.ts#L3170
And there's no documentation stated where and how the links `obsidian://` and `obsidian://local/` would work and should processed. I have to use debugger and console.log to test and then find out how the function would work.

Is there any plan or roadmap to state when the script would being update to a more detailed version for developers to be able to reference?

lishid:
> Actually, function getAbstractFileByPath() and other related file modification functions has no detailed description, such as:
https://github.com/obsidianmd/obsidian-api/blob/763a243b4ec295c9c460560e9b227c8f18d8199b/obsidian.d.ts#L3170

Yeah that's my bad - I haven't had a chance to go through all the previous code to add docs to each function. We will try to do that bit by bit.

> And there's no documentation stated where and how the links obsidian:// and obsidian://local/ would work and should processed. I have to use debugger and console.log to test and then find out how the function would work.

This is on purpose - `app://local` links is only an implementation detail for the desktop app because of how Electron works - those links fail to work on other platforms (Android/iOS) because Android uses `content://` with some weird prefix, and iOS uses `file:///var/...path/to/vault`. 

Your use case should only need to interact with the abstract file system as we expose through Vault. `getResourcePath` is normally only used with browser native things, such as `` paths, `<audio>` and `<video>`, etc.

nekomeowww:
> > Actually, function getAbstractFileByPath() and other related file modification functions has no detailed description, such as:
> > https://github.com/obsidianmd/obsidian-api/blob/763a243b4ec295c9c460560e9b227c8f18d8199b/obsidian.d.ts#L3170
> 
> Yeah that's my bad - I haven't had a chance to go through all the previous code to add docs to each function. We will try to do that bit by bit.
> 
> > And there's no documentation stated where and how the links obsidian:// and obsidian://local/ would work and should processed. I have to use debugger and console.log to test and then find out how the function would work.
> 
> This is on purpose - `app://local` links is only an implementation detail for the desktop app because of how Electron works - those links fail to work on other platforms (Android/iOS) because Android uses `content://` with some weird prefix, and iOS uses `file:///var/...path/to/vault`.
> 
> Your use case should only need to interact with the abstract file system as we expose through Vault. `getResourcePath` is normally only used with browser native things, such as `` paths, `<audio>` and `<video>`, etc.

I will update my code later this week to fit the requirements.

> This is on purpose - app://local links is only an implementation detail for the desktop app because of how Electron works - those links fail to work on other platforms (Android/iOS) because Android uses content:// with some weird prefix, and iOS uses file:///var/...path/to/vault.

So this will need more documentations to reference, otherwise developers might not be able to be aware of those things. (especially when the plugin developers have finished their Pull Request to join community-plugin.json and used the **incompatible** APIs to implement their feature for the new versions, there's no code review or suggestion at all)

Even though, I have tested on Linux/macOS/Windows/iPadOS, `app://local` path works fine. Maybe those incompatiblities would come out on Android platform or so.

nekomeowww:
@lishid Updated.

> https://github.com/nekomeowww/obsidian-asciinema-player/blob/02dbb044f75f26e2064b586978e230ea31dcd3ab/src/main.ts#L89
Your plugin should take care of cleaning up itself and removing any additional scripts/styles loaded when it's disabled. Override the onunload() function to perform the cleanup. (Tip: You can keep a reference to each script/style tag you create in your plugin's class to use in your unload function)

Unload is probably impossible for `asciinema-player`, they used the custom element API (ref: https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements) which is unremovable or deletablt with in a window. 
That's why I use: 
```JavaScript
!document.getElementById('asciinema-player-js') && (document.createElement('asciinema-player').constructor === HTMLUnknownElement || document.createElement('asciinema-player').constructor === HTMLElement
```
to determine whether the script file has been injected or not.

> https://github.com/nekomeowww/obsidian-asciinema-player/blob/02dbb044f75f26e2064b586978e230ea31dcd3ab/src/main.ts#L124
This is rather inefficient - use vault.getAbstractFileByPath instead.

I get `null` when I use `vault.getAbstractFileByPath()` when I put a relative path (such as `assets/asciinema-player.cast`) with it.

lishid:
> Unload is probably impossible for asciinema-player, they used the custom element API (ref: https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements) which is unremovable or deletablt with in a window.
That's why I use:

Ah yes, custom elements do kinda suck in that regard. You can probably still remove the <script> and <style> tags from head though? (Also based on my inspection your code won't load the elements if enableOfflineSupport is turned off?

> I get null when I use vault.getAbstractFileByPath() when I put a relative path (such as assets/asciinema-player.cast) with it.

getAbstractFileByPath just looks up the file in a map indexed by the `file.path`, so you should be getting the exact same thing (or null) if you try to iterate through all files and look for `file.path === mypath`.

ericaxu:

PR #579, opened 2021-10-25
Add Graph Analysis plugin

HEmile:

## Repo URL

Link to my plugin: https://github.com/SkepticMystic/graph-analysis
This is a collaborative effort with @SkepticMystic !

## Release Checklist

- [x] I have tested this on Windows, macOS, and Android
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
> We have changed the checkCallback, is this what you mean? SkepticMystic/graph-analysis@32bd6cf

Not quite. You want to perform your conditional check first, regardless of whether `checking` is true or false. What you're doing right now is only running the conditional check when `checking` is true, so in the case where your conditional fails `this.app.workspace.getLeavesOfType(VIEW_TYPE_GRAPH_ANALYSIS).length !== 0`, if `checking` was false, then it was never checked and you'll call `this.initView` even though there's already a view.

> I think this happens because we use Svelte.. it requires that we remove inlineSourceMap from package.json. Not sure if that causes the issue, but it is necessary for Svelte (Marcus Docs)

Maybe you're uploading the file that your dev command is generating? Marcus's docs configures the `build` command to generate source maps without sources (inside `rollup.config.js`).

SkepticMystic:
@lishid I think we implement the checkCallback properly now: https://github.com/SkepticMystic/graph-analysis/commit/73e86c63c9ade879efbe030879694d20fcf6dbdf
And I used `npm run build` to build the final release, I think it doesn't have the sources anymore: https://github.com/SkepticMystic/graph-analysis/commit/e9e2a0f7aa3252b3e09a5ec0f0cf5aba7c1a5b79

lishid:
Weird, I can still reduce the bundle size from 1.72mb down to 891kb when removing the last line... maybe try not outputting any sourcemaps at all?

lishid:
Either way, nothing blocking here, just a matter of squeezing the bundle size. Will merge.

SkepticMystic:
Huh, alright. I will try building without sourcemaps today.
Thank you for all the feedback, Licat :)

lishid:
Happy to help!

PR #581, opened 2021-10-25
Add Custom Attachment Location plugin

RainCat1998:

## Repo URL

Link to my plugin: [https://github.com/RainCat1998/obsidian-custom-attachment-location](https://github.com/RainCat1998/obsidian-custom-attachment-location)

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/RainCat1998/obsidian-custom-attachment-location/blob/30a6f9f31e84411bcb7b2659c0fb786bc495facc/main.ts#L45
I'm not sure this is a great idea. Here you will permanently overwrite these settings, since there are many cases where the app will exit without unloading your plugin. Obsidian will also save these settings to disk once you modify them.

https://github.com/RainCat1998/obsidian-custom-attachment-location/blob/30a6f9f31e84411bcb7b2659c0fb786bc495facc/main.ts#L82
You should use `this.handlePaste.bind(this)` when registering this callback so that you can use `this` to reference your plugin here.

Also, `app` doesn't refer to anything - you should use `this.app` instead. Your code only works because we set a global debugging variable `app`, which shouldn't be normally used.

https://github.com/RainCat1998/obsidian-custom-attachment-location/blob/30a6f9f31e84411bcb7b2659c0fb786bc495facc/main.ts#L119
You can import `moment` from `obsidian` to get a properly typed moment object from the app's library.

lishid:
https://github.com/RainCat1998/obsidian-custom-attachment-location/blob/30a6f9f31e84411bcb7b2659c0fb786bc495facc/main.ts#L78
Oh also, async paste handlers won't work because you must set `event.preventDefault` before your handler function finish synchronous execution. You should go through your `clipBoardItems` first, and set `preventDefault` properly before firing off any async tasks such as storing the file.

RainCat1998:
> https://github.com/RainCat1998/obsidian-custom-attachment-location/blob/30a6f9f31e84411bcb7b2659c0fb786bc495facc/main.ts#L45 I'm not sure this is a great idea. Here you will permanently overwrite these settings, since there are many cases where the app will exit without unloading your plugin. Obsidian will also save these settings to disk once you modify them.
> 
> https://github.com/RainCat1998/obsidian-custom-attachment-location/blob/30a6f9f31e84411bcb7b2659c0fb786bc495facc/main.ts#L82 You should use `this.handlePaste.bind(this)` when registering this callback so that you can use `this` to reference your plugin here.
> 
> Also, `app` doesn't refer to anything - you should use `this.app` instead. Your code only works because we set a global debugging variable `app`, which shouldn't be normally used.
> 
> https://github.com/RainCat1998/obsidian-custom-attachment-location/blob/30a6f9f31e84411bcb7b2659c0fb786bc495facc/main.ts#L119 You can import `moment` from `obsidian` to get a properly typed moment object from the app's library.


I have updated my code, but this plugin still need to overwrite settings and call some fuctions which is not exported in obsidian-api. Otherwise I will rebuild many functions which is already implimented in obsidian (such as saving attachment and generating md links). I think it is a temporary solution before newer obsidian-api release.

lishid:
https://github.com/RainCat1998/obsidian-custom-attachment-location/blob/9a13396257917ae49befd783e62346ab84fb2e13/main.ts#L91
These `ts-ignore` shouldn't be necessary anymore.

> I have updated my code, but this plugin still need to overwrite settings and call some fuctions which is not exported in obsidian-api. Otherwise I will rebuild many functions which is already implimented in obsidian (such as saving attachment and generating md links). I think it is a temporary solution before newer obsidian-api release.

That's fine - but I highly suggest adding a big red text at the top of your plugin's README file to let the user know that this plugin will override the two config options - and it will likely not be able to restore it given how the setting system works. This way you will avoid having angry users who wastes hours to find out why their settings don't get saved.

PR #583, opened 2021-10-26
added obsidian-crypto-lookup plugin to community-plugins.json

kinabalu:

## Repo URL

Link to my plugin: https://github.com/kinabalu/obsidian-crypto-lookup

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/kinabalu/obsidian-crypto-lookup/blob/bfb2916d03930ae46194c1ede53a0b6673619acb/main.ts#L81

Actually, I don't recommend `await`ing the preloadCurrencies at all, since the plugin loading process delays Obsidian's loading process until it's done before presenting the workspace to the user. Since this could depend on reading files and especially loading a network location, I highly suggest running this in parallel but not awaiting it (call the function but don't `await`).

In fact, I would go one step further and suggest you to lazy-load them - since presumably a user with your plugin may not use your command until later, you can probably only load it when `this.currency` isn't yet loaded.

kinabalu:
for now I've commented out that block of code as it is unused and will be used in a future feature and fixed the conflilcts with base branch.

lishid:
Sounds good!

PR #589, opened 2021-10-28
Add Extended Multi Cursor plugin

ollyfg:
Simpler than `advanced-cursors`, this plugin brings some sublime text keybindings and allows click-to-deselect on existing selections.

Related: https://forum.obsidian.md/t/ability-to-multi-cursor-select-with-keyboard/7665/11

Sorry I can't test on platforms other than MacOS I'm stuck at home due to covid. Isn't the point of Electron + JS to be platform agnostic unless you get into some really advanced stuff? Happy to wait until I've tested on all platforms though, but it'll be a few months.

## Repo URL

Link to my plugin: https://github.com/ollyfg/obsidian-multi-cursor

## Release Checklist

- [ ] I have tested this on Windows, macOS, and Linux _(if applicable)_ (Nope! Only have macOS available...)
  - [X] `main.js`
  - [X] `manifest.json`
  - [x] `styles.css` _(optional)_ (N/A)

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2731,5 +2731,13 @@
         "description": "Use multiple cursors even more powerfully.",
         "author": "SkepticMystic",
         "repo": "SkepticMystic/advanced-cursors"
+    },
+    {
+        "id": "obsidian-multi-cursor",
+        "name": "Extended Multi Cursor",
+        "description": "This plugin enhances the default multi-cursor implementation with Sublime Text style keybindings. It also allows you to click on an existing selection to de-select it.",
+        "author": "Oliver Fawcett",
+        "repo": "ollyfg/obsidian-multi-cursor",
+        "branch": "master"
     }
 ]


Taitava:
I have now tested version `1.0.1` on:
- Windows 10
- Ubuntu 20.04 (Linux)

Works on both of them! 🌞

lishid:
https://github.com/ollyfg/obsidian-multi-cursor/blob/92abd8442b7a8b8da3667bfeb52e68660f5b47dc/main.ts#L20

Please rename your class and interface names from the sample plugin to your actual plugin's name.

https://github.com/ollyfg/obsidian-multi-cursor/blob/92abd8442b7a8b8da3667bfeb52e68660f5b47dc/main.ts#L22
Don't keep references to CodeMirror editors as this may cause memory leaks when editor views are opened and closed.

Instead, you can find the latest active CodeMirror instances using `workspace.iterateCodeMirrors(callback)` when unloading your plugin to unhook events.

https://github.com/ollyfg/obsidian-multi-cursor/blob/92abd8442b7a8b8da3667bfeb52e68660f5b47dc/main.ts#L57
This will not work on mobile at all - you won't be getting an instance of CodeMirror.Editor because the mobile app does not use CodeMirror 5. Maybe you should just set `isDesktopOnly` in your manifest.json so as to only work with the CM5 editor.

ollyfg:
Hi there!

Thanks for the feedback, I can't wait for some good docs around the API interface!

I'm withdrawing this plugin so we can all pull behind https://github.com/SkepticMystic/advanced-cursors - no point re-inventing the wheel, and the unique bits of this plugin should be fairly easy to port across.

PR #594, opened 2021-10-28
Add AnkiBridge

JeppeKlitgaard:

## Repo URL

Link to my plugin: https://github.com/JeppeKlitgaard/ObsidianAnkiBridge

## Release Checklist

  - [X] `main.js`
  - [X] `manifest.json`


lishid:
https://github.com/JeppeKlitgaard/ObsidianAnkiBridge/blob/bc3f3403b74cb8c5f1031d09bbe8fcf0fafc5592/src/main.ts#L70
Your plugin does not properly unload this function.

https://github.com/JeppeKlitgaard/ObsidianAnkiBridge/blob/bc3f3403b74cb8c5f1031d09bbe8fcf0fafc5592/src/main.ts#L292
Not sure if concurrent processing is a good idea. I know it can perform faster, but I'd generally advise to use sequential processing for batch processes to avoid race conditions and resource contention.

https://github.com/JeppeKlitgaard/ObsidianAnkiBridge/blob/bc3f3403b74cb8c5f1031d09bbe8fcf0fafc5592/src/utils/grammar.ts#L3
This function doesn't seem to need to be async, along with the whole call tree from [here](https://github.com/JeppeKlitgaard/ObsidianAnkiBridge/blob/bc3f3403b74cb8c5f1031d09bbe8fcf0fafc5592/src/main.ts#L27)

JeppeKlitgaard:
Thank you very much for having a look at this. I will fix these soon.

JeppeKlitgaard:
@lishid 

1 and 3 are fixed in 0.2.20.

I am on the fence about 2, since there shouldn't be any race conditions or contention since each promise runs on a separate file. Will change this in the future and compare performance benefit vs the additional code complexity.

lishid:
Sorry for getting back to you this late!

https://github.com/JeppeKlitgaard/ObsidianAnkiBridge/blob/b294381ea1ebd42221cbd17286b45bc6d66ae143/src/main.ts#L168
I think you forgot to restore the previous callback function when tearing down the save command override.

JeppeKlitgaard:
No worries! My term is ending soon and I plan to make some more improvements to AnkiBridge over the break and also improve the documentation a fair bit.

I will fix that callback soon, probably this weekend. Thank you so much for having a look and for making Obsidian so fantastic!

JeppeKlitgaard:
@lishid 

Would it be possible to add a callback for an "explicit save" or "user-initiated save" event?

This hacky approach I am using is taken from the `linter` plugin, but it doesn't easily scale and multiple plugins using the same approach override each other.

I think there is a use-case for an event that is triggered whenever the user explicitly requests a save (i.e. through CTRL + S, the menu, the save file command, etc). Typically this will be to signal to plugins that "I have changed something in the file, please consider the changes".

These should not be triggered by automatic saves, which in many cases could lead to UX and performance issues.

Currently I plan to remove the sync-on-save functionality.

lishid:
Yeah I can look into that later - for now I don't think we will facilitate plugins from messing with the document data on save because a defective plugin or an edge case the plugin author failed to consider could be an easy way to introduce data loss.

JeppeKlitgaard:
Yea, that would probably be quite complicated as well. I was more thinking of the event as something that would poke plugins and say "Hey, the user just explicitly saved."

Messing with data could then happen after saving, if needed. Once a plugin has changed the data it wants, it could trigger an implicit save.

lishid:
Ah I see what you mean - I would still consider making it a synchronous editor operation to have plugins execute before the file gets saved, to avoid possible race conditions between the editor itself and plugins trying to change the data.

Another potential issue is that plugins might do some kind of re-scan of the entire file when that happens, to parse out things. Obsidian's editor saves every 2 seconds, and so providing this kind of hook could end up making the app consume a lot of CPU.

JeppeKlitgaard:
You're probably right that it might make sense for the plugins to be able to change the data before the file is saved to disk.

The thought behind distinguishing "explicit" (user-initiated, for example CTRL + S, "save file" command, etc) and "implicit" (automatic saves by Obsidian) was to avoid this being triggered for every automatic save. It would for example be silly to have AnkiBridge sync with Anki every 2 seconds, but certainly it is muscle memory for me to press "CTRL + S" when I have made a significant change that could justify a resync.

lishid:
I think while you want your plugin to behave that way, it's not clear to a clueless user what the distinction between ctrl+s vs autosave. I'm totally with you on the muscle memory of ctrl-s (I actually do that accidentally to web apps often and triggers the "save webpage" UI more often than I'd like).

I think implementing something that only works when ctrl+s happens is going to be more confusing (to the average user) and you are likely to get a lot of bug reports that your functionality "isn't working" because the user doesn't understand the difference.

lishid:
IMO implementing a command is probably your best bet. It's by no means ideal, especially when multiple plugins want to do something before saving, but for now it's the best workaround.

JeppeKlitgaard:
For now the solution I have ended up going with is simply removing "sync-on-save" functionality, as I don't like the way it ended up.

There is already a command to do the syncing, so users can use that.

I will probably document an alternative approach, which I am currently using. It involves making a custom macro using https://github.com/phibr0/obsidian-macros and then binding it to CTRL + S. This allows me to trigger both `linter`, `AnkiBridge`, and Obsidian save with one command, which is pretty nifty.

JeppeKlitgaard:
In the just-released v.0.2.21 the sync-on-save functionality has been removed (thus resolving the last outstanding feedback).

I am working on writing tests for the parsers after which I will probably add:
- Better documentation/demos
- Image support
- Cloze support

In that order.

I hope to get audio and video support done soonish too.

Up to you whether it is worthy of being on the community plugin list yet or whether you want to hold off until more features have been added. 

Thank you so much for having a look and being so helpful!

PR #596, opened 2021-10-29
Community Plugin: Lineup Builder

James-Fallon:

## Repo URL

Link to my plugin: https://github.com/James-Fallon/obsidian-lineup-builder

## Release Checklist

- [x] I have tested this on Windows, macOS, and iOS
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/James-Fallon/obsidian-lineup-builder/blob/33b52fd983c708e24d79329a359f7b2feb90b329/src/svg-builder.ts#L26
Here you run the risk of a DOM injection attack - you are directly putting in user-provided strings to the HTML text that is then used to render the DOM. You should either sanitize all user inputs before it makes it into the HTML string, or use DOM building functions to create elements and append text into it.

James-Fallon:
@lishid Thanks for the comment, good spot. I have implemented encoding of user input on your suggestion, as seen in this commit: https://github.com/James-Fallon/obsidian-lineup-builder/commit/548a9251c6ed0b3200ba273cfe54eb37b411ebff

I've since added a new release version 1.0.1 with the fix and updated the manifest.json and version file accordingly.

lishid:
Sorry it took so long for the review!

PR #604, opened 2021-10-30
Add "Luna: Dark Mode Switcher"

selfire1:

## Repo URL

Link to my plugin: https://github.com/selfire1/obsidian-luna-dark-mode

## Release Checklist

- [x] I have tested this on Windows, macOS, and Linux _(if applicable)_ (Not tested on Windows!)
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2738,5 +2738,12 @@
         "description": "Translate code blocks into simple English using Machine Learning with the Stenography API",
         "author": "bramses",
         "repo": "bramses/stenography-obsidian"
+    },
+    {
+        "id": "obsidian-luna-dark-mode",
+        "name": "Luna: Dark Mode Switcher",
+        "description": "Automate switching to dark and light mode.",
+        "author": "Joschua",
+        "repo": "selfire1/obsidian-luna-dark-mode"
     }
 ]


lishid:
https://github.com/selfire1/obsidian-luna-dark-mode/blob/4f4e09579905fcf30bd336a850b029040c34ea95/main.ts#L120
This is a really poor (read: difficult to understand) way of using `var` to jump up the closure block.

https://github.com/selfire1/obsidian-luna-dark-mode/blob/4f4e09579905fcf30bd336a850b029040c34ea95/main.ts#L162
This also is not unloaded correctly.

Honestly with the way you're managing the intervals, why not register a global interval in your onload which checks the config and determines how to set the current mode, rather than having different intervals running different callbacks?

https://github.com/selfire1/obsidian-luna-dark-mode/blob/4f4e09579905fcf30bd336a850b029040c34ea95/main.ts#L155
This is redundant from the next line, right?

https://github.com/selfire1/obsidian-luna-dark-mode/blob/4f4e09579905fcf30bd336a850b029040c34ea95/main.ts#L305
Just use `changeTheme` instead of `setTheme` which takes care of setting the config and triggering the event for you.

selfire1:
Hi @lishid,

thanks for taking the time to go through the plugin. I've rewritten the code and internal logic based on your comments (and learned a lot about writing code on the way!).

### 1 + 2: Interval logic
I wasn't able to set the interval as a global variable, so I instead wrote a function. [See Line 98](https://github.com/selfire1/obsidian-luna-dark-mode/blob/069e08e3d55b56fa869939ce5f59c383a1aa462f/main.ts#L98). It gets called to `set` and `clear` a singular interval.

### 3: Redundant line
You're totally right, I deleted the line in question.

### 4: `changeTheme` instead of `setTheme`
Done! See [L238](https://github.com/selfire1/obsidian-luna-dark-mode/blob/069e08e3d55b56fa869939ce5f59c383a1aa462f/main.ts#L238).

As far as I can tell, the remarks were resolved. Currently there is quite a bit of `console.log` in the code which I would clean up once the plugin passes approval.

Thanks again for your time!

lishid:
Sorry it took so long for the review!

https://github.com/selfire1/obsidian-luna-dark-mode/blob/8b17420524ad4c999d71c839553808894e2e6e04/main.ts#L99
This doesn't work either, because the variable here is bound to each individual execution of the function, so in essence, when you call `startTimeInterval(false)` it will have nothing in this variable.

When I said global interval, I actually meant single interval - have your plugin register an interval in the `onload` function regardless of what mode the configuration is. Have the interval registered with Obsidian using `this.registerInterval(setInterval(callback, timeout))`, then inside the callback function, check what mode the plugin is configured for.

ericaxu:

PR #605, opened 2021-10-30
add copy-publish-url

kometenstaub:

## Repo URL

Link to my plugin: https://github.com/kometenstaub/copy-publish-url

## Release Checklist

- [x] I have tested this on macOS and Linux _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/kometenstaub/copy-publish-url/blob/5a65f4467aa60622ea8405a9cbd982688393f6ed/src/main.ts#L40
You should change this `callback` to a `checkCallback` with this condition.

https://github.com/kometenstaub/copy-publish-url/blob/5a65f4467aa60622ea8405a9cbd982688393f6ed/src/main.ts#L47
Triple equals to false means it can't be undefined, by definition, so you just need to check if it's `=== false`.

kometenstaub:
I have implemented the `checkCallback` and removed the unnecessary check.

PR #607, opened 2021-10-30
Community Plugin: Language Translator

twentytwokhz:

## Repo URL

Link to my plugin: https://github.com/twentytwokhz/language-translator

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/twentytwokhz/language-translator/blob/a77c5461559f5f8c9f135d6dcf96eb1d1d8f8575/src/main.ts#L58
We don't recommend setting default hotkeys because it might conflict with something the user already set for something else.

Taitava:
> https://github.com/twentytwokhz/language-translator/blob/a77c5461559f5f8c9f135d6dcf96eb1d1d8f8575/src/main.ts#L58 We don't recommend setting default hotkeys because it might conflict with something the user already set for something else.

Not only related to this particular plugin, but my opinion is that it would be nice if Obsidian could prevent introducing a new hotkey, if a user has already defined the hotkey to do something else. This would concern both plugins and new versions of Obsidian that introduce new default hotkeys. I have opened [a feature request about it in the forums](https://forum.obsidian.md/t/omit-new-default-hotkeys-if-a-hotkey-is-already-in-use/26380).

Now that you mentioned that assigning default hotkeys is not recommended, I'd also like to suggest that this recommendation would be present in the API documentation in the following places:

- https://github.com/obsidianmd/obsidian-api/blob/5687a75ecfdb78b6a832127cf6f9229e27e196a0/obsidian.d.ts#L1296-L1298
- https://github.com/obsidianmd/obsidian-api/blob/5687a75ecfdb78b6a832127cf6f9229e27e196a0/obsidian.d.ts#L466-L469

Thanks. 🙂

lishid:
Agreed yeah, we'll figure something out!

lishid:
Ok, I have implemented both of your suggestions and they will be live in the app and API in the next release.

PR #608, opened 2021-10-30
Automatically validate newly added Plugins.

phibr0:
This GitHub Action will run when a Pull Request is made that modifies the `community-plugins.json`. It will check for simple Errors, like having mismatching ID's. In its current implementation the Action won't fail, but instead add a Comment to the PR explaining the Errors like so:

image

### DIFF
--- a/.github/workflows/validate_plugin.yml
+++ b/.github/workflows/validate_plugin.yml
@@ -1,0 +1,61 @@
+name: Plugin Validation
+
+on:
+  pull_request:
+    branches:
+      - "master"
+    paths:
+      - 'community-plugins.json'
+
+jobs:
+  plugin-validation:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+      - run: npm install axios
+      - uses: actions/github-script@v5
+        with:
+          script: |
+            const fs = require('fs');
+            const axios = require('axios');
+            const author = context.payload.sender.login
+            const plugins = JSON.parse(fs.readFileSync('community-plugins.json', 'utf8'));
+            const plugin = plugins[plugins.length - 1];
+
+            let reply = `### Hello ${plugin.author}! <a href="https://obsidian.md"><img align="right" height="30px" src="https://user-images.githubusercontent.com/59741989/139557624-63e6e31f-e617-4041-89ae-78b534a8de5c.png"/></a>\n**I found the following Errors in your Plugin, ${plugin.name}:**\n\n`;
+            if (context.payload.pull_request.changed_files > 1) {
+                reply += `:x: You modified other Files.\n`
+            }
+
+            await axios.get(`https://api.github.com/repos/${plugin.repo}`)
+              .then(async () => {
+                await axios.get(`https://api.github.com/repos/${plugin.repo}/releases`)
+                  .then(async (releases) => {
+                    if(!releases.data[0].assets.find(p => p.name == "main.js")) {
+                      reply += 'Your latest Release is missing the `main.js` File.\n';
+                    }
+                    if(!releases.data[0].assets.find(p => p.name == "manifest.json")) {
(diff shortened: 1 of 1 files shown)


phibr0:
What other Errors should it be able to detect?

.github/workflows/validate_plugin.yml
```diff
@@ -0,0 +1,71 @@
...
+              reply += `\n---\n<sup>This check was done automatically and might not be correct. If something went wrong, please ping \`@phibr0\`.`;
+              return reply;
+            }
+            
+            if (context.payload.pull_request.changed_files > 1) {
+                addError('You modified other Files.');
+            }
+
+            await axios.get(`https://api.github.com/repos/${plugin.repo}`)
+              .then(async () => {
```
lishid, line 44: I think these then/catch chains can be simplified since you're already using await anyway.

.github/workflows/validate_plugin.yml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          node-version: 14
+      - run: npm install axios
+      - uses: actions/github-script@v5
+        with:
+          script: |
+            const fs = require('fs');
+            const axios = require('axios');
+            const author = context.payload.sender.login
+            const plugins = JSON.parse(fs.readFileSync('community-plugins.json', 'utf8'));
```
lishid, line 25: Maybe also validate this? I'm not sure if there's a way but sometimes plugin authors don't add their plugin last. I was hoping there'd be a way to tell based on the PR but I don't think that's possible?

lishid (changes requested):

.github/workflows/validate_plugin.yml
phibr0: One could check if the author is the same as the author in the repo field. I think that would be good enough, right?

lishid: Could add that as a warning - sometimes people do use a different name as their github username. Another thing I had in mind is just validating the JSON which we have a check for as well (here maybe just add a try-catch).

phibr0: What I meant is using the author thats inside the repo string (`<author>/<repo>`). That should be the same as the one opening the PR right?

lishid: Ohh yes you're right! Most of the time it matches. I've only seen once or twice where one collaborator owns the repo and another collaborator submits the PR, but that should be fine.

claremacrae:
@joethei pointed this PR out in relation to #923, which arose as a result of a user finding that a plugin didn't work in the plugin store...

What would be the minimum necessary for this PR to be activated as an experiment?

lishid:
I can pull it in and do some work on it.

lishid:
@phibr0 can you invite me to your patch branch

phibr0:
Uhh, the branch doesn't exist anymore it seems

lishid:
Ok, I got a version but I don't know what's the best way to test... Any recommendations? Should I just push it and see what happend?

phibr0:
I always tested it using my fork, and creating PR's on that one.

claremacrae:
Nice... I was going to ask if the code could be run on the command line, and test on some files with the kinds of breakages it's coded for.... But I see that the code is entirely in the YAML, so that's not going to immediately work.

claremacrae:
I think what argentum did for some checks in the Hub was first to turn them on, but not configure the repo to fail PRs if the tests failed..

That way, it's not a problem if there's an error in the action.

And then after a few PRs to give you confidence, you can then configure the repo to require the checks to pass before a PR can be merged....

claremacrae:
You could always prime a few people to submit broken PRs too!

claremacrae:
In summary, I would say yes, just push it!

liamcain:
Hi @phibr0! Thanks again for doing all the legwork with this! We ended up using a [forking this](https://github.com/obsidianmd/obsidian-releases/blob/master/.github/workflows/validate-plugin-entry.yml) and testing it in a separate pull request to resolve some permission issues, but since then it has been working beautifully.

PR #610, opened 2021-10-31
Add Sapphire theme

gracejoseph1236:
# I am submitting a new Community Theme

## Repo URL

Link to my theme: https://github.com/gracejoseph1236/obsidian-sapphire

## Theme checklist

- [x] My repo contains all required files (please *do not* add them to this `obsidian-releases` repo)
  - [x] `obsidian.css`
  - [x] The screenshot file
- [x] I have indicated which modes (dark, light, or both) are compatible with my theme
- [x] If my repo is not using the `master` branch, please indicate in `community-themes.json` with `"branch": "main"`.

### DIFF
--- a/community-css-themes.json
+++ b/community-css-themes.json
@@ -724,5 +724,13 @@
         "screenshot": "screenshot.png",
         "modes": ["dark"],
         "branch": "main"
+    },
+    {
+        "name": "Sapphire",
+        "author": "Piglet1236",
+        "repo": "gracejoseph1236/obsidian-sapphire",
+        "screenshot": "example.png",
+        "modes": ["dark"],
+        "branch": "main"
     }
 ]


ericaxu:
This is too similar to Carnelian and Ruby themes. Have you considered merging them somehow?

gracejoseph1236:
I did consider merging them but not exactly sure what to do

ericaxu:
I'm afraid just changing the accent color isn't enough for a new theme in the store. For example you can make your theme work together with Style Setting plugin and let the user define which color they like, instead of submitting slight variations of the same theme to offer the additional color choices.

I hope that makes sense!

gracejoseph1236:
Yeah I was actually thinking of doing that. If I do end up doing it I'll
probably include Emerald and the other two themes, but in that case should
I remove te other themes?

ericaxu:
> Yeah I was actually thinking of doing that. If I do end up doing it I'll probably include Emerald and the other two themes, but in that case should I remove te other themes?

Yes, if you could consolidate them into a single theme it would be great!

PR #614, opened 2021-10-31
add linked-data-vocabularies

kometenstaub:
# I am submitting a new Community Plugin

## Repo URL

Link to my plugin: https://github.com/kometenstaub/obsidian-linked-data-vocabularies

## Release Checklist

- [x] I have tested this on Windows and Linux
- [x] My GitHub release contains all required files
  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css` _(optional)_
- [x] GitHub release name matches the exact version number specified in my manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
- [x] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [x] README clearly describes the plugins purpose and provides clear usage instructions.
- [x] I have added a license in the LICENSE file.

PR #616, opened 2021-10-31
add linked data helper

kometenstaub:

## Repo URL

Link to my plugin: https://github.com/kometenstaub/linked-data-helper

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3044,5 +3044,13 @@
         "author": "Matter",
         "description": "Official Matter Obsidian integration.",
         "repo": "getmatterapp/obsidian-matter"
+    },
+    {
+        "id": "linked-data-helper",
+        "name": "Linked Data Helper",
+        "description": "Generate the necessary data for Linked Data Vocabularies.",
+        "author": "kometenstaub",
+        "repo": "kometenstaub/linked-data-helper",
+        "branch": "main"
     }
 ]


kometenstaub:
Please only merge this together with #614 since this plugin is only useful for generating the needed data for the other plugin.

lishid:
https://github.com/kometenstaub/linked-data-helper/blob/e7247a953640d45bcc1256157cf3c9cce36dc158/src/methods/methods-loc.ts#L20

This doesn't seem necessary. 

https://github.com/kometenstaub/linked-data-helper/blob/e7247a953640d45bcc1256157cf3c9cce36dc158/src/methods/methods-loc.ts#L27
I am very confused by what's happening here.

https://github.com/kometenstaub/linked-data-helper/blob/e7247a953640d45bcc1256157cf3c9cce36dc158/src/methods/methods-loc.ts#L55
Why not just use the vault API?

lishid:
I find it weird that your code is littered with `@ts-ignore`. There's definitely something wrong with your type definitions here.

It's also quite weird how you're mixing NodeJS's file system APIs along side Obsidian's Vault and Adapter API - if you have specific needs you can use Node but generally you should stick to one.

kometenstaub:
> https://github.com/kometenstaub/linked-data-helper/blob/e7247a953640d45bcc1256157cf3c9cce36dc158/src/methods/methods-loc.ts#L20
> This doesn't seem necessary.

If I don't do it like that, I get this error:
`Property 'getBasePath' does not exist on type 'DataAdapter'.`

> https://github.com/kometenstaub/linked-data-helper/blob/e7247a953640d45bcc1256157cf3c9cce36dc158/src/methods/methods-loc.ts#L27
> I am very confused by what's happening here.

I actually didn't need that, I already had that path from where I called that function, thanks!

> https://github.com/kometenstaub/linked-data-helper/blob/e7247a953640d45bcc1256157cf3c9cce36dc158/src/methods/methods-loc.ts#L55
> Why not just use the vault API?

For the record (because already discussed on Discord), the file that is parsed has a size of 2.5GB.

> I find it weird that your code is littered with @ts-ignore. There's definitely something wrong with your type definitions here.

I've changed them to `@ts-expect-error` and explained why they are there or have improved the typings.
Some things were properly typed, but TS couldn't infer some things from conditions, so I assigned them to variables (instead of getting the values again from the object with the key) or changed the conditions.
See the file [here](https://github.com/kometenstaub/linked-data-helper/blob/main/src/methods/methods-loc.ts).


> It's also quite weird how you're mixing NodeJS's file system APIs along side Obsidian's Vault and Adapter API - if you have specific needs you can use Node but generally you should stick to one.

Node is used for reading the file that is outside of the vault. The generated files will be placed inside of the vault (so that it also works on mobile), so I'm using the vault API for that. That has the benefit that, if users want to change the path for writing these files in the vault, that they only have to specify the path from the vault root instead of the absolute path.

lishid:
> If I don't do it like that, I get this error:
Property 'getBasePath' does not exist on type 'DataAdapter'.

Ah you're right, I was thinking of something else that didn't apply here. The notice should probably say that this plugin only works on the desktop app though.

> For the record (because already discussed on Discord), the file that is parsed has a size of 2.5GB.

Yup ok that makes sense.

> Node is used for reading the file that is outside of the vault. The generated files will be placed inside of the vault (so that it also works on mobile), so I'm using the vault API for that. That has the benefit that, if users want to change the path for writing these files in the vault, that they only have to specify the path from the vault root instead of the absolute path.


Ok that's fair.

kometenstaub:
> Ah you're right, I was thinking of something else that didn't apply here. The notice should probably say that this plugin only works on the desktop app though.

But it won't even be installable on mobile because of the desktop flag in the manifest?

lishid:
Yeah that should be the case, I'm just thinking about the error message you're showing when something unexpected happens

kometenstaub:
Ok, I'll think about how to word it better.

kometenstaub:
> Yeah that should be the case, I'm just thinking about the error message you're showing when something unexpected happens

Actually I'm not using that function anymore, so there's no message to be displayed.

I also rebased it.

PR #619, opened 2021-10-31
add Auto Class plugin

OfficerHalf:

## Repo URL

Link to my plugin: https://github.com/OfficerHalf/obsidian-auto-class

## Release Checklist

  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2855,5 +2855,13 @@
         "author": "Tim Hor & Andrew Brown",
         "description": "Manipulate sentences as a unit of movement",
         "repo": "timhor/obsidian-sentence-navigator"
+    },
+    {
+        "id": "auto-class",
+        "name": "Auto Class",
+        "description": "Automatically apply CSS classes to markdown views based on a note's path.",
+        "author": "OfficerHalf",
+        "repo": "OfficerHalf/obsidian-auto-class",
+        "branch": "main"
     }
 ]


lishid:
https://github.com/OfficerHalf/obsidian-auto-class/blob/6634bde1ac6f6d3004f496bd41222a7e2882de42/src/plugin.ts#L92
The official API for this is `MarkdownView.getMode()`.

https://github.com/OfficerHalf/obsidian-auto-class/blob/6634bde1ac6f6d3004f496bd41222a7e2882de42/src/plugin.ts#L110
This isn't really a good way to keep track of things because you now run the risk of memory leaks by holding onto MarkdownView's that are no longer used by the app. I recommend using a global `WeakMap<MarkdownView, string[]>` and then iterate through all current MarkdownView in the workspace, rather than keeping a list of previously seen MarkdownViews.
Just gotta make sure that you properly remove the previous classes before applying new ones.

OfficerHalf:
Done and done, released as version 1.5.3. I wasn't sure the best way to get a reference to all current markdown views to loop through them (no response in the dev channel yet) so it looks like this right now:
`const views = this.app.workspace.getLeavesOfType('markdown').map((leaf) => leaf.view) as MarkdownView[];`

I can adjust that if there's a better way but it functions and no more memory leak. Good catch - I figured it was fine since I wasn't saving them and the list would get cleared when Obsidian closed, but I hadn't thought about garbage collection.

OfficerHalf:
Updated to resolve conflicts and slightly adjusted the description as the scope has expanded since this was first released. 😄

lishid:
https://github.com/OfficerHalf/obsidian-auto-class/commit/627bc7b8529ea027745f5e14eae942d06705a9eb#diff-57716bc53661ccdb4571d571f1a5493294848965ae6068dbb3d9a1167d2cc1a6R134

Easier if you just use `forEach` on the list of leafs rather than mapping it to `leaf.view` and then forEach on the resulting list. Could also add an instanceof check for type coercion.

OfficerHalf:
Makes sense. Refactored it.

PR #633, opened 2021-11-08
add obsidian-downfile

WHG555:
Repo URL
Link to my plugin: https://github.com/WHG555/obsidian-downfile

Release Checklist

- [x]  I have tested this on Window

 My GitHub release contains all required files
 main.js
 manifest.json
 styles.css
 GitHub release name matches the exact version number specified in my manifest.json (Note: Use the exact version number, don't include a prefix v)
 The id in my manifest.json matches the id in the community-plugins.json file.
 README clearly describes the plugins purpose and provides clear usage instructions.
 I have added a license in the LICENSE file.

这个插件是给在中国的ob使用人员来使用，方便我们下载插件、主题。同时添加了一个CSS样式的库，可以更方便的分享，下载CSS样式。

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2797,6 +2797,13 @@
         "author": "RainCat1998",
         "repo": "RainCat1998/obsidian-custom-attachment-location"
     },
+     {
+        "id": "obsidian-downfile",
+        "name": "资源库",
+        "author": "whghcyx",
+        "description": "down plugin,theme,css in china",
+        "repo": "WHG555/obsidian-downfile"
+    },
     {
         "id": "cryptsidian",
         "name": "Cryptsidian",


lishid:
There's a lot of common mistakes in this plugin. It would help a lot if you can read this document and revise the plugin: https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md

- Avoid using NodeJS's fs directly since your plugin doesn't need anything specific to it, just stick to using the adapter API.
- In general, use one single forward slash to delimit folders, like `/`. In your plugin you sometimes use `//` and sometimes use `\\`.
- Maybe it makes more sense to use a GitHub proxy rather than having you manually copy over/update all the plugins to your repos?

lishid:
Also I would recommend renaming the plugin to something more in-line with what it does (to facilitate downloading themes/plugins for Chinese users who aren't able to access GitHub). `obsidian-downfile` doesn't sound like it has anything to do with that.

ericaxu:

PR #634, opened 2021-11-09
Added CookLang plugin

deathau:

## Repo URL

Link to my plugin: https://github.com/deathau/cooklang-obsidian

## Release Checklist

  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/main.ts#L35
Instead of using activeLeaf (without a null check too!) you should use `getLeaf()`.

https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/main.ts#L44
Here you can use `getLeaf(true)`

https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/main.ts#L58
Here you should always perform the `file.extension === 'md'` check regardless of whether `checking` is true or false. In your case, if your function was called with `checking = false` then this callback will fail to properly handle non-md files.

https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/main.ts#L70
Use `getConfig('newFileLocation')` instead which gives a default fallback.

https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/main.ts#L75
Use getActiveFile and test for null.

https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/main.ts#L101
Use `instanceof` instead of `as`.

https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/main.ts#L111
Normally, obsidian icons should be resized to `0 0 100 100`.

https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/cookView.ts#L21
This should just be `this.contentEl` (it's defined on the super-class View).

https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/cookView.ts#L61
Test for null activeLeaf and also use `instanceof` check.

https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/cookView.ts#L62
The best way to change modes normally is to use `getViewState`, modify the mode, then `setViewState` it back.

https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/cookView.ts#L149
Use `element.empty()`

https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/cookView.ts#L198
You can use `parentElement.createSpan()` to create elements directly as a child of the parent element. You can also use `createEl({cls: 'css-class', text: 'inner text'})` to make your code less verbose.

deathau:
I'm going through and making all the changes suggested. Most of them make a lot more sense than what I was doing, lol.

I do have some specific responses, though:
https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/main.ts#L70
`getConfig()` is not exposed, either, correct? I couldn't find it in the ts definitions...

https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/main.ts#L111
I understand that `0 0 100 100` is the norm, but are there any specific reasons? I basically just grabbed this icon from CookLang themselves, and the whole point of SVG is that it scales automatically without me having to do the work of scaling up all the coordinate values, right?

https://github.com/deathau/cooklang-obsidian/blob/c2d3d6622c26b94f732f9e2b231b9c91209fefb1/src/cookView.ts#L62
Are there any resources (or good examples you can think of) for using `getViewState` and `setViewState`? I initially figured that they _would_ be the best way to change modes, but I couldn't figure out if there was an event or something that got triggered when the view state changed...

I've now made all the suggested changes except the SVG thing and the state stuff. I'll wait for your response on those.

lishid:
> getConfig() is not exposed, either, correct? I couldn't find it in the ts definitions...

Correct, and this is the right API to use once we expose it.

> I understand that 0 0 100 100 is the norm, but are there any specific reasons? I basically just grabbed this icon from CookLang themselves, and the whole point of SVG is that it scales automatically without me having to do the work of scaling up all the coordinate values, right?

Hmm I'm not exactly sure I understand how svg works but if it's working then let's leave it :)

> Are there any resources (or good examples you can think of) for using getViewState and setViewState? I initially figured that they would be the best way to change modes, but I couldn't figure out if there was an event or something that got triggered when the view state changed...

I think the docs are a bit scarce on them at the moment, but they will trigger the setState/getState functions on your view which is likely the event you're thinking of.

PR #635, opened 2021-11-09
remove obsidian-todo-txt from community plugins

trashhalo:
There isn't a good way in the obsidian api right now to implement a new text format that has similar behavior to MarkdownSourceView respecting the various editor settings. Lots of custom logic for token parsing is baked into hypermd codemirror mode. example [[foo]] printing to be clickable internal link. As an implementation I originally extended from that view and modify the parts I needed for todo-txt but its constantly breaking as changes go out.

I give up.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1110,13 +1110,6 @@
         "branch": "main"
     },
     {
-        "id": "todo-txt",
-        "name": "Todo.txt support",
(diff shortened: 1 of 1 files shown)


lishid:
Aww that's sad... It does make sense though, we've spent a lot of time on the editor and it's pretty difficult to replicate - if you come back to it one day, let me know how it goes!

PR #651, opened 2021-11-21
add obsidian-tressel community plugin

aseem-thakar:

## Repo URL

Link to my plugin: https://github.com/aseem-thakar/obsidian-tressel/

## Release Checklist
  - [x]  Windows
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/aseem-thakar/obsidian-tressel/blob/4d8e9e6ac7e447883a3e116a570b79e61e669e92/main.ts#L1
Does this work on mobile devices? I'm guessing it's co-built for node and web, but it's probably using fetch or XMLHttpRequest under the hood.

I would recommend using the request API that Obsidian exposes, which ignores any CORS restrictions, but if your server responds to CORS properly then either should work.

https://github.com/aseem-thakar/obsidian-tressel/blob/4d8e9e6ac7e447883a3e116a570b79e61e669e92/main.ts#L56
Don't recommend keeping a reference of this in your plugin class. You just need an inline variable for this use case.

https://github.com/aseem-thakar/obsidian-tressel/blob/4d8e9e6ac7e447883a3e116a570b79e61e669e92/main.ts#L64
Highly recommend using the `vault` API instead of the adapter API when possible - here you can use `getAbstractFileByPath` and check for whether the result is `instanceof TFolder`.

https://github.com/aseem-thakar/obsidian-tressel/blob/4d8e9e6ac7e447883a3e116a570b79e61e669e92/main.ts#L87
Same here, you can use `Vault.create` instead.
Also, you should use `await` here instead of `.then`.

https://github.com/aseem-thakar/obsidian-tressel/blob/4d8e9e6ac7e447883a3e116a570b79e61e669e92/main.ts#L120
Same as above.

https://github.com/aseem-thakar/obsidian-tressel/blob/4d8e9e6ac7e447883a3e116a570b79e61e669e92/main.ts#L157
Same for async-await instead of .then.

lishid:
Ah it seems like you do control the API end of things, so feel free to do it either way.

I'm fine having the plugin merged as it is, let me know if you want that to happen or if you want to address the suggestions before I do that.

aseem-thakar:
> https://github.com/aseem-thakar/obsidian-tressel/blob/4d8e9e6ac7e447883a3e116a570b79e61e669e92/main.ts#L1 Does this work on mobile devices? I'm guessing it's co-built for node and web, but it's probably using fetch or XMLHttpRequest under the hood.
> 
> I would recommend using the request API that Obsidian exposes, which ignores any CORS restrictions, but if your server responds to CORS properly then either should work.
> 
> https://github.com/aseem-thakar/obsidian-tressel/blob/4d8e9e6ac7e447883a3e116a570b79e61e669e92/main.ts#L56 Don't recommend keeping a reference of this in your plugin class. You just need an inline variable for this use case.
> 
> https://github.com/aseem-thakar/obsidian-tressel/blob/4d8e9e6ac7e447883a3e116a570b79e61e669e92/main.ts#L64 Highly recommend using the `vault` API instead of the adapter API when possible - here you can use `getAbstractFileByPath` and check for whether the result is `instanceof TFolder`.
> 
> https://github.com/aseem-thakar/obsidian-tressel/blob/4d8e9e6ac7e447883a3e116a570b79e61e669e92/main.ts#L87 Same here, you can use `Vault.create` instead. Also, you should use `await` here instead of `.then`.
> 
> https://github.com/aseem-thakar/obsidian-tressel/blob/4d8e9e6ac7e447883a3e116a570b79e61e669e92/main.ts#L120 Same as above.
> 
> https://github.com/aseem-thakar/obsidian-tressel/blob/4d8e9e6ac7e447883a3e116a570b79e61e669e92/main.ts#L157 Same for async-await instead of .then.

Awesome thanks for the advice - I'm defs going to act on all of those points in my next update

I'm cool with merging it in now as is and then refining those later :)
Thanks

PR #653, opened 2021-11-25
Add Calibre Plugin

caronchen:

## Repo URL

Link to my plugin: https://github.com/caronchen/obsidian-calibre-plugin

## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2899,5 +2899,12 @@
         "description": "This plugin enables you to increase/decrease the size of an image by scrolling",
         "author": "Nico Jeske",
         "repo": "nicojeske/mousewheel-image-zoom"
+    },
+    {
+        "id": "obsidian-calibre-plugin",
+        "name": "Calibre",
+        "description": "This plugin allows you to access your calibre libraries and read books directly in Obsidian.",
+        "author": "caronchen",
+        "repo": "caronchen/obsidian-calibre-plugin"
     }
 ]


pjeby:
It would be helpful if your README would explain the purpose of the plugin in more detail.  It says it will "Allow you to access your calibre libraries and read books directly in Obsidian", but it does not give any indication as to *why* thiis would be a good idea, vs. e.g. using a browser window, manually creating an iframe in an Obsidian document, or using the [URI Commands](https://github.com/kzhovn/uri-commands-obsidian) plugin to launch Calibre URLs or the content server. 

The documentation also doesn't mention that the plugin will create a `CALIBRE.md` file in your vault root, and that you need to not have a note with a conflicting name.  (Which on Mac and Windows can be a note like `Calibre.md`.)  Looking at the plugin's source code, it appears as though the plugin also does not update this file when the URL is changed, so I don't think the plugin will work correctly if somebody actually needs to configure a custom URL or port.  (And if it *did* update the file, it would overwrite or erase anything the user had in a Calibre.md file, if any, which is something there should be a warning for, at least in the documentation.)

caronchen:
@pjeby 

Thanks for your review and suggestions.

Fixed already. Here is diff.

https://github.com/caronchen/obsidian-calibre-plugin/commit/38240026dcd0d408113d40881f46e5019959d3d0#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

Thanks very much.

pjeby:
Rather than using the cache `resolved` event, I would suggest simply creating the file in the `onLayoutReady` handler; the resolved event runs after every change to every file in the vault, so this is an excessive/unnecessary frequency of running.

I also think the warning about CALIBRE.md should be much more prominent, since the plugin will delete the contents of this file, causing data loss.  Your current warning says they shouldn't *create* it, but doesn't take into account that if they have an *existing* note named `calibre.md` or `Calibre.md` in the vault root, its contents will still be erased, and if they have an existing note of that name elsewhere in their vault, installing this plugin may *break existing links to that note*.

To avoid users losing data, I would suggest doing one or more of the following:

* Move the filename information to the top of the README as something they need to check *before* installing the plugin if they have a note named any variation of `Calibre`, `calibre`, etc. anywhere in their vault.
* Allow users to configure the path, and don't write the file until after they've chosen a location
* Default to a name with less likelihood of collision, not located in the root (e.g. `obsidian-calibre-plugin/Calibre-View.md`)
* Check the contents of the file before modifying it, to ensure its contents weren't previously written by the plugin, and display an error message instead of overwriting it if the contents don't match
* Don't use a note at all, and instead create a view type that directly renders the required HTML

Obviously the easiest change would be just the documentation, and some of these things are incompatible with each other; the list is just to give you an idea of some of the possibilities for how to avoid the issue.  Creating a view type might actually be second easiest, as it will let you not need to have a note at all, thereby avoiding all the related issues of file protection and documentating the issue.

In addition, a view can be written to save the current URL of the iframe it creates, so that each view instance can retain its position in the Calibre browsing interface across Obsidian workspace changes and restarts.  This would allow you to have panes open to specific books without losing your position when using the Obsidian core "workspaces" plugin, or if you close and reopen a vault.  Views can also customize their leaf/pane "more options" menu with additional operations.

caronchen:
@pjeby 

Fixed by:

* Move `How it Works` to the top of the README.
* Change container file name to `@CALIBRE.Container.md`, still in the root of vault.
* Add identification string into file, and check before modify the container file by the string.
* Create `@CALIBRE.Container.md` file when clicking the ribbon icon or change address in plugin settings.

Thanks a lot.

pjeby:
Looks good to me, that addresses the concerns I had.  There may be comments from other reviewers before the plugin is published, though.

caronchen:
@pjeby Is there anything else I can do before publishing? And how soon will it be published?

pjeby:
As I said above, there may be comments from other reviewers before the plugin is published.  Publishing is done by the Obsidian core developers and there appear to be up to 43 other pull requests ahead of this one in the queue.  (Though a lot of those are stalled waiting for the submitter to update or resolve issues, so it's probably a lot less than 43 in practice.)

lishid:
https://github.com/caronchen/obsidian-calibre-plugin/blob/add6eb1755592b24d9d225973488005ddc6a1141/src/main.ts#L64
`getViewState` makes the view serializes its whole view state. Rather than doing that you can get the file path by using `leaf instanceof MarkdownView && leaf.view.file.path === MY_PATH`

https://github.com/caronchen/obsidian-calibre-plugin/blob/add6eb1755592b24d9d225973488005ddc6a1141/src/main.ts#L88
I understand what you're trying to do here with case insensitivity, but I'd still suggest just using one universal casing and not allow the user to use other ways for the file.

https://github.com/caronchen/obsidian-calibre-plugin/blob/add6eb1755592b24d9d225973488005ddc6a1141/src/tools.ts#L12
That's a huge icon! The icon itself is almost 90kb (which is 70% of your plugin's final js file). I think that's fine but the engineer in me keeps want to know why it's so big.

https://github.com/caronchen/obsidian-calibre-plugin/blob/add6eb1755592b24d9d225973488005ddc6a1141/src/tools.ts#L7

Aaaaah I see what you're trying to do. This is a really bad way to implement your functionality and honestly that's why there's so many hacks around the vault that you're trying to do.

I'd highly suggest you to use a proper `ItemView` instead, like pjeby's suggestion:

> Don't use a note at all, and instead create a view type that directly renders the required HTML

You just need an iframe pane that points to a server location. Whatever you're trying to do (with the global resize hooks to resize the generated div, creating a hardcoded file to render it in markdown) is a lot more convoluted for a simple problem.

ericaxu:

PR #654, opened 2021-11-26
Create Please add repo metadata at https://github.com/obsidianmd/obsidian-releases/

dandv:
image


ericaxu:
Added, thanks for the suggestion.

PR #656, opened 2021-11-29
Please make the repo structure more accessible to newcomers

dandv:
I'm new to Obsidian. I'm trying to find the repo with the Electron app source code to file an issue. It doesn't seem to be in this org, unless I'm missing something.

This repo comes closest, but it's rather obscure to newcomers:

1. The first line of the README is "Until Silver piggybacked the community themes onto it. And also community plugins. And then snippets.". Am I supposed to know who "Silver" is?
2. There's no repo description in the upper-right sidebar. Or website. Or tags.
3. There's no Issues tab. Or explanation of where to file issues.


phibr0:
Obsidian isn't open-source. As the Readme says this Repo was created to distribute the Obsidian Versions and now also Community Themes/Plugins/Snippets. I agree that there could be some more info though. In the meantime, please file any Bug-Reports on the official [Forum](https://forum.obsidian.md/).

ericaxu:
@dandv Thanks for the suggestion! Due to legacy reasons the README was quite informal (it was pretty private at the beginning) and we never remembered to update it until now.

Made an attempt to make it more clear: https://github.com/obsidianmd/obsidian-releases/commit/4b2e29197428db0f45d900deab568450259036c2

PR #658, opened 2021-11-29
Update community-plugins.json

horriblename:

## Repo URL

Link to my plugin:https://github.com/horriblename/preview-mode-keybinds-obsidian

## Release Checklist
- [] I have tested the plugin on
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  ~~Android~~ _(if applicable)_
  - [ ]  ~~iOS~~ _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
  - [ ] ~~`styles.css`~~ _(optional)_

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2136,7 +2136,7 @@
         "id": "obsidian-codemirror-options",
         "name": "CodeMirror Options",
         "author": "nothingislost",
-        "description": "Configure various CodeMirror options such as text selection behavior and syntax highlighting",
+        "description": "Enhance Obsidian's desktop edit mode with features such as WYSIWYG / Live Preview, Syntax Highlighting, and more.",
         "repo": "nothingislost/obsidian-codemirror-options",
         "branch": "main"
     },


lishid:
I think the naming of the plugin can be improved. Since it's mostly for navigation, I'm thinking of something like "Preview mode navigation" and you can mention vim keybinds in your plugin description.

Also keep in mind that we will be soon renaming "Preview mode" to "Reading mode", once Live Preview is released, to minimize possible user confusion.

https://github.com/horriblename/preview-mode-keybinds-obsidian/blob/ece79b4f21a287afd0b1007b31113df23fb72e0e/main.ts#L44
This isn't a good way to register for hotkeys. Doing it this way would result in multiple event handlers hooked, which means anytime there's a layout change (such as splitting a pane, closing a pane), your commands will run n+1 times.

What you should do, if you insist on using keydown events, is to hook it globally (like on window) on plugin load.

https://github.com/horriblename/preview-mode-keybinds-obsidian/blob/ece79b4f21a287afd0b1007b31113df23fb72e0e/main.ts#L56
You can check for `view.getMode() === 'preview'` to check if the pane is in preview mode here.

https://github.com/horriblename/preview-mode-keybinds-obsidian/blob/ece79b4f21a287afd0b1007b31113df23fb72e0e/main.ts#L58
Here you can check if the document is focused on an input, by checking `document.activeElement instanceof HTMLInputElement`.

https://github.com/horriblename/preview-mode-keybinds-obsidian/blob/ece79b4f21a287afd0b1007b31113df23fb72e0e/main.ts#L90
This is not needed.

ericaxu:

PR #659, opened 2021-11-29
Page Count Plugin

ReaderGuy42:
Shows page count to one decimal place in status bar.
https://github.com/ReaderGuy42/Obsidian-Page-Count

## Repo URL

Link to my plugin: https://github.com/ReaderGuy42/Obsidian-Page-Count

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2900,4 +2900,13 @@
         "author": "Nico Jeske",
         "repo": "nicojeske/mousewheel-image-zoom"
     }
+ 
+     {
+        "id": "obsidian-Page-Count",
+        "name": "Page Count",
+        "description": "Adds the current note's page count (with one decimal place) to Obsidian's status bar.",
+        "author": "Ash Cull",
+        "repo": "ReaderGuy42/Obsidian-Page-Count"
+    }
+ 
 ]


liamcain:
Hi @ReaderGuy42, it looks like there was an issue with this commit: [645073c2f](https://github.com/ReaderGuy42/Obsidian-Page-Count/commit/645073c2f0fbeaa4bdd237ae267ce1e42633b739). Your typescript file got replaced with the transpiled javascript code that the compiler spits out. This makes the code a lot harder to read and will make it difficult for anyone to properly review.

Is there a different typescript file that's not uploaded to Github that you're using or have you been manually modifying the generated `main.js` file?

ReaderGuy42:
Hi :) Honestly I'm very new to this and don't know anything about any of this. I edited the original Reading Time plugin, but just changing some things in the `main.js` file. In the tutorials on publishing plugins I saw that I was supposed to be editing `main.ts` so I did that, and after sorting out some issues with `npm` I got it to work; `npm run dev` worked, but I'm not sure how or why. Again on Discord someone said that something there wasn't done correctly, but I don't understand what exactly. (https://discord.com/channels/686053708261228577/840286264964022302/914947872997404712)
I can send the original `main.js` file that I edited if that's relevant.

I was just trying to create a plugin that did a page count, and it does work locally, so I thought someone else trying to do the same thing might profit from it.
Thanks!

liamcain:
No problem at all! I'm impressed you got it working just by modifying the main.js file since that's not exactly human readable.

I think a better starting point for you might be the Reading Time plugin's `main.ts` file then. https://github.com/avr/obsidian-reading-time/blob/main/main.ts

Basically, most Obsidian plugins are written using Typescript, but since Obsidian can't actually read typescript files, they need to be _transpiled_ (turned into) a javascript file. So in a way `main.js` is just the output file that's not meant to be read by humans.

So a good workflow for you to follow might be starting with the Reading Time plugin, then make your changes to `main.ts`. When you want to test your changes in Obsidian, run `npm run build` which will generate the new `main.js` file. The `main.js` file doesn't actually need to be added to your git repository. Instead, Obsidian will look for it in the Github Releases section.

Let me know if any of that was confusing or didn't make sense!

ReaderGuy42:
Thanks for the clear explanation! So I've downloaded the files from AVR's repo and edited some of the `main.ts` file. However, the most important part, i.e. the actual calculation of words per page does not appear here, rather only in the `main.js` file in the release. Can I still copy that stuff over from the .js to the .ts file?

liamcain:
> So I've downloaded the files from AVR's repo and edited some of the main.ts file. However, the most important part, i.e. the actual calculation of words per page does not appear here, rather only in the main.js file in the release. 

Sorry, I'm not exactly sure I understand. That logic should _start_ in the `main.ts` file and then make its way into the `main.js` file automatically after you run `npm run build`.

> Also, I think maybe I was understanding something wrong, because it seems that main.js is readable

It's readable in the sense that it's valid javascript code but it wouldn't exactly be fun to review.

ReaderGuy42:
I think (!) I have it figured out now. I created a new node-module by copying the one Reading Time was using and adapting the stuff I needed. Now when I do `npm run dev` it takes my changed code automatically. I've created a new release with the new `main.js` and also uploaded the new `main.ts` and manifest. Thanks for the help!

lishid:
Ok whatever happened to this file, made it very difficult to read - it's also based on very old versions of the app's API which is no longer recommended to be used.

Highly suggest you put the code through an auto-formatter to add proper indentations. If you can find the old `main.ts` back and re put your modifications in there instead of getting it off the poorly re-constructed `main.js` that'd be even better.

https://github.com/ReaderGuy42/Obsidian-Page-Count/blob/d93ed335d686503554b76db30b9b27ca33e20879/main.ts#L3
I could not find this dependency anywhere. I think you must have it somewhere in your local dev computer but did not commit it to the github repo.

https://github.com/ReaderGuy42/Obsidian-Page-Count/blob/d93ed335d686503554b76db30b9b27ca33e20879/main.ts#L27
This isn't the API you should use anymore. Plus you aren't unhooking the change function properly when your plugin unloads, which means a user with your plugin cannot disable your plugin properly.

The new API we recommend is `Workspace.on('editor-change')`, and given the intensity of your calculations, you should probably make it debounced so that you don't re-compute the number of words in the document on every single keypress.

ReaderGuy42:
Sorry for the delay. I really don't understand most of what you've said. I was pretty much just making stuff up as I went along. Therefore I'd like to open this up to anyone who's interested in the plugin, if they'd like to take over.
Thanks :)

lishid:
Sounds good. I think there are a couple of other plugins that also provide page information (Get Info, Info Panel), so let's close down this PR for now until someone wants to give it another go. I do agree that having it in the status bar could be helpful.

ReaderGuy42:
Great, thank you!

PR #660, opened 2021-11-30
Add obsidian-graphviz plugin

QAMichaelPeng:

## Repo URL

Link to my plugin: https://github.com/QAMichaelPeng/obsidian-graphviz

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3098,5 +3098,13 @@
         "description": "Create markdown-based presentations in Obsidian",
         "repo": "MSzturc/obsidian-advanced-slides",
         "branch": "main" 
+    },
+    {
+        "id": "obsidian-graphviz",
+        "name": "Obsidian Graphviz",
+        "description": "Render Graphviz Diagrams",
+        "author": "Feng Peng",
+        "repo": "QAMichaelPeng/obsidian-graphviz",
+        "branch": "main"
     }
 ]


lishid:
https://github.com/QAMichaelPeng/obsidian-graphviz/blob/9af1664dd308496621ec491c95f8f4bc8ac48bde/src/main.ts#L20
Since your code runs asynchronously, you should return a promise or use async-await properly here. You can also use `processors.imageProcessor.bind(processors)` directly without having an intermediate function. Just make `imageProcessor` return a promise which resolves when your png rendering is complete (you can promisify the exec/fs actions pretty easily).

Instead of using a base64 URI, you can use a blob URI which is more efficient:
https://stackoverflow.com/a/21206824

QAMichaelPeng:
Thanks for the feedback. Will address it in this week.

QAMichaelPeng:
Most issues pr comments were resolved.
One remaining  is return promise from GraphvizPlugin:onload().

Since typescript will return a promise by default for async, do we need to explicitly return a promise? Could you share an example? I've checked variant plugins and didn't find one returning a promise in onload.

Thanks.

lishid:
Yeah that looks good. One more thing you should do is to convert the `.then()` and `.catch` to await: https://github.com/QAMichaelPeng/obsidian-graphviz/blob/9c32378a15adc5b306405c1c75524f16455f35c3/src/processors.ts#L72

> Since typescript will return a promise by default for async, do we need to explicitly return a promise? Could you share an example? I've checked variant plugins and didn't find one returning a promise in onload.

Yeah, an async function basically returns a promise. In fact, async-await is just some syntactic sugar that automatically wraps your code in different chunks of callback functions with promises and `.then()` and `.catch()`. It's really the new way of writing promises that makes them so much easier to read (just like synchronous code!)

There are some examples at https://developers.google.com/web/fundamentals/primers/async-functions

QAMichaelPeng:
@lishid , thanks for the comments. Code updated.

PR #661, opened 2021-12-01
Add Image Window plugin

valentine195:

This is a simple plugin that adds the ability to open image files in the vault in a new Obsidian BrowserWindow instance.

## Repo URL

Link to my plugin: https://github.com/valentine195/obsidian-image-window

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/valentine195/obsidian-image-window/blob/3a33dd64873230ab23ec9ff661d4c21972a2d2d7/src/main.ts#L65
You can use `getResourcePath` instead which will give you an `app://local` path on desktop, which resolves to the file:/// path without requiring you to disable webSecurity on the BrowserWindow.

lishid:
You are also running the risk of HTML DOM injection by using an un-sanitized input string, should consider HTML encode the result of the link.

valentine195:
> https://github.com/valentine195/obsidian-image-window/blob/3a33dd64873230ab23ec9ff661d4c21972a2d2d7/src/main.ts#L65 You can use `getResourcePath` instead which will give you an `app://local` path on desktop, which resolves to the file:/// path without requiring you to disable webSecurity on the BrowserWindow.

Thanks, changed. Didn't like disabling the webSecurity, glad there's a workaround :)

> You are also running the risk of HTML DOM injection by using an un-sanitized input string, should consider HTML encode the result of the link.

I added in the DOMPurify santizer (based on a comment you made in the Discord a few weeks ago). Does this work for you?

https://github.com/valentine195/obsidian-image-window/blob/49bfc08e55938b97e5e908903e7cd10bda1aeb38/src/main.ts#L120

lishid:
We do expose the DOM sanitizer in the plugin API now, but I think it's overkill - you only have a single path to protect and it's probably fine to just escape entities (`&<>"`)

PR #664, opened 2021-12-03
Add TTS plugin

joethei:

## Repo URL

Link to my plugin: https://github.com/joethei/obsidian-tts

## Release Checklist
  - [X]  Windows
  - [X]  macOS
  - [X]  Linux
  - [X]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


lishid:
https://github.com/joethei/obsidian-tts/blob/2b879928d2319118ceed4eabc64b8bf592b0f202/src/main.ts#L22
You can throw an error here and Obsidian should automatically cancel the plugin loading. You're also continuing execution here instead of returning...

https://github.com/joethei/obsidian-tts/blob/2b879928d2319118ceed4eabc64b8bf592b0f202/src/main.ts#L32
Here you'll want to perform this check regardless of the value of `checking`, because there are possible cases where your callback is called when `checking` is false (where you'd pass null to `play`). You should also check for a truthy value rather than `undefined` because it may return `null` instead.

joethei:
Thanks for spotting the first one, have addressed your comments now.
The one thing I am not too happy about is the double negation I had to use here https://github.com/joethei/obsidian-tts/blob/23cc414b8e4707b0bff52ad7babb59a9fad79fdb/src/main.ts#L34, if you know a better way lmk.

lishid:
> The one thing I am not too happy about is the double negation I had to use here https://github.com/joethei/obsidian-tts/blob/23cc414b8e4707b0bff52ad7babb59a9fad79fdb/src/main.ts#L34, if you know a better way lmk.

Double negation works fine (we use it often too) to convert things into booleans. Here specifically you can return true because of the boolean logic two lines above, but I have a suggestion that applies to all of your `checkCallback`s. Specifically, your checkCallbacks should perform the check and also return true/false regardless of the checking variable.

The other commands (Stop playback, pause playback) are also missing the `isSpeaking` check when it's called - I don't know if that's going to be a problem when you call `stop` when it's not speaking.

Try to follow the code flow in the guide my Marcus here: https://marcus.se.net/obsidian-plugin-docs/guides/commands#conditional-commands
Doing it this way makes sure that you are covered in both cases.

```ts
checkCallback(checking: boolean) => {
  if (condition satisfied) {
    if (!checking) {
      // Perform action
    }
    return true;
  }
  // optionally return false
}
```

joethei:
> The other commands (Stop playback, pause playback) are also missing the `isSpeaking` check when it's called - I don't know if that's going to be a problem when you call `stop` when it's not speaking.

There is a check in the service for this when calling the method.

Also, TIL of `editorCallback`, seems to make the play command a lot easier.

Adressed those issues now.

lishid:
> There is a check in the service for this when calling the method.

Yeah I'm guessing it's probably ok regardless. The recommended function would be something like this:

```ts
if (this.ttsService.isSpeaking()) {
    if (!checking) this.ttsService.stop();
    return true;
}
```

PR #667, opened 2021-12-05
Add Get Info plugin

chetachiezikeuzor:
# Get Info Plugin

## Repo URL

Link to my plugin: https://github.com/chetachiezikeuzor/Get-Info-Plugin

Get Info Demo Header

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/chetachiezikeuzor/Get-Info-Plugin/blob/e42b74d59bc3549b6148691f4cbdd35b0399b8b3/src/plugin/main.ts#L41
Please don't do this - just setup your status bar icon right when your plugin initializes. If plugins use timeouts to try and beat each other to a position then it'll be a very bad game to play. Eventually we may give a way for the user to reorganize the ribbon and status bar, but in the meantime I suggest avoiding hacks like this.

https://github.com/chetachiezikeuzor/Get-Info-Plugin/blob/e42b74d59bc3549b6148691f4cbdd35b0399b8b3/src/plugin/main.ts#L73

https://github.com/chetachiezikeuzor/Get-Info-Plugin/blob/e42b74d59bc3549b6148691f4cbdd35b0399b8b3/src/plugin/main.ts#L78
Here you should just return if file is null. You can also reassign `getActiveFile` to file instead of making a new variable called `fileData` which is just a file. Then you won't need the conditional `?.extension` either.

https://github.com/chetachiezikeuzor/Get-Info-Plugin/blob/e42b74d59bc3549b6148691f4cbdd35b0399b8b3/src/plugin/main.ts#L81
The `?` here is unnecessary, and it expands into this ugly expression `fileCache = fileCache === null || fileCache === void 0 ? void 0 : fileCache.replace(...);`

https://github.com/chetachiezikeuzor/Get-Info-Plugin/blob/e42b74d59bc3549b6148691f4cbdd35b0399b8b3/src/plugin/main.ts#L88
This doesn't do anything since the result isn't assigned to the file contents. Also you should probably use a regular if clause instead of a ternary operator.

https://github.com/chetachiezikeuzor/Get-Info-Plugin/blob/e42b74d59bc3549b6148691f4cbdd35b0399b8b3/src/plugin/main.ts#L101
Here you are calling getWordCount 3 times! Which recomputes the word count of the document 3 times. Should store this in a variable and reuse the result.

lishid:
https://github.com/chetachiezikeuzor/Get-Info-Plugin/blob/f2ee5251e80254eef808f67571fc08d10de7ef6b/src/plugin/main.ts#L73
This doesn't seem used (and is probably a really bad idea for performance reasons; you don't want to re-compute the word count for every character typed).

chetachiezikeuzor:
Ok fixed the status bar item ➡️
https://github.com/chetachiezikeuzor/Get-Info-Plugin/blob/06b443a885c9cf64b5e230fc0fb9bd1eb51405ac/src/plugin/main.ts#L23

Reassigned to file ➡️
https://github.com/chetachiezikeuzor/Get-Info-Plugin/blob/06b443a885c9cf64b5e230fc0fb9bd1eb51405ac/src/plugin/main.ts#L70

`?` have been removed ➡️
https://github.com/chetachiezikeuzor/Get-Info-Plugin/blob/06b443a885c9cf64b5e230fc0fb9bd1eb51405ac/src/plugin/main.ts#L73

Changed to `if` statement ➡️
https://github.com/chetachiezikeuzor/Get-Info-Plugin/blob/06b443a885c9cf64b5e230fc0fb9bd1eb51405ac/src/plugin/main.ts#L80

My bad 😓! Assigned to variable now! ➡️
https://github.com/chetachiezikeuzor/Get-Info-Plugin/blob/06b443a885c9cf64b5e230fc0fb9bd1eb51405ac/src/plugin/main.ts#L83

I removed the re-computing function! Not sure why I left that one. ➡️
https://github.com/chetachiezikeuzor/Get-Info-Plugin/blob/06b443a885c9cf64b5e230fc0fb9bd1eb51405ac/src/plugin/main.ts#L68

lishid:
Looks great!

[This](https://github.com/chetachiezikeuzor/Get-Info-Plugin/blob/06b443a885c9cf64b5e230fc0fb9bd1eb51405ac/src/plugin/main.ts#L82) is probably not necessary and doesn't do anything anyway.

PR #668, opened 2021-12-05
Added the obsidian-success-plan plugin (REDO)

joshwingreene:

## Repo URL

Link to my plugin: https://github.com/thex3family/obsidian-success-plan

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2928,5 +2928,13 @@
         "description": "Structured plugin. Create hierarchy in notes using \".\"",
         "author": "dobrovolsky",
         "repo": "dobrovolsky/obsidain-structure"
+    },
+    {
+      "id": "obsidian-success-plan",
+      "name": "Success Plan",
+      "author": "Joshwin Greene",
+      "description": "Manage your tasks, projects, key results, and goals within Obsidian. This plugin follows #theGamificationProject's Success Plan Framework.",
+      "repo": "thex3family/obsidian-success-plan",
+      "branch": "main"
     }
 ]


lishid:
https://github.com/thex3family/obsidian-success-plan/blob/c5595eda82d055da30de1ac53c5852fb9d0aa6c9/src/main.ts#L49

A couple of suggestions here:

1. Instead of detaching and re-creating your view, you should probably reuse it?
2. Keep a reference to your leaf so you don't have to get it again after.

https://github.com/thex3family/obsidian-success-plan/blob/c5595eda82d055da30de1ac53c5852fb9d0aa6c9/src/main.ts#L137

There's gotta be a better way to reload the view (like what you did for the previous point, detach and re-create) so the user doesn't have to manually re-enable the plugin.

https://github.com/thex3family/obsidian-success-plan/blob/c5595eda82d055da30de1ac53c5852fb9d0aa6c9/src/main.ts#L21
I don't think this variable is useful here.

https://github.com/thex3family/obsidian-success-plan/blob/c5595eda82d055da30de1ac53c5852fb9d0aa6c9/src/view.tsx#L31
Use `this.contentEl`, same for below.

https://github.com/thex3family/obsidian-success-plan/blob/c5595eda82d055da30de1ac53c5852fb9d0aa6c9/src/ReactApp.tsx#L13
I'm not sure I understand the Provider architecture in React, but it seems like this is awfully a lot of work to pass one variable to your react app, requiring two additional files. Plus you're passing through the `settings` variable directly!

https://github.com/thex3family/obsidian-success-plan/blob/c5595eda82d055da30de1ac53c5852fb9d0aa6c9/src/ReactApp.tsx#L196
Note that this function could be pretty expensive if you were to update the view often - perhaps implement some sort of cache based on each file's last modified time (`file.stat.mtime`) so that you don't re-parse every file whenever the UI needs updating?

ericaxu:

PR #670, opened 2021-12-05
Add Link Favicons Plugin

joethei:

## Repo URL

Link to my plugin: https://github.com/joethei/obsidian-link-favicon

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`
- [ X] I have read the tips in https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md and have self-reviewed my plugin to avoid these common pitfalls.


lishid:
https://github.com/joethei/obsidian-link-favicon/blob/e5d2b0c395bf2875234883b2a237751c7543272d/src/main.ts#L22
You may want to try-catch this since it could throw an exception.

https://github.com/joethei/obsidian-link-favicon/blob/e5d2b0c395bf2875234883b2a237751c7543272d/src/main.ts#L21
You can cast your link element to an `HTMLAnchorElement` as a better way to mitigate this typescript warning.

https://github.com/joethei/obsidian-link-favicon/blob/e5d2b0c395bf2875234883b2a237751c7543272d/src/main.ts#L25
Could consider using a switch instead, or at least use else-ifs.

You also don't have a failsafe for when provider isn't in your list (which normally shouldn't happen).

joethei:
I have addressed these issues now.

lishid:
https://github.com/joethei/obsidian-link-favicon/blob/a5541fc7386a372b4c0a9fe25443058d58abbf0e/src/main.ts#L40
You caught the error but don't stop execute here, which will cause an exception to happen when you try to query `domain.hostname` next - probably should `continue` when an error is caught.

Should probably also try-catch on the url provider. My opinion would be to put the whole chunk of code in your try-catch (for each link).

https://github.com/joethei/obsidian-link-favicon/blob/a5541fc7386a372b4c0a9fe25443058d58abbf0e/src/provider.ts#L4
This may just be a language/preference thing but I would just make this a simple map rather than an object-class with static fields and a provider collection:

```ts
interface IconProvider {
	id: string;
	name: string;
	url: (domain: string, settings: FaviconPluginSettings) => Promise<string>;
}
let providers: Record<string, IconProvider> = {
  'google': {id: 'google', name: 'Google', url: ...},
  // others
};

// get
providers[id];

// iterate
for (let id in providers) {
  if (providers.hasOwnProperty(id)) {    }
}
```

joethei:
I addressed these comments now,

> This may just be a language/preference thing but I would just make this a simple map rather than an object-class with static fields and a provider collection:

Yeah, that is way more elegant that what I have done.

PR #672, opened 2021-12-05
obsidian-attachment-name-formatting

JYC333:

## Repo URL

Link to my plugin: https://github.com/JYC333/obsidian-attachment-name-formatting

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/JYC333/obsidian-attachment-name-formatting/blob/9832caae0408ace3ae780bc5f4000cc927b3de66/main.ts#L10
I'm not familiar with any image format named `.git`. We also recently added support for `.webp`.

https://github.com/JYC333/obsidian-attachment-name-formatting/blob/9832caae0408ace3ae780bc5f4000cc927b3de66/main.ts#L47
Should avoid using `var` as it's from the older javascript world. Use `let` or `const` instead.

https://github.com/JYC333/obsidian-attachment-name-formatting/blob/9832caae0408ace3ae780bc5f4000cc927b3de66/main.ts#L10
You can't store RegExp objects in settings because they are serialized to JSON.

https://github.com/JYC333/obsidian-attachment-name-formatting/blob/9832caae0408ace3ae780bc5f4000cc927b3de66/main.ts#L54
This only works if the link is exactly the same file path, with no heading/blockid/page subpaths - you need to resolve the link using metadataCache.getFirstLinkpathDest. You should also use `parseLinktext` first to split out the file path from the `#heading` subpath.

https://github.com/JYC333/obsidian-attachment-name-formatting/blob/9832caae0408ace3ae780bc5f4000cc927b3de66/main.ts#L58
This has a lot of ways to fail as well. `item.link` may not be exactly the same as the path, and a `replace` will disregard where you're replacing. If you just want the file's parent folder then you can use `path.substring(0, path.length - filename.length)`.

https://github.com/JYC333/obsidian-attachment-name-formatting/blob/9832caae0408ace3ae780bc5f4000cc927b3de66/main.ts#L85
Should use `Vault.getAbstractFileByPath` instead of iterating through every single file in the vault.

https://github.com/JYC333/obsidian-attachment-name-formatting/blob/9832caae0408ace3ae780bc5f4000cc927b3de66/main.ts#L65
You need to await for this operation to finish before attempting the next operation.

JYC333:
I correct all the comments and please take a look again.
Thanks!

lishid:
Looks fine. One more thing:

https://github.com/JYC333/obsidian-attachment-name-formatting/blob/6f801ca17eca22e7ff6060963bef5b148e1252cb/main.ts#L10
This isn't user configurable, and you probably don't want it to be configurable as a regex string anyway because there are lots of ways users can screw this up. You're also testing the regex on the linktext, instead of the actual file name.

What I recommend you to do here is to have a simple array of extensions lowercased. Then get the linked files and test the `attachmentFile.extension` against your list using `list.contains(extension)`.

JYC333:
Thanks for the comment and Happy New Year!

I remove the configurable extensions from settings and hardcoded them. Please take a look again,

Thanks!

PR #673, opened 2021-12-06
added obsidian-regex-replace

Gru80:

## Repo URL

Link to my plugin: https://github.com/Gru80/obsidian-regex-replace

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2928,5 +2928,12 @@
         "description": "Structured plugin. Create hierarchy in notes using \".\"",
         "author": "dobrovolsky",
         "repo": "dobrovolsky/obsidain-structure"
+    },
+    {
+        "id": "obsidian-regex-replace",
+        "name": "Regex Find and Replace",
+        "description": "Provides a Find/Replace dialog which optionally supports regular expressions and scope (full document or text selection).",
+        "author": "Martin Eder",
+        "repo": "Gru80/obsidian-regex-replace"
     }
 ]


lishid:
https://github.com/Gru80/obsidian-regex-replace/blob/3b89f675c038cc3708bd4a9677dd438ef22a510c/src/main.ts#L171
I would suggest just automatically detect whether the user has a selection instead of providing a toggle for it. If there's no selection then use full document.

https://github.com/Gru80/obsidian-regex-replace/blob/3b89f675c038cc3708bd4a9677dd438ef22a510c/src/main.ts#L173
Should store `editor.getValue` in a variable so you don't have to get it multiple times.

Gru80:
Thanks a lot for the review! Since this is my first attempt of a valuable contribution in the domain of TypeScript/JavaScript any feedback is highly appreciated, I'll consider your suggestions accordingly.

lishid:
No worries!

https://github.com/Gru80/obsidian-regex-replace/blob/aa742c1f578aee246099a684e532be3c506ad551/src/main.ts#L176
Here you are attempting a match, then a replace, based off the same regex - this means the code would have to run the regex twice. Instead, what you can do is to perform the replacement and test whether the final result is equal to the original text, and only call `setValue` if the result changed.

https://github.com/Gru80/obsidian-regex-replace/blob/aa742c1f578aee246099a684e532be3c506ad551/src/main.ts#L198
Here you are calling `.split()` twice, which is also double work for something that you can throw in a variable to reuse.

Gru80:
any further steps needed from my side in order to proceed with the pull-request?

lishid:
> https://github.com/Gru80/obsidian-regex-replace/blob/aa742c1f578aee246099a684e532be3c506ad551/src/main.ts#L176
Here you are attempting a match, then a replace, based off the same regex - this means the code would have to run the regex twice. Instead, what you can do is to perform the replacement and test whether the final result is equal to the original text, and only call setValue if the result changed.

I don't think you addressed this one, should be pretty easy to do.

Gru80:
Thanks a lot for the quick response,
Actually I also had a look at the proposed change for the redundant regex-execution - but actually I'm afraid I'd loose some information when following the suggested approach:
Because I'm not using `match()` just for checking if there is _any_ occurance at all, but also to be able to inform the user how many replacements have been made (determined by the length of the array which is returned by `match()`). Since `replace()` just returns the new string with the replaced content, I think I would have no possibility to get information regarding the _number of performed replacements_ this way. And from the User-Interaction point of view I think it is helpfull to be informed about how many changes have been made.

lishid:
Ah you're looking for `rresult.length`.

In this case, what you could do is to use the `replace(regex, callback)` syntax where in the callback you can increment a number and then return the `replaceString`. That should achieve your desired goal and only run the search once.

Gru80:
splendid hint, thanks for that! I wasn't aware of the `replace`-signature with the callback-argument, that should indeed enable a more elegant solution with just one regex-execution - I'll check that out

PR #674, opened 2021-12-06
Add Generic Initiative Tracker for TTRPGs to the Plugin List

beaushinkle:

## Repo URL

Link to my plugin: https://github.com/beaushinkle/obsidian-generic-initiative-tracker

## Release Checklist
  - [X]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -2928,5 +2928,13 @@
         "description": "Structured plugin. Create hierarchy in notes using \".\"",
         "author": "dobrovolsky",
         "repo": "dobrovolsky/obsidain-structure"
+    },
+    {
+      "id": "generic-initiative-tracker",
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/beaushinkle/obsidian-generic-initiative-tracker/blob/c820bcf29cbd0e9186ed3899df35f5804c4a08bd/src/main.ts#L225
This `?` is not necessary. I also suggest putting `this.view` into a variable the first time you use it in each function, because it is a multi-step getter property that iterates through every leaf in the app, and you repeat `this.view` often in each function that makes it perform a lot of unnecessary duplicate computation for every access to the property.

PR #679, opened 2021-12-08
Add Matter to community plugins

hunterclarke:

Hi all! I'm with the [Matter](https://hq.getmatter.app) team. We've received dozens of requests for an Obsidian integration. We can't wait to get this to all of our users. Let me know if there is anything else you need from us.

## Repo URL
Link to my plugin: [obsidian-matter](https://github.com/getmatterapp/obsidian-matter)

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3030,5 +3030,12 @@
         "description": "See the favicon for a linked website.",
         "author": "Johannes Theiner",
         "repo": "joethei/obsidian-link-favicon"
+    },
+    {
+        "id": "matter",
+        "name": "Matter",
+        "author": "Matter",
+        "description": "Official Matter Obsidian integration.",
+        "repo": "getmatterapp/obsidian-matter"
     }
 ]


lishid:
https://github.com/getmatterapp/obsidian-matter/blob/8c2f5791325340d8b690697567ee5f81220ff7a3/src/main.ts#L38
I would advise running this after `Workspace.onLayoutReady()` so that Obsidian has a chance to pick up the vault files. Also awaiting a long-running process in your plugin's onload function will delay Obsidian's own app loading causing the app to show the loading bar while your plugin does the work behind the scene. Should probably not await for this to finish and just run it in parallel. (Plus if anything fails here your plugin will be deactivated).

Everything else looks fine.

hunterclarke:
@lishid I've addressed your feedback in the master branch and marked a new release (1.0.1). Let me know if there anything else you need.

lishid:
> I was unsure what you were referring to regarding Workspace.onLayoutReady(), I'm not seeing a hook in the API for executing plugin code after. Let me know if running the sync in parallel is insufficient.

There are cases where your plugin may be initialized before the vault has scanned all the files. If your plugin depends on knowing files in the vault then you should delay until the app has finished initial loading.

hunterclarke:
> There are cases where your plugin may be initialized before the vault has scanned all the files. If your plugin depends on knowing files in the vault then you should delay until the app has finished initial loading.

Got it. Thanks for explaining and for the quick turnaround time!

PR #683, opened 2021-12-09
Add labels to and close stale PRs

argenos:


lishid:
I'll leave off any automation for now since I know some people really hate that, thanks for the PR though!

.github/workflows/stale.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v4
+        with:
+          stale-pr-message: "Hi there, as this PR has not seen any activity in the last 30 days, it will be closed in 30 days unless there are any updates."
+          close-pr-message: "Hi there, to keep things tidy, we're closing PRs after two months of inactivity.\nFeel free to ask us to re-open it when you're ready to continue. Thanks for your understanding!"
+          days-before-stale: 30
+          days-before-close: 30
```
lishid, line 17: Is it intentional to have both stale and close set at 30 days? Feels kinda confusing if we're only using it to close it down.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,19 @@
+name: 'Close stale PRs'
+on:
+  schedule:
+    - cron: "27 7 * * *"
```
lishid, line 4: What's this time for?

.github/workflows/stale.yml
joethei: Yes that's intentional, after 30 days the PR will get the `stale` label, after another 30 days it will get closed, if there is no activity.

.github/workflows/stale.yml
joethei: The job is set to run at 07:27 every day.
The exact numbers aren't that important here, I just chose some random ones.

PR #685, opened 2021-12-12
add local plantuml provider

joethei:

## Repo URL

Link to my plugin: https://github.com/joethei/obsidian-local-plantuml

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3060,5 +3060,12 @@
         "author": "kometenstaub",
         "repo": "kometenstaub/obsidian-linked-data-vocabularies",
         "branch": "main"
+    },
+     {
+        "id": "local-plantuml",
+        "name": "Local PlantUML Provider",
+        "description": "Render your PlantUML diagrams locally. ",
+        "author": "Johannes Theiner",
+        "repo": "joethei/obsidian-local-plantuml"
     }
 ]


lishid:
https://github.com/joethei/obsidian-local-plantuml/blob/ec2d4a2db2e2592262732a8e8610bfafe3bc0e88/main.ts#L115
Your plugin doesn't seem to do anything in Obsidian as far as I can see in the code. Am I missing something?

There are two functions to generate diagrams through an external process call, but I don't see them used anywhere.

joethei:
Yes, this plugin is useless by itself, the methods are being called by the plantuml plugin if this plugin is installed.

Without this plugin a remote server is used for rendering. 

This separation is to keep the main plugin mobile compatible.

lishid:
Oh I see! You can use `require()` directly in your code selectively so that you don't need a second plugin.

Instead of importing the NodeJS packages directly at the top of the file, you could instead import it inside a function like this:

```ts
function doSomeNodeStuff() {
  let fs = require('fs');
  ..
}
```

Since the require will never be called unless you call the function, it will work on mobile as long as you don't call the function (test for `Platform.isDesktopApp`)

joethei:
Ah, that works perfectly, thank you for letting me know of this.

PR #686, opened 2021-12-13
File Info Panel Plugin added to community-plugins.json

CattailNu:
Added a File Info Panel entry to community-plugins.json

## Repo URL

Link to my plugin:
https://github.com/CattailNu/obsidian-file-info-panel-plugin

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/chetachiezikeuzor/Get-Info-Plugin
You should totally collaborate with @chetachiezikeuzor since both of your plugins are so similar.

https://github.com/CattailNu/obsidian-file-info-panel-plugin/blob/84c4756a8ccb0f79b7995b27a2a232e87a85ba50/main.ts#L23
Don't import path because that will make your plugin only work on the desktop app. Instead, just join the path using `/` and then apply `normalizePath` from `obsidian`.

https://github.com/CattailNu/obsidian-file-info-panel-plugin/blob/84c4756a8ccb0f79b7995b27a2a232e87a85ba50/main.ts#L91
Use the new `editor-changed` event on `Workspace` instead of this one, which will not work with the new editor.

https://github.com/CattailNu/obsidian-file-info-panel-plugin/blob/84c4756a8ccb0f79b7995b27a2a232e87a85ba50/main.ts#L103
You register a lot of events throughout the app and call this function repeatedly. I would highly suggest debouncing this function so that you don't end up re-computing statistics hundreds of times per second and causing unnecessary CPU usage.

This is especially true since you're calling `updateDisplay` every time which re-builds the entire DOM of your view.

https://github.com/CattailNu/obsidian-file-info-panel-plugin/blob/84c4756a8ccb0f79b7995b27a2a232e87a85ba50/main.ts#L72
I don't think you need to hook up to **all** of these events. Some of them are especially noisy when some things happens on the file system, for example, renaming a file could cause modification of dozens of other files due to links updating. You shouldn't be updating statistics dozens of time when this happens. In fact, you are probably OK to use `MetadataCache`'s `resolved` event to replace all the vault events except delete. In either case, you should debounce your update function as well.

CattailNu:
I cannot find a working example of: "Use the new editor-changed event on Workspace instead of this one, which will not work with the new editor."  The events definitely need to be sorted out.  The project has scope-creeped to include [selected] word counts/word usage statistics.

lishid:
I don't remember seeing a sample out there but you can try and play with this API here: https://github.com/obsidianmd/obsidian-api/blob/master/obsidian.d.ts#L3680

CattailNu:
Tried that, but it actually fires off more than the other events did.  I still need to debounce and I have a "selected text" issue now that I don't know how to solve.

lishid:
> Tried that, but it actually fires off more than the other events did. I still need to debounce and I have a "selected text" issue now that I don't know how to solve.

Yes that is correct, it does fire fairly often. You'll need some debouncing regardless though.

This new event supports both the legacy editor (cm5) and the new editor (cm6). What do you mean by "I have a "selected text" issue now that I don't know how to solve."? If you're referring to a lack of selection change events, you can get it by hooking a global `window.addEventListener('selectionchange')` and registering that event in your plugin to have it unload properly.

CattailNu:
I don't know the syntax to use to register that event.  (new to typescript)  I also need to figure out debouncing yet.  I appreciate the feedback.  Thank you!

CattailNu:
this.registerDomEvent(window, 'selectionchange', (event) => {
			console.log("selectionchange");
			this.requeryStats();
		});

Put this in the main.ts with the other event registrations.  The console.log never fires.

lishid:
Ah it seems like you have to hook it as a capture event, which isn't supported by registerDomEvent. Apologies!

lishid:
You can still use something like

```ts
let callback = (evt: Event) => {
  console.log(evt);
};
window.addEventListener('selectionchange', callback, true);
this.register(() => window.removeEventListener('selectionchange', callback, true));

CattailNu:
That worked! Thanks!
From your initial comments, with my replies preceded with *:
---------------
https://github.com/chetachiezikeuzor/Get-Info-Plugin
You should totally collaborate with @chetachiezikeuzor since both of your plugins are so similar.
* Scope creep has pushed it to no longer similar. (added word usage statistics, selection).

https://github.com/CattailNu/obsidian-file-info-panel-plugin/blob/84c4756a8ccb0f79b7995b27a2a232e87a85ba50/main.ts#L23
Don't import path because that will make your plugin only work on the desktop app. Instead, just join the path using / and then apply normalizePath from obsidian.
* Changed.

https://github.com/CattailNu/obsidian-file-info-panel-plugin/blob/84c4756a8ccb0f79b7995b27a2a232e87a85ba50/main.ts#L91
Use the new editor-changed event on Workspace instead of this one, which will not work with the new editor.
* Added.

https://github.com/CattailNu/obsidian-file-info-panel-plugin/blob/84c4756a8ccb0f79b7995b27a2a232e87a85ba50/main.ts#L103
You register a lot of events throughout the app and call this function repeatedly. I would highly suggest debouncing this function so that you don't end up re-computing statistics hundreds of times per second and causing unnecessary CPU usage.
* Debounced.  I think...

This is especially true since you're calling updateDisplay every time which re-builds the entire DOM of your view.

https://github.com/CattailNu/obsidian-file-info-panel-plugin/blob/84c4756a8ccb0f79b7995b27a2a232e87a85ba50/main.ts#L72
I don't think you need to hook up to all of these events. Some of them are especially noisy when some things happens on the file system, for example, renaming a file could cause modification of dozens of other files due to links updating. You shouldn't be updating statistics dozens of time when this happens. In fact, you are probably OK to use MetadataCache's resolved event to replace all the vault events except delete. In either case, you should debounce your update function as well.
* This should only be grabbing the currently open file's statistics, not rolling through all files in the vault. I'm hoping this is a non-issue.
---------------

lishid:
https://github.com/CattailNu/obsidian-file-info-panel-plugin/blob/fde733ea2b44d17786223ec220579d2c5963ae2f/main.ts#L156
Should check for null here. (I know in theory it shouldn't because you're checking the active file, but there may still be ways this can get you null, for example, other plugins replacing `.md` views)

> This should only be grabbing the currently open file's statistics, not rolling through all files in the vault. I'm hoping this is a non-issue.

Yeah, now that you have it debounced it should be fine. I was just saying that this event is fired for every file change, even if it's not the active file. You should test if the `file` passed in from the event is the active file being looked at, so you don't re-compute when any random file is changed in the vault. This can happen for a variety of reasons like using Sync, or using another app on the same vault. (Ex: Using vscode to run a global search & replace)

Ideally you only need to trigger a re-compute when the specific file you have open in view has changed.

CattailNu:
From your comments, with my replies preceded with *:
Should check for null here. (I know in theory it shouldn't because you're checking the active file, but there may still be ways this can get you null, for example, other plugins replacing .md views)
* Done.  I wasn't sure how javascript handles && for if (v && v.file && ... ) so I just split them out.

This should only be grabbing the currently open file's statistics, not rolling through all files in the vault. I'm hoping this is a non-issue.
Yeah, now that you have it debounced it should be fine. I was just saying that this event is fired for every file change, even if it's not the active file. You should test if the file passed in from the event is the active file being looked at, so you don't re-compute when any random file is changed in the vault. This can happen for a variety of reasons like using Sync, or using another app on the same vault. (Ex: Using vscode to run a global search & replace).

I understand you aren't computing for every file - I'm saying that you're re-computing the current file even if a different file was changed.

Ideally you only need to trigger a re-compute when the specific file you have open in view has changed.
* I hadn't realized it worked like this.  Thank you for the explanation!  Changed.

lishid:
>  I wasn't sure how javascript handles && for if (v && v.file && ... ) so I just split them out.

`if (v && v.file && v.file === file)` will work just fine!

> I hadn't realized it worked like this. Thank you for the explanation! Changed.

Comparing the file pointer should work too, instead of comparing the file paths.

PR #690, opened 2021-12-17
add quoth entry to community plugins

erykwalder:

## Repo URL

Link to my plugin: https://github.com/erykwalder/quoth

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:

https://github.com/erykwalder/quoth/blob/a6a51f5fad175584d025f50f409b144419d084a9/src/selection.ts#L11
Why can't you do this in `getSelectedRange` instead? I feel like having this event fire on every selection change is kinda unnecessary.

https://github.com/erykwalder/quoth/blob/a6a51f5fad175584d025f50f409b144419d084a9/src/copyReference.ts#L36
I would recommend moving this check to where you hook the event, so you don't end up registering a useless event.

lishid:
You may also want to consider adding a bit more to the plugin description to facilitate searching for your plugin.

erykwalder:
> https://github.com/erykwalder/quoth/blob/a6a51f5fad175584d025f50f409b144419d084a9/src/selection.ts#L11 Why can't you do this in `getSelectedRange` instead? I feel like having this event fire on every selection change is kinda unnecessary.

That's how it started, but pulling up the command palette on desktop changes the selection once, and on mobile twice. So this listens in a way that ignores selection changes on the way to the command palette. Definitely open to other ideas though, in case I've overlooked something.

> https://github.com/erykwalder/quoth/blob/a6a51f5fad175584d025f50f409b144419d084a9/src/copyReference.ts#L36 I would recommend moving this check to where you hook the event, so you don't end up registering a useless event.

Definitely would be my preference too, and I can move the isMobile part for sure. One thing I couldn't figure out how to do was unregister the event if someone toggles off the setting. Maybe I can declare some things that aren't in the API to unregister it.

> You may also want to consider adding a bit more to the plugin description to facilitate searching for your plugin.

Good call!

PR #692, opened 2021-12-19
Add obsidian-completr

tth05:

## Repo URL

Link to my plugin: https://github.com/tth05/obsidian-completr

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


WhiteNoiseDev:
Hi 
there are already a few plugins that accomplish this. Not sure if you are willing to join forces with them.
- https://github.com/tadashi-aikawa/obsidian-various-complements-plugin
- https://github.com/yeboster/autocomplete-obsidian
- https://github.com/joeyuping/quick_latex_obsidian

tth05:
Thanks for hints, here's my reasoning for why I made this. 

1. [yeboster/autocomplete-obsidian](https://github.com/yeboster/autocomplete-obsidian)
I used this before and it's buggy, laggy and missing a few features I like. Also hasn't been updated for a while. I looked at the codebase and wasn't happy with the implementation and it seemed unnecessary complex in some parts, so that's why I decided to make my own plugin. 
2. [tadashi-aikawa/obsidian-various-complements-plugin](https://github.com/tadashi-aikawa/obsidian-various-complements-plugin)
I overlooked this (probably because of the name), but it's really not what I'm looking for when it comes to auto-completion. I guess people can use this if they need support for more complex languages.
3. [joeyuping/quick_latex_obsidian](https://github.com/joeyuping/quick_latex_obsidian)
I'd say this complements other completion plugins nicely, but is vastly different in what it tries to achieve.

lishid:

https://github.com/tth05/obsidian-completr/blob/b0dde7f56f6e21f1b41e2853a5a19a40dc8e6a3a/src/main.ts#L74
You may want to run these async tasks actually asynchronously instead of delaying the loading of the entire app and other plugins. Just purposefully not awaiting them, or moving them to another function that you don't await for would solve this.

https://github.com/tth05/obsidian-completr/blob/b0dde7f56f6e21f1b41e2853a5a19a40dc8e6a3a/src/main.ts#L30
Your auto-completion mechanism only works for the legacy editor and only on the desktop app. Perhaps consider using the new editor autocomplete mechanism here: https://github.com/obsidianmd/obsidian-api/blob/c148c24b30dc68d9835820b4169b95d3a038a0f6/obsidian.d.ts#L2381

tth05:
> Your auto-completion mechanism only works for the legacy editor and only on the desktop app. Perhaps consider using the new editor autocomplete mechanism here

I didn't see Live Preview was released. Did you mean to link `Plugin#registerEditorExtension` instead of `Plugin#registerEditorSuggest`? I'm guessing the things I do in `iterateCodeMirrors` should be done using extensions instead now with CodeMirror 6?

lishid:
> I didn't see Live Preview was released. Did you mean to link `Plugin#registerEditorExtension` instead of `Plugin#registerEditorSuggest`? I'm guessing the things I do in `iterateCodeMirrors` should be done using extensions instead now with CodeMirror 6?

I think your plugin may be better suited for using `registerEditorSuggest`, because it provides auto-completion which is what EditorSuggest is exactly designed for, and uses the shared `Editor` interface that works on both legacy and new editors. Internally, we use it for link auto-completion and tag auto-completion. It does come with some limitations like less control from your side on when it can be triggered, but I still think it's worth looking into.

Otherwise, you could consider maintaining both a cm5 version like right now, and add a cm6 version with `registerEditorExtension`, though you will then have to maintain two different versions that are API incompatible.

tth05:
> I think your plugin may be better suited for using registerEditorSuggest, because it provides auto-completion which is what EditorSuggest is exactly designed for, and uses the shared Editor interface that works on both legacy and new editors. Internally, we use it for link auto-completion and tag auto-completion. It does come with some limitations like less control from your side on when it can be triggered, but I still think it's worth looking into.

I'm already registering an `EditorSuggest` one line above the one you linked:
https://github.com/tth05/obsidian-completr/blob/b0dde7f56f6e21f1b41e2853a5a19a40dc8e6a3a/src/main.ts#L29

I've also made a new release and fully migrated to codemirror 6. I also removed the two `await`s when the plugin loads :)

lishid:
Looks fine. A couple of things I want to mention:

- Inserting <style> elements into anywhere but <head> is not DOM spec compliant. Although it kinda works, you should add it to the head element (and remember to remove it when unloading) instead. (Or just add it to your css file). Either way, right now you are not cleaning it up properly when destroying/unloading, and also not checking whether it exists already.

https://github.com/tth05/obsidian-completr/blob/b38b8058258494cc291d5a5abdbc48d4d20d9c70/src/main.ts#L117
Here you should use the MarkdownView StateField rather than attempting to get the current active view from the workspace: 
https://github.com/obsidianmd/obsidian-api/blob/master/obsidian.d.ts#L953

https://github.com/tth05/obsidian-completr/blob/b38b8058258494cc291d5a5abdbc48d4d20d9c70/src/main.ts#L130
Here I think if you return false (and don't preventDefault), cm6 will continue and let other things handle the event.

tth05:
> Inserting <style> elements into anywhere but is not DOM spec compliant. Although it kinda works, you should add it to the head element (and remember to remove it when unloading) instead. (Or just add it to your css file). Either way, right now you are not cleaning it up properly when destroying/unloading, and also not checking whether it exists already.

Thanks for the hint. My original intention was to allow customization of these colors, but I guess the same thing can be done anyway by providing custom css if a user wants to. I've removed the code and just added the classes to `styles.css`.

> Here you should use the MarkdownView StateField rather than attempting to get the current active view from the workspace

Fixed it. Didn't know about the new API.

> Here I think if you return false (and don't preventDefault), cm6 will continue and let other things handle the event.

That is correct, but special formatting features like lists will not automatically get a new line with the `-` prefix and auto-indentation. 
It instead produces just a new line. My code ensures the behavior is the same as when pressing `Enter` normally. Here you can see what happens if I just return:
Lumpy_Iberianemeraldlizard

lishid:
> That is correct, but special formatting features like lists will not automatically get a new line with the - prefix and auto-indentation.
It instead produces just a new line. My code ensures the behavior is the same as when pressing Enter normally. Here you can see what happens if I just return:

I see. I don't know why it would do that since you're already registering with Prec.highest, but I guess if there's no other way then leave it like that for now.

PR #703, opened 2021-12-28
Added obsidian-timestamper Plugin

Gru80:
Added obsidian-timestamper plugin, a plugin to insert customized or predefined time- and date-stamps at the current cursor position

## Repo URL

Link to my plugin:
https://github.com/Gru80/obsidian-timestamper

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  Android
  - [x]  iOS 
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3098,5 +3098,12 @@
         "description": "Create markdown-based presentations in Obsidian",
         "repo": "MSzturc/obsidian-advanced-slides",
         "branch": "main" 
+    },
+    {
+        "id": "obsidian-timestamper",
+        "name": "TimeStamper",
+        "description": "Insert a customized or predefined time- or date-stamp at the current cursor position.",
+        "author": "Martin Eder",
+        "repo": "Gru80/obsidian-timestamper"
     }
 ]


Gru80:
I've renamed the default branch from `main` to `master` in order to be in line with the current obsidian workflow

lishid:
https://github.com/Gru80/obsidian-timestamper/blob/b795760bbe674c780cb0d017f4e74c8e328c1bd5/src/main.ts#L12
Have you considered using `moment`, which comes with Obsidian by default (`import {moment} from 'obsidian';`) and is used throughout the app elsewhere for date formatting (like templates and daily notes)?

Gru80:
> https://github.com/Gru80/obsidian-timestamper/blob/b795760bbe674c780cb0d017f4e74c8e328c1bd5/src/main.ts#L12
> Have you considered using `moment`, which comes with Obsidian by default (`import {moment} from 'obsidian';`) and is used throughout the app elsewhere for date formatting (like templates and daily notes)?

Thanks a lot for your feedback!
Actually I have considered `moment` indeed (although I have to confess I was not aware that it currently comes with Obsidian).
The reason I have chosen `datetime` is because it is extremely lightwight and independent - from the [website of moment.js](https://momentjs.com/docs/#/-project-status/) I've learned that its longer-term future is kind of doubtful, so whenever Obsidian would consider replacing it at any point in time, the plugin would not be affected and still work.

Please let me know your thoughts on that, if it matters I can adapt the plugin in order to use `moment` instead.

Gru80:
I understand your line of argumentation and it sounds reasonable to me, so I just committed changes for a new version 1.1.0 of the plugin which uses `moment.js` instead of `dateformat`

Gru80:
@lishid - the new release 1.1.0 wich uses `moment` is now tested and [published](https://github.com/Gru80/obsidian-timestamper/releases/tag/1.1.0)

PR #706, opened 2021-12-30
Add Import Foundry and Import JSON plug-ins

farling42:

## Repo URL

Link to my plugins:
https://github.com/farling42/obsidian-import-foundry
https://github.com/farling42/obsidian-import-json

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3106,5 +3106,19 @@
         "author": "Feng Peng",
         "repo": "QAMichaelPeng/obsidian-graphviz",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-import-json"
+        "name": "JSON Importer",
+        "author": "farling42",
+        "description": "Import a JSON file containing an array of data, creating notes from a Handlebars template file",
+        "repo": "farling42/obsidian-import-json"
+    },
+    {
+        "id": "import-foundry"
+        "name": "Foundry world Importer",
+        "author": "farling42",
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/farling42/obsidian-import-foundry/blob/9949fd00009c4ef305e7ae3cfe5e71d25a22ab43/main.ts#L117
This fails to properly await for the results one at a time (it will run everything in parallel and ignore the process). Use a for loop here instead.

https://github.com/farling42/obsidian-import-foundry/blob/9949fd00009c4ef305e7ae3cfe5e71d25a22ab43/main.ts#L248
I highly suggest avoid mixing `async` `await` and the old callback style `.then()/.catch()` usage. Best to stick to the new async syntax.

lishid:
https://github.com/farling42/obsidian-import-json/blob/6d038f40cb9ae7ad33dad7480f630ad3bded01b0/main.ts#L123
Same problem here.

farling42:
Both plugins have been updated to remove foreach() and then() loops

farling42:
> I noticed that one of your plugin has an `obsidian-` prefix in the ID and the other doesn't. I don't think it matters but may want to normalize that just for consistency.

Oh yeah. I forgot to remove that in the plugin information. I use it as a prefix for the repository name 😁

PR #707, opened 2021-12-30
Add Rant-Lang plugin

lanice:

## Repo URL

Link to my plugin: https://github.com/lanice/obsidian-rant/

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3106,5 +3106,13 @@
         "author": "Feng Peng",
         "repo": "QAMichaelPeng/obsidian-graphviz",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-rant",
+        "name": "Rant-Lang",
+        "author": "Leander Neiß",
+        "description": "Thin wrapper around the high-level procedural templating language Rant",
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/lanice/obsidian-rant/blob/85e05ac5bb2fcb4f4a0a2b6852b1710f1f081356/src/main.ts#L136
The active view may not always be the pane that's calling the processor - there are cases where your processor is called from a non active pane. Either way, I don't think this should be necessary as your processor should be properly unloaded from the view's lifecycle.

https://github.com/lanice/obsidian-rant/blob/85e05ac5bb2fcb4f4a0a2b6852b1710f1f081356/src/main.ts#L23
You may want to add an unload function to cleanup everything here.

lanice:
Thank you for the review comments! I will work on them.

lishid:
Great work!

PR #714, opened 2022-01-05
Add Binary File Manager to community plugins

qawatake:

## Repo URL

Link to my plugin: https://github.com/qawatake/obsidian-binary-file-manager-plugin

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/qawatake/obsidian-binary-file-manager-plugin/blob/7aad9d3fb41c374d08375f24670388a003ba845d/src/Generator.ts#L33
This function shouldn't be async.

https://github.com/qawatake/obsidian-binary-file-manager-plugin/blob/7aad9d3fb41c374d08375f24670388a003ba845d/src/main.ts#L65
And this shouldn't need to await this. (there's a couple of more places)

PR #716, opened 2022-01-05
Add own plugin

liammagee:

## Repo URL

Link to my plugin: https://github.com/liammagee/obsidian-topic-linking

## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


lishid:
https://github.com/liammagee/obsidian-topic-linking/blob/93fe52f4c295beda8e212722b2431a34baccce9d/src/pdf.ts#L364
Couple of issues with the way you're handling here:
- You shouldn't rely on callback function style array iteration since it will run everything in parallel. Instead, you should use a regular for loop.
- debounce does not return a Promise, so waiting for it does not do anything. I assume your intention here is to wait for the processing to complete.


Try something like this, which would greatly simplify your code:
```ts
for (let file of files) {
  await this.processPDF(vault, settings, fileToProcess);
}
```

https://github.com/liammagee/obsidian-topic-linking/blob/93fe52f4c295beda8e212722b2431a34baccce9d/src/bookmark.ts#L33
https://github.com/liammagee/obsidian-topic-linking/blob/93fe52f4c295beda8e212722b2431a34baccce9d/src/bookmark.ts#L35
Same here, don't use `.map` or `.forEach` in async functions unless you intentionally want to run everything in parallel (you may run into resource contention when doing that).
I recommend using a for loop and awaiting the result of each step, rather than reading the file contents of all files into memory first, and then process them one by one.

https://github.com/liammagee/obsidian-topic-linking/blob/93fe52f4c295beda8e212722b2431a34baccce9d/src/topic.ts#L105
Same here as well.

liammagee:
Thanks, I've modified things as suggested.

lishid:
Here I suggest you do the following:
https://github.com/liammagee/obsidian-topic-linking/blob/b1dafe923ed4d682f2c1c70a0676d5103b2c1df2/src/bookmark.ts#L33

```ts
        for (let file of files) {
            let contents = await vault.cachedRead(file);
            let links: string[] = contents.match(/https*:\/\/[^ )]*/g);
            ...
        }
```
Doing it like this will avoid having to hold the contents of all the files in memory all at once in an array.

PR #723, opened 2022-01-09
add new plugin: para-shortcuts

gOATiful:

## Repo URL

Link to my plugin: https://github.com/gOATiful/para-shortcuts

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
Love your username!

https://github.com/gOATiful/para-shortcuts/blob/89db92b94af4d34829ece4189bcaaad0b3df0b37/src/main.ts#L3
Weird that your imports all have `.js`?

https://github.com/gOATiful/para-shortcuts/blob/89db92b94af4d34829ece4189bcaaad0b3df0b37/src/main.ts#L7
Should avoid using NodeJS API if you intent for your plugin to run on mobile. For joining paths you can simply use `/` in between and then run `normalizePath` from `obsidian` to have it normalized to Obsidian internal paths. For dirname probably just do some string operations.

https://github.com/gOATiful/para-shortcuts/blob/89db92b94af4d34829ece4189bcaaad0b3df0b37/src/main.ts#L212
you can use `moment` from `obsidian` for a proper date format.

https://github.com/gOATiful/para-shortcuts/blob/89db92b94af4d34829ece4189bcaaad0b3df0b37/src/main.ts#L56
Recommend using `async`/`await` instead of using `.then()` and `.catch()`. (Applies to other places further down as well)

https://github.com/gOATiful/para-shortcuts/blob/89db92b94af4d34829ece4189bcaaad0b3df0b37/src/main.ts#L124
Here you don't need to wrap this in a Promise, as you're in an async function inside anyway. I also don't recommend using a `while` clause to achieve this because you run the risk of getting stuck (for example, renameFile errors out but createFolder always suceeds).

lishid:
https://github.com/gOATiful/para-shortcuts/blob/89db92b94af4d34829ece4189bcaaad0b3df0b37/src/main.ts#L186
Same here, the promise is not necessary and you can simplify it down to just the for loop without try-catch.

gOATiful:
@lishid thank you for your review! 
I have addressed the following issues:

- [x] `Weird that your imports all have .js?` - fixed
- [x] `run normalizePath from obsidian` - changed
- [x] `moment from obsidian` - changed
- [x] `the risk of getting stuck` - changed control flow, so it can not happen
- [x] `the promise is not necessary` - changed the occurrences accordingly

I also created a new release:
https://github.com/gOATiful/para-shortcuts/releases/tag/1.0.1

lishid:
Looks great!

PR #724, opened 2022-01-09
Add ObsidianTweaks

JeppeKlitgaard:

## Repo URL

Link to my plugin: https://github.com/JeppeKlitgaard/ObsidianTweaks

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


JeppeKlitgaard:
Resubmitting this since it appears that some people are actually using it!

The Obsidian API now exposes the functionality needed to play around with selections and words such that ObsidianTweaks no longer needs to resort to the underlying CodeMirror instance.

lishid:
One thing you might want to look into is using `editorCallback` rather than `callback` in your command definitions, see examples from here: https://marcus.se.net/obsidian-plugin-docs/guides/commands#editor-commands

Otherwise it looks pretty straightforward.

lishid:
Also I suggest tweaking the description in the listing so users can actually know what your plugin does

JeppeKlitgaard:
Good point - I will probably 'tweak' to better match the current implementation which is mostly focused on improved toggle behaviour and some code editor commands

JeppeKlitgaard:
Both of these suggestions have been implemented in the now released version 0.2.7

lishid:
Do a new PR to update your description in this repo as well.

PR #727, opened 2022-01-11
Add Obsidian Sortable Plugin.

alexandru-dinu:

## Repo URL

Link to my plugin: https://github.com/alexandru-dinu/obsidian-sortable

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/alexandru-dinu/obsidian-sortable/blob/ca4e6a9c6d3703ab5541b3d3d1224f3c4b84c545/src/main.ts#L19
I recommend removing all the setting-related parts from your plugin given that you aren't using them.

https://github.com/alexandru-dinu/obsidian-sortable/blob/ca4e6a9c6d3703ab5541b3d3d1224f3c4b84c545/src/main.ts#L23
When hooking this as a global event, you risk interfering with other parts of the app UI as well as plugins, such as plugin modals, settings tab, etc.

I think at the very least, you'll want to verify that this event is fired inside a Markdown rendering context with `element.matchParent('.markdown-preview-view')`.

https://github.com/alexandru-dinu/obsidian-sortable/blob/ca4e6a9c6d3703ab5541b3d3d1224f3c4b84c545/src/sortable.ts#L38
Rather than using a DOM attribute to identify your table, you should use a `WeakMap<HTMLTableElement, TableState>`, which is properly garbage collected when the table is removed from DOM. With that setup, you no longer need to generate any ID, or modify the DOM itself for identification purposes.

https://github.com/alexandru-dinu/obsidian-sortable/blob/ca4e6a9c6d3703ab5541b3d3d1224f3c4b84c545/src/sortable.ts#L26
Should be `HTMLElement` instead of `HTMLInputElement`.

alexandru-dinu:
Hi,

Thank you for the suggestions.

Using a `WeakMap<HTMLTableElement, TableState>` does really simplify the logic. 
This solves one problem I had with a use case where the users sorts a table, then changes to a different files.
I understand that using `WeakMap` I can simply let GC took action, as the reference to the table key is now gone.

However, there is still one quirk that I am aware of:
If a user sorts some tables in a file, then unloads the plugin, I would like to reset the tables to their default state.
This was easily achievable by traversing the `tableStates` map and resetting each table value.
If I understand correctly, there's no way to iterathe through a `WeakMap`.

Therefore, I assume that I have to get all the `HTMLTableElement`s in the active panes, check if they exist in the map, and if so, delete them.

Is this the preferred way to approach this case?

Thank you!

lishid:
Yeah, unfortunately we are not allowed to traverse through WeakMaps so your only bet is to query it from DOM. One slight problem is that if the DOM is out of view and not attached (like say, in a hidden pane) it would not be caught, but I think that's the best we can do for now without risking memory leaks. Disabling the plugin isn't something that happens often, and tables gets re-rendered when users switch files so it's probably not a big deal.

alexandru-dinu:
Thank for the suggestions -- I pushed the changes to the repository just now.

Here is the commit permalink:
https://github.com/alexandru-dinu/obsidian-sortable/commit/d83b928e13d8107e4c08b42d33dae5f6fe91d221

lishid:
Thanks for the quick patch!

PR #729, opened 2022-01-13
Add link-to-obsidian

moelody:

## Repo URL

Link to my plugin: https://github.com/moelody/link-to-obsidian

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3216,5 +3216,12 @@
         "author": "farling42",
         "description": "Imports your journal entries from your selected Foundry VTT world into a folder within your Obsidian Vault",
         "repo": "farling42/obsidian-import-foundry"
+    },
+    {
+        "id": "link-to-obsidian",
+        "name": "Link Server",
+        "author": "moelody",
+        "description": "This plugin will open a reverse proxy server at port 3333 to get wikiLink information Obsidian API.",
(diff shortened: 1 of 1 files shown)


lishid:
I think your plugin ID should be tweaked to match your plugin name. "link-to-obsidian" doesn't really describe your plugin well. "link server" is a better name and maybe even better with "link info server".

https://github.com/moelody/link-to-obsidian/blob/fbe9d4241604a580d2a87bff72787265c8a8a95d/src/proxy/server.ts#L33
Should use `readCached` for better performance.

https://github.com/moelody/link-to-obsidian/blob/fbe9d4241604a580d2a87bff72787265c8a8a95d/src/proxy/server.ts#L36
I think the `?` is not needed here because you already test for `file` earlier.

https://github.com/moelody/link-to-obsidian/blob/fbe9d4241604a580d2a87bff72787265c8a8a95d/src/proxy/server.ts#L46
Should test for `adapter instanceof FilesystemAdapter` first and then use `adapter.getBasePath()` instead to get the basePath.

PR #732, opened 2022-01-14
Add elf

mirnovov:

## Repo URL

Link to my plugin: https://github.com/mirnovov/elf

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3216,5 +3216,13 @@
         "author": "farling42",
         "description": "Imports your journal entries from your selected Foundry VTT world into a folder within your Obsidian Vault",
         "repo": "farling42/obsidian-import-foundry"
+    },
+    {
+        "id": "elf",
+        "name": "Enhanced Live Preview Features",
+        "author": "mirnovov",
+        "description": "Incorporates limited Reading View functionality into Live Preview mode",
+        "repo": "mirnovov/elf",
+        "branch": "main"
     }
 ]


pjeby:
What is this line intended to do?

https://github.com/mirnovov/elf/blob/0208e2f3c726e66167e78dc8c47429869757db2d/main.ts#L67

Note that the debounce() function doesn't actually do anything, nor arrange for anything to happen.  It returns a function you can call, and *that function* keeps state to end up being called later.  Generally speaking, you want to save the debounced function somewhere, so you can ensure that the *same* function is called each time.  The code as written neither saves the result of debounce() nor ever calls it.

mirnovov:
@pjeby My intent is to wait until updates have stopped for x amount of time before calling it.

The correct way of using `debounce` is something like this, right?

```js

update(update: ViewUpdate): void {
    if (this.live && (update.docChanged || update.viewportChanged)) { 
        this.debouncedRegisterLinks(update.view);
    }
    else if (this.live) {
        setTimeout(() => this.addEventHandlers(update.view), 10);
    }
}

debouncedRegisterLinks = debounce(this.registerLinks, 200);
	
registerLinks(view: EditorView): void {
```

mirnovov:
Anyway, after implementing the debounce function correctly, I've found it causes my `Link` classes to get out of sync with the editor state, so I've dropped it for now, and I've force-updated my initial commit to reflect that

pjeby:
> The correct way of using debounce is something like this, right?

Almost.  If you want it to wait until things have *stopped*, you also need to pass `true` as the third argument.  Otherwise, the debounce operates as a rate limiter, such that it will ensure a minimum 200 ms between calls to the underlying function,  To get it to be "200ms of silence", you need the third argument to be true.

>  I've force-updated my initial commit to reflect that

No need to force-update, that's what revision control is for -- to let other people learn from your mistakes.  :wink:

lishid:
Thanks for submitting your plugin. Before I do an in-depth review, let's talk about its features just in case it's subject to change:

> Some link interactions (including clicking without holding down Ctrl/Cmd). Links can still be edited when the cursor is within their bounds.

We're actually planning to change this behavior to be default soon. It's been reported many times the current behavior isn't desirable so we're looking to fix that soon.

> Unfocusing on note open. This also includes deselecting text where possible so it doesn't show the Markdown syntax.

This has actually been implemented in Obsidian v0.13.20.

> Minor rendering changes to appear visually closer to Reading mode.

I'm not sure what this entails, but I haven't seen anything else.


I really appreciate your work attempting to fix what we didn't do well in core. Given that we'll be implementing these features soon, do you still want to enlist and maintain your plugin?

mirnovov:
> I'm not sure what this entails, but I haven't seen anything else.

It hides/de-emphasises comments and escaped characters when they've not in the focused line. Since this is just CSS, it can be done via a theme or snippet; it isn't really worth having a whole plugin solely for it.

> I really appreciate your work attempting to fix what we didn't do well in core. Given that we'll be implementing these features soon, do you still want to enlist and maintain your plugin?

Nah, if you guys are going to implement everything, there's no real point. I developed the plugin with the expectation that this might happen, so I have no issues in that regard; if anything it's less work for me. Some of my workarounds would have most likely hit their limitations sooner or later anyway. If there's any Live Preview additions people want in the future, I may revive the plugin, but for now I'll kill it.

lishid:
Ok makes sense, I can definitely see there's a lot of friction to apply the hack, so it's better if we support this from core.

PR #734, opened 2022-01-15
add plugin: weather-fetcher

fyears:

## Repo URL

Link to my plugin: https://github.com/fyears/weather-fetcher

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3211,6 +3211,13 @@
         "repo": "farling42/obsidian-import-json"
     },
     {
+        "id": "weather-fetcher",
+        "name": "Weather Fetcher",
+        "author": "fyears",
+        "description": "Fetch and insert current weather into the editor of Obsidian.",
+        "repo": "fyears/weather-fetcher"
+    },
+    {
         "id": "import-foundry",
         "name": "Foundry world Importer",
         "author": "farling42",
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3211,17 +3211,17 @@
         "repo": "farling42/obsidian-import-json"
     },
     {
-        "id": "weather-fetcher",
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/fyears/weather-fetcher/blob/a7f487dc8ae34155e89da1c00db1c366bff91c55/main.ts#L48
Don't do this in the production version of your plugin. This will end up with a lot of disk writes for logs which isn't desirable for end users.
It also contains a memory leak (your log array keeps growing bigger). You should use `adapter.append` instead.

https://github.com/fyears/weather-fetcher/blob/a7f487dc8ae34155e89da1c00db1c366bff91c55/main.ts#L132
https://github.com/fyears/weather-fetcher/blob/a7f487dc8ae34155e89da1c00db1c366bff91c55/main.ts#L92
This function doesn't need to be async.

https://github.com/fyears/weather-fetcher/blob/a7f487dc8ae34155e89da1c00db1c366bff91c55/main.ts#L106
Your settings object is rather simple and doesn't need to be deep cloned. Either way, totally recommend avoiding external dependencies like lodash as imports as much as you can.

lishid:
Looks good. One more thing I'd recommend:

https://github.com/fyears/weather-fetcher/releases/tag/0.0.3
Your release seems to include the entire feather-icons library which bloats your plugin to 138kb (90% of which is feather-icons and its dependencies).

Maybe you can copy the svg file of the icon you're using and hardcode it instead of importing it from feather-icons?

fyears:
Oh no that’s bad! I thought there would be some tree shaming… I would fix it in next version. Thanks!

PR #736, opened 2022-01-21
Add Core Search Assistant to community plugins

qawatake:

## Repo URL

Link to my plugin: https://github.com/qawatake/obsidian-core-search-assistant-plugin

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/qawatake/obsidian-core-search-assistant-plugin/blob/8df64a98148a1513ed6db7c33ac3a4c26ce96ffa/src/components/CardView.ts#L9
The way you're managing leaves here will cause issues with the rest of the API, for example, `Workspace.iterateLeaves`.
Instead, you should inherit the `WorkspaceParent` abstract class. I know the documentation on it is sparse but the three class fields that you may be interested are:

```
	allowSingleChild: boolean = false; // Automatically close when last child is removed
	autoManageDOM: boolean = true; // Set to false if you want to manually manage child/parent DOM attachment
	children: WorkspaceItem[] = []; // The list of child items
```

qawatake:
@lishid 

Thank you for your review!
Let me ask you three questions.

1. I manage leaves in a similar way that you pointed out in the following components. Should I modify them too?
  - https://github.com/qawatake/obsidian-core-search-assistant-plugin/blob/8df64a98148a1513ed6db7c33ac3a4c26ce96ffa/src/components/PreviewModal.ts#L99-L106
  - https://github.com/qawatake/obsidian-core-search-assistant-plugin/blob/8df64a98148a1513ed6db7c33ac3a4c26ce96ffa/src/components/WorkspacePreview.ts#L46-L55

2. What should I pass args to constructor of a descendant of `WorkspaceParent`? Is there a problem with the code below?
  ```ts
  export class PreviewGenerator extends WorkspaceParent {
	app: App;

	constructor(app: App) {
		super(app);
		this.app = app;
	}
}
  ```
3. How should I add leafs to a descendant of `WorkspaceParent`? `insertChild`?

lishid:
Hmm I've just reviewed the internal code and I realized that we currently don't have a way to register plugin-managed WorkspaceParents for serialization and deserialization, which means it won't save properly. Unfortunately, the code there is still a bit immature because there hasn't been strong use-cases for other kinds of workspace parents yet.

I guess keeping what you do right now is probably best you can do. The consequences would be less destructive IMO.

Should we proceed with the merge?

qawatake:
@lishid 
> Hmm I've just reviewed the internal code and I realized that we currently don't have a way to register plugin-managed WorkspaceParents for serialization and deserialization, which means it won't save properly. Unfortunately, the code there is still a bit immature because there hasn't been strong use-cases for other kinds of workspace parents yet.

I see.  I hope that some way to register plugin-managed WorkspaceParents will be implemented.

BTW, I'm replacing the way of displaying previews as follows.
Is it OK?

*BEFORE*:
```ts
const previewContainerEl = this.createPreviewContainerEl(item, id);
const leaf = new (WorkspaceLeaf as any)(this.app) as WorkspaceLeaf;
leaf.openFile(item.file, { state: { mode: 'preview' } });
previewContainerEl.appendChild(leaf.containerEl);
this.leafs.push(leaf);
```

*AFTER*:
```ts
const previewContainerEl = this.createPreviewContainerEl(item, id);
const leaf = new (WorkspaceLeaf as any)(this.app) as WorkspaceLeaf;
const previewView = new MarkdownView(leaf).previewMode;
previewView.view.file = item.file; // necessary to remove error message
previewView.set(content, false); // load content
previewContainerEl.appendChild(previewView.containerEl);
this.leafs.push(leaf);
```

Let me explain the reason for this replacement ↓.
I'd like to catch the 'layout-change' event.
The former code includes `leaf.openFile`, which triggers the 'layout-change' events.
This is not good for me because I cannot distinguish normal 'layout-change' events from ones triggered by `leaf.openFile`.
In my experiment, the latter code did not trigger events.

lishid:
I would not suggest doing that change because you're making a lot of assumptions on the internals of the MarkdownView. For example, you are not correctly setting the internal data fields related to the file, which could cause it to accidentally wipe the file empty should something triggers a save.

`openFile` internally goes through the proper lifecycle of opening the file, hooking up events, etc.

qawatake:
@lishid 

I see. Thank you.
So I will revert codes and try to manage 'layout-change' events in other ways.

I'll let you know when it's done.

qawatake:
@lishid 

I changed the code about rendering markdown views.
Could you check it again?
https://github.com/qawatake/obsidian-core-search-assistant-plugin/blob/ebd7e843315ad0991c5aebaccb4a70f58b0d011f/src/MarkdownViewRenderer.ts

qawatake:
@lishid 

BTW, does `new WorkspaceLeaf(app)` have any side effect on CPU usage?
The CPU usage after using `new WorkspaceLeaf(app)` is higher by 3-5 % than the one after using `workspace.getLeaf(true)` even though I detach leaves.

lishid:
> BTW, does new WorkspaceLeaf(app) have any side effect on CPU usage?
The CPU usage after using new WorkspaceLeaf(app) is higher by 3-5 % than the one after using workspace.getLeaf(true) even though I detach leaves.

I am not aware of anything that could affect this. If it's indeed the case you can find out more using the devtools performance profiler.

https://github.com/qawatake/obsidian-core-search-assistant-plugin/blob/ebd7e843315ad0991c5aebaccb4a70f58b0d011f/src/MarkdownViewRenderer.ts#L83
I don't recommend using `view.modes` directly, it's an internal implementation that you should be touching. To set the mode, you can use `setState({mode: 'source'})`. For highlights and scroll you can use `view.editor` rather than `view.modes.source.editor`.

qawatake:
@lishid 

Thank you for your review!

> I don't recommend using view.modes directly, it's an internal implementation that you should be touching. To set the mode, you can use setState({mode: 'source'}). For highlights and scroll you can use view.editor rather than view.modes.source.editor.

I modified the codes as below.
https://github.com/qawatake/obsidian-core-search-assistant-plugin/compare/main...ft-safer-api

> I am not aware of anything that could affect this. If it's indeed the case you can find out more using the devtools performance profiler.

I investigated the problem more by using devtools, and found that there was no problem. The process after using my plugin was about the same as the one after normally opening files in new panes. Sorry for bothering you.

PR #739, opened 2022-01-21
add lock screen plugin

ericbiewener:

## Repo URL

Link to my plugin: https://github.com/ericbiewener/obsidian-lock-screen-plugin

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3216,5 +3216,12 @@
         "author": "farling42",
         "description": "Imports your journal entries from your selected Foundry VTT world into a folder within your Obsidian Vault",
         "repo": "farling42/obsidian-import-foundry"
+    },
+    {
+        "id": "obsidian-lock-screen-plugin",
+        "name": "Lock Screen",
+        "author": "Eric Biewener",
+        "description": "Protect your vault with a lock screen.",
+        "repo": "ericbiewener/obsidian-lock-screen-plugin"
     }
 ]


ericbiewener:
> I have added a license in the LICENSE file.

Can someone clarify this instruction?

kometenstaub:
That means that you added a license to your project:

- [Options](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository)
- [License chooser](https://choosealicense.com/)
- [How to add a license to a repository](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository)

ericbiewener:
Gotcha, wasn't sure if there was something in the obsidian-releases repo that needed to have license information.

lishid:
https://github.com/ericbiewener/obsidian-lock-screen-plugin/blob/c5f18d048f4ed46dbed1d7e90ed5f46bf33547ed/src/main.ts#L11
This shouldn't need to await.

https://github.com/ericbiewener/obsidian-lock-screen-plugin/blob/c5f18d048f4ed46dbed1d7e90ed5f46bf33547ed/src/main.ts#L15
This doesn't seem to do anything.

https://github.com/ericbiewener/obsidian-lock-screen-plugin/blob/c5f18d048f4ed46dbed1d7e90ed5f46bf33547ed/src/main.ts#L1
Should be just `import {Plugin} from obsidian`.

https://github.com/ericbiewener/obsidian-lock-screen-plugin/blob/c5f18d048f4ed46dbed1d7e90ed5f46bf33547ed/src/main.ts#L17
Should implement an onunload function to cleanup any leftover DOM elements when the plugin is disabled.

ericbiewener:
> Should be just `import {Plugin} from obsidian`.

Why is that? The resulting esbuild code is equivalent.

[I have addressed the other comments.](https://github.com/ericbiewener/obsidian-lock-screen-plugin/commit/268175d70f4202f27aae5f5d3225e7186ec0e184)

lishid:
That's fine if you want to leave it like that. I've just seen most conventions to prefer the dereferenced version instead.

lishid:
In regards to the patch: https://github.com/ericbiewener/obsidian-lock-screen-plugin/commit/268175d70f4202f27aae5f5d3225e7186ec0e184

I think this adds a possibility of memory leak in regards to the container elements being kept in memory even after the lock screen is closed.

ericbiewener:
> I think this adds a possibility of memory leak in regards to the container elements being kept in memory even after the lock screen is closed.

Yikes, good call. I've updated the code to simply track string ids in order to re-query the node for removal upon cleanup rather than retaining a reference to the node. See [here](https://github.com/ericbiewener/obsidian-lock-screen-plugin/blob/main/src/show-lock-screen.ts#L12-L16) and [here](https://github.com/ericbiewener/obsidian-lock-screen-plugin/blob/main/src/dom-service.ts).

lishid:
LGTM. Perhaps clear the `ids` list once you finish the cleanup in case the component gets reused (it probably won't though).

ericbiewener:
> LGTM. Perhaps clear the `ids` list once you finish the cleanup in case the component gets reused (it probably won't though).

Done in a patch release.

PR #740, opened 2022-01-21
Snippetor: Create and tweak common snippets

ebullient:

## Repo URL

Link to my plugin: https://github.com/ebullient/obsidian-snippetor

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3223,5 +3223,13 @@
         "description": "Formatting attachments name (filename attachmentType indexNumber.xxx)",
         "author": "JYC333",
         "repo": "JYC333/obsidian-attachment-name-formatting"
+    },
+    {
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/ebullient/obsidian-snippetor/blob/d6217e1f412d4d014c09d2953b7430805c573d0c/src/snippetor-CreateCheckboxesModal.ts#L23
Might want to add some more error handling here so the promise doesn't hang forever (say, a crash happens in the onClose handler).

lishid:
I didn't get a chance to go through everything but it looks like you know what you're doing 🤣

PR #745, opened 2022-01-24
Adding Multi-Column Markdown to community-plugins.

ckRobinson:

## Repo URL

Link to my plugin: [Here](https://github.com/ckRobinson/multi-column_markdown)

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/ckRobinson/multi-column_markdown/blob/69ad38ab5d6f92f6896a848a3deba82bd814649c/src/main.ts#L34
Here you don't need to use an `editorCheckCallback`, and the third parameter is already the markdown view (which you don't really use anyway). Just use a regular `editorCallback`

lishid:
Hmm why not just use the repo name "multi-column-markdown" instead of "multi-column_markdown"?

ckRobinson:
You're right that would make it more consistent. I believe the "_" was inserted automatically by github when creating the repo with a space in the title and I just didn't notice or think of that. If I was to update the repo url should I open a new PR with the change?

lishid:
Yup!

ckRobinson:
PR created #769. Thanks!

PR #748, opened 2022-01-25
add obsidian save as gist plugin

ghedamat:

## Repo URL

https://github.com/ghedamat/obsidian-save-as-gist

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/ghedamat/obsidian-save-as-gist/blob/f0aa15da903048a363fe595d3b897f3bdf3c1bf0/main.ts#L24
This might return null - you should use a checkCallback to prevent your command from showing up in the list when that happens. You also aren't testing for `file.extension === 'md'`, so a user might trigger this on, say, a pdf file, and you will process it just fine.

At this point, probably easiest to just use editorCallback and use `view.file` for the active file, or `editor.getValue()`.

ghedamat:
hey @lishid thanks for the feedback


I pushed a new version and these are the changes

https://github.com/ghedamat/obsidian-save-as-gist/commit/f49a66d62d90ddbc5ae8382e6d1bcc1635841e99


let me know if I understood your suggestions correctly!

lishid:
One possibility if you wish to consider is to merge both commands into a single one: If the user selected something (`editor.getSelection() !== ''`) then use that selection, and if not, use the whole file.

ghedamat:
Good call @lishid !

Now that I use the same api it might be nicer to do that.

I wonder if it's less confusing for users though. 

Will ponder!

Thanks again for the help!

ghedamat:
@devnoname120 that's a good idea! are you interested in helping with a pr? or could you open an issue on https://github.com/ghedamat/obsidian-save-as-gist so we can track this idea?

devnoname120:
@ghedamat kinda overwhelmed with my own open-source projects right now, so I won't be able to meaningfully contribute. :(

ghedamat:
No worries! I know that feeling! Thanks for the suggestion!

PR #752, opened 2022-01-26
Add Obsidian Notes Publish

SaraVieira:

## Repo URL

https://github.com/SaraVieira/obsidian-notes-publish-plugin

Link to my plugin:

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

I have used `innerHTML` in my plugin but it takes no user input, all the code is static, let me know if changing will be better


lishid:
Sorry I missed this, I've gotten a few community moderators who are concerned with the name "Publish Notes", mind if we change it to something like `Publish Notes with Airtable`?

SaraVieira:
> Sorry I missed this, I've gotten a few community moderators who are concerned with the name "Publish Notes", mind if we change it to something like `Publish Notes with Airtable`?

I don't think that would make sense tbh, it's already in the description and it would have to be changed since my idea would be to add support for Postgres and mongodb databases

ryanjamurphy:
Just reiterating: the conflict between "Obsidian Publish" and "Obsidian Notes Publish" will be very confusing, so I think we need to find another title for this.

> [There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors.](https://twitter.com/codinghorror/status/506010907021828096?lang=en)

I think the key is that Obsidian's core plugins have a fairly generic name, owing to the fact that Obsidian is usually attached to them to designate their "official" status/to differentiate them. (There are a few community plugins that also call themselves "Obsidian ______", but I think this is a practice we should be discouraging.)

On the other hand, a unique and easily-identifiable name is a good thing for the plugin. It makes it easy to talk about it and find it on the plugin gallery.

A few possibilities:
- PubNote
- ShareLink or Share Link (as you've already written on your site's docs)
- VieiraPublish
- PTN: Publish This Note

These might be dumb, just trying to offer some suggestions to get the creative juices flowing.

I like the idea of future support for other databases btw!

SaraVieira:
Hey!!

Sorry for the delay and I completely get it, they should have more generic names otherwise it's hard to find the official plugins 

Will change everything today to be PubNote and ping you back here :)

Thank you for taking the time :)

lishid:

PR #753, opened 2022-01-26
Add frontmatter tag suggest plugin

jmilldotdev:

## Repo URL

Link to my plugin: https://github.com/jmilldotdev/obsidian-frontmatter-tag-suggest

## Release Checklist
- [ x ] I have tested the plugin on
  - [ x ]  Windows
  - [ x ]  macOS
  - [ ]  Linux
  - [ x ]  iOS _(if applicable)_
- [ x ] My GitHub release contains all required files
  - [ x ] `main.js`
  - [ x ] `manifest.json`
- [ x ] GitHub release name matches the exact version number specified in my manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
- [ x ] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [ x ] My README.md describes the plugin's purpose and provides clear usage instructions.
- [ x ] I have read the tips in https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md and have self-reviewed my plugin to avoid these common pitfalls.
- [ x ] I have added a license in the LICENSE file.
- [ x ] My project respects and is compatible with the original license of any code from other plugins that I'm using.


pjeby:
How is this different from Obsidian's existing autocompletion of tags in the front matter?  The README does not explain.

jmilldotdev:
I updated the README:

"This plugin is useful since Obsidian only shows the tag autocomplete with a # character, but if you use a # character in the `tags:` frontmatter, the file will not be considered under that tag. This plugin removes the need to have to go back and delete the # character on any stock autocompleted tags."

pjeby:
Ah!  Nice.  I misread your source and thought it was inserting the suggestion with a `#` in front.  But in any case, most users won't read the source to figure things like that out.  :wink:

lishid:
https://github.com/jmilldotdev/obsidian-frontmatter-tag-suggest/blob/33fd9957ed126d3ccbec6c09764f2e5d90840076/main.ts#L44
Should do this in your `getSuggestions` function. Plus since you aren't even reusing `this.tags` anywhere else it can probably stay as an inline variable.

https://github.com/jmilldotdev/obsidian-frontmatter-tag-suggest/blob/33fd9957ed126d3ccbec6c09764f2e5d90840076/main.ts#L73
Is this cast necessary? As far as I remember this was already typed as `Editor`.

PR #757, opened 2022-01-29
Add "Divide and Conquer" to the Community Plugins

chrisgrieser:

## Repo URL
https://github.com/chrisgrieser/obsidian-divide-and-conquer

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
- [x] My project respects and is compatible with the original license of any code from other plugins that I'm using. I have given proper attribution to these other projects in my `README.md`.


lishid:
https://github.com/chrisgrieser/obsidian-divide-and-conquer/blob/6770615142ae5e9ca25e9f7372531530c8a79eaa/main.ts#L85
You should run these async and await each of them. Right now you just start enabling/disabling all plugins at the same time and they may run into concurrency issues. Plus some plugins take longer to enable/disable and you may cause issues by ignoring when they finish.

lishid:
Instead of forEach, just use a plain for loop and await the result of each call.

chrisgrieser:
got it. Should be fixed now.

PR #759, opened 2022-01-30
Add Live2D Widget

moelody:

## Repo URL

Link to my plugin: https://github.com/moelody/live2d-widget-obsidian

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
- What does your plugin do? A screenshot would be nice, I had to do a search what _live2d_ even is.

- https://github.com/moelody/live2d-widget-obsidian/blob/73ee23e6d70e4f1e2636b3b7e338a07fb3a55226/src/settings.ts#L199-L226
Why do you load live2d the settings, yet you unload it in the main class? (you also have some minor code duplication there, which I guess is unnecessary)

- your method for loading the settings menu is interesting, it took me a minute to understand what you are doing.
- This method of putting the description and the name of an setting in seperate places can lead to problems further down the line when you add some new settings and mess up the ordering.
- It would be more user friendly if you used non abbreviatied names in your settings, I can only guess that 'vOffset' means 'vertical offset'

- You request javascript from multiple online servers
[Please dont assume "Internet everywhere"](https://forum.obsidian.md/t/plugin-theme-css-authors-please-don-t-assume-internet-everywhere/21967?u=joethei), this can break your plugin

- Why do you load `express` and `get-port` in your `package.json`?

lishid:
Suggest adding some screenshots to your README.md to help understand what Live2D is.

https://github.com/moelody/live2d-widget-obsidian/blob/73ee23e6d70e4f1e2636b3b7e338a07fb3a55226/src/main.ts#L22
This will improperly remove all scripts from the head element, including ones added by other plugins. 

https://github.com/moelody/live2d-widget-obsidian/blob/73ee23e6d70e4f1e2636b3b7e338a07fb3a55226/src/settings.ts#L224
You should import this script and bundle it with your plugin instead of having it dynamically loaded from the internet.

lishid:
@joethei Thanks for the help!

moelody:
Thanks for the advise. I'll try load scripts on worker.

moelody:
I'm done. Please take a look！

joethei:
- The plugin is marked as mobile compatible but uses express, which is using node.js API's
- I don't understand why you included minified dependencies in the repo(`lib` folder and potentially `src/lodash.ts`) instead of including them in the `package.json`, please clarify.

- <https://github.com/moelody/live2d-widget-obsidian/blob/b70726bbefc93254e5facffc2eaf71330f3cd533/src/settings.ts#L307-L311> this can be replaced with a call to the `debounce` method the obsidian API provides: <https://github.com/obsidianmd/obsidian-api/blob/71b3b362bb1f0e8f3c13f1a1d92f65427d2ad477/obsidian.d.ts#L678-L695>

- What is the reasoning behind the injection of the scripts using the express server ? I feel like this could be achieved with the obsidian API.

- The release does not contain the required files, see the README on this repo for details

lishid:
> I'm done. Please take a look！

Ok I double checked and I see what you're trying to do.

That's not really what I'm suggesting - you should be able to directly embed the Live2D script in your plugin's code in the build process, rather than having to host a local server, which in its own is already a bad idea (users will be prompted that your plugin wants to listen to a port), and also will not work on mobile.

It's also highly inefficient to host such a server just to serve a few static files.

lishid:

PR #760, opened 2022-01-30
Add Tag & Word cloud plugin

joethei:

## Repo URL

Link to my plugin: https://github.com/joethei/obsidian-tagcloud

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


lishid:
https://github.com/joethei/obsidian-tagcloud/blob/52dd7c4adf06bce50ef8ca530d9e4ed0c7110143/src/main.ts#L170
Recommend using cachedRead to get better performance here.

https://github.com/joethei/obsidian-tagcloud/blob/52dd7c4adf06bce50ef8ca530d9e4ed0c7110143/src/main.ts#L238
Hmm reading and parsing every markdown file in the vault is quite an expensive operation to be done on vault startup every time. Perhaps it would be better to lazily compute it when you have a block being rendered?

joethei:
> Hmm reading and parsing every markdown file in the vault is quite an expensive operation to be done on vault startup every time. Perhaps it would be better to lazily compute it when you have a block being rendered?

I explicitly chose to do this as a vault wide scan is really time intensive(takes ~25 minutes in my test vault with ~12.000 files, so not normal use)

I have implemented your other suggestion, currently testing it for any differences.

lishid:
> I explicitly chose to do this as a vault wide scan is really time intensive(takes ~25 minutes in my test vault with ~12.000 files, so not normal use)

Oh wow! Would it make more sense to cache your results to a file (and identified by file path/size/mtime) and re-compute only changed files from last run? I would also suggest running the profiler to see what in your processing pipeline takes so long.

lishid:
If I had to guess, it's your replace chain, which generates a lot of intermediate large strings - might be better to split the file by line and run the replacements instead.

joethei:
I did improve the performance a bit(down to 14 minutes now), and added a cache,
so the data will only be calculated if there where changes.

PR #761, opened 2022-02-01
adding obsidian-excel-to-markdown-table to plugins

ganesshkumar:

## Repo URL

Link to my plugin: https://github.com/ganesshkumar/obsidian-excel-to-markdown-table

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/ganesshkumar/obsidian-excel-to-markdown-table/blob/3fae8466c2eedf0613fab22e253528e40828fa60/src/main.ts#L9
Ctrl+Shift+V conflicts with the browser engine's "paste as plaintext". You should also use `Mod` instead of `Ctrl` if you want it to become `Cmd` on macOS.

ganesshkumar:
@lishid , what is the suggested option for the conflict with browser engine's paste as plaintext. Can it be Alt+Mod+V?

lishid:
Mod-Alt-V might work, or you can leave it blank and let the user assign something of their preference or only use it from the command palette.

There's also another option - if you want, you can rewrite the plugin to use the Workspace's `editor-paste` event, and try to detect whether the paste is from excel (possibly run a regex for some excel-specific `mso-` tags from the `text/html` blob). That way, your plugin will integrate directly into the paste handler for the editor and won't require an additional hotkey.

ganesshkumar:
Thanks for you input, @lishid . I guess I will add Mod-Alt-V and let the user reassign if needed. This is the same behavior found in excel to markdown plugin in VS Code as well.

ganesshkumar:
Thanks for you input, @li. I guess I will add Mod-Alt-V and let the user reassign if needed. This is the same behavior found in excel to markdown plugin in VS Code as well.

ganesshkumar:
@lishid , I have 
- updated the hotkeys to `Mod-Alt-V`
- added a note in the README.md on how to change the hotkeys.
- updated the version and made a release (0.1.1)

lishid:
I still think it's worth considering using the `editor-paste` event as it will be more integrated, but I think it's already pretty useful as it is right now.

ganesshkumar:
Definitely. Will try to integrate with `editor-paste` event and update the plugin soon 👍

PR #764, opened 2022-02-03
Add power search plugin

aviral-batra:

## Repo URL

Link to my plugin: https://github.com/aviral-batra/obsidian-power-search

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
Given that your plugin is designed to search Anki notes, I think it would make more sense to name it more appropriately to avoid disappointing users looking for ways to search for notes within Obsidian.

https://github.com/aviral-batra/obsidian-power-search/blob/6df53d1b8bd8ed33941ee9efae4f7b8cf4160f4b/src/main.ts#L18
Please see this tip: https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-managing-references-to-custom-views

https://github.com/aviral-batra/obsidian-power-search/blob/6df53d1b8bd8ed33941ee9efae4f7b8cf4160f4b/src/search.ts#L66
This may return null and cause an exception.

aviral-batra:
> Given that your plugin is designed to search Anki notes, I think it would make more sense to name it more appropriately to avoid disappointing users looking for ways to search for notes within Obsidian.

Yeah, however, the very next thing I am going to do is make a generalised inheritable/extendable index class to add a variety of types of indexes. Querying the obsidian vault is next on my list so I will include this functionality! I will probably do it tomorrow. I'll also write an api so that others can create their own custom indexes. Anki and Obsidian search will be the "core" searchable  indexes for now. Then each source will have its own toggle in the settings so that it can be added/removed to the content to be searched. 

> https://github.com/aviral-batra/obsidian-power-search/blob/6df53d1b8bd8ed33941ee9efae4f7b8cf4160f4b/src/main.ts#L18 Please see this tip: https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-managing-references-to-custom-views
> 
> https://github.com/aviral-batra/obsidian-power-search/blob/6df53d1b8bd8ed33941ee9efae4f7b8cf4160f4b/src/search.ts#L66 This may return null and cause an exception.

Whoops, overlooked these and will fix now.

aviral-batra:
@lishid I have now added this functionality!

lishid:
Sounds good!

PR #766, opened 2022-02-04
Add obsidian-relativenumber plugin

thisdotrob:

## Repo URL

[Link to my plugin](https://github.com/thisdotrob/obsidian-relativenumber-plugin)

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3348,5 +3348,12 @@
         "description": "Trigger shortcuts in Apple's Shortcuts app from Obsidian with custom commands.",
         "repo": "macstories/obsidian-shortcut-launcher",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-relativenumber",
+        "name": "Relativenumber (relative line numbers)",
+        "author": "Rob Stevenson",
+        "description": "Displays relative line numbers in the editor's gutter.",
+        "repo": "thisdotrob/obsidian-relativenumber-plugin"
     }
 ]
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/thisdotrob/obsidian-relativenumber-plugin/blob/9ff5d3ecf4652a17afcf6d68c783ba6adb69bb20/main.ts#L53
What's this for? Your function is already a shorthand `this`-bound function, it shouldn't require tweaks to unknown internal objects from the event emitter.

https://github.com/thisdotrob/obsidian-relativenumber-plugin/blob/9ff5d3ecf4652a17afcf6d68c783ba6adb69bb20/main.ts#L65
You are not disabling the legacy extension when your plugin is disabled (through the `onunload` function).

thisdotrob:
> https://github.com/thisdotrob/obsidian-relativenumber-plugin/blob/9ff5d3ecf4652a17afcf6d68c783ba6adb69bb20/main.ts#L53 What's this for? Your function is already a shorthand `this`-bound function, it shouldn't require tweaks to unknown internal objects from the event emitter.
You are correct it is not needed anymore. At one point when developing that bit of the code for [this PR against the original relative line numbers plugin](https://github.com/nadavspi/obsidian-relative-line-numbers/pull/4) I was getting undefined `this` inside the `config-changed` handler, but after moving the code about into it's current state it is no longer needed.

I've [removed](https://github.com/thisdotrob/obsidian-relativenumber-plugin/commit/07826ce7b05280c1e04a941fc8b32548daab0f17) this config listening code anyhow as it seems unnecessary and means the legacy and non-legacy modes are inconsistent in their treatment of the 'Show line numbers' main setting.

That commit also fixes the unloading.

PR #770, opened 2022-02-05
Add persistent-graph plugin

Sanqui:

## Repo URL

Link to my plugin: https://github.com/Sanqui/obsidian-persistent-graph

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
  - I'm accessing `Workspace.activeLeaf` directly because I don't know a better way to query for an active graph view.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3413,5 +3413,12 @@
         "author": "Alex Bieg",
         "description": "A command palette that does all of the things you want it to do.",
         "repo": "AlexBieg/obsidian-better-command-palette"
+    },
+    {
+        "id": "persistent-graph",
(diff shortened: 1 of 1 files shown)


Sanqui:
My plugin is using internal APIs but perhaps it's interesting enough to include, if not just close it :)

lishid:
Looks good.

https://github.com/Sanqui/obsidian-persistent-graph/blob/4abb998c17fd77c065daba72b378a4b351962b4d/main.ts#L18
Just make sure you are checking for null activeLeaf which can happen sometimes.

PR #772, opened 2022-02-07
Add auto-note-mover plugin

farux:

## Repo URL

Link to my plugin: https://github.com/farux/obsidian-auto-note-mover
## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3420,5 +3420,13 @@
         "author": "Rob Stevenson",
         "description": "Displays relative line numbers in the editor's gutter.",
         "repo": "thisdotrob/obsidian-relativenumber-plugin"
+    },
+    {
+        "id": "auto-note-mover",
+        "name": "Auto Note Mover",
+        "author": "faru",
+        "description": "Auto Note Mover will automatically move the active notes to their respective folders according to the rules.",
+        "repo": "farux/obsidian-auto-note-mover",
+        "branch": "main"
     }
 ]


pjeby:
This looks quite useful!

A few review comments, though:

- Don't use `vault.adapter` to check the existence of files: use `vault.getAbstractFileByPath()` and check if the result is a `TFolder` or `TFile`.   The Obsidian vault tracks the existence of all files and folders in the vault that aren't under a directory whose name begins with `.` (which Obsidian ignores, except for its own plugins and settings).

  (Basically, you should never use `vault.adapter` at all unless you are doing something special with the `.obsidian` directory or other "hidden" files/folders, which is something almost no plugins ever need to do.)

- `parseFrontMatterEntry` is not an async function, you don't have to `await` it -- which means `isFmDisable` doesn't need to be async, and in turn, does not need to be awaited by its caller, which in turn no longer needs to be async either.

- You should probably check in the rename event that only the file's basename has changed; otherwise if somebody moves their note away from its automatic placement, it will be immediately moved back!

- Your `fileCheck()` function makes 10 assignments to `this` properties, that should probably be replaced with `const` assignments.  For example, replace `this.regex = new RegExp(...)` with `const regex = new RegExp(...)` and remove the `this.` from future `this.regex` usage.  The way you are doing this right now is storing these values on the plugin even after the function is finished running.  Worse, if another run of this function began, it could end up reading the values from a different call to the function!  (It's unlikely, because of the debounce, but that just leads to another problem...)

- Your debounce design will work fine when someone is just editing or renaming a file, but if they move a large directory (or if Obsidian rebuilds its index), it will receive possibly hundreds of "rename" or "changed" events in a short time.  The debounce will lead to nearly all of these being ignored, because debounce doesn't keep track of the arguments and debounce them separately.  That is, if fileCheck() is called hundreds of times with different files, it will only process *one* of those files when it actually runs.  If you want it to actually process each file,  you need to instead have your event handlers add the file to a `Set` and call a debounced function that will run through and check each file in the set.

farux:
Hi, thank you @pjeby for the review.

- I stopped using vault.adapter and changed it to vault.getAbstractFileByPath().

- Removed await and async from isFmDisable and caller.

- Changed "this" in fileCheck() to const.

This is the first time I have used `Set`, but I think it probably worked.
The original plan for this plugin was to move only the note currently being edited.
My concern at the time was that it would be called all the time while I was typing the text and would keep looping, which would slow down the performance.
So I did some research and found out about `debounce`.
Anyway, now it can move multiple files.

I'm a beginner and I'm still learning.
I've looked at the api but can't figure out how to do this.

> check in the rename event that only the file's basename has changed

If you have a concrete code example, it might help me understand.
I assumed that if the user did not want to move the note temporarily, they would switch the trigger to manual.
For now, I have added a command to switch the trigger.

pjeby:
> My concern at the time was that it would be called all the time while I was typing the text and would keep looping, which would slow down the performance.

The `changed` event on the metadata cache only occurs after a save, and saves are about 2 seconds apart, so your debouncing would've needed to be longer than 2 seconds to have any effect.  You were waiting for 1 second of silence, which will happen once every two seconds while the user is typing.

> If you have a concrete code example, it might help me understand.  I assumed that if the user did not want to move the note temporarily, they would switch the trigger to manual.  For now, I have added a command to switch the trigger.

What I mean is that if you rename a folder, or move a file to another folder, it will fire a rename event.  And if a folder is renamed or moved, the rename event will fire for *every file that was in the original folder or a sub-folder thereof*.  But these moves and renames are probably things your plugin doesn't care about.

The `rename` event actually provides both a file object as the first parameter, and the file's old `path` before renaming as the second parameter.  So if you compared `oldPath.split("/").pop()` to the file's current `basename + extension`, you could skip processing the file if they were the same, because the basename wasn't changed and so your rules would not be moving it someplace different than where it was last moved.  (Unless the rules had been changed, of course.)

If you did this, it would also fix the problem of getting lots of events for different files, and you could possibly drop the debouncing part.  The only other event that might happen frequently is the `change` event, in the case where a lot of files were changed outside Obsidian or it has to rebuild its index for some reason.  But debouncing those events wouldn't really resolve much.

So if you wanted to, you could just filter the rename events to make sure the filename is changing, and let Obsidian's save-debouncing keep you from pinging the same file more than once every 2 seconds.  Processing change events when Obsidian is reindexing might slow down the reindexing, but it's not a common operation by any means -- usually it happens if a new release of Obsidian changes the index format, or if your index files get corrupted, or you copy your vault to another machine, etc.

> This is the first time I have used Set, but I think it probably worked.

Not quite.  What I meant was for you to have a property on `this`, where `registerCheck()` would *add* items to the set, but the debounced function `fileCheck` would loop to *remove* them, so that even if `registerCheck()` was called many times before `fileCheck` got to run, `fileCheck` would have a list of all the files that `registerCheck` was called on.

That way, even if every file was being `changed` by an index rebuild, they would all be added to the set, then when the debounced fileCheck function was called, it would see all of them in the set to process.  But if you take the approach of *not* debouncing and just letting Obsidian save handle things (and avoid calling the function for rename events that don't change the file's basename), you can drop the Set thing (as well as the debounce) and just perform the operation right from the event.

All in all, I think it would be a very good idea to not process rename events that don't change the basename, or else someone could inadvertently trigger a resort of a whole lot of files just by renaming the folder that contains them.

farux:
Hi. I added a filter for basename to the beginning of fileCheck.

After that, I took a debounce of fileCheck and got the error only during obsidian boot.

I got an underfind error with `isFmDisable(frontmatter)` and `cacheTag.find()`, so I added a boolean check.
The error disappeared, but I'm not sure what's going on.

I added `app.vault.on('modify'` because `app.metadataCache.on('changed'` couldn't catch the event when the tag wrangler plugin changed the tags of multiple files at the same time, but both were triggered almost simultaneously while editing the text, so I used `window.setTimeout` to delay the trigger of `app.vault.on('modify'` for 1 second. Is there a smarter way to do this?

pjeby:
> I got an underfind error with isFmDisable(frontmatter) and cacheTag.find(), so I added a boolean check.
The error disappeared, but I'm not sure what's going on.

If the metadata cache didn't have a fileCache entry for the file, you would get an error when you tried to access its `.frontmatter`.  I'm not sure why this would happen at startup, precisely; perhaps Licat can chime in on that point.

> Is there a smarter way to do this?

Don't track the `modify` event, only `changed`.  The `changed` event is the one that matters, because that tells you when the fileCache has been updated to match the file's contents.  I don't know what you mean by "couldn't catch the event" when tag wrangler changed multiple files: that should only have happened if you were debouncing and didn't have a proper queue set up.

At this point, I would suggest replacing lines 51-85 with:

```ts
                this.registerEvent(
			this.app.vault.on('create', fileCheck)
		);
		this.registerEvent(
			this.app.metadataCache.on('changed', fileCheck)
		);
		this.registerEvent(
			this.app.vault.on('rename', fileCheck)
		);
```

and moving the "Automatic" trigger check to inside fileCheck.  That is, get rid of fileSet, registerCheck, and the `modify` event hook.  They are all unnecessary.  The above *should* work correctly even with Tag Wrangler: i.e. if you rename a tag to match a move rule, it should move all the files with the renamed tag.

If you want to verify it is working, you can put a `debugger` line in your fileCheck and open the developer console before doing a tag rename or any other sort of change.  Obsidian will pause and let you use the debugger to look at variable contents or step through the actions it's doing so you can see how it's working.  This will work even at startup if you leave the console open and restart Obsidian.  (You should remove the `debugger` statement when you're done testing, of course.)

farux:
> moving the "Automatic" trigger check to inside fileCheck. That is, get rid of fileSet, registerCheck, and the modify event hook. 

I think I got it right.
I couldn't figure out how to identify that the caller is a command in the Automatic trigger check.
For now, I threw in a variable to identify it.

I used the debugger and found that when loading the vault during the obsidian boot, all folders and files in the vault were assigned to 'file' one after another, and when a file was assigned, the frontmatter undefined error appeared.


Is this because the fileCheck is in onload()?　Should I put it outside?

pjeby:
So, you don't need the fileSet thing or the loop any more: those were only needed for debouncing.

> Is this because the fileCheck is in onload()?　Should I put it outside?

No.  It's because apparently you're getting `changed` events during vault startup, but for some reason the fileCache is empty at that point.  Which is weird, but this should fix it...  change the registerEvent lines to:

```ts
                this.app.workspace.onLayoutReady(()=> {
                    this.registerEvent(this.app.vault.on('create', file => fileCheck(file)));
		    this.registerEvent(this.app.metadataCache.on('changed', file => fileCheck(file)));
		    this.registerEvent(this.app.vault.on('rename', (file, oldPath) => fileCheck(file, oldPath));
                });
```

This will delay listening to any events until after the vault is loaded, and also prevent any extraneous values being passed in to fileCheck now that you have other parameters.  (Some of the events (like `changed`) will pass other parameters besides the file.)

I would suggest at this point that since you have a manual/automatic switch, you don't actually need an enable/disable switch as well; in manual mode people can just not use the command if they don't want to move anything, or disable the plugin altogether if they aren't using it.

I think this is probably the last feedback I have for you; that doesn't mean Lishid won't have any for you as well, but things are looking pretty good now.  I specifically volunteered to help review this because I'm thinking I'll want to use it myself. :grin:

farux:
> This will delay listening to any events until after the vault is loaded,

It worked! No more errors.
If the callback is an arrow function, the type of the `file` is `TAbstractFile` and it cannot be passed to `fileCheck`, so I did this.

```
			this.registerEvent(this.app.vault.on('create', fileCheck));
			this.registerEvent(this.app.metadataCache.on('changed', fileCheck));
			this.registerEvent(this.app.vault.on('rename', fileCheck));
		});

```

> you don't actually need an enable/disable switch

It makes sense. I removed the switch.

I also wanted to add an Auto/Manual trigger  indicator to the status bar. `onload()` and the `command` will run the indicator function, but I can't find a way to run the indicator function when the trigger is changed in the `SettingTab`, so I'm holding off.

> I specifically volunteered to help review this because I'm thinking I'll want to use it myself.

You have been very kind to share this with me and it has been very helpful.
I can't complete it without you.
Thank you so much for your help.

pjeby:
> If the callback is an arrow function, the type of the file is TAbstractFile and it cannot be passed to fileCheck

Make fileCheck take a TAbstactFile, and in the body, do a `if ( !(file instanceof TFile) ) return;`  In general, when you have a TAbstractFile, you can get TypeScript to recognize it as a TFile or TFolder by doing an instanceof check.  TypeScript will then implicitly cast it to the correct type.

farux:
Oh, I see.
It's done.

pjeby:
Great!  Don't forget to remove the fileSet and the loop, though: they're not doing anything but the rename check, which can be done without the loop or the Set.  The fileSet and loop were only ever needed for debouncing, and there's no debouncing any more.

farux:
Oops! I fixed it.

PR #774, opened 2022-02-07
Add Social Plugin

ckemmler:

## Repo URL

Link to my plugin: https://github.com/Sekund/sekund-plugin-react

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/Sekund/sekund-plugin-react/blob/65453bb2c4839503d96af68add6873495101be2a/src/main.tsx#L221
Here you should use `workspace.getLeaf()` to get an unpinned leaf to open your note. This function gives you a leaf while respecting the pinned status of the pane.

https://github.com/Sekund/sekund-plugin-react/blob/65453bb2c4839503d96af68add6873495101be2a/src/main.tsx#L250
These events aren't tracked for removal when your plugin is disabled. You can use `Plugin.registerDomEvent` to track them for auto removal on plugin disable.

https://github.com/Sekund/sekund-plugin-react/blob/65453bb2c4839503d96af68add6873495101be2a/src/main.tsx#L359
I'm assuming you want to deal with markdown files here - better to use `workspace.getActiveViewOfType(MarkdownView)`. You can also use `view.getState()` instead of `leaf.getViewState()`.

https://github.com/Sekund/sekund-plugin-react/blob/65453bb2c4839503d96af68add6873495101be2a/src/main.tsx#L129
Here you should use `leaf.view.getMode() === 'preview'` which is less expensive than serializing the entire view state just to get the current mode.

ckemmler:
Hey @lishid thank you for your review. Those issues are now fixed.

lishid:
https://github.com/Sekund/sekund-plugin-react/commit/c17d6f882cd2f7619a13e07bf6431c1395770371#diff-1cd8b18798a1a103bfe13bef54354c1f3a3bea29a31c8eea1a0c67a3a839b811R129
You should use `leaf.view instanceof MarkdownView && leaf.view.getMode() === 'Preview'` instead. Always do an `instanceof` check instead of using `as any`.

https://github.com/Sekund/sekund-plugin-react/commit/c17d6f882cd2f7619a13e07bf6431c1395770371#diff-1cd8b18798a1a103bfe13bef54354c1f3a3bea29a31c8eea1a0c67a3a839b811R221
Here you might create new leaves twice. `getLeaf()` also gives you a leaf guaranteed, so you should always get a leaf back.

https://github.com/Sekund/sekund-plugin-react/commit/c17d6f882cd2f7619a13e07bf6431c1395770371#diff-1cd8b18798a1a103bfe13bef54354c1f3a3bea29a31c8eea1a0c67a3a839b811R363
I don't think setting the mode here like that works. Actually I think my bad - you're probably better off using getViewState and setViewState on `view.leaf`.

ckemmler:
That all makes sense. Thank you @lishid. Done

lishid:
Looks good. I need to fix your PR to move your plugin to the last row of the json file as well as follow the spacing convention before merging, however I don't have commit right to your PR branch - can you fix it yourself or give me permission in the PR? Thanks.

ckemmler:
I just added you as a collaborator on my fork. Let me know if that works for you.

PR #775, opened 2022-02-07
Add a custom release branch for plugin

xldenis:
I keep forgetting that merging to master breaks plugin installs, so I'm setting up a custom `releases` branch from which obsidian can pull to install.


lishid:
Actually, in newer versions of Obsidian the branch tag is no longer used and github's default branch (`HEAD`) is now used.

For a workflow that does not disrupt plugin installs, I think the most recommended approach is to make a `dev` branch which is what you work on (and merge to), and only merge to master/main when doing a release.

xldenis:
Ah, that's too bad but no big issue

PR #780, opened 2022-02-10
Add Sidekick

hadynz:

## Repo URL
Link to my plugin: https://github.com/hadynz/obsidian-sidekick

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3458,10 +3458,10 @@
         "repo": "aviral-batra/obsidian-power-search"
     },
     {
-        "id": "obsidian-kindle-plugin",
-        "name": "Autosuggest",
-        "description": "Autosuggest new connections by highlighting text that could match existing links and tags",
+        "id": "obsidian-sidekick",
+        "name": "Sidekick",
+        "description": "A companion to identify hidden connections that match your tags and pages",
         "author": "Hady Osman",
-        "repo": "hadynz/obsidian-autosuggest"
+        "repo": "hadynz/obsidian-sidekick"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3456,5 +3456,12 @@
         "author": "Aviral Batra",
         "description": "Searches notes based on current line",
         "repo": "aviral-batra/obsidian-power-search"
+    },
+    {
+        "id": "obsidian-kindle-plugin",
+        "name": "Autosuggest",
+        "description": "Autosuggest new connections by highlighting text that could match existing links and tags",
+        "author": "Hady Osman",
+        "repo": "hadynz/obsidian-autosuggest"
     }
 ]


community-plugins.json
```diff
@@ -3456,5 +3456,12 @@
         "author": "Aviral Batra",
         "description": "Searches notes based on current line",
         "repo": "aviral-batra/obsidian-power-search"
+    },
+    {
+        "id": "obsidian-kindle-plugin",
+        "name": "Autosuggest",
```
liamcain, line 3462: At least for me, the word "autosuggest" usually refers to the typeahead component that you see here:

image

A more descriptive name might help avoid confusion and help with discovery. Something like `Highlight Connections` or `Highlight unlinked mentions` (that one's pretty boring sounding, I know)

liamcain: At least in the React.js world, [autosuggest](https://react-autosuggest.js.org/) is pretty common, plus the Obsidian API calls them "Suggests."

Regardless, I like the new name and the plugin looks awesome!

hadynz:
Hey @pjeby. I've seen you do some amazing code reviews for newly submitted plugins. Would love it if you have sometime to have a look at what I have in this plugin and share with me any tips.

The plugin is very much in alpha/beta, but I am sharing it now to start curating feature requests from users. So it is still in active development.

pjeby:
So, my first comment is that I don't think Sidekick is a helpful name at all, as it says nothing about what the plugin does.  To be fair, my ideas for an alternative are not at all better.  About the cleverest ideas I have are Link Spotter or Missing Links (because it shows you the links you are missing).  :wink:

A good rule of thumb for names is that if somebody describes the problem they're having, and another person says, "Oh, you need the X plugin", hearing X should make the first person go, "Ah!  Of course, that's just what I need".  So in this case, "Sidekick" makes the first person go, "I need the what now?"  Whereas if the problem is that "It's such a pain linking and tagging my notes", the answer "Link Spotter" or "Missing Links" fits the, "Ah, yes, something to spot the links for me" or "Ah, yes, I'm definitely missing links."  I don't think either of these is a *great* name, but you can at least see what I'm getting at as a rule for naming.  My own plugins are not that great at this necessarily, especially since some have grown away from their original purpose.  But it's always something to aim for.

As for the code itself, the first problem I see is that it will likely perform very poorly on large vaults by causing memory thrashing.  Specifically, it is creating page objects for every note every time a file changes in a vault.  In a vault with thousands of notes, this will create and destroy thousands of objects every couple of seconds while someone is typing.  Rather than recreate these objects each time, you could keep a WeakMap from files to their PageIndex (for example), so that at least you wouldn't be creating all those objects every time.

(Beyond that, simply iterating all the files in the vault will be a noticeable performance hit in larger vaults, and there isn't any reason to do that every time a file is modified, vs. only updating the alias and tag info for the specific file modified.)

But also, the `modify` event is not what you want to monitor.  `modify` means the file has changed on disk but does not mean it has been updated in the metadata cache.  What you actually want is to monitor the metadata cache's `changed` event which will not only tell you what file has changed, but also give you its metadata and text.  You can then use that to update your index for the specific file.  (You will also want to monitor vault `rename` events that change a file's basename, as these affect your index and you are currently not tracking that as far as I could tell in a cursory sweep.)

The next place where performance will be bad on large vaults is the actual highlighting process, which is doing a linear search through the massive amount of text it compiles...  after first doing another linear search through it to cut down references already present in the current file.

Ironically, you could perhaps do away with all your indexes and just directly search the metadata cache and have better performance, since that would only be one loop over the data, and not create and destroy so many objects in order to loop over them multiple times.  It would also always be up to date and not need to listen to any events, since your searching would always be on the current cache state.  (Not that I'm recommending that approach, mind you...  but it might still be better performance-wise.)

I am not really sure what to suggest here, though, because changing the architecture of how this works is obviously a big ask, yet I don't see how this could perform well on even a vault with only a few thousand notes while someone is actively typing.  My gut feeling is that it needs a smarter sort of index updating, and perhaps a btree or some other structure to index the text lookups so that word and phrase lookups will be fast, rather than just searching linearly.  And it would definitely need to not rebuild the entire index on every modification to the vault.

Also, at this point I have not reviewed the entire code base -- these problems were enough to make me stop because if they aren't fixed, people with larger vaults won't really be able to use the plugin anyway, unless they have lots of CPU and RAM to burn.  I would suggest doing some testing with a larger vault, like at least 5000 notes with random names to see what I mean.  I haven't tried it on mine (which is about 2500), but the Dataview plugin had some similar issues with excessive updating and it would make typing lag like crazy when it was looping over the entire vault every couple of seconds the way your plugin is doing.  There are some differences in what was being done so I don't know if yours would be as bad at my vault size, but it's really more a question of how large a vault creates how much of a lag rather than "will it be a problem at all".  (It may also take several more minutes of typing before you'd hit a lag, as the data structures you're thrashing are probably a lot smaller than the ones Dataview was.)

Finally, on the features front, I would mention that simply matching note names isn't something that'll work for everyone: lots of people use Zettelkasten prefixes or dates in note names, which means they won't ever be suggested by the plugin unless they added aliases for them.  But doing something about that is also obviously a lot more complex, unless it amounts to "strip things that look like Zettelkasten prefixes and skip notes whose names match the daily note pattern".

All in all, I'm intrigued by the plugin's promise, and wish I could use it myself, but I'm also fairly sure I would not be able to even without the potential performance issues, due to my heavy use of zettelkasten prefixes and lack of pre-existing aliases for plenty of files.  That being said, it's not necessarily an obstacle to publishing in the plugin store, though I'd suggest making note of the size limits in your README in that case, after first testing with some larger vaults and actually typing some large notes in said vaults for a half hour or so.  If you can find real users with *really* large vaults (10-30k) to play with this and give you feedback, even better.

lishid:
https://github.com/hadynz/obsidian-sidekick/blob/4f61555a7268d8806aeb35e23a869311102fae44/src/index.ts#L23

See this new tip for a better way to manage cm6 extensions: https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#how-to-changereconfigure-your-cm6-extensions

https://github.com/hadynz/obsidian-sidekick/blob/4f61555a7268d8806aeb35e23a869311102fae44/src/indexing/indexer.ts#L13
You are leaking this event registration which is not un-registered when your plugin is disabled. I would suggest moving this code to your plugin main class instead and making your indexAll function public. You should also debounce this because there are cases where `modify` is called many times within a second (for example, when updating links in a dozen of files as a result of a file rename).

https://github.com/hadynz/obsidian-sidekick/blob/4f61555a7268d8806aeb35e23a869311102fae44/src/indexing/indexer.ts#L25
Your indexing function is fairly computationally heavy; I would suggest storing index results by file's mtime and size, and only re-compute for a file when its mtime/size has been changed from last time.

https://github.com/hadynz/obsidian-sidekick/blob/4f61555a7268d8806aeb35e23a869311102fae44/src/search/index.ts#L16
Would recommend limiting the number of results to avoid your search function computing and displaying thousands of results needlessly.

hadynz:
@pjeby Thank you so much for your time to write-up this very thorough review. Without going too much in-depth in the code, it was still very, very valuable.

I must admit, I am yet to optimise the plugin for Performance at this stage. Instead I was focusing on getting basic MVP functionality out for now. But appreciate as an owner of large vaults, your focus on non-functionals right off the cuff. 

The low hanging fruit takeaways that I got from you are:

* Re-think indexing. Probably as you suggested avoid trying to build the entire index, and be smarter about only updating indices for files that have been modified

* Spot on about me listening to the wrong events! Thanks for that... you are correct, I am really only interested in updates to a file's cache and a file's name

* Highlight portions of the editor view that are visible instead of the entire document (like [some plugins do][1])

> Ironically, you could perhaps do away with all your indexes and just directly search the metadata cache and have better performance, since that would only be one loop over the data, and not create and destroy so many objects in order to loop over them multiple times. It would also always be up to date and not need to listen to any events, since your searching would always be on the current cache state. (Not that I'm recommending that approach, mind you... but it might still be better performance-wise.)

This is not a bad idea at all, at least for the purpose of initial MVP. The need to async index right now is actually limited.

I am doing some research now around using NLP methods that increases the intelligence of the match making. At a very basic level, at least using [stemming][1] to suggest matches to the user. Once I lock in an approach, it will largely determine and guide the productionised method of indexing (if any) that I would be using. 

I'll look to your idea of simplifying things and removing indexing for now and see how it performs on large vaults to get a feel.

> lots of people use Zettelkasten prefixes or dates in note names, which means they won't ever be suggested by the plugin unless they added aliases for them

Good call. It is possible that this plugin would not be most suitable for users with Zettelkasten workflows. That said, I'll try and look around for some published Zettelkasten vaults and see if there is a way to attempt to support them by looking at `#h1` headers for examples.

Thanks a tonne again!

[1]: https://github.com/nothingislost/obsidian-dynamic-highlights/blob/ad10ca240977578e25f3061bdb8a5141c91011d5/src/highlighters/selection.ts#L117
[2]: https://en.wikipedia.org/wiki/Stemming

Krammy:
I would call this _Link Suggester_ rather than Sidekick.

Sidekick gives no indication of what the plug-in is for.

lishid:
Any updates on this PR? Do you still intend to add it to the plugin repo?

hadynz:
There's been a lot of work on the plugin since the last review. But I am yet to update this branch.

I am going for three weeks on annual leave so will continue active development of the plugin which has ceased for a couple of months now.

I'm happy for you to close this if you'd prefer. I aim to have an update though on this in 3 weeks.

lishid:
Ok we can leave this open since you're still working on it.

ericaxu:
@hadynz Following up - do you still intend to add it to the plugin store?

hadynz:
Hi @ericaxu. I am happy for this PR to be closed. I am not actively working on this plugin at the moment (but will aim to come back to it one day).

PR #782, opened 2022-02-11
Add remember-file-state plugin

ludovicchabant:

## Repo URL

Link to my plugin: https://github.com/ludovicchabant/obsidian-remember-file-state

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/ludovicchabant/obsidian-remember-file-state/blob/1fa0ec435b289302d3248ede4085e3b46b96dae3/src/main.ts#L61

Here you will keep a reference to this view forever, even if the view has been detached/deleted from the workspace. Not sure if there's a better way to do this that would avoid this problem though.

https://github.com/ludovicchabant/obsidian-remember-file-state/blob/1fa0ec435b289302d3248ede4085e3b46b96dae3/src/main.ts#L133
Missing null check for this view.

https://github.com/ludovicchabant/obsidian-remember-file-state/blob/1fa0ec435b289302d3248ede4085e3b46b96dae3/src/main.ts#L33
Probably better to keep an object keyed by the file path, which would make lookups O(1).

ludovicchabant:
Hi! Thanks for the review and advice -- this is my first foray into Obsidian development so any help is welcome.

- I think I found a way to fix the first issue by generating unique IDs for views, so I don't need to associate the unhook callbacks to the actual pointers. See [this commit][1] (which also fixed another dangling reference bug).
- Second issue is fixed (see [this commit][2]).
- Third issue is also fixed (see [this commit][3]).

Thanks again!

[1]: https://github.com/ludovicchabant/obsidian-remember-file-state/commit/54e9fdb23d3be6a323ec722b1f7a235dd113e1ac
[2]: https://github.com/ludovicchabant/obsidian-remember-file-state/commit/5a3a7a20d0ca4a8a728f29c6e29cd631d476dccf
[3]: https://github.com/ludovicchabant/obsidian-remember-file-state/commit/f31243944d80718d4369bff8d0a1c8334a639b34

lishid:
https://github.com/ludovicchabant/obsidian-remember-file-state/commit/f31243944d80718d4369bff8d0a1c8334a639b34#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR33
You can type this as `Record<string, RememberedFileState>` which would help with type checks later on.

https://github.com/ludovicchabant/obsidian-remember-file-state/commit/54e9fdb23d3be6a323ec722b1f7a235dd113e1ac#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR149
I think you misunderstood what I meant - registered callbacks will be dereferenced when your plugin unloads. I went back to re-read your old code and noticed that you are deleting all references when the view is unloaded, which should actually properly dereference everything. I don't think the change to using IDs do anything (your `_viewUninstallers` object still hold a reference to the `uninstall` function which holds a reference to the view itself. Instead, I suggest using a `Map<View, uninstaller>` as you can't use object keys in plain objects.

ludovicchabant:
> You can type this as `Record<string, RememberedFileState>` which would help with type checks later on.

Thanks again for the review. [This is fixed][1].

[1]: https://github.com/ludovicchabant/obsidian-remember-file-state/commit/bba891d726558baef247427ec08715cc8d1b213f

> I don't think the change to using IDs do anything (your `_viewUninstallers` object still hold a reference to the `uninstall` function which holds a reference to the view itself.

Nicely spotted. In fact, there's no need to capture the view in these callbacks, since by definition the view is the `this` object at this point, so I there's [a simple fix for it][2], I think. The view IDs don't help per se, but I think they make it cleaner by ensuring that if there's a bug, there's less chance of leaking anything.

[2]: https://github.com/ludovicchabant/obsidian-remember-file-state/commit/d56257b72d95d951e7559cb92ff06a8c570fdd41

PR #786, opened 2022-02-13
Add Path Title plugin

jdeal:

## Repo URL

Link to my plugin: https://github.com/jdeal/obsidian-path-title-plugin

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/jdeal/obsidian-path-title-plugin/blob/948fc84216ae1b016daeae44b893dafaa3c0ba3d/src/plugin.ts#L53
The rename event can sometimes be called many times consecutively (when renaming a folder with lots of subfolders) - you should debounce the `setPaneTitles` function.

https://github.com/jdeal/obsidian-path-title-plugin/blob/948fc84216ae1b016daeae44b893dafaa3c0ba3d/src/plugin.ts#L144
This should be reused - put it in a variable and reference it again.

I also suggest you detaching the element instead of just setting its display to none.

https://github.com/jdeal/obsidian-path-title-plugin/blob/948fc84216ae1b016daeae44b893dafaa3c0ba3d/src/plugin.ts#L166
Here you are not removing the elements you added (`.path-title-plugin-path-title-container`).

https://github.com/jdeal/obsidian-path-title-plugin/blob/948fc84216ae1b016daeae44b893dafaa3c0ba3d/src/plugin.ts#L61
I feel like the execution complexity here doesn't warrant the extra effort for memoizing path replacements - if you update paths on a reasonable debounce I recommend removing the memoization and just call the function for every file view.

jdeal:
Hey @lishid, thanks for the suggestions! Sorry for the slow reply, but I made all those changes, which you can see in a [comparison here](https://github.com/jdeal/obsidian-path-title-plugin/compare/1.0.1...1.0.4). Let me know if there's anything else I should change.

lishid:
https://github.com/jdeal/obsidian-path-title-plugin/blob/7be09b6d58de956bf02dc847521fbc78f953d08e/src/plugin.ts#L158
Possible crash here if find returns null.

jdeal:
Thanks for another catch @lishid! [Fixed in a subsequent build](https://github.com/jdeal/obsidian-path-title-plugin/compare/1.0.4...1.0.6), along with another potential null I found. And thanks for merging anyway!

PR #790, opened 2022-02-18
Add obsidian-advanced-selection plugin

anselmwang:

## Repo URL

Link to my plugin: https://github.com/anselmwang/obsidian-advanced-selection

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [ ] `main.js`
  - [ ] `manifest.json`
- [ ] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3507,5 +3507,12 @@
         "description": "A patch for Obsidian's built-in CodeMirror Editor to support Japanese word splitting",
         "repo": "sonarAIT/cm-japanese-patch",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-advanced-selection",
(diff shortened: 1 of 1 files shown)


lishid:
The plugin code looks fine but I think calling it "Advanced selection" may cause users to over-expect from the plugin - I would recommend naming it something like "Mark and select"

lishid:
Looks good, sorry for the late reply - do you want to rename the github repo as well before we commit this in?

anselmwang:
I am new to GitHub and am not aware that a repository can be renamed. This is so handy :) 

Just update, thanks!

PR #796, opened 2022-02-21
Add digital garden plugin to plugin list

oleeskild:

## Repo URL

Link to my plugin: https://github.com/oleeskild/obsidian-digital-garden

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/oleeskild/obsidian-digital-garden/blob/1f4ae2b5c2f6b12ef9427939d4884a7c579ede7a/main.ts#L48
This may return null - you should probably use an `editorCallback` instead which gives you the editor, the markdownView using which you can get the `markdownView.file`.

https://github.com/oleeskild/obsidian-digital-garden/blob/1f4ae2b5c2f6b12ef9427939d4884a7c579ede7a/main.ts#L151
This may return null, and you should pass in the sourcePath here instead of trying to get it from the active file, which may have changed since the initial call of your function as your function is async.

https://github.com/oleeskild/obsidian-digital-garden/blob/1f4ae2b5c2f6b12ef9427939d4884a7c579ede7a/main.ts#L154
Images may not always be PNGs. You should check the file extension first.

https://github.com/oleeskild/obsidian-digital-garden/blob/1f4ae2b5c2f6b12ef9427939d4884a7c579ede7a/main.ts#L125
Same comments as above. You should also use `cachedRead` here for better performance.

Also as a note, given that you're processing images first, transclusions after, your images in transcluded files won't be rendered.

oleeskild:
Thanks for a thorough review!

I've pushed fixes for all the comments you've made except for the first one.
Correct me if I'm wrong, but wouldn't using editorCallback hide the command when in Preview Mode? I think that would be confusing to users. I've instead added a null check, which shows a Notice to the user if no file is active: [https://github.com/oleeskild/obsidian-digital-garden/blob/1c898e98a155cd29868b8eede37882b59d56c4da/main.ts#L48](https://github.com/oleeskild/obsidian-digital-garden/blob/1c898e98a155cd29868b8eede37882b59d56c4da/main.ts#L48)

What do you think? Is this an acceptable solution?

lishid:
Sorry for the late reply! For 1 yes, your are right and you should keep this - the only thing you want to make sure is to test whether the current file is actually a markdown file `(file.extension === 'md')` so you don't accidentally attempt to publish a different file type and parse it as markdown.

PR #797, opened 2022-02-22
Update community-plugins.json

geoffreysflaminglasersword:

## Repo URL

Link to my plugin: https://github.com/geoffreysflaminglasersword/obsidian-living-graph

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/geoffreysflaminglasersword/obsidian-living-graph/blob/b9547dbcc13c9616bd8ae9dc670e81b3470c2f05/main.ts#L364
Suggest removing console logs when you distribute the plugin to avoid clogging up the console.

https://github.com/geoffreysflaminglasersword/obsidian-living-graph/blob/b9547dbcc13c9616bd8ae9dc670e81b3470c2f05/main.ts#L371
This shouldn't be necessary since you're already using onLayoutReady.

https://github.com/geoffreysflaminglasersword/obsidian-living-graph/blob/b9547dbcc13c9616bd8ae9dc670e81b3470c2f05/main.ts#L385
Don't save settings when your plugin unloads, instead, save it when the setting is changed (which you already do anyway).

https://github.com/geoffreysflaminglasersword/obsidian-living-graph/blob/b9547dbcc13c9616bd8ae9dc670e81b3470c2f05/main.ts#L399
I think there's a chance you leak memory here - you should clear your intervals regardless since you're about to dereference the array.

geoffreysflaminglasersword:
@lishid good looking out, my bad. Everything should be good, though I do actually need the layout change listener as people open new graphs

lishid:
Oh yeah totally, I just meant this specific line that's calling `refreshLeaves` in the `onload` function that isn't necessary because you're calling it again in onLayoutReady:

https://github.com/geoffreysflaminglasersword/obsidian-living-graph/blob/c09cdb789c9d3f8b7b680779d33974a193c39c7b/main.ts#L369

PR #798, opened 2022-02-22
Add Hotkeys Chords Plugin

trenta3:

## Repo URL

Link to my plugin: https://github.com/trenta3/obsidian-hotkeys-chords

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [X]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


lishid:
General coding tips:
- Please auto-format your file to normalize the use of spaces and tabs.
- Please use `let` or `const` instead of `var.

https://github.com/trenta3/obsidian-hotkeys-chords/blob/7709ba53ddd119f4f4f338b1651562d8fcf01a6d/main.ts#L77
You can't use class objects in settings, since they are serialized to and from JSON which loses the object. Should probably find another way to have your object format reconstructed.

https://github.com/trenta3/obsidian-hotkeys-chords/blob/7709ba53ddd119f4f4f338b1651562d8fcf01a6d/main.ts#L106
I don't think you should register a cm6 plugin, instead you want it to work globally. In this case I think you should use `window.addEventListener('keydown', callback, {capture: true})` and remember to unregister it when your plugin unloads.

https://github.com/trenta3/obsidian-hotkeys-chords/blob/7709ba53ddd119f4f4f338b1651562d8fcf01a6d/main.ts#L144
Won't this capture every single keystroke?

https://github.com/trenta3/obsidian-hotkeys-chords/blob/7709ba53ddd119f4f4f338b1651562d8fcf01a6d/main.ts#L315
I don't think you need a class here, just use a regular function?

https://github.com/trenta3/obsidian-hotkeys-chords/blob/7709ba53ddd119f4f4f338b1651562d8fcf01a6d/main.ts#L341
Move this line above so that in case the callback function errors out you don't get stuck with the listener.
You should also add a global escape hatch in case something happens (like your dialog got accidentally closed) - for example, the `ESC` key might be a good way to break out and unhook this event handler.

trenta3:
Hi! Thank you for taking the time to properly review the PR. I will send an updated version of it in a couple of days, meanwhile let me address some points:

> https://github.com/trenta3/obsidian-hotkeys-chords/blob/7709ba53ddd119f4f4f338b1651562d8fcf01a6d/main.ts#L77 You can't use class objects in settings, since they are serialized to and from JSON which loses the object. Should probably find another way to have your object format reconstructed.

I'm reconstructing the objects when loading the settings in https://github.com/trenta3/obsidian-hotkeys-chords/blob/7709ba53ddd119f4f4f338b1651562d8fcf01a6d/main.ts#L94-L100 so that this causes no problem. I properly tested the plugin and this works correctly.
 
> https://github.com/trenta3/obsidian-hotkeys-chords/blob/7709ba53ddd119f4f4f338b1651562d8fcf01a6d/main.ts#L106 I don't think you should register a cm6 plugin, instead you want it to work globally. In this case I think you should use `window.addEventListener('keydown', callback, {capture: true})` and remember to unregister it when your plugin unloads.

Thank you for the suggestion, I will check if it works, even though I remember trying something similar and having to fight with it not working.

> https://github.com/trenta3/obsidian-hotkeys-chords/blob/7709ba53ddd119f4f4f338b1651562d8fcf01a6d/main.ts#L144 Won't this capture every single keystroke?
Yes it does, but if at least a partial match isn't found the variable is restored to an empty array in https://github.com/trenta3/obsidian-hotkeys-chords/blob/7709ba53ddd119f4f4f338b1651562d8fcf01a6d/main.ts#L165-L168
Capturing every keystroke is essential for the plugin to work since we need to match sequences of them, and this was a way to simplify the logic of checking for partial matches. I have to say that in my experience this produces no lags whatsover since checking for partial matches happens really fast.

> https://github.com/trenta3/obsidian-hotkeys-chords/blob/7709ba53ddd119f4f4f338b1651562d8fcf01a6d/main.ts#L315 I don't think you need a class here, just use a regular function?
You are definitely right on this one. I didn't notice it.

> https://github.com/trenta3/obsidian-hotkeys-chords/blob/7709ba53ddd119f4f4f338b1651562d8fcf01a6d/main.ts#L341 Move this line above so that in case the callback function errors out you don't get stuck with the listener. You should also add a global escape hatch in case something happens (like your dialog got accidentally closed) - for example, the `ESC` key might be a good way to break out and unhook this event handler.
You are right for this one. Concerning the second part, I already took care of the Escape key in https://github.com/trenta3/obsidian-hotkeys-chords/blob/7709ba53ddd119f4f4f338b1651562d8fcf01a6d/main.ts#L323-L326

As I said, I will be rebasing the PR in a couple of days.

trenta3:
I implemented the suggested changes and updated the artifacts in the release. I tested the plugin and it still works well locally.

lishid:
Thanks, I must have missed all the places you were already addressing my concerns 🤣 .

PR #801, opened 2022-02-25
Add obsidian-kobo-highlights-importer-plugin

bitwiseops:

## Repo URL

Link to my plugin: https://github.com/bitwiseops/obsidian-kobo-highlights-import

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/bitwiseops/obsidian-kobo-highlights-import/blob/003b1f028463e99971935195822e602ff7f0f017/main.ts#L35
I recommend setting this up lazily, only when your command is ran. This would speed up the app's boot up process as well as reduce resources needed if your plugin isn't being actively used this session.

https://github.com/bitwiseops/obsidian-kobo-highlights-import/blob/003b1f028463e99971935195822e602ff7f0f017/main.ts#L109
I highly recommend building the whole string in memory and writing it once at the end. Doing it like this would incur many small individual disk writes which is inefficient and also causes things like sync services to repeatedly sync the partial file.

Plus, I don't see the need for `FileSystemAdapter` here since the methods you're calling aren't specific to that one.

bitwiseops:
Hello!

I've just updated my plugin following your suggestions. You can now check if it is okay.

PR #805, opened 2022-02-27
Add TaskPlanner Plugin

eoghano4321:

## Repo URL

Link to my plugin: https://github.com/eoghano4321/TaskPlanner

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3552,5 +3552,13 @@
         "description": "Command palette without unwanted commands",
         "repo": "qawatake/obsidian-command-palette-minus-plugin",
         "branch": "main"
+    },
+    {
+	"id": "obsidian-task-planner-plugin"
+    "name": "Task Planner",
+    "author": "Eoghan O'Mahony",
+	"description": "Task Planning and Project Manager Plugin for Obsidian"
+    "repo": "eoghano4321/TaskPlanner",
+	"branch": "master"
     }
 ]


lishid:
https://github.com/eoghano4321/TaskPlanner/blob/d695a43e36baaa10fa010b3d915774961eecaa82/src/main.ts#L24
Don't use this - it's a debugging variable that's set for convenience for debugging in the console, but not designed to be used for actual production code. Use the reference you get from `MyPlugin.app`.

https://github.com/eoghano4321/TaskPlanner/blob/d695a43e36baaa10fa010b3d915774961eecaa82/src/main.ts#L66
You should run this code in `Workspace.onLayoutReady`.

https://github.com/eoghano4321/TaskPlanner/blob/d695a43e36baaa10fa010b3d915774961eecaa82/src/main.ts#L79
From a UX perspective I feel like this isn't really necessary.

https://github.com/eoghano4321/TaskPlanner/blob/d695a43e36baaa10fa010b3d915774961eecaa82/src/main.ts#L106
Ok I highly recommend not doing this every 4 seconds. You should instead, watch for file change events and react to the ones that specify the file you're interested in.

Constantly reading file from disk (and perform potentially expensive parsing on them) leads to a resource hog where your plugin does a lot of wasted work (when the file hasn't changed) even if the app is completely idle and nothing is happening.


https://github.com/eoghano4321/TaskPlanner/blob/d695a43e36baaa10fa010b3d915774961eecaa82/src/Parser.ts#L33
You should look for the file using `Vault.getAbstractFileByPath` and then read it with `Vault.cachedRead` instead. By doing that, it would remove the need to re-read the file from disk if it's already cached in memory.

https://github.com/eoghano4321/TaskPlanner/blob/d695a43e36baaa10fa010b3d915774961eecaa82/src/Parser.ts#L137
Same as above.

https://github.com/eoghano4321/TaskPlanner/blob/d695a43e36baaa10fa010b3d915774961eecaa82/src/Parser.ts#L143
A lot of this doesn't require the `await` keyword here.

https://github.com/eoghano4321/TaskPlanner/blob/d695a43e36baaa10fa010b3d915774961eecaa82/src/CreateTaskNote.ts#L63
Here and many more - I see that you're using the `adapter` layer for simplicity - I would highly recommend using `Vault` instead of adapter for all your file creation/existence use because every `adapter` call translates to an underlying file system call, whereas `Vault` is a highly optimized, in-memory mirror of the state on disk, which means a lot of operations will not require an OS level disk read, leading to better performance and less resource usage.

eoghano4321:
I think that I have fixed these issues in the code. I have created a new 0.1.3 release

lishid:
https://github.com/eoghano4321/TaskPlanner/blob/52be1fed6ee8233d94c47fff28661781c3487791/src/main.ts#L67
Should still use onLayoutReady for this.

https://github.com/eoghano4321/TaskPlanner/blob/52be1fed6ee8233d94c47fff28661781c3487791/src/main.ts#L85
This isn't really necessary, plus it'll disappear like immediately because the timeout is in milliseconds.

https://github.com/eoghano4321/TaskPlanner/blob/52be1fed6ee8233d94c47fff28661781c3487791/src/main.ts#L112
This is probably even worse than doing it every 4 seconds - you'll basically run this code for every keystroke the user types.   You're looking for a debounce function, not setTimeout which just delays execution instead of combining multiple function calls into one.

Try using `import {debounce} from 'obsidian';`, the syntax is 
```js
update_tasks = debounce(() => this.parser.update_act_tasks(), 4000);
```

https://github.com/eoghano4321/TaskPlanner/blob/52be1fed6ee8233d94c47fff28661781c3487791/src/main.ts#L113
You also shouldn't be doing this kind of checks, and constructing the file name string twice.

I would suggest using the `Vault.on('change')` event instead, and check if the file in question matches the path you're looking for, then running the debounced function, instead of using `input` and `change`.

https://github.com/eoghano4321/TaskPlanner/blob/52be1fed6ee8233d94c47fff28661781c3487791/src/Parser.ts#L31
You really should make this string concatenation into some kind of helper function so you don't have to copy paste this everywhere in your code.

https://github.com/eoghano4321/TaskPlanner/blob/52be1fed6ee8233d94c47fff28661781c3487791/src/Parser.ts#L35
This is just totally unnecessary and bad for performance - you're asking the vault to generate a list of all files just to see if this variable is in that list. The vault literally just gave you the file from `getAbstractFileByPath`, which means if the file returned exists it must have been in the vault.

Here you should check instead if `task_file instanceof TFile` (the file is not null and is an instance of a file, and not a folder)

https://github.com/eoghano4321/TaskPlanner/blob/52be1fed6ee8233d94c47fff28661781c3487791/src/Parser.ts#L60
await not needed here.

https://github.com/eoghano4321/TaskPlanner/blob/52be1fed6ee8233d94c47fff28661781c3487791/src/CreateTaskNote.ts#L64
Same as above.

https://github.com/eoghano4321/TaskPlanner/blob/52be1fed6ee8233d94c47fff28661781c3487791/src/CreateTaskNote.ts#L66
Again, you keep re-constructing the file paths here. You can use `task_file.path` instead of reconstructing it. Remember that each time you construct the file path, you're creating a new Date object, converting it into a Moment object (which isn't necessary since you can just do `moment()` instead of `moment(new Date())`.

lishid:

PR #806, opened 2022-02-27
Text generation plugin

nhaouari:

## Repo URL

Link to my plugin:
[https://github.com/nhaouari/obsidian-textgenerator-plugin/](https://github.com/nhaouari/obsidian-textgenerator-plugin/)

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/nhaouari/obsidian-textgenerator-plugin/blob/6e0a0743128b5763ef589a4d803e4f7fadb73f18/main.ts#L2
Bogus import here, probably unintentional.

https://github.com/nhaouari/obsidian-textgenerator-plugin/blob/6e0a0743128b5763ef589a4d803e4f7fadb73f18/main.ts#L49
This should be imported from `obsidian`, the window.request function is for debugging purposes and isn't intended to be used in production code.

https://github.com/nhaouari/obsidian-textgenerator-plugin/blob/6e0a0743128b5763ef589a4d803e4f7fadb73f18/main.ts#L55
getActiveFile may return null causing a crash here.

https://github.com/nhaouari/obsidian-textgenerator-plugin/blob/6e0a0743128b5763ef589a4d803e4f7fadb73f18/main.ts#L59
Don't mutate shared cache objects. Make a copy if you absolutely have to, but you can easily skip the `position` parameter in your iteration `if (key === 'position') continue;`

https://github.com/nhaouari/obsidian-textgenerator-plugin/blob/6e0a0743128b5763ef589a4d803e4f7fadb73f18/main.ts#L77
Oh yeah this is bad - activeLeaf can be null, activeLeaf.view may not be a MarkdownView so it may be something else without a currentMode. Plus you're repeating this really long object chain.

Instead you should use `workspace.getActiveViewOfType(MarkdownView)`, check if that returns non null, and then use the `view.editor` from the returned view.
(See tip: https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-accessing-workspaceactiveleaf-directly)

https://github.com/nhaouari/obsidian-textgenerator-plugin/blob/6e0a0743128b5763ef589a4d803e4f7fadb73f18/main.ts#L99
Same as above.

https://github.com/nhaouari/obsidian-textgenerator-plugin/blob/6e0a0743128b5763ef589a4d803e4f7fadb73f18/main.ts#L133
Actually, you should use an `editorCallback` here and just pass the editor object along your function calls. This makes it for a consistent experience, rather than your various functions try to get the current active file/leaf constantly from Workspace.

https://github.com/nhaouari/obsidian-textgenerator-plugin/blob/6e0a0743128b5763ef589a4d803e4f7fadb73f18/main.ts#L220
Forgot to rename the class here 😅

nhaouari:
Hi,

Thank you for raising these issues and helping to improve the quality of the codebase. I have fixed all of them.

Best regards

nhaouari:
> https://github.com/nhaouari/obsidian-textgenerator-plugin/blob/33f8e745feb1098e719860f54b195880c7ccb271/main.ts#L89 Given that you're not using the activeView for anything here (since you already have a reference to the editor) you can drop this.

OK, I have modified the functions that use the editor.

PR #808, opened 2022-03-01
Add Creases plugin

liamcain:

## Repo URL

Link to my plugin:
https://github.com/liamcain/obsidian-creases

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/liamcain/obsidian-creases/blob/09ae6194032e629a63f10c7ac7a9a94697f80d90/src/index.ts#L38
Editor commands should always check regardless of whether it's checking or not, to avoid race conditions where your function is called after the checked condition has changed.

Example of how it should be handled: https://marcus.se.net/obsidian-plugin-docs/guides/commands#editor-commands

https://github.com/liamcain/obsidian-creases/blob/09ae6194032e629a63f10c7ac7a9a94697f80d90/src/index.ts#L70
Same as above.

https://github.com/liamcain/obsidian-creases/blob/09ae6194032e629a63f10c7ac7a9a94697f80d90/src/index.ts#L124
Missing null check on view.

https://github.com/liamcain/obsidian-creases/blob/09ae6194032e629a63f10c7ac7a9a94697f80d90/src/index.ts#L285
Same here.

liamcain:
> Editor commands should always check regardless of whether it's checking or not, to avoid race conditions where your function is called after the checked condition has changed

Aah, good point. Fixed!

And added null checks for `view`. I was confused why Typescript let me get away with that but silly me, I didn't have `strictNullChecks` enabled on my project _tsk tsk_.

---

Thanks for the review and for looking the other way at all the hacks 🙈

PR #810, opened 2022-03-02
Add Obsidian Dynamic Embed plugin

dybrawin:

## Repo URL

Link to my plugin: [Obsidian Dynamic Embed](https://github.com/dybrawin/obsidian-dynamic-embed)

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3552,5 +3552,13 @@
         "description": "Command palette without unwanted commands",
         "repo": "qawatake/obsidian-command-palette-minus-plugin",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-dynamic-embed",
+        "name": "Obsidian Dynamic Embed",
+        "author": "Ivaylo Dimitrov Dabravin",
+        "description": "Embed snippets, templates and any linkable by delegating the current scope to the embedded file, treating it as content instead of a reference.",
+        "repo": "dybrawin/obsidian-dynamic-embed",
+        "branch": "main"
     }
 ]


liamcain:
https://github.com/dybrawin/obsidian-dynamic-embed/blob/441d8b9f1858273d8c0038e0af142b46894792bb/main.ts#L15 Instead of fetching all the files and matching by basename, you should use [getFirstLinkpathDest](https://github.com/obsidianmd/obsidian-api/blob/71b3b362bb1f0e8f3c13f1a1d92f65427d2ad477/obsidian.d.ts#L2039) which will properly support absolute links to files as well as resolve ambiguous links according to how users would expect.

dybrawin:
@liamcain Cheers for the heads up, missed that dedicated method.
- [Master is updated](https://github.com/dybrawin/obsidian-dynamic-embed/blob/e4c61f141db28c9edbbee5c1f8bb87b70bc63b60/main.ts#L15)
- [New release (1.0.2) is up](https://github.com/dybrawin/obsidian-dynamic-embed/releases/tag/1.0.2)

dabravin:
@lishid Thanks for the tip. I saw some strange behavior with the cachedRead, views not updating in the case of embedded scripts (e.g. dataview query showing outdated/irrelevant/wrong inlinks). I still observe that with the standard read method, but looks less frequent, let me know what you think.
As for the file extension, great tip, preparing a new release, will follow up with a comment and update on the link syntax.

lishid:
`cachedRead` should always get the most recent version of the file, except very specific cases in race conditions where there's a split second delay between other apps making changes to the file and Obsidian getting the file changed event from the OS. This almost never happens and if is happening to you, it must be a different problem.

Also you should make your function async and `await` for the read, rather than using .then() because it won't be awaited for on the caller of your code block processor.

dabravin:
@lishid The blocking call was the culprit behind outdated embeds/views, cheers for pointing that out.
Did extensive testing with `cachedRead` and can confirm it doesn't introduce any regression compared to the basic `read`.
- [Master is updated](https://github.com/dabravin/obsidian-dynamic-embed/tree/master)
- [New release (1.1.0) is up](https://github.com/dabravin/obsidian-dynamic-embed/releases/tag/1.1.0)

Release notes :
> Changing markdown conversion to blocking call
Embedded file reading is now using the cachedRead API method
Added an error case for non-markdown embedded files
Update of usage doc, not supporting links to blocks in files
General code refactoring and improving readability and quality

PR #814, opened 2022-03-05
Add Card View Switcher to community plugins

qawatake:

## Repo URL

Link to my plugin: https://github.com/qawatake/obsidian-card-view-switcher-plugin

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3552,5 +3552,13 @@
         "description": "Command palette without unwanted commands",
         "repo": "qawatake/obsidian-command-palette-minus-plugin",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-card-view-switcher-plugin",
+        "name": "Card View Switcher",
+        "author": "qawatake",
+        "description": "Quick switcher with card view",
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/qawatake/obsidian-card-view-switcher-plugin/blob/9362d62a2b569aa7ff2ba721a5e077b71733f578/src/main.ts#L21
Here you run into a memory leak where every invocation of your command creates a new switcher and it is added as a child component of your plugin - you should store a copy of it once it's created and reuse it on subsequent triggers, or removeChild when your Switcher object is gone.

qawatake:
@lishid 

Thank you for reviewing!
I have fixed codes. Could you check it again?

https://github.com/qawatake/obsidian-card-view-switcher-plugin/blob/013444638f4b377c313dab9a7b68f41aa933d79a/src/main.ts#L42-L47

PR #815, opened 2022-03-05
Add Steemit plugin to community plugins

anpigon:

## Repo URL

Link to my plugin: https://github.com/anpigon/obsidian-steemit-plugin

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
Hi @anpigon, it looks like your pull request caused the whole file to get reformatted. Do you mind fixing it so that only your additions are shown in the diff?

Also, could you add a link to the plugin in the pull request description? It makes it a bit easier for passersby to check out what the plugin is.

anpigon:
@liamcain Thank you for the comment. I fixed it.

lishid:
https://github.com/anpigon/obsidian-steemit-plugin/blob/f56c43d1e83c9ac5ba56bfa6cee57f40692772ab/src/steemit-client.ts#L19
Recommend using `cachedRead` for better performance.

https://github.com/anpigon/obsidian-steemit-plugin/blob/f56c43d1e83c9ac5ba56bfa6cee57f40692772ab/src/steemit-client.ts#L20
Recommend using `metadataCache.getFileCache(file)` and using the resulting frontmatter instead of bringing your own frontmatter+yaml parser, which is bloating your plugin to over 1MB.

anpigon:
> Recommend using `metadataCache.getFileCache(file)` and using the resulting frontmatter instead of bringing your own frontmatter+yaml parser, which is bloating your plugin to over 1MB.

@lishid Thanks for the advice. I've fixed it all. 
- [x] Using `cachedRead` for better performance. [05a536a](https://github.com/anpigon/obsidian-steemit-plugin/commit/05a536a5d570393ffcfa1e0db048bcb3e369c211)
- [x] Using `metadataCache.getFileCache(file)` instead of gray-matter package. [2a289fd](https://github.com/anpigon/obsidian-steemit-plugin/commit/2a289fd790b8d4bfa318c379bf55dd74478ad90c)

lishid:
I did a quick survey of your release file and I think a huge portion of it is just lodash and bluebird, but it seems like that's from the upstream library so I don't suppose we have a way to replace them with modern js versions.

PR #816, opened 2022-03-05
Add Kindle Export Plugin

SimeonLukas:

## Repo URL

Link to my plugin: https://github.com/SimeonLukas/obsidian-kindle-export

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3552,5 +3552,13 @@
         "description": "Command palette without unwanted commands",
         "repo": "qawatake/obsidian-command-palette-minus-plugin",
         "branch": "main"
-    }
+    },
+    {
+	"id": "obsidian-kindle-export"
+    "name": "Kindle Export",
+    "author": "Simeon Stanek",
+	"description": "Export your Notes to your Kindle. Includes embedded Notes and Images ![[]]"
+    "repo": "SimeonLukas/obsidian-kindle-export",
+	"branch": "main"
+}
 ]


SimeonLukas:
Hi Obsidian Team,

I have a little Plugin, which I wrote.
I hope it is ok.
I think it is really usefull.

Greetings,
Simeon

lishid:
https://github.com/SimeonLukas/obsidian-kindle-export/tree/main/node_modules
Uhhh really don't recommend committing the node_modules folder to your git.

https://github.com/SimeonLukas/obsidian-kindle-export/blob/2b4a35ddaa0e7814a8768d15f28d8cae59d66250/main.ts#L56
this may return null causing a crash later.

https://github.com/SimeonLukas/obsidian-kindle-export/blob/2b4a35ddaa0e7814a8768d15f28d8cae59d66250/main.ts#L63
Recommend using `cachedRead` instead of read for better performance.

https://github.com/SimeonLukas/obsidian-kindle-export/blob/2b4a35ddaa0e7814a8768d15f28d8cae59d66250/main.ts#L69
This should reuse the dokument variable.

https://github.com/SimeonLukas/obsidian-kindle-export/blob/2b4a35ddaa0e7814a8768d15f28d8cae59d66250/main.ts#L148
Here you are using Buffer which is a NodeJS API, which only exists in the desktop app. I recommend using a browser-compatible ArrayBuffer to Base64 function which would work across all platforms, but you can also just set your plugin to only work on desktop [here](https://github.com/SimeonLukas/obsidian-kindle-export/blob/2b4a35ddaa0e7814a8768d15f28d8cae59d66250/manifest.json#L9).

SimeonLukas:
Thank you very much for your answer.
I try to fix this.

So I have much to learn. 😅

Greetings

SimeonLukas:
Hi lishid,

I made the Changes you told me.
Thank you very much for your help and advice.

The Buffer API works for Android Devices. On IOS I can't test.

Greetings

lishid:
https://github.com/SimeonLukas/obsidian-kindle-export/blob/35093060f055dfc264622e81d9a841438213db57/main.ts#L68
May have missed this one to `cachedRead` instead of `read`.

SimeonLukas:
I fixed it!
Thank you very much! 💯 
This feels great 😊.

PR #818, opened 2022-03-05
add obsidian-pdf-annotation, a Plugin to extract PDF Annotations

munach:

## Repo URL

Link to my plugin: https://github.com/munach/obsidian-pdf-annotations

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [X]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


munach:
Dear reviewer, 
thanks a lot for reviewing my plugin. I hope you'll find everything as expected. 
kind regards
Franz

lishid:
https://github.com/munach/obsidian-pdf-annotations/blob/ab3ce95351a741d2f269474ca71242217c2617c4/main.ts#L234
Here you should always check these conditions even if `checking` is false. Here's an example of  how to do it: https://marcus.se.net/obsidian-plugin-docs/guides/commands#conditional-commands

https://github.com/munach/obsidian-pdf-annotations/blob/ab3ce95351a741d2f269474ca71242217c2617c4/main.ts#L247
getActiveFile may return null causing a crash here.

I suggest writing an anonymous async function here rather than mixing async/await with `.then`.

https://github.com/munach/obsidian-pdf-annotations/blob/ab3ce95351a741d2f269474ca71242217c2617c4/main.ts#L289
Don't need to write the settings back if they haven't changed from default values, right?

munach:
Dear lishid, 
thanks a lot for your suggestions. In fact, the refactoring of the .then notations removed some duplicated code. 
kind regards 
Franz

lishid:
Seems like your pull request is kinda broken - can you fix that on your branch? At the very least it should not modify the releases and stats files.

PR #819, opened 2022-03-07
Add Search Everywhere plugin to community plugins

Mom0aut:

## Repo URL

Link to my plugin: https://github.com/Mom0aut/obsidian-search-everywhere.git

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/Mom0aut/obsidian-search-everywhere/blob/c9b115c16dffd53cd373debc247641916557cc74/main.ts#L3
https://github.com/Mom0aut/obsidian-search-everywhere/blob/c9b115c16dffd53cd373debc247641916557cc74/main.ts#L77
Seems unnecessary since you aren't storing any settings for now? Maybe remove all of this code and get it back from the sample plugin later once you add settings. Or maybe don't have settings at all.

https://github.com/Mom0aut/obsidian-search-everywhere/blob/c9b115c16dffd53cd373debc247641916557cc74/main.ts#L65
See this tip: https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#class-names

https://github.com/Mom0aut/obsidian-search-everywhere/blob/c9b115c16dffd53cd373debc247641916557cc74/main.ts#L32
This is terrible haha. The user may have reconfigured that hotkey for something else.

I recommend using `app.commands.executeCommandById('global-search:open')` with a `// @ts-ignore` for the missing commands definition.

Mom0aut:
@lishid thank you for your feedback, i fixed the issues.

lishid:
https://github.com/Mom0aut/obsidian-search-everywhere/blob/945658b1de693b5b17a1a663e4eb9ccb6d5f9cc2/main.ts#L26

Seems like you're missing the reference to `this` here - even though it works, it's probably referring to the global debug variable `this` which is not intended for production use.

You should pass `this.app` from inside your plugin to this function.

Mom0aut:
@lishid should be fixed now.

lishid:
Ah just what I was afraid.

https://github.com/Mom0aut/obsidian-search-everywhere/blob/8a673bd56533d8ade5ac8679427df29ca841ee0c/main.ts#L34
Here you're creating a brand new function, which isn't the same reference as the function you passed into it the first time when registering the event.

You can either use `this.registerDomEvent(window, 'keydown', callback)` to have it automatically unhook, or keep a reference to the original function created by the anonymous function syntax.

Mom0aut:
@lishid thank you for your feedback it should be fixed now.

lishid:
I would consider adding `isDoublePress = false` if the key isn't Shift, to avoid quickly typing two capital letters like `<shift>S<shift>T` to trigger your sequence.

You can also replace your setTimeout with a better mechanism to just record the last Shift key's press time with `Date.now()` and compare the second press's time with the first, that way you won't need to register a global timer.

One more thing I would recommend is that since your code is so simple, I would consider merging all your code into a single function in the registered event:

```ts
let lastKeydownTime = 0;
this.registerDomEvent(window, 'keydown', (event) => {
	if (event.key !== "Shift") {
		// Clears the keypress to break the sequence
		lastKeydownTime = 0;
		return;
	}
	if (Date.now() - lastKeydownTime < 500) {
		lastKeydownTime = 0;
		// @ts-ignore
		app.commands.executeCommandById('global-search:open')
	} else {
		lastKeydownTime = Date.now();
	}
})
```

Mom0aut:
thank you for your suggestion i changed the double press detection, it is much simpler now thank you :D

PR #820, opened 2022-03-07
Add Obsidian-CircuitJS, plugin to embed CircuitJS circuit simulations into notes

StevenGann:

## Repo URL

[LINK](https://github.com/StevenGann/obsidian-circuitjs)

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


StevenGann:
I do not have access to a MacOS device to test on, but considering how simple the plugin currently is I don't expect any issues. Presently, all it really does is detect specific code blocks and replace them with an iframe of CircuitJS with the circuit code as a URL parameter.

As for Android and iOS, I also lack any iOS devices but it seems to work on my Android device, but usability isn't great on my 4" phone screen so I wouldn't recommend it.

lishid:
https://github.com/StevenGann/obsidian-circuitjs/blob/361c7616ccbea28944ce52d83a8bba90955fd73a/main.ts#L22
This doesn't seem necessary? It also never clears once the rendering is done.

https://github.com/StevenGann/obsidian-circuitjs/blob/361c7616ccbea28944ce52d83a8bba90955fd73a/circuitRenderer.ts#L33
This is prone to a DOM injection attack - please avoid using innerHTML when possible. You can build up these elements with our helper function `createEl`.

See this tip: https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml

https://github.com/StevenGann/obsidian-circuitjs/blob/361c7616ccbea28944ce52d83a8bba90955fd73a/circuitRenderer.ts#L38
Since your post processor doesn't have anything to unload, you don't really need to implement a full MarkdownRenderChild. You can simply just render your iframe in the post processor function.

StevenGann:
Thank you for the feedback. The MarkdownRenderChild and statusBarItemEl are from the plugin I was working off of as an example. I'll remove the statusBarItemEl line but do I need to refactor to not implement a MarkdownRenderChild? I'm still trying to decide where to go with this plugin and I don't know if I'll have something to unload later.

As for the potential for a DOM injection, all user input is not only LZIP compressed but also Base64 encoded, so if someone worked out a functional DOM injection attack it'd be pretty impressive. That said, I was also simply unaware of any alternatives (New to Obsidian, TypeScript, and web development in general) and will definitely refactor that part accordingly. I just don't want you thinking I hadn't thought that part through.

Last question: Do I need to open a new PR after I make these changes, or will new commits automatically be added to this PR? This is the first time I've done a real PR flow on Github, I'm more accustomed to BitBucket.

lishid:
> The MarkdownRenderChild and statusBarItemEl are from the plugin I was working off of as an example. I'll remove the statusBarItemEl line but do I need to refactor to not implement a MarkdownRenderChild? I'm still trying to decide where to go with this plugin and I don't know if I'll have something to unload later.

I think removing the MarkdownRenderChild would make your code a lot cleaner as your post processor could literally be a single line that creates the DOM elements for the iframe. Right now there's a lot of boilerplate around it.

> As for the potential for a DOM injection, all user input is not only LZIP compressed but also Base64 encoded, so if someone worked out a functional DOM injection attack it'd be pretty impressive. That said, I was also simply unaware of any alternatives (New to Obsidian, TypeScript, and web development in general) and will definitely refactor that part accordingly. I just don't want you thinking I hadn't thought that part through.

`this.settings.editable` is what can be injected with - even though it's typed as a boolean in code, it may be a string in the JSON file. Either way, it's always a good idea to avoid innerHTML in cases like this.

> Last question: Do I need to open a new PR after I make these changes, or will new commits automatically be added to this PR? This is the first time I've done a real PR flow on Github, I'm more accustomed to BitBucket.

No need to do anything here - this PR is only here to add your plugin into the repo and doesn't actually contain anything from your plugin anyway.

StevenGann:
Thank you.
Any chance there's an example of using `createEl()`, `createDiv()` and `createSpan()`? I looked at the Obsidian API docs but honestly 
`function createDiv(o?: DomElementInfo | string, callback?: (el: HTMLDivElement) => void): HTMLDivElement;` 
...doesn't really help me understand how to use this function, and GitHub search, Bing, and Google are all coming up dry.

lishid:
The typical use case is `parent.createEl('a', {cls: 'css-class', text: 'innerText'})`.
`DomElementInfo` should tell you what can be set in the parameter.

StevenGann:
I really appreciate the feedback. I'm pretty new to this whole technology stack.
[16c2918](https://github.com/StevenGann/obsidian-circuitjs/commit/16c29189cdb97c7575f0f64c51087e6c35ca85bc)
I've implemented most of your recommendations here, as well as some further improvements. I appreciate that I could probably eliminate the MarkdownRenderChild but unless that'll prevent admission into the community catalogue I'd rather keep it structured in this way.

lishid:
Sure, I have no problem with that.

PR #824, opened 2022-03-09
Add File Cleaner plugin

Johnson0907:

## Repo URL

Link to my plugin: https://github.com/Johnson0907/obsidian-file-cleaner

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/Johnson0907/obsidian-file-cleaner/blob/a02b5ba31ebbfa1cb02fe75116d814bb90b35696/main.ts#L21
You can use the new for-of loops here like this:

```ts
for (let file of files) {
	if (file.stat.size === 0) {
		cleanFiles.push(file);
	}
}
```

https://github.com/Johnson0907/obsidian-file-cleaner/blob/a02b5ba31ebbfa1cb02fe75116d814bb90b35696/main.ts#L30
Same as above.

https://github.com/Johnson0907/obsidian-file-cleaner/blob/a02b5ba31ebbfa1cb02fe75116d814bb90b35696/main.ts#L32
Here you should add `await`, and you will also need to make your function `async () => {`.

Johnson0907:
Thanks for your guidance, the code is updated.

lishid:
https://github.com/Johnson0907/obsidian-file-cleaner/blob/46c5fc0816b4cc6c26df5a05a25c8e5ee2581b56/main.ts#L32
Don't make it run async like that because it won't be awaited for properly. You should add the async keyword to the top level function here: https://github.com/Johnson0907/obsidian-file-cleaner/blob/46c5fc0816b4cc6c26df5a05a25c8e5ee2581b56/main.ts#L15

Johnson0907:
I have updated the code, thanks.

lishid:
Looks good. You should avoid using `var` and always use `let` or `const`.

Johnson0907:
I am not a computer major, this is my first time learning to program to achieve my own needs.
I got so much from obsidian. Thank you very much for your guidance.

PR #827, opened 2022-03-09
Submit 'Novel Word Count' plugin

isaaclyman:
This plugin solves a real problem for me and makes Obsidian much better as a novel-writing tool. It's my first Obsidian plugin and I realize it may not be the most stable thing around. I'm very much open to feedback.

## Repo URL

Link to my plugin: https://github.com/isaaclyman/novel-word-count-obsidian

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


isaaclyman:
FYI: Just published 2.0.0 - discovered a bug that needed fixing, also added a command and a button in the plugin settings to recount all files in the vault.

eleanorkonik:
It looks like you forked the template rather than "using" the template, so the repo is a copy of the template repo instead of being a child. This means, for example, that people can't file issues on the plugin repo. You might be able to go into the repo settings and turn on issues and turn off the fact that it's a template repo.

PR #832, opened 2022-03-12
New plugin submission

mtopping:
https://github.com/mtopping/obsidian-raindrop/releases/tag/1.0.0

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3736,5 +3736,12 @@
         "author": "Johnson0907",
         "description": "Help you to clean empty files and unused attachments in the vault.",
         "repo": "Johnson0907/obsidian-file-cleaner"
+    },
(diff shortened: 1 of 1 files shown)


lishid:
Your plugin code looks mostly ok so this should be quick and easy.

https://github.com/mtopping/obsidian-raindrop/blob/1e061acb20294100a96e295d81927579cab460be/main.ts#L120
Here this might cause a reference leak since code blocks are internally managed by the rendering page. Once the code block rendering disappears, you still retain a reference to this object which causes a memory leak.

https://github.com/mtopping/obsidian-raindrop/blob/1e061acb20294100a96e295d81927579cab460be/src/raindrop.ts#L19
Should just use regular await instead of mixing awaits and `.then` here.

PR #833, opened 2022-03-12
Add Telegraph Publish Plugin

reorx:

## Repo URL

Link to my plugin: https://github.com/reorx/obsidian-telegraph-publish

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/reorx/obsidian-telegraph-publish/blob/3ae43ae0b11e11e448d8540152ba3c3b0daed6f4/main.ts#L171
Missing null check here.

https://github.com/reorx/obsidian-telegraph-publish/blob/3ae43ae0b11e11e448d8540152ba3c3b0daed6f4/main.ts#L177
I don't think this is what you want to do here. Instead you'll want to unpack the object you got from `getViewState().state`.

https://github.com/reorx/obsidian-telegraph-publish/blob/3ae43ae0b11e11e448d8540152ba3c3b0daed6f4/main.ts#L234
Recommend keeping async/await instead of then/catch (consider using try-catch here instead of .catch)

https://github.com/reorx/obsidian-telegraph-publish/blob/3ae43ae0b11e11e448d8540152ba3c3b0daed6f4/main.ts#L317
https://github.com/reorx/obsidian-telegraph-publish/blob/3ae43ae0b11e11e448d8540152ba3c3b0daed6f4/main.ts#L344
Vulnerability due to HTML injection from fileTitle and other injected variables.

Actually, you can consider using our built in DOM element helper shortcut:

`parentEl.createEl('type', {cls: 'css-class', text: 'innerText'})/createDiv/createSpan({options})`

reorx:
It's worthing mentioning that in [this commit](https://github.com/reorx/obsidian-telegraph-publish/commit/e36c4a920ecf83e259369dcdfe3d94ec73e90672)
I removed the code that gets HTML content from preview element,
because I found that Obsidian uses a progressive rendering mechanism in preview mode,
while scrolling, only the visible part with a small margin is rendered to DOM elements,
this makes it impossible to get full HTML content for long documents.

lishid:
> Missing null check here.

I was trying to say that `view` could be null here, causing an exception to be thrown when you access `view.file`.

PR #837, opened 2022-03-14
Add Daily Notes Viewer Plugin

Johnson0907:

## Repo URL

Link to my plugin: https://github.com/Johnson0907/obsidian-daily-notes-viewer

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/Johnson0907/obsidian-daily-notes-viewer/blob/dd893fc30d180b231e54ee5f53490096339e486e/src/index.ts#L59
You should make sure that `createOrUpdateViewer` is debounced, so you don't end up doing it multiple times per second when a series of events happen close together.

lishid:
https://github.com/Johnson0907/obsidian-daily-notes-viewer/blob/dd893fc30d180b231e54ee5f53490096339e486e/src/util.ts#L53
I also suggest doing a `cachedRead` here and only write the file if it has changed - this would avoid an extra disk write.

Johnson0907:
The source code has been updated.

lishid:
https://github.com/Johnson0907/obsidian-daily-notes-viewer/blob/720f6e48acef67f40fd9784fcb231be0a52d6087/src/index.ts#L59
This is incorrect - and it doesn't work. You need to create the function itself as debounced, and from here you just need to call the debounced function.

https://github.com/Johnson0907/obsidian-daily-notes-viewer/blob/720f6e48acef67f40fd9784fcb231be0a52d6087/src/util.ts#L50
Here you are creating the string `${beginning}\n${fileText}` three times - should set it into a variable first (before the if-else condition) and reuse the variable.

Johnson0907:
Thanks for your guidance, the source code has been updated.

lishid:
Hmmmm

https://github.com/Johnson0907/obsidian-daily-notes-viewer/blob/c65798ebcdeed0355b2ca8405ad637f85fa6dc51/src/index.ts#L101
That's still not quite it - you're creating a timer variable here that's always 0, and then clearing that, which doesn't have an effect. But we are getting close.

You need to do this:

```ts
delayCreateOrUpdateViewer = debounce(() => createOrUpdateViewer(this.app, this.settings), 1000);
```

Johnson0907:
I've updated the source code, but I'm not sure if it's correct. Please check it out. Thanks again.

lishid:
You can import `debounce` from `obsidian`.

Johnson0907:
Thanks for the reminder, I have updated the source code.

lishid:
Hmm ok looks good to merge. One last thing I think we should improve - the plugin's description says: "Help you to review your daily notes like in roam research"

I think a lot of users may not know what that means (or haven't used Roam), so perhaps change the description so it better explain what this plugin exactly does instead?

lishid:
The description is also in this PR so let me know when you get it changed and I can have it merged!

PR #839, opened 2022-03-14
Add "Lapel" to the community plugins list

liamcain:

## Repo URL

Link to my plugin: https://github.com/liamcain/obsidian-lapel

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/liamcain/obsidian-lapel/blob/141e155ea0cfe9a0a0fae399a4cbff5573c071bd/src/headingWidget.ts#L42

Don't use type.name, that's auto generated and the generation algorithm isn't guaranteed. Instead, you should use the proper `node.prop(tokenClassNodeProp)` (or `lineClassNodeProp`) from my fork of `@codemirror/stream-parser`.

liamcain:
> Don't use type.name, that's auto generated and the generation algorithm isn't guaranteed. Instead, you should use the proper node.prop(tokenClassNodeProp) (or lineClassNodeProp) from my fork of @codemirror/stream-parser

Almost positive I remember you saying exactly that in `#plugin-dev`... Still made that mistake 🤦 . Anyway, fixed!

PR #843, opened 2022-03-16
Add "Screwdriver" to the community plugins list.

vrtmrz:

## Repo URL

Link to my plugin: [obsidian-screwdriver](https://github.com/vrtmrz/obsidian-screwdriver)

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/vrtmrz/obsidian-screwdriver/blob/50a0c7310d9a4b19e0b73ef01fc72ff543b6bac2/main.ts#L42
Maybe write an array and do a filename split to get extension and check if extension is in your array?

https://github.com/vrtmrz/obsidian-screwdriver/blob/50a0c7310d9a4b19e0b73ef01fc72ff543b6bac2/main.ts#L57
I don't think this is a good way to do it. I would recommend using the btoa function instead.

https://github.com/vrtmrz/obsidian-screwdriver/blob/50a0c7310d9a4b19e0b73ef01fc72ff543b6bac2/main.ts#L184
Use `string[]` and `array.join('')` with your strings for better performance - doing this kind of string appends in JavaScript causes progressively larger strings to be created and stored in memory which is bad performance for the garbage collector.

https://github.com/vrtmrz/obsidian-screwdriver/blob/50a0c7310d9a4b19e0b73ef01fc72ff543b6bac2/main.ts#L267
These functions don't need to exist if you don't need them.

vrtmrz:
Thank you for merging!
I'll fix the problems you advised me in a few days.

PR #850, opened 2022-03-19
add obsidian-quiet-outline to community-plugins.json

guopenghui:

## Repo URL

Link to my plugin: https://github.com/guopenghui/obsidian-quiet-outline

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/guopenghui/obsidian-quiet-outline/blob/b632c71d8a46ac466de7997959af914c0e4a9214/src/plugin.ts#L31
Don't do this - you would cause the app to unnecessarily load the mathjax module even when no mathjax is used by the user.

lishid:
https://github.com/guopenghui/obsidian-quiet-outline/blob/b632c71d8a46ac466de7997959af914c0e4a9214/src/plugin.ts#L34
You shouldn't need to do this if you set your CSS properly. Seems like you're doing a lot of work to find out about the theme... Would be easy to do as a simple css selector.

https://github.com/guopenghui/obsidian-quiet-outline/blob/b632c71d8a46ac466de7997959af914c0e4a9214/src/plugin.ts#L140
You are not cleaning up this MutationObserver when your plugin is unloaded. That would cause a memory leak due to this being bound on document.body.

https://github.com/guopenghui/obsidian-quiet-outline/blob/b632c71d8a46ac466de7997959af914c0e4a9214/src/plugin.ts#L73
You should use the heading information in Obsidian's metadata cache instead - it has the actual markdown parsed headings with the content text and document position. This way you don't risk computing based on wrong parsing information (such as code blocks possibly having `~~~` as the delimiter.

https://github.com/guopenghui/obsidian-quiet-outline/blob/b632c71d8a46ac466de7997959af914c0e4a9214/src/plugin.ts#L101
Debounce takes an integer for milliseconds, this should be 300 instead.

https://github.com/guopenghui/obsidian-quiet-outline/blob/b632c71d8a46ac466de7997959af914c0e4a9214/src/plugin.ts#L107
This isn't necessary.

https://github.com/guopenghui/obsidian-quiet-outline/blob/b632c71d8a46ac466de7997959af914c0e4a9214/src/Outline.vue#L134
You should use [Obsidian's proper sanitizer](https://github.com/obsidianmd/obsidian-api/blob/c01fc3074deeb3dfc6ee02546d113b448735b294/obsidian.d.ts#L2682) to avoid DOM injection - your security check can easily by bypassed with `<div onload="javascript code"></div>`.

lishid:
https://github.com/guopenghui/obsidian-quiet-outline/commit/a771c34a5f7a02812025ecbdb3b42646836b44f5#diff-e04e17442fb95f758ea1b52e11a812bb8dcffb844997fb4ccf95b422a5d64d8fR118
That's a good first step, however you should run the sanitizer last (after you run the stuff through marked).

guopenghui:
> [guopenghui/obsidian-quiet-outline@a771c34#diff-e04e17442fb95f758ea1b52e11a812bb8dcffb844997fb4ccf95b422a5d64d8fR118](https://github.com/guopenghui/obsidian-quiet-outline/commit/a771c34a5f7a02812025ecbdb3b42646836b44f5#diff-e04e17442fb95f758ea1b52e11a812bb8dcffb844997fb4ccf95b422a5d64d8fR118) That's a good first step, however you should run the sanitizer last (after you run the stuff through marked).

There is a problem: 
If use sanitizer last, `<mjx-container>` (generated by renderMathjax) will be removed and latex formula doesn't look right.  I think it may be enough to sanitize the user input part (`option.label`) only.

lishid:
Ah yes you are right, mathjax will get stripped. I think you can keep mathjax out of the sanitizer.

PR #853, opened 2022-03-21
Add obsidian-desmos to community plugins

Nigecat:

## Repo URL

Link to my plugin: [github.com/Nigecat/obsidian-desmos](https://github.com/Nigecat/obsidian-desmos)

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
In general, you should avoid relying on NodeJS API because your plugin won't run on mobile.

https://github.com/Nigecat/obsidian-desmos/blob/719b077c749349ad78b95ba5604ae8b9aaabeeba/src/main.ts#L15
You can use async-await for this instead of mixing with the old .then promise style.

https://github.com/Nigecat/obsidian-desmos/blob/719b077c749349ad78b95ba5604ae8b9aaabeeba/main.js#L367
Yeah here you should really switch to using async-await.

https://github.com/Nigecat/obsidian-desmos/blob/719b077c749349ad78b95ba5604ae8b9aaabeeba/main.js#L389
Please use the vault/adapter API - here for displaying images you should use `vault.getResourcePath(tfile)` or `adapter.getResourcePath(pathstring)`.

https://github.com/Nigecat/obsidian-desmos/blob/719b077c749349ad78b95ba5604ae8b9aaabeeba/src/renderer.ts#L194
This isn't unloaded when your plugin unloads. In fact, you will register a bunch of these event handlers for each render action.

Nigecat:
@lishid thanks for the feedback,

I have removed all usage of the NodeJS API in [0e5756d](https://github.com/Nigecat/obsidian-desmos/commit/0e5756d0eaba564e9cd8fbaf1a564cbd51be206d) and tested the plugin on Android. This also uses the vault/adapter API instead of the node API. 

I wasn't sure if the usage of `await` in the onload method would have any impact on load times, this appears not be the case after some testing and this has been fixed in [e2c9a18](https://github.com/Nigecat/obsidian-desmos/commit/e2c9a183330763e1f189c668b301fe7a00ceaff6).

As for the method in the renderer, I can't use the async-await syntax there (as far as I know) as the promise gets resolved from a callback, and using an async function plus a return statement wouldn't return the value to the correct place. 

The event handlers were unloaded [here](https://github.com/Nigecat/obsidian-desmos/blob/e2c9a183330763e1f189c668b301fe7a00ceaff6/src/renderer.ts#L162) when the render message returns, however they weren't removed if an error returned (and still bound multiple handlers if there were more than one graph). This has been fixed in [5da468c](https://github.com/Nigecat/obsidian-desmos/commit/5da468c7f27d2c4d034a3b74ab2c1c6bc2f34596) to use a single global handler which is removed on unload.

lishid:
> As for the method in the renderer, I can't use the async-await syntax there (as far as I know) as the promise gets resolved from a callback, and using an async function plus a return statement wouldn't return the value to the correct place.

Ah you are right, my bad!

Otherwise, looks good to me. One minor suggestion is to consider using Component as a parent class for Renderer and just add it as a child to your plugin - then you can use onload/onunload to automatically manage the enable/disable state propagation with your window.message event.

PR #856, opened 2022-03-21
Add Obsidian Keep plugin

Ellpeck:

## Repo URL

Link to my plugin: https://github.com/Ellpeck/ObsidianKeep

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

## A Note
I'd like to point out here that the way I am making this plugin happen is extremely [cursed](https://github.com/Ellpeck/ObsidianKeep/blob/master/main.ts#L105-L117) in [places](https://github.com/Ellpeck/ObsidianKeep/blob/master/main.ts#L132-L140), so I fully understand if it can't be added to the community plugin list right now. There's a short explanation as to why all of this is being done in [the README](https://github.com/Ellpeck/ObsidianKeep/blob/master/README.md) as well.

As stated in the README, I plan on improving this compatibility beyond just displaying Google Keep directly once the [official Google Keep API](https://developers.google.com/keep) is farther along and if/once it exits Enterprise-only mode. Though, since I don't know when that is happening, and since I find this plugin extremely useful for personal use, I thought I'd still try to submit it officially!

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3728,6 +3728,6 @@
         "name": "Obsidian Keep",
         "author": "Ellpeck",
         "description": "An Obsidian plugin that displays Google Keep in the app"
-        "repo": "Ellpeck/https://github.com/Ellpeck/ObsidianKeep"
+        "repo": "Ellpeck/ObsidianKeep"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3722,5 +3722,12 @@
         "author": "Johnson0907",
         "description": "Help you to clean empty files and unused attachments in the vault.",
         "repo": "Johnson0907/obsidian-file-cleaner"
+    },
+    {
+        "id": "obsidian-keep"
+        "name": "Obsidian Keep",
+        "author": "Ellpeck",
+        "description": "An Obsidian plugin that displays Google Keep in the app"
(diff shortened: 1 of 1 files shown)


lishid:
I highly highly don't recommend using BrowserView. You say that iframes don't work - is it because of X-Frame-Options, or is it because of Google's login prevention? Is there nothing else we can explore to make this work with iframes?

lishid:
I think we have some form of `X-Frame-Options` bypass in the electron request handler, but I think some sites add a meta tag in the HTML which we aren't stripping.

Ellpeck:
It looks like I might be able to use [onHeadersReceived](https://www.electronjs.org/docs/latest/api/web-request#webrequestonheadersreceivedfilter-listener) to strip the `X-Frame-Options` and/or `Content-Security-Policy` headers, but I'm not sure if that's possible to do from within a plugin?

I'll have to investigate that.

lishid:
That's what we're using to strip x-frame-options, but I suspect you may be running into more than that since we already strip it.

Hop on discord and DM me `Licat` from the obsidian discord and we can try to work something out

Ellpeck:
Closing this for now, as I've decided to expand on the plugin and rename it. Thanks for all the help, both here and on Discord! ❤️

PR #865, opened 2022-03-27
Add obsidian-text-toolbar plugin

farux:

## Repo URL

Link to my plugin: https://github.com/farux/obsidian-text-toolbar

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3838,5 +3838,13 @@
         "author": "Johnson0907",
         "description": "Help you to view some recent daily notes on one page.",
         "repo": "Johnson0907/obsidian-daily-notes-viewer"
+    },
+    {
+        "id": "obsidian-text-toolbar",
+        "name": "Text Toolbar",
+        "author": "faru",
+        "description": "A customizable toolbar that allows CSS snippets, markdowns, and built-in macros to be applied to selected text.",
+        "repo": "farux/obsidian-text-toolbar",
+        "branch": "main"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3844,7 +3844,7 @@
         "name": "Zotero Desktop Connector",
         "author": "mgmeyers",
         "description": "Insert citations, bibliographies, and notes directly from Zotero desktop.",
-        "repo": "mgmeyers/obsidian-zotero-desktop-connector",
+        "repo": "mgmeyers/obsidian-zotero-desktop-connector"
     },
     {
         "id": "obsidian-text-toolbar",


farux:
Sorry. I don't know what to do.

adifyr:
Hi @farux. In your `community-plugins.json` file, at line no. 3847, there's a comma at the end of `"repo": "mgmeyers/obsidian-zotero-desktop-connector",`. You need to remove that comma and commit it again.

farux:
@adifyr Thank you!

lishid:
https://github.com/farux/obsidian-text-toolbar/blob/833e17d45a9b88c9e41e103f5a1990d4f53c543f/utils/ToolBarBuilder.ts#L6
This shouldn't inherit Plugin.
https://github.com/farux/obsidian-text-toolbar/blob/833e17d45a9b88c9e41e103f5a1990d4f53c543f/utils/Action.ts#L5
Same here, and probably for all your other classes - you should only have a single class extending `Plugin` in your entire bundle.

https://github.com/farux/obsidian-text-toolbar/blob/833e17d45a9b88c9e41e103f5a1990d4f53c543f/utils/ToolBarBuilder.ts#L25
You don't need to register events for components defined in your own closure, which will be detached/removed from DOM when your plugin unloads.

https://github.com/farux/obsidian-text-toolbar/blob/833e17d45a9b88c9e41e103f5a1990d4f53c543f/utils/ToolBarBuilder.ts#L66
These aren't removed when your plugin unloads.

https://github.com/farux/obsidian-text-toolbar/blob/833e17d45a9b88c9e41e103f5a1990d4f53c543f/utils/Action.ts#L7
Missing null check for `view`.

https://github.com/farux/obsidian-text-toolbar/blob/833e17d45a9b88c9e41e103f5a1990d4f53c543f/main.ts#L36
Should use `view.getMode()` instead of asking it to serialize the full state just to get the mode from it.

https://github.com/farux/obsidian-text-toolbar/blob/833e17d45a9b88c9e41e103f5a1990d4f53c543f/main.ts#L39
Ok so I've gone around to various places in your plugin - you should manage your state internally, rather than constantly querying everything from the DOM and making various assumptions on where/how your DOM elements are located.

Your toolbar itself should manage its own attach/detach state instead of having it attached to body randomly here, and detached somewhere else. Same for the DOM event handlers: events that go on your custom elements should be created/attached once when the element is created. Events that go on global objects like document or window should have clear lifecycle management, rather than being gated by `getElementByClassName`.

Here's an example that's bad:
- You attach a global keydown listener [here](https://github.com/farux/obsidian-text-toolbar/blob/833e17d45a9b88c9e41e103f5a1990d4f53c543f/main.ts#L48), but the removal is nowhere to be found in this file. Instead it's located in some "utils" file [here](https://github.com/farux/obsidian-text-toolbar/blob/833e17d45a9b88c9e41e103f5a1990d4f53c543f/utils/Utils.ts#L107), but only after a `document.getElementsByClassName('text-tool-bar')[0]` check. To someone reviewing your code, that means constantly chasing around the codebase for lifecycle management, and possible failure modes (such as your `text-tool-bar` component not being properly attached to `body`, causing a global keydown event leak).

For a better way to structure your code, you should look into having symmetrical functions such as `show()` and `hide()` that does exactly the opposite of each other, such as:

show: attach element to body, register event listerners, etc.
hide: detach element regardless of whether the element is attached, unregister event listeners regardless of element states, etc.

You should avoid randomly querying elements from the DOM as well. When you create an element, keep a reference to it and pass it along. It's a bad practice to assume things like [this](https://github.com/farux/obsidian-text-toolbar/blob/833e17d45a9b88c9e41e103f5a1990d4f53c543f/utils/Utils.ts#L204) which may crash for example, if your element isn't in the DOM for some reason, or if another plugin happens to use the same class name.

https://github.com/farux/obsidian-text-toolbar/blob/833e17d45a9b88c9e41e103f5a1990d4f53c543f/utils/Utils.ts#L424
This kind of element/children management is basically unreadable. What does the code try to do with these elements? There's type casts everywhere. Is this where you display your buttons for each action? Why not just keep them in some kind of data structure like:

```ts
class Button {
  el: HTMLElement;
}
class ButtonsContainer {
  buttons: Button[][];
}
```

https://github.com/farux/obsidian-text-toolbar/blob/833e17d45a9b88c9e41e103f5a1990d4f53c543f/utils/Utils.ts#L115
Shouldn't really do this here. You don't even check if `.cm-content` is from a markdown view, or the active markdown view.

PR #866, opened 2022-03-28
add Control Characters plugin

joethei:

## Repo URL

Link to my plugin: https://github.com/joethei/obsidian-control-characters

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


lishid:
https://github.com/joethei/obsidian-control-characters/blob/b13701423da103332bed873d42b7152fd2e9ba2b/src/main.ts#L29
Use `vault.getConfig('legacyEditor')` instead.

https://github.com/joethei/obsidian-control-characters/blob/b13701423da103332bed873d42b7152fd2e9ba2b/src/NewLineDecoration.ts#L35
I would highly recommend looking into using `Decoration.mark` instead of `Decoration.replace` - replace widgets are a lot more heavy weight, both performance wise from a DOM perspective, and also from how it's being managed inside CM6 internally.

We've also noticed a lot of issues involving replacement widgets with IMEs (input methods such as chinese, japanese, etc).

lishid:
Oh also, would recommend using an auto formatter to normalize whitespace in your code...

joethei:
I have updated the plugin with mark decorations.
Thanks for the suggestion, it fixed another small issue I was having.

PR #871, opened 2022-04-03
added obsidian-list-modified

franciskafieh:

## Repo URL

Link to my plugin:
https://github.com/franciskafieh/obsidian-list-modified

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/franciskafieh/obsidian-list-modified/blob/bbd60aeb42a2f113a9495a477e1f14180af68e62/src/main.ts#L42
This isn't needed since you already registered the event with the component system.

https://github.com/franciskafieh/obsidian-list-modified/blob/bbd60aeb42a2f113a9495a477e1f14180af68e62/src/main.ts#L51
You can just use `this.app` here.

https://github.com/franciskafieh/obsidian-list-modified/blob/bbd60aeb42a2f113a9495a477e1f14180af68e62/src/main.ts#L85
Possible race condition here. You read the file into a class-shared field, and fire off a function with no parameters. What if another modify event happens at the same time? Your modify events aren't awaited for by anything, therefore they are all ran concurrently. This means there's a chance something happens here and the reference to `dailyFileContent`, `dailyFileCache`, `activeFileCache`, and `dailyFile` gets changed from another run of the modify function.

At the very least, you should use function-scope variables and pass them around to various helper functions, to ensure you're always dealing with the same set of files. For better safety, you should also serialize your `modify` event handlers so at most one is running at any time (You can use a promise-based queue where you always await for the last execution with `prevPromise.then(nextCallback, nextCallback)` ).

https://github.com/franciskafieh/obsidian-list-modified/blob/bbd60aeb42a2f113a9495a477e1f14180af68e62/src/main.ts#L57
You shouldn't cast here - instead you should do an `instanceof` check.

franciskafieh:
All these issues should be fixed. feel free to let me know if there's anything else you would like me to do! thanks so much!!

lishid:
https://github.com/franciskafieh/obsidian-list-modified/blob/9490de023fdd762d5d00fee5bd2285b3c6c46a74/src/main.ts#L74
This is quite inefficient because this function will go through every single link in the whole vault. You should use `MetadataCache.resolvedLinks` instead to check for connectedness between two notes.

franciskafieh:
Will fix, thanks!

PR #872, opened 2022-04-03
Update community-plugins.json

matthewwong525:

## Repo URL

Link to my plugin: https://github.com/fleetingnotes/fleeting-notes-obsidian

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
See tip here: https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#class-names

Is "Fleeting notes" some kind of online service, or another app? I can't seem to find much about it because I've seen the term "fleeting notes" used as a generic term. Would be useful to document which service this is.

From code it seems to be some random mongoDB instance? https://github.com/fleetingnotes/fleeting-notes-obsidian/blob/564e30c10d5e4bdbb52d55559c44f9c2036a1b61/main.ts#L138

matthewwong525:
"Fleeting Notes" is an application. For more info visit this website: https://fleetingnotes.app/

I've also updated the class names and released a new version under 0.0.2 release

lishid:
Ah ok, that's so confusing! Honestly I hate to give this kind of advice because it doesn't sound nice but the name pick here is terrible 😢 

I feel like once we merge this plugin, you'll get a ton of people downloading the plugin expecting it to "help them write fleeting notes in Obsidian" and discover that it's just a plugin to sync from an outside app called "fleeting notes".

I don't mind merging such a plugin, but I would recommend renaming the plugin to something like "Fleeting notes sync" and specify in the description that this is for syncing with an app called "Fleeting notes".

Also, personally, I would totally recommend avoiding such a generic name for a product in the first place.

lishid:
Hmm seems like I don't have permission to fix the merge conflict - can you give me permission for merging or fix the conflict from your end? `branch: master` is also not required since that's the default.

PR #874, opened 2022-04-04
Submitting scambier/obsidian-auto-updated-date

scambier:

## Repo URL

Link to my plugin: https://github.com/scambier/obsidian-auto-updated-date

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3838,5 +3838,12 @@
         "description": "Insert citations, bibliographies, and notes directly from Zotero desktop.",
         "repo": "mgmeyers/obsidian-zotero-desktop-connector",
         "branch": "main"
+    },
+    {
+        "id": "scambier.automatic-updated-date",
+        "name": "Automatic updated header",
+        "author": "scambier",
+        "description": "Automatically adds a front matter `updated` field to your notes",
+        "repo": "scambier/obsidian-auto-updated-date"
     }
 ]


lishid:
https://github.com/scambier/obsidian-auto-updated-date/blob/8f9ffc98f6083147e0f9ba8b5a768f12c65d2dfa/src/main.ts#L89
Possible race condition - check for whether your plugin got disabled before adding the event back.

Honestly I don't think what you're doing here works. The `modify` event is fired from every single possible source - even from Obsidian Sync and third party sync tools. That means you'll create yourself a modify loop with two devices modifying the file, syncing the new frontmatter to the other device, and triggering your plugin again on the other device.

To properly achieve what you want to do, you'll have to resort to monkey-patching the save function in Markdown View.

scambier:
Thanks for the feedback. I've been trying all evening, but all solutions were buggy and incomplete. At best, I can write the correct data to the disk but the view isn't updated, and I get error messages when closing files (due to monkey-patching).

I'll get back to it when (if) there's a proper API to do this kind of work.

lishid:
Sounds good, thanks for trying and sorry about the frustration.

PR #875, opened 2022-04-04
added doubleshift

Qwyntex:

## Repo URL

Link to my plugin: https://github.com/Qwyntex/doubleshift

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
Possible collaboration effort with this other plugin? https://github.com/mom0aut/obsidian-search-everywhere

Qwyntex:
> Possible collaboration effort with this other plugin? https://github.com/mom0aut/obsidian-search-everywhere

I actually got the idea from this plugin and used its keylistener as a starting point. So i guess you could say that it is a collaboration. Should i add that in my readme then?

lishid:
Ah ok! I guess you should at least credit the original author.

https://github.com/Qwyntex/doubleshift/blob/d65974cd05b841a6997cbbe3ac1ef3f8fb19d795/main.ts#L5
Here you shouldn't store a typed object - it will be destructively converted to JSON string and re-parsed when serialized to disk.

Instead you should store the plugin's ID here as a string.

Qwyntex:
It doesn't show my gifs in the description do you know why that is?

lishid:
You should use the raw.githubusercontent.com links for them to work properly I think

PR #879, opened 2022-04-06
add Google Task plugin

YukiGasai:

## Repo URL

Link to my plugin: https://github.com/YukiGasai/obsidian-google-tasks/

## Release Checklist
  - [x]  Windows
  - [ ]  macOS I have no way of testing on Mac :(
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/YukiGasai/obsidian-google-tasks/blob/6b04cfa453504f6f38505248a7ef3ab21ff99bf6/src/GoogleTasksPlugin.ts#L3
You should import this from obsidian rather than including a copy in your plugin.

https://github.com/YukiGasai/obsidian-google-tasks/blob/6b04cfa453504f6f38505248a7ef3ab21ff99bf6/src/GoogleTasksPlugin.ts#L59
This can't be directly casted like that, you should perform an instanceof check instead.

https://github.com/YukiGasai/obsidian-google-tasks/blob/6b04cfa453504f6f38505248a7ef3ab21ff99bf6/src/GoogleTasksPlugin.ts#L85
You need to perform this check regardless of the value of `checking`. There are times where your command will be executed with `checking = false` and your check should fail and stop the command here. (Same for all your other commands)

https://github.com/YukiGasai/obsidian-google-tasks/blob/6b04cfa453504f6f38505248a7ef3ab21ff99bf6/src/GoogleTasksPlugin.ts#L88
Should use `await` here instead of `then`.

YukiGasai:
Thanks for response, I updated the project to fix all the issues.

PR #884, opened 2022-04-08
Add  KR Book Info Plugin

kmsk99:

## Repo URL

Link to my plugin: https://github.com/kmsk99/kr-book-info-plugin

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/kmsk99/kr-book-info-plugin/blob/1259cba3d3df1cbf9f2e5f32ee843e042a6df2ef/main.ts#L15
This might return null.

https://github.com/kmsk99/kr-book-info-plugin/blob/1259cba3d3df1cbf9f2e5f32ee843e042a6df2ef/main.ts#L32
Recommend sanitizing the new title in case it contains punctuations or any illegal characters in the file system.

https://github.com/kmsk99/kr-book-info-plugin/blob/1259cba3d3df1cbf9f2e5f32ee843e042a6df2ef/src/crawler.ts#L1
Recommend using Obsidian's built-in `requestUrl` function instead of using an external library.

https://github.com/kmsk99/kr-book-info-plugin/blob/1259cba3d3df1cbf9f2e5f32ee843e042a6df2ef/src/crawler.ts#L2
Recommend using the DOMParser API (https://developer.mozilla.org/en-US/docs/Web/API/DOMParser) instead of using an external library.

https://github.com/kmsk99/kr-book-info-plugin/blob/1259cba3d3df1cbf9f2e5f32ee843e042a6df2ef/src/crawler.ts#L127
Here you should create an object and use `stringifyYAML` to serialize the YAML to avoid issues with formatting characters. https://github.com/obsidianmd/obsidian-api/blob/fc616fa86533d31b1d68b45887f4c9c46da389bf/obsidian.d.ts#L2987

kmsk99:
1. null checked
2. sanitizing filename checked
3. use requestUrl instead of axious checked
4. use DOMParser instead of cheerio checked
5. inserted yaml with stringifyYAML checked

Thanks for commenting.
Checked all of issues, I hope it will be fine.

lishid:
https://github.com/kmsk99/kr-book-info-plugin/blob/99b3f9578283fa9940f41bcbbebd47c441c6a9bd/main.ts#L14
One more thing I forgot - you should check that the file's extension is `md` just to avoid accidentally processing attachments like PDFs and images.

kmsk99:
Fixed extension issue. thanks a lot.

PR #885, opened 2022-04-08
Add Paste Image Rename

reorx:

## Repo URL

Link to my plugin: https://github.com/reorx/obsidian-paste-image-rename

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/reorx/obsidian-paste-image-rename/blob/b60f2d7cded3b8f1fab9e05c37bc43147af2dc43/src/main.ts#L122
You might also want to check if the user is using markdown links instead of wikilinks.

https://github.com/reorx/obsidian-paste-image-rename/blob/b60f2d7cded3b8f1fab9e05c37bc43147af2dc43/src/main.ts#L184
Input variables needs to be sanitized here.

https://github.com/reorx/obsidian-paste-image-rename/blob/b60f2d7cded3b8f1fab9e05c37bc43147af2dc43/src/main.ts#L168
You can get a lot of the properties you're using `path` for from the file: file.basename, file.extension, file.parent.path. If you absolutely need to use `path` from nodejs then you must mark your plugin as desktop only in your manifest.json file because it won't work on mobile.

https://github.com/reorx/obsidian-paste-image-rename/blob/b60f2d7cded3b8f1fab9e05c37bc43147af2dc43/src/main.ts#L142
This has a memory leak here where you keep the reference of all the modals that has ever existed, even after they were closed.

reorx:
Thanks for the careful and professional code review as always.

- [x] https://github.com/reorx/obsidian-paste-image-rename/blob/b60f2d7cded3b8f1fab9e05c37bc43147af2dc43/src/main.ts#L122
    You might also want to check if the user is using markdown links instead of wikilinks.
    - add handling useMarkdownLinks in this commit https://github.com/reorx/obsidian-paste-image-rename/commit/7a62a266a2bd5192b49fe326e028acf11d5e11f6
- [x] https://github.com/reorx/obsidian-paste-image-rename/blob/b60f2d7cded3b8f1fab9e05c37bc43147af2dc43/src/main.ts#L184
    Input variables needs to be sanitized here.
    - add sanitization for the two values at the source of user input, thus the string should be safe here https://github.com/reorx/obsidian-paste-image-rename/commit/c30db95c8681d6d27bf7bf1f75dd41552f99c730
- [x] https://github.com/reorx/obsidian-paste-image-rename/blob/b60f2d7cded3b8f1fab9e05c37bc43147af2dc43/src/main.ts#L168
    You can get a lot of the properties you're using path for from the file: file.basename, file.extension, file.parent.path. If you absolutely need to use path from nodejs then you must mark your plugin as desktop only in your manifest.json file because it won't work on mobile.
    - Yes, replaced path module occurrences with TFile attributes and self-implemented functions https://github.com/reorx/obsidian-paste-image-rename/commit/a667fa161b1c08bc72453dbf06a4957ffaa8870c
- [x] https://github.com/reorx/obsidian-paste-image-rename/blob/b60f2d7cded3b8f1fab9e05c37bc43147af2dc43/src/main.ts#L142
    This has a memory leak here where you keep the reference of all the modals that has ever existed, even after they were closed.
    - remove the modal on closing it https://github.com/reorx/obsidian-paste-image-rename/commit/c473ce18fbb168a503d5f25172eeaa23f0b6eac9

lishid:
https://github.com/reorx/obsidian-paste-image-rename/blob/7a62a266a2bd5192b49fe326e028acf11d5e11f6/src/main.ts#L148
You can use this API to generate a markdown link: https://github.com/obsidianmd/obsidian-api/blob/fc616fa86533d31b1d68b45887f4c9c46da389bf/obsidian.d.ts#L1090

https://github.com/reorx/obsidian-paste-image-rename/blob/b60f2d7cded3b8f1fab9e05c37bc43147af2dc43/src/main.ts#L184
Here you are sanitizing for filenames but you aren't sanitizing when constructing the regex (special regex characters like `.` or `*`).

reorx:
Updated:
- [use generateMarkdownLink to generate markdown links](https://github.com/reorx/obsidian-paste-image-rename/commit/d1095f6103493089c71570c3547a31f964aa0b25)
- [escape user input string before constructing RegExp](https://github.com/reorx/obsidian-paste-image-rename/commit/48239de4dbddbf1e3e6a467a70c53e44bd204904)

> Here you are sanitizing for filenames but you aren't sanitizing when constructing the regex (special regex characters like . or *).

Can't believe I missed that 😂, but "escape" seems a better word that "sanitize" when it come to the preprocessing for regex string, sanitizing should be idempotent while escaping is mostly not.

lishid:
> Can't believe I missed that 😂, but "escape" seems a better word that "sanitize" when it come to the preprocessing for regex string, sanitizing should be idempotent while escaping is mostly not.

Good point, sorry for the confusion!

PR #887, opened 2022-04-08
Adding the Day Planner with Templates Plugin

amykhar:

## Repo URL

Link to my plugin: https://github.com/amykhar/obsidian-day-planner-with-templates

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/amykhar/obsidian-day-planner-with-templates/blob/07b797d3b5006cb84a9fc60a2fe57dd9a979d826/src/main.ts#L100
I know this is similar in the original day planner but I highly recommend against re-reading the day planner file from disk every 2 seconds and do a full parsing on it. You should instead react to `vault.on('modified')` event and only re-parse the daily note when the changed file is the daily note.

https://github.com/amykhar/obsidian-day-planner-with-templates/blob/07b797d3b5006cb84a9fc60a2fe57dd9a979d826/src/main.ts#L144
Should just use `view.file` here.

https://github.com/amykhar/obsidian-day-planner-with-templates/blob/07b797d3b5006cb84a9fc60a2fe57dd9a979d826/src/main.ts#L39
Settings are loaded from disk as a JSON object, and not as a class object. You should not store or use typed objects here because it will become regular js objects on subsequent loads. Same applies [here](https://github.com/amykhar/obsidian-day-planner-with-templates/blob/07b797d3b5006cb84a9fc60a2fe57dd9a979d826/src/settings.ts#L8). (You should just type them as interfaces instead of classes)

https://github.com/amykhar/obsidian-day-planner-with-templates/blob/07b797d3b5006cb84a9fc60a2fe57dd9a979d826/src/planner-md.ts#L34
Since you're working with the editor already, I suggest using the editor's file and contents instead of pulling together some information from the file (`vault.read`/`vault.write`) and some from the editor (`getCursor().line`). For example, here you can simply do `editor.replaceSelection(contents)` instead of reading and writing the file.

https://github.com/amykhar/obsidian-day-planner-with-templates/blob/07b797d3b5006cb84a9fc60a2fe57dd9a979d826/src/planner-md.ts#L110
Missing null check for the return object of `getActiveViewOfType`, and also should use `view.file` instead of `view.getState().file`.

https://github.com/amykhar/obsidian-day-planner-with-templates/blob/07b797d3b5006cb84a9fc60a2fe57dd9a979d826/src/status-bar.ts#L152
innerHTML causes DOM injection vulnerability: https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml

ericaxu:

PR #888, opened 2022-04-10
Add Notion-Like Tables Plugin

trey-wallis:

## Repo URL

Link to my plugin:
https://github.com/trey-wallis/obsidian-notion-like-tables

## Release Checklist
  - [X]  Windows
  - [X]  macOS
  - [X]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


lishid:
https://github.com/trey-wallis/obsidian-notion-like-tables/blob/a6394b7d1e352da2c32069b975ef8955d9ac2b93/main.ts#L66
editor-change is an event that is fired very frequently (per keystroke) and what you're doing here risks being a bit too frequent and could potentially lag out the typing experience. I would suggest using the `vault.on('modified')` event instead and also put a debounce on it.

https://github.com/trey-wallis/obsidian-notion-like-tables/blob/a6394b7d1e352da2c32069b975ef8955d9ac2b93/main.ts#L176
Yeah... that's really hacky and I wouldn't recommend doing something like that. You risk breaking people's workspaces since leaves are keyed by ID and generating new leaves causes things like undo-close-pane and pane navigation histories to malfunction.

Instead, you should use `MarkdownPreviewView.rerender` to ask for reading mode panes to re-render everything.

https://github.com/trey-wallis/obsidian-notion-like-tables/blob/a6394b7d1e352da2c32069b975ef8955d9ac2b93/src/app/services/dataUtils/index.tsx#L61
Should use `cachedRead` here for better performance.

https://github.com/trey-wallis/obsidian-notion-like-tables/blob/a6394b7d1e352da2c32069b975ef8955d9ac2b93/src/app/services/dataUtils/index.tsx#L111
Missing await.

trey-wallis:
Hi @lishid. Thank you for the feedback! Could you give an example of how to use `MarkdownPreviewView.rerender` with the post processor?

trey-wallis:
In regard to use `editor-change`, I tried using `vault.modify` but it doesn't work with my caching system. Let me explain.

The first time the `registerMarkdownPostProcessor` is called I check the settings for my app data using a CRC-32 hash generated from the element table (it is just the table data hashed together). If the hash doesn't exist, then the app instantiates new data. When a user enter data into my app table, a new CRC-32 hash is produced and that information is saved in the settings file. Then I write the new data to the markdown file. From there, the `registerMarkdownPostProcessor` will be called again (since the file data changed), my app will reload and look for data in the settings file that matches the table element hash (which is the same as what we just saved).

The issue arises when a user edits the source markdown directly. When they type a key, I need to update the settings file with a new hash and data, merging any old data with the new changes. This works fine with the normal editing/reading modes. With live preview, when I use `vault.modify` there can be a delay between when the modify callback is called and when the rendering of my app occurs. For example, if I edit the word "cat" to "cats" and then quickly click outside of the markdown, my table will be rendered. But the modify callback might take around 1-2s to be called. What happens in this case, is that I get the updated data in the new table element, the element gets hashed, but nothing is found in the settings file, so new data is instantiated. From there, the modify callback is called and then it tries to update data that doesn't exist anymore (since I have garbage collection for old data).

I don't really see a way around this besides using the `editor-change` to make sure that the data gets updated in the cache before the markdownPostProcessor reloads

trey-wallis:
@lishid Besides these questions I have, I have made the suggested changes that you indicated. What further actions do I need to take to get my plugin approved?

lishid:
> Also, is there currently a way to tell if the `this.registerMarkdownPostProcessor` is returning an element that is in reading or editing mode (with live preview)? I am thinking it would be good to make my app read only in live preview mode. So if a user clicks on the app, no operations are preformed and Obsidian just shows the markdown.

Ah I see what you mean. I don't think there's any major differences between the two but if you check for `containerEl.matchParent('.cm-preview-code-block')` that should tell you whether it's in Live Preview.

lishid:
Oh and you need to put your plugin at the end of the file.

PR #893, opened 2022-04-13
Update community-plugins.json

ocapraro:

## Repo URL

Link to my plugin: https://github.com/ocapraro/obsidian-math-plus

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/ocapraro/obsidian-math-plus/blob/86b7f8a472166bd34853159ebb6ae4e24e56c4d8/ExcalidrawCanvas.jsx#L9
Don't think this is used.

https://github.com/ocapraro/obsidian-math-plus/blob/86b7f8a472166bd34853159ebb6ae4e24e56c4d8/ExcalidrawCanvas.jsx#L142
When is this interval cleared?

https://github.com/ocapraro/obsidian-math-plus/blob/86b7f8a472166bd34853159ebb6ae4e24e56c4d8/parser.ts#L117
Unused debugging statement?

https://github.com/ocapraro/obsidian-math-plus/blob/86b7f8a472166bd34853159ebb6ae4e24e56c4d8/main.ts#L54
Suggest putting vault and adapter to variables to improve code readability.

ocapraro:
Ok, should be updated now. Before the interval was only being cleared when the user presses the save drawing button, but now it is also cleared if the block ever gets un-rendered.

lishid:
Maybe auto saving every second is a bit excessive? Any way to check whether the drawing has changed before saving?

ocapraro:
Good point, that has now been added!

lishid:
Hmm you're using local storage for it, but you aren't cleaning it up after use... Maybe just store it in an in-memory Record<string, string> object at the root of your file on line 16?

ocapraro:
I'm using local storage so that the data is saved between renders, or if the application is closed. I'm not exactly sure how I could accomplish the same effect by storing it in memory.

lishid:
Hmm I'm not sure exactly what you're trying to do there - are you using localStorage to store the full image data? Wouldn't that not sync to other devices and get lost if one were to clear Obsidian's cache directory? (Also you aren't taking into account that multiple Obsidian vaults open will share the local storage object, so you're potentially overwriting/sharing with another vault if that vault also runs the same plugin.

ocapraro:
Ah, I was not taking that into account, thanks for letting me know! I've now phased out local storage, and am instead saving the data to json files.

lishid:
Looks good so far, I would still recommend avoid reading the file once per second to check if the file has changed. At the very least you should keep an in-memory copy of last-saved state when the canvas is visible to compare with so you don't have to constantly re-read the file from disk.

ocapraro:
Makes sense, just updated it.

lishid:
https://github.com/ocapraro/obsidian-math-plus/blob/84d0f70b93046eefda3faa72f07e3f60aecb8240/main.ts#L148
I still don't think you are properly managing the lifecycle of this react component - you aren't un-mounting it anywhere are you? If that's the case then you are likely still leaking memory via `window.addEventListener("hashchange")` and `setInterval`.

ocapraro:
Good catch, it's unmounting properly now

PR #894, opened 2022-04-14
feat: Add obsidian-link-keeper

luhaifeng666:

## Repo URL

Link to my plugin: https://github.com/luhaifeng666/obsidian-link-keeper

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3895,5 +3895,12 @@
         "author": "Aditya Majethia",
         "description": "Chat View lets you quickly and easily create elegant Chat UIs in your Markdown Files.",
         "repo": "adifyr/obsidian-chat-view"
+    },
+    {
+        "id": "obsidian-link-keeper",
+        "name": "Link Keeper",
+        "author": "Haifeng Lu",
+        "description": "This is a plugin for Obsidian to manage your frequently used links. ",
+        "repo": "luhaifeng666/obsidian-link-keeper"
     }
 ]


lishid:
https://github.com/luhaifeng666/obsidian-link-keeper/blob/9ac7b74bde4d7816af14833902ec1829aad90938/main.ts#L15
Is there any reason why your plugin doesn't just save its links into the settings object? You're only saving a few name/url pairs anyway?

https://github.com/luhaifeng666/obsidian-link-keeper/blob/9ac7b74bde4d7816af14833902ec1829aad90938/main.ts#L103
Should use `app.commands.executeCommandById(id)` instead.

luhaifeng666:
> https://github.com/luhaifeng666/obsidian-link-keeper/blob/9ac7b74bde4d7816af14833902ec1829aad90938/main.ts#L15 Is there any reason why your plugin doesn't just save its links into the settings object? You're only saving a few name/url pairs anyway?
> 
> https://github.com/luhaifeng666/obsidian-link-keeper/blob/9ac7b74bde4d7816af14833902ec1829aad90938/main.ts#L103 Should use `app.commands.executeCommandById(id)` instead.

Because I wanna use it with [efficient-tools](https://www.npmjs.com/package/efficient-tools). They can share a set of configurations.

luhaifeng666:
> https://github.com/luhaifeng666/obsidian-link-keeper/blob/9ac7b74bde4d7816af14833902ec1829aad90938/main.ts#L15 Is there any reason why your plugin doesn't just save its links into the settings object? You're only saving a few name/url pairs anyway?
> 
> https://github.com/luhaifeng666/obsidian-link-keeper/blob/9ac7b74bde4d7816af14833902ec1829aad90938/main.ts#L103 Should use `app.commands.executeCommandById(id)` instead.

Because I wanna use it with [efficient-tools](https://www.npmjs.com/package/efficient-tools). They can share a set of configurations.

luhaifeng666:
> https://github.com/luhaifeng666/obsidian-link-keeper/blob/9ac7b74bde4d7816af14833902ec1829aad90938/main.ts#L15 Is there any reason why your plugin doesn't just save its links into the settings object? You're only saving a few name/url pairs anyway?
> 
> https://github.com/luhaifeng666/obsidian-link-keeper/blob/9ac7b74bde4d7816af14833902ec1829aad90938/main.ts#L103 Should use `app.commands.executeCommandById(id)` instead.

Thanks for you support~

lishid:
I think your plugin is rather unconventional, and it seems more like a plugin for your personal use case? If that's the case then you don't have to submit it to the plugin store to be able to use it yourself.

The reason why I say it's unconventional is that your plugin is very simple (manage an array of links), but still relies on external data sources (file in user's home folder). If you want other users to use this plugin then you should at least explain that
your plugin stores its data outside of the app and in the user's home directory, and is meant to be used with your `efficient-tools` project.

Otherwise, I don't see any reason why a user would use this instead of a favorites.md file with a bunch of links.

PR #896, opened 2022-04-14
Add drag-n-drop plugin

artem-barmin:

## Repo URL

Link to my plugin: https://github.com/artem-barmin/obsidian-block-drag-n-drop

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3895,5 +3895,12 @@
         "author": "Aditya Majethia",
         "description": "Chat View lets you quickly and easily create elegant Chat UIs in your Markdown Files.",
         "repo": "adifyr/obsidian-chat-view"
+    },
+    {
+      "id": "obsidian-drag-n-drop-plugin",
+      "name": "Drag-n-Drop for blocks",
+      "author": "Artem Barmin",
+      "description": "Allow moving/copying/and creation embeds for blocks with drag-n-drop just like Logseq or Roam",
+      "repo": "artem-barmin/obsidian-block-drag-n-drop"
     }
 ]


lishid:
https://github.com/artem-barmin/obsidian-block-drag-n-drop/blob/2412983a5b042d55b4c91de6a6399a2fa66e9174/main.ts#L48
Wouldn't this grab the content of potentially a different pane? You should probably run the querySelector in the view's containerEl.

https://github.com/artem-barmin/obsidian-block-drag-n-drop/blob/2412983a5b042d55b4c91de6a6399a2fa66e9174/main.ts#L163
Not sure why you're disabling all the type checks, but you're missing all the typing information here and things like `event.target.cmView.editorView` are very arbitrary. Definitely recommend turning on typechecks to help you properly get things that are exposed through the API so you don't end up depending on internal references that are likely to change in the future.

(I bet you'll end up with a lot of errors when you enable type checking...)

https://github.com/artem-barmin/obsidian-block-drag-n-drop/blob/2412983a5b042d55b4c91de6a6399a2fa66e9174/main.ts#L349
You can combine these 3 into a single `registerEditorExtension([ first, second, third ])`.

artem-barmin:
1 and 3 is not a problem - will fix it

2 - enabling typing is also not a problem, will fix it

But things like `event.target.cmView.editorView` for my point of view is necessary, or please suggest my way how to get target editor from `drop` or `dragenter` event. Obsidian seems don't have API like: `DOM element(event.target)` → `Pane(or Leaf in Obsidian terminology)`

artem-barmin:
@lishid added type declarations where possible. Only places that left: direct access to Codemirror6 editor, because `Editor` wrapper from Obsidian API does not provide needed functionality like: `domAtPos`. I will be able to fix remaining in case when Obsidian API will be extended OR Obsidian will add method to get underlying Codemirror instance

artem-barmin:
Regarding this https://github.com/artem-barmin/obsidian-block-drag-n-drop/blob/2412983a5b042d55b4c91de6a6399a2fa66e9174/main.ts#L48 - it does not break anything because I'm copying tab-size property that are the same across different panes, so any pane is ok

artem-barmin:
I put `ts-ignore` in 6 places to be able to find them in case when API will change

lishid:
You're right, I don't see a better way for now without directly accessing the editor. I guess you should mention (or test for) that your plugin only supports the new editor (as opposed to the Legacy editor).

artem-barmin:
Thank you! Added note to README

PR #899, opened 2022-04-17
add Obsidian Columns plugin

tnichols217:

## Repo URL

Link to my plugin:
https://github.com/tnichols217/obsidian-columns

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
I'm not too sure I understood what `md` is for - maybe if you included some examples in the readme it would help.

lishid:
I also don't think it's a great idea to use `md` as a code block since users may be using it for markdown examples and now they get rendered instead of staying a source block.

tnichols217:
> I also don't think it's a great idea to use `md` as a code block since users may be using it for markdown examples and now they get rendered instead of staying a source block.

ah, i could just change it to col-md or something, youre right.

PR #900, opened 2022-04-17
Added Notion API To Plugins List

LastKnightCoder:

## Repo URL

Link to my plugin: https://github.com/LastKnightCoder/obsidian-notion

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3895,5 +3895,12 @@
         "author": "Aditya Majethia",
         "description": "Chat View lets you quickly and easily create elegant Chat UIs in your Markdown Files.",
         "repo": "adifyr/obsidian-chat-view"
+    },
+    {
+        "id": "obsidian-notion-plugin",
+        "name": "Notion API",
+        "author": "lastknightcoder",
+        "description": "Using Notion API in Obsidian",
+        "repo": "LastKnightCoder/obsidian-notion"
     }
 ]


lishid:
https://github.com/LastKnightCoder/obsidian-notion/blob/959933a34f7b95e6fd01ba0bfee7da33d9b568e2/main.ts#L27
Should use `await` here instead of `.then` and `.catch`

https://github.com/LastKnightCoder/obsidian-notion#introduction
I think given that your plugin inserts embeds from Notion, it should be named "obsidian-notion-embed" or something like that. "Obsidian Notion" is such a generic name, it doesn't really describe your plugin well.

LastKnightCoder:
> https://github.com/LastKnightCoder/obsidian-notion/blob/959933a34f7b95e6fd01ba0bfee7da33d9b568e2/main.ts#L27 Should use `await` here instead of `.then` and `.catch`
> 
> https://github.com/LastKnightCoder/obsidian-notion#introduction I think given that your plugin inserts embeds from Notion, it should be named "obsidian-notion-embed" or something like that. "Obsidian Notion" is such a generic name, it doesn't really describe your plugin well.

PR #911, opened 2022-04-20
Add entry for Obsidian Variables plugin

jffaust:

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3997,5 +3997,12 @@
         "author": "twibiral",
         "description": "Allows to execute code snippets within a note.",
         "repo": "twibiral/obsidian-execute-code"
+    },
+    {
+        "id": "obsidian-variables",
+        "name": "Variables",
+        "author": "Jean-Frédéric Faust",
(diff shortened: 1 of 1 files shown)


lishid:
https://github.com/jffaust/obsidian-variables/blob/01c5e5af061c3d8cea1f60c1a02e4e1568342396/src/main.ts#L36

Yeah no that's just a bad idea. You will cause a lot of issues by doing this.

1. Any elements that have event listeners (for example, the copy code block button from code blocks) will be re-generated and lose all of the event listeners. This will inevitably break a lot of core and other plugin functionality.
2. Any blocks created by plugins will be mis-referenced (because you re-created new ones here using innerHTML) such that updates to them will stop working (things like dataview tables, etc).
3. You still run the risk of DOM injection because you aren't sanitizing the resulting string (only sanitizing the partial replacement value is not sufficient).
4. This is terrible for performance. Not only are you replacing (re-parsing the HTML) many times depending on how many variables were set in your plugin, but just re-parsing the HTML for every single rendered block is a bad practice.

PR #912, opened 2022-04-20
Add Obsidian-Reset-Checkboxes Plugin

branwall:

## Repo URL

[Link to my plugin](https://github.com/branwall/obsidian-reset-checkboxes)

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3997,5 +3997,12 @@
         "author": "twibiral",
         "description": "Allows to execute code snippets within a note.",
         "repo": "twibiral/obsidian-execute-code"
-    }
+    },
+    {
+        "id": "obsidian-reset-checkboxes",
+        "name": "Reset Checkboxes",
+        "author": "Branwall",
+        "description": "Adds commands to check or reset all checkboxes in the open file.  Useful for checklists.",
+        "repo": "branwall/obsidian-reset-checkboxes"
+    }  
 ]


lishid:
https://github.com/branwall/obsidian-reset-checkboxes/blob/f3a10dd1f7dc4b8442dba2cd6e5a66230a7faca5/main.ts#L13
Ok here I would suggest running your code on a line-by-line basis instead of using `getValue()` because when you replace the whole file's content with `setValue()`, it causes line and position associations (like folding, etc) to be completely broken, which means when you do this, the user will lose all of their editor folding. Other plugins that rely on text positions and markers will also lose positions, and may or may not be able to deal with that gracefully. Lastly this is bad for performance because it forces the editor to re-parse the entire file.

What you can do is to get the number of lines, iterate through each line, and issue a `replaceRange` only when you find a match. The downside with doing that is that the undo history will now undo each line at a time.

For a perfect implementation, you should add your changed ranges to an array and use the editor API "transaction" with an array of `changes`, which will performs all changes at the same time in a single undo commit:
https://github.com/obsidianmd/obsidian-api/blob/036708710c4a4b652d8166c5929d5ba1ffb7fb91/obsidian.d.ts#L836

ericaxu:

PR #914, opened 2022-04-21
Add "Plugins Galore" to Community Plugins

dylanpizzo:

## Repo URL

Link to my plugin: https://github.com/dylanpizzo/obsidian-plugins-galore

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/dylanpizzo/obsidian-plugins-galore/blob/8e485bc32a122407041a8d714aeebbec0d228788/src/util/pluginActions.ts#L55

I would recommend against storing files in other plugin folders, instead you can just save that data inside your plugin's data file (how most other plugins store their settings). This would allow users who decide to stop using your plugin to simply delete your plugin's folder and not have to manually clean up each plugin for a `.galore` file.

dylanpizzo:
I think that didn't occur to me because, as you said, most plugins use that for settings :smiley:. Great suggestion! Done.

PR #915, opened 2022-04-21
Add Auto Card Link

nekoshita:

## Repo URL

Link to my plugin: https://github.com/nekoshita/obsidian-auto-card-link

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
In your readme file, you should mention which online service you're using to generate the link details, to avoid being criticized by privacy-concerned users.

https://github.com/nekoshita/obsidian-auto-card-link/blob/6aa33c95fc768a4bc05779c5f60550be456eb2f5/src/main.ts#L146
The editor object is passed to the `editor-paste` event, so you shouldn't have to get it manually.

https://github.com/nekoshita/obsidian-auto-card-link/blob/6aa33c95fc768a4bc05779c5f60550be456eb2f5/src/main.ts#L157
You should also check whether there are files in the clipboard - some browsers will paste images with the URL in the `text/plain` field but you're only catching it if the URL has an image extension.

https://github.com/nekoshita/obsidian-auto-card-link/blob/6aa33c95fc768a4bc05779c5f60550be456eb2f5/src/settings.ts#L15
You shouldn't put objects here along with your settings, because the config object is serialized to JSON and deserialized from string. Why not just store them as a global constant like `export const MY_REGEX = /..../;`?

https://github.com/nekoshita/obsidian-auto-card-link/blob/6aa33c95fc768a4bc05779c5f60550be456eb2f5/src/code_block_processor.ts#L2
Should use `parseYAML` from `obsidian` instead of bringing your own copy of a YAML parser.

https://github.com/nekoshita/obsidian-auto-card-link/blob/6aa33c95fc768a4bc05779c5f60550be456eb2f5/src/code_block_generator.ts#L67
Consider using `requestUrl` from `obsidian`.

nekoshita:
@lishid Thanks you very much for the review!
I fixed all the problems.

> In your readme file, you should mention which online service you're using to generate the link details, to avoid being criticized by privacy-concerned users.

https://github.com/nekoshita/obsidian-auto-card-link/pull/20
This plugin does not use online service any more.


> The editor object is passed to the editor-paste event, so you shouldn't have to get it manually.

https://github.com/nekoshita/obsidian-auto-card-link/pull/19/commits/89473de595ecd4431358d2930fbf2c26bdc7ab80

> You should also check whether there are files in the clipboard - some browsers will paste images with the URL in the text/plain field but you're only catching it if the URL has an image extension.

https://github.com/nekoshita/obsidian-auto-card-link/commit/3baca5cb49fc2caf6148e0095a1182d7ac160c0d

> You shouldn't put objects here along with your settings, because the config object is serialized to JSON and deserialized from string. Why not just store them as a global constant like export const MY_REGEX = /..../;?

https://github.com/nekoshita/obsidian-auto-card-link/pull/19/commits/7ca3305fef3b9d51a388a69894107f2e37713246

> Should use parseYAML from obsidian instead of bringing your own copy of a YAML parser.

https://github.com/nekoshita/obsidian-auto-card-link/pull/19/commits/e2a32e7892106ee076f1d0cd2eebe11f81f60eca

> Consider using requestUrl from obsidian.

https://github.com/nekoshita/obsidian-auto-card-link/pull/19/commits/2c586c4fde27c6b85646dce4141afc460266a617

PR #918, opened 2022-04-23
added autoMOC to community plugins

dalcantara7:

## Repo URL

Link to my plugin: https://github.com/dalcantara7/obsidian-auto-moc

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/dalcantara7/obsidian-auto-moc/blob/2ccb1b0ea26c21c17bb972eecccfc25c74af320f/main.ts#L65
This isn't necessary.

https://github.com/dalcantara7/obsidian-auto-moc/blob/2ccb1b0ea26c21c17bb972eecccfc25c74af320f/main.ts#L38
You should just use an `editorCallback` instead.

https://github.com/dalcantara7/obsidian-auto-moc/blob/2ccb1b0ea26c21c17bb972eecccfc25c74af320f/main.ts#L87
`file` might be null and could also be an attachment image/video/pdf. Should check for that.

https://github.com/dalcantara7/obsidian-auto-moc/blob/2ccb1b0ea26c21c17bb972eecccfc25c74af320f/main.ts#L98
That's terrible. Don't do that. You can get global linking information from `metadataCache.resolvedLinks` to help you avoid doing this really hacky thing.

https://github.com/dalcantara7/obsidian-auto-moc/blob/2ccb1b0ea26c21c17bb972eecccfc25c74af320f/main.ts#L129
You should use `app.fileManager.generateMarkdownLink` to create links with the user's preferred format. Should also check that the current view is not null first.
Also you're not passing `this` properly. You should pass the app property through your functions, or put your these functions in your plugin's class instead of being naked outside.

dalcantara7:
I have resolved the issues mentioned and pushed new code to the master branch. Do I need to create a new release and submit a new PR?

lishid:
Keep using this PR is fine, I will fix any merge conflict at the end when we're ready to merge!

https://github.com/dalcantara7/obsidian-auto-moc/blob/2cf2348a7d1128037a7501428ee14b09ebd0623c/main.ts#L58
Here I think you're passing this parameter wrong. `sourcePath` means from where this link resolves, and is used to properly determine the relative path to generate. You're passing the linked file instead of the source file.

Also, see this tip: https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-iterating-all-files-to-find-a-file-by-its-path

dalcantara7:
I resolved the sourcePath issue. I also attempted to modify the search code using the getAbstractFile method from the tip, but that documentation seems to be outdated. The closest function matching that I was able to find is getAbstractFileByPath() which returns TAbstractFile. generateMarkdownLink() expects a TFile, not TAbstractFile, but it seems to work even passing a TAbstractFile in. 

Am I missing something?

lishid:
When you get a TAbstractFile, it can either be a TFolder or a TFile (makes sense, because you're giving the API a path and looking for something at that path).

What you do with TypeScript is that if you check for `if(file instanceof TFile)`, then inside the if condition, it will automatically cast the variable to a `TFile`.

```ts
let file = this.app.vault.getAbstractFileByPath(path);
if (file instanceof TFile) {
    this.app.fileManager.generateMarkdownLink(file, sourcePath);
}
```

PS. Thanks for catching the error in the docs, I have updated it to say `getAbstractFileByPath` instead.

dalcantara7:
Thanks for the explanation. It might be good to add that explanation to the docs (esp for first time Obsidian plugin and Typescript devs). Specifically that inside the if check, it casts it to TFile. Would make it a little more clear.

I have updated the check in my code and the changes are on master.

lishid:
https://github.com/dalcantara7/obsidian-auto-moc/blob/3c7ff24b796c1ded4ce6ad80f04fe3ba124bfa76/main.ts#L65
Here I don't think you should fail and return - this would interrupt the for loop. Instead I would suggest just ignoring it altogether and move on.

https://github.com/dalcantara7/obsidian-auto-moc/blob/3c7ff24b796c1ded4ce6ad80f04fe3ba124bfa76/main.ts#L55
Here I recommend putting this outside your for loop to avoid running this multiple times. You also need to do a null check on the view.

Also you should be able to get the active file path directly using `view.path` instead of `getActiveFile().path`.

Actually, looking at your code as a whole, you should probably run `getActiveViewOfType` in `runAutoMOC`, and pass the `editor` to your `addMissingLinks` function.

lishid:
> It might be good to add that explanation to the docs (esp for first time Obsidian plugin and Typescript devs).

Will do, thanks for the feedback!

dalcantara7:
Good catch on the redundant get in the for loop. Thanks.

I have made the changes you suggested except for getting the path from "view". view is a variable I declared that came from this.app.workspace, but that doesn't expose a path (as far as I could tell). I left in getActiveFile() unless there is some other way around it.

lishid:
MarkdownView exposes a `file` property (via one of its parent classes), which is the markdown file that is open. `getActiveFile` literally just gives you the `view.file`.

lishid:
Ok code looks good. Before I merge though you need to move your plugin to the end of the list.

PR #919, opened 2022-04-24
Add Calibre Plugin

caronchen:

## Repo URL

Link to my plugin: https://github.com/caronchen/obsidian-calibre-plugin

## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


lishid:
https://github.com/caronchen/obsidian-calibre-plugin/blob/4b4024120c000374d59f61b0f2470ef6ace4a286/src/main.tsx#L49

Here you should use `workspace.splitActiveLeaf(direction)` instead because `activeLeaf` can be null. You should also keep a reference to this leaf to use with `revealLeaf` later.

https://github.com/caronchen/obsidian-calibre-plugin/blob/4b4024120c000374d59f61b0f2470ef6ace4a286/src/CalibreView.tsx#L28
Should set the sandbox property of iframes - By default, obsidian sets the following:
`allow-forms allow-presentation allow-same-origin allow-scripts allow-modals`

https://github.com/caronchen/obsidian-calibre-plugin/blob/4b4024120c000374d59f61b0f2470ef6ace4a286/src/CalibreView.tsx#L23
Honestly I don't think it makes sense to include ReactDOM just to render a single iframe here and a div later. You should just use Obsidian's DOM shortcuts:

```ts
let iframe = this.contentEl.createEl('iframe');
iframe.setAttribute('sandbox', 'allow-forms allow-presentation allow-same-origin allow-scripts allow-modals');
iframe.src = this.settings.address;
```

caronchen:
> https://github.com/caronchen/obsidian-calibre-plugin/blob/4b4024120c000374d59f61b0f2470ef6ace4a286/src/main.tsx#L49
> 
> Here you should use `workspace.splitActiveLeaf(direction)` instead because `activeLeaf` can be null. You should also keep a reference to this leaf to use with `revealLeaf` later.
> 
> https://github.com/caronchen/obsidian-calibre-plugin/blob/4b4024120c000374d59f61b0f2470ef6ace4a286/src/CalibreView.tsx#L28 Should set the sandbox property of iframes - By default, obsidian sets the following: `allow-forms allow-presentation allow-same-origin allow-scripts allow-modals`
> 
> https://github.com/caronchen/obsidian-calibre-plugin/blob/4b4024120c000374d59f61b0f2470ef6ace4a286/src/CalibreView.tsx#L23 Honestly I don't think it makes sense to include ReactDOM just to render a single iframe here and a div later. You should just use Obsidian's DOM shortcuts:
> 
> ```ts
> let iframe = this.contentEl.createEl('iframe');
> iframe.setAttribute('sandbox', 'allow-forms allow-presentation allow-same-origin allow-scripts allow-modals');
> iframe.src = this.settings.address;
> ```

Thanks for your CR, fixed now.
https://github.com/caronchen/obsidian-calibre-plugin/commit/6500d9cb66a62654ae24bdd7477328c7b4e8103d

PR #920, opened 2022-04-24
New plugin: Omnisearch

scambier:

## Repo URL

Link to my plugin: https://github.com/scambier/obsidian-omnisearch

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/scambier/obsidian-omnisearch/blob/c82a4d18add9d847c0a3b30a949c965f8bf9e014/src/main.ts#L24
You should just use `editorCallback` here.

https://github.com/scambier/obsidian-omnisearch/blob/c82a4d18add9d847c0a3b30a949c965f8bf9e014/src/modals.ts#L19
I would recommend using the `scope` object on the modal for these for consistency with obsidian's internals. I also don't feel like these should go to a global event bus? IMO it should work within your modal's code.

lishid:
Oh yeah forgot to check, your plugin ID contains a `.`, while it's technically not disallowed I would suggest for consistency to use `-` instead.

scambier:
Hi @lishid, thanks for that code review. I've fixed the issues, except for the event bus. I agree that it's a bit of an overkill solution (though the event bus itself is [super minimal](https://github.com/scambier/obsidian-omnisearch/blob/627c7f33d62430f653eb4b13d3a51273b68a1ad6/src/event-bus.ts#L3)), but I find it easier to manage events this way.

It allows me to cleanly (de)activate some events depending on the context, like when you go from the Vault modal to the Infile modal: both modals are actually stacked, I can disable the Vault modal events, and reactivate them when you go back with escape ([source ref](https://github.com/scambier/obsidian-omnisearch/blob/627c7f33d62430f653eb4b13d3a51273b68a1ad6/src/components/ModalInFile.svelte#L30-L43)).

Fixes:
- https://github.com/scambier/obsidian-omnisearch/commit/e7b552dc922ca627a5423b41c9f24a06b692fad7
- https://github.com/scambier/obsidian-omnisearch/commit/627c7f33d62430f653eb4b13d3a51273b68a1ad6

lishid:
Sounds good!

scambier:
@lishid thanks for the approval :) There is just an issue with the `repo` field: it should be `scambier/obsidian-omnisearch`, and **not** `scambier/obsidian-releases`. Sorry for the mistake!

PR #921, opened 2022-04-25
New Plugin: Folder Index

turulix:

## Repo URL

Link to my plugin: https://github.com/turulix/obsidian-folder-index

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/turulix/obsidian-folder-index/blob/8d9bfadc01dbb3c252c32326f6a5bc70750799b6/src/modules/FolderNoteModule.ts#L27
Here you should check whether `evt.target instanceof HTMLElement` and if not just return.

https://github.com/turulix/obsidian-folder-index/blob/8d9bfadc01dbb3c252c32326f6a5bc70750799b6/src/main.ts#L34
You should unload the `folderNodeModule` in your onunload.

https://github.com/turulix/obsidian-folder-index/blob/8d9bfadc01dbb3c252c32326f6a5bc70750799b6/src/modules/FolderNoteModule.ts#L82
activeLeaf might be null or it might be pinned, instead you should use `getLeaf()`.

https://github.com/turulix/obsidian-folder-index/blob/8d9bfadc01dbb3c252c32326f6a5bc70750799b6/src/modules/FolderNoteModule.ts#L109
Be careful here because this event is fired for every file and folder inside a nesting structure when you rename the top-level folder.

https://github.com/turulix/obsidian-folder-index/blob/8d9bfadc01dbb3c252c32326f6a5bc70750799b6/src/main.ts#L35
Good effort on reverse engineering the graph module! I think given the way you're overwriting the graph could be problematic with future updates though.

I would suggest having the graph component disabled by default and making the setting a global killswitch to not even load the graph related things.

turulix:
Thank  you for the suggestions. I will fix them when i have time  :)

turulix:
The problems should be fixed now :)

lishid:
https://github.com/turulix/obsidian-folder-index/blob/8962faf7b72254dc558a3e27e8ec9a3594f6d661/src/main.ts#L65
Missing a null check for graphManipulator here. Otherwise everything else looks good!

lishid:
Oh before I merge, you need to move your plugin to be the last item in the list.

turulix:
All done :)

PR #922, opened 2022-04-26
Add 'Paste image Png to Jpeg' Plugin

musug:

## Repo URL https://github.com/musug/obsidian-paste-png-to-jpeg

Link to my plugin:

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
- [x]My GitHub release contains all required files
  - [x] `main.js`
  - [x] `manifest.json`


claremacrae:
Hello. Someone will probably ask you to edit the description to fill in the Link to your plugin…

musug:
> Hello. Someone will probably ask you to edit the description to fill in the Link to your plugin…

Sorry, I forgot to add it, now it's done

lishid:
https://github.com/musug/obsidian-paste-png-to-jpeg/blob/1c23a5179ec0df3957c1360bd07f91a0c018c48a/src/main.ts#L48
Use `Date.now()`

https://github.com/musug/obsidian-paste-png-to-jpeg/blob/1c23a5179ec0df3957c1360bd07f91a0c018c48a/src/main.ts#L119
Here you can use `file.basename` and `file.extension`.

https://github.com/musug/obsidian-paste-png-to-jpeg/blob/1c23a5179ec0df3957c1360bd07f91a0c018c48a/src/main.ts#L140
Don't need to use the `.then`.

https://github.com/musug/obsidian-paste-png-to-jpeg/blob/1c23a5179ec0df3957c1360bd07f91a0c018c48a/src/main.ts#L198
You should use `generateMarkdownLink` regardless because it takes care of properly resolving your image from the file it's embedding from.

https://github.com/musug/obsidian-paste-png-to-jpeg/blob/1c23a5179ec0df3957c1360bd07f91a0c018c48a/src/main.ts#L172
If you need an editor instance, you should check for it earlier, and pass it along. You should also use the same sourcePath as `view.file.path`.

https://github.com/musug/obsidian-paste-png-to-jpeg/blob/1c23a5179ec0df3957c1360bd07f91a0c018c48a/src/utils.ts#L168
I think you should just use canvas for this instead of having to import the library. The library itself just uses canvas to do this anyway: https://github.com/fengyuanchen/compressorjs/blob/21c2bb19d6c39723c812953391caad49661f4a66/src/index.js#L294

You could even reuse the same canvas to do both conversion and compression. I also think you should be able to draw the ArrayBuffer directly onto the canvas without having to convert it into a Blob, then to a data url.

musug:
Hi @lishid ,
Thanks for the review 

>https://github.com/musug/obsidian-paste-png-to-jpeg/blob/1c23a5179ec0df3957c1360bd07f91a0c018c48a/src/main.ts#L198
You should use generateMarkdownLink regardless because it takes care of properly resolving your image from the file it's embedding from.

Here we can't get the correct return value from 'this.app.fileManager.generateMarkdownLink' function, for example, the correct value should be ! [[image/test-1.jpeg]] , but the return value is ! [[test-1.jpeg]], the previous image/ is missing
I suspect that the underlying implementation is based on name, not path, so I have to start with `! [[${ file.path}]]` to achieve
https://github.com/musug/obsidian-paste-png-to-jpeg/blob/a8c34910a524abf29624fe42f40586b96b722bd7/src/main.ts#L121

Other than that, everything else has been taken care of

Thanks again

lishid:
> Here we can't get the correct return value from 'this.app.fileManager.generateMarkdownLink' function, for example, the correct value should be ! [[image/test-1.jpeg]] , but the return value is ! [[test-1.jpeg]], the previous image/ is missing

Actually, `generateMarkdownLink` generates `[[test-1.jpeg]]` by default because the link style is "shortest unique path" by default. If you change your link style to absolute path then it would generate the long form. This is the link that gets generated if the user uses the auto-complete from `[[`.

musug:
@lishid 
> Actually, generates by default because the link style is "shortest unique path" by default. If you change your link style to absolute path then it would generate the long form. This is the link that gets generated if the user uses the auto-complete from . (All this to say that it's done on purpose)`generateMarkdownLink``[[test-1.jpeg]]``[[`

There is a bug, if there are multiple subdirectories, image/test.jpge, image/1/test.jpge, image/2/test.jpeg, there is no way to distinguish which image is which, I think, a parameter should be added here to decide whether to return "shortest unique path"

lishid:
> There is a bug, if there are multiple subdirectories, image/test.jpge, image/1/test.jpge, image/2/test.jpeg, there is no way to distinguish which image is which, I think, a parameter should be added here to decide whether to return "shortest unique path"

If there are multiple files with the same name then the engine would automatically generate absolute paths to distinguish them. Either way, this is a user preference thing, and if you always want absolute paths from `generateMarkdownLink` you can totally do that by setting your personal link format to "Absolute path". I understand that you may not like it as it introduces possibility of ambiguity, and this is definitely something we are aware of, but the reason why we offer the ability to use the "shortest unique path" is because shorter links look nice and less cluttered.

Just to make sure we are on the same page, the shortest unique path algorithm here means:
- If there's a single file with this name then use the filename itself.
- If there's multiple files with the same name then use the absolute path.

Of course, all the link resolution system in Obsidian is designed to properly resolve these links, and in case of conflict (say you rename a file and causes a global duplicate), Obsidian will automatically update the links to absolute to avoid ambiguity.

musug:
@lishid
 It's true what you said, I've modified it
Thank you for solving my doubts

lishid:
Were you able to take a look at this suggestion? It seems like you removed the compression part of the code but you are still importing Compressorjs.

> https://github.com/musug/obsidian-paste-png-to-jpeg/blob/1c23a5179ec0df3957c1360bd07f91a0c018c48a/src/utils.ts#L168
I think you should just use canvas for this instead of having to import the library. The library itself just uses canvas to do this anyway: https://github.com/fengyuanchen/compressorjs/blob/21c2bb19d6c39723c812953391caad49661f4a66/src/index.js#L294
> 
> You could even reuse the same canvas to do both conversion and compression. I also think you should be able to draw the ArrayBuffer directly onto the canvas without having to convert it into a Blob, then to a data url.

lishid:
Everything else looks good!

musug:
@lishid Changed

PR #924, opened 2022-04-27
Update community-plugins.json

YTolun:

## Repo URL

Link to my plugin: https://github.com/YTolun/obsidian-filename-emoji-remover

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lishid:
https://github.com/YTolun/obsidian-filename-emoji-remover/blob/e4cd31b289cdff3b317a2be49de5af00ca57e7e6/src/main.ts#L17
Here you should check that `file instanceof TFile` before proceeding, otherwise you're trying to process folders as well.

When you do typecheck it as `TFile`, you can directly use its `file.basename` and `file.extension` instead of having to split like that which doesn't work if the file has additional dots in the file name.

https://github.com/YTolun/obsidian-filename-emoji-remover/blob/e4cd31b289cdff3b317a2be49de5af00ca57e7e6/src/main.ts#L24
Should generate this regex once and reuse it throughout.

https://github.com/YTolun/obsidian-filename-emoji-remover/blob/e4cd31b289cdff3b317a2be49de5af00ca57e7e6/src/main.ts#L28
Here isn't it best to just use `filename = filename.replace(regex, '');` instead of looping the match results? I would even take this one step further and perform the replacement without testing for a match first, and check if the new file name is identical to the old one to decide whether to rename it.

https://github.com/YTolun/obsidian-filename-emoji-remover/blob/e4cd31b289cdff3b317a2be49de5af00ca57e7e6/src/main.ts#L22
I feel like it would be a bit more efficient to just use `lastIndexOf('/')`. You can also use `file.parent.path`.

https://github.com/YTolun/obsidian-filename-emoji-remover/blob/e4cd31b289cdff3b317a2be49de5af00ca57e7e6/src/main.ts#L59
Here I would suggest running the `off` routine regardless of `toggle` first, this would allow you to prevent a situation where the function gets called twice with `true` and you register two identical events.

YTolun:
Hello, thank you for your very insightful and just as helpful review. I have improved the code quality based on your feedback and created a new release in the repo. Do I need to do anything else?

liamcain:
I noticed a couple of small things but otherwise it looks good!

---

https://github.com/YTolun/obsidian-filename-emoji-remover/blob/ef6a35f4fe3b691aecbb4e6e9740eeecb94338c2/src/main.ts#L54 I don't think you need the `await` inside the map since you're mapping the files to Promises and the outer `Promise.all` will wait for each of those promises.

https://github.com/YTolun/obsidian-filename-emoji-remover/blob/ef6a35f4fe3b691aecbb4e6e9740eeecb94338c2/src/settings.ts#L27 you might want to add a loading state or maybe show a Notice() after renaming all the files, just in case this takes a while to rename all the files.

liamcain (approved):

community-plugins.json
```diff
@@ -4091,5 +4091,12 @@
...
         "branch": "main"
-    }
+    },
+    {
+        "id": "obsidian-filename-emoji-remover",
+        "name": "Filename Emoji Remover",
+        "author": "Yüksel Tolun",
+        "description": "This is a simple plugin to automatically remove emojis from filenames. Main purpose is to get rid of Dropbox sync issues for Readwise imported content",
+        "repo": "YTolun/obsidian-filename-emoji-remover"
+    },
```
liamcain, line 4101:
```suggestion
    }
```

liamcain (changes requested):

YTolun:
Thanks for the suggestion, I committed the trailing comma change. I'll be implementing the other improvements over the weekend. Cheers!

lishid:
https://github.com/YTolun/obsidian-filename-emoji-remover/blob/ef6a35f4fe3b691aecbb4e6e9740eeecb94338c2/src/main.ts#L23
Here you can write it better without creating the intermediate array:
```ts
const {basename: filename, basename: oldFilename, extension: FileExtension} = file;
```

PR #925, opened 2022-04-28
Add a new plugin: Styled Container

LastKnightCoder:

## Repo URL

Link to my plugin: https://github.com/LastKnightCoder/obsidian-styled-container

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


LastKnightCoder:
> This plugins seems to do the same thing as
> 
> * the default [Callouts](https://help.obsidian.md/How+to/Use+callouts)
> * as well as the [Admonition](https://github.com/valentine195/obsidian-admonition) plugin
> 
> Please clarify what your plugin does differently.
> 
> You are also adding your other plugin to the list again, resulting in the merge conflict.

They are similar in function, the biggest difference may be in the style, I think this style of mine is more modern and can provide another option for the user.

Should I merge master first and create a new pull request? I don't know much about it.

lishid:
> They are similar in function, the biggest difference may be in the style, I think this style of mine is more modern and can provide another option for the user.

Wouldn't it be much easier to apply CSS to the native callouts instead of doing all this work to create new code blocks?

The "Admonition" plugin used to do exactly what you do with code blocks, and there's a number of downsides from doing it that way:
- You lose spellchecking ability when editing because code blocks aren't spellchecked.
- `[[wikilinks]]` are not considered part of the document and so backlinks won't work, and the links won't be automatically updated when you rename the destination files causing broken links.
- It's more difficult to nest fenced code blocks in another compared to callouts (you have to add one more tick to every parent block).
- When using other apps, fenced code blocks fall back to raw markdown whereas callouts fall back to a quote block with parsed and formatted markdown.

LastKnightCoder:
> > They are similar in function, the biggest difference may be in the style, I think this style of mine is more modern and can provide another option for the user.
> 
> Wouldn't it be much easier to apply CSS to the native callouts instead of doing all this work to create new code blocks?
> 
> The "Admonition" plugin used to do exactly what you do with code blocks, and there's a number of downsides from doing it that way:
> 
> * You lose spellchecking ability when editing because code blocks aren't spellchecked.
> * `[[wikilinks]]` are not considered part of the document and so backlinks won't work, and the links won't be automatically updated when you rename the destination files causing broken links.
> * It's more difficult to nest fenced code blocks in another compared to callouts (you have to add one more tick to every parent block).
> * When using other apps, fenced code blocks fall back to raw markdown whereas callouts fall back to a quote block with parsed and formatted markdown.

Thank you for your comments!

I wrote this plugin because to implement such a style requires writing HTML code and the ability to not use Markdown in HTML code, like this:

```html
<div class="custom-block tip">
<p class="custom-block-title">TIP(or another title)</p>
My Content(But can't use markdown)
</div>
```

So I wrote a plugin to make it easier to do this.

I didn't think that much about the plugin when I wrote it, internal linking and spell checking were not my goals and I would stress that internal linking and spell checking should not be used in the plugin.

As for nesting, there are nesting issues with blockquote, 50 steps forward, 100 steps back, and as for migration, once you use a plugin there are bound to be migration issues, and that risk should be considered by the user.

This plugin is good enough for me and I probably won't consider modifying it, but I can close the Pull Request if you think there are still problems.

Thanks again for your review.

lishid:
I'm pretty sure Obsidian's native callouts generate some similar HTML that can be similarly styled via custom CSS. Additionally, the Admonitions plugin also supports the fenced code block syntax.

Either way - If this works for you then that's perfectly fine, though I don't think it makes sense to submit to the plugin repository since it's quite unlikely that a user would prefer to use this plugin over Admonition or native callouts.

I think the best course of action here is to close the PR for now. You are, of course, totally free to continue to develop and use your plugin without having it in the official repo!

PR #926, opened 2022-04-28
Add Release Timeline to community-plugins.json

cakechaser:

## Repo URL

Link to my plugin: https://github.com/cakechaser/obsidian-release-timeline

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
Really cool plugin!

---

Since this plugin requires dataview to be installed, I recommend also adding a check and warning the user that it's a requirement. Otherwise users might run into unexpected errors. Maybe something like:

```
const isDataviewInstalled = !!getAPI();
if (!isDataviewInstalled) {
  new Notice("The . The Release-Timeline plugin requires Dataview to properly function.");
}
```

Just make sure not to put it directly in the `onLoad` is it's not guaranteed this plugin will be loaded after dataview. Inside `onLayoutReady` should work though.

https://github.com/cakechaser/obsidian-release-timeline/blob/10251e87b0303e9e405d08c9d9ed1da337a0a96d/main.ts#L28 You don't want to be querying the whole document here. I would suggest changing this to:

```ts
let matches = el.querySelectorAll(".td-first, .td-next");
```

https://github.com/cakechaser/obsidian-release-timeline/blob/10251e87b0303e9e405d08c9d9ed1da337a0a96d/functions.ts#L6 This is really just a matter of code style, but I would recommend not wrapping your helper functions inside a class. You can just keep them as top-level functions and pass in the plugin instance to the functions that need it. Reason being:
- You might want to split this file into different files based on purpose at some point (`parse.ts`, `timelineRender.ts` for example)
- These functions are less portable if the plugin gets bigger and you need to use the function in different parts of the code.
- They are slightly harder to test, if you wanted to add unit tests down the road

Anyway, not suggesting you need to refactor it, but wanted to offer some unsolicited advice!

cakechaser:
Hi @liamcain, thank you very much for the feedback!

I've added the Dataview notice and fixed the html query from your 2nd suggestion.

Tried to refactor the helper functions from your 3rd suggestion, but ran into some issues, and it was taking a long time for me to figure this out, since it's my first JS project.
So I left them as is for now, hopefully it's not an issue. Nevertheless, thank you very much for the suggestion, I'll try to refactor them for one of the next releases.

liamcain (approved):
Looks great! Nice job. This will be merged in shortly.

PR #927, opened 2022-04-29
Add Pandoc Reference List plugin

mgmeyers:

## Repo URL

Link to my plugin: https://github.com/mgmeyers/obsidian-pandoc-reference-list

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/main.ts#L68 This is redundant, `onLayoutReady` will evaluate the callback immediately if the layout is already ready.

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/main.ts#L59 Licat will be mad if he sees this. Instead of assuming there's only one view for your plugin, I recommend fetching all matching views as needed:

```ts
const views = this.app.workspace.getLeavesOfType(viewType).map(leaf => leaf.view);
```

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/main.ts#L171 why not just use Obsidian's exported debounce function with resetTimer=True?

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/main.ts#L98-L109 Since `onLoad` is blocking, maybe move this into `onLayoutReady` to avoid slowing down startup.

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/viewManager.ts#L24 metadata can be null here

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/viewManager.ts#L42 metadata can be null here

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/view.ts#L70 typo

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/regExps.ts#L2 👀 

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/helpers.ts#L9 why not just use `navigator.clipboard.writeText` instead?

mgmeyers:
> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/main.ts#L68 This is redundant, `onLayoutReady` will evaluate the callback immediately if the layout is already ready.

✅ 

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/main.ts#L59 Licat will be mad if he sees this. Instead of assuming there's only one view for your plugin, I recommend fetching all matching views as needed:

Hmm, I'm not sure if there's a way for users to instantiate another view, though. Though I may be wrong. This code is a copy and paste from your calendar plugin: https://github.com/liamcain/obsidian-calendar-plugin/blob/master/src/main.ts#L40

I similarly check if a view exists here:

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/main.ts#L90 

and here:

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/main.ts#L125

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/main.ts#L171 why not just use Obsidian's exported debounce function with resetTimer=True?

Just more of a personal preference. It's a little more clear to me whats going on while I'm scanning code when I see the `setTimeout`.

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/main.ts#L98-L109 Since `onLoad` is blocking, maybe move this into `onLayoutReady` to avoid slowing down startup.

✅ I removed the `await` here so `fixPath` won't block:
https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/main/src/main.ts#L95

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/viewManager.ts#L24 metadata can be null here

✅ 

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/viewManager.ts#L42 metadata can be null here

✅ 

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/view.ts#L70 typo

✅ 

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/regExps.ts#L2 👀

🤕 

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/helpers.ts#L9 why not just use `navigator.clipboard.writeText` instead?

I want to write both text and HTML to the clipboard at the same time. I can do this with the `navigator.clipboard` api, but its a lot more clunky: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write#examples

Since this is a desktop-only plugin, I though I'd save myself some trouble and use the cleaner electron api

liamcain:
> Hmm, I'm not sure if there's a way for users to instantiate another view, though. Though I may be wrong. This code is a copy and paste from your calendar plugin: https://github.com/liamcain/obsidian-calendar-plugin/blob/master/src/main.ts#L40

Haha that's how I know Licat will be mad. He yelled at me about that😅 It's fixed in the calendar beta branch but people have still been copying that pattern.

It's not a great pattern since views are closable, so you don't want to be holding a reference to the closed view and prevent it from being garbage collected.

> I want to write both text and HTML to the clipboard at the same time

Ah, okay good to know. I was guessing it had to due with `navigator` requiring document focus.

mgmeyers:
> > Hmm, I'm not sure if there's a way for users to instantiate another view, though. Though I may be wrong. This code is a copy and paste from your calendar plugin: https://github.com/liamcain/obsidian-calendar-plugin/blob/master/src/main.ts#L40
> 
> Haha that's how I know Licat will be mad. He yelled at me about that😅 It's fixed in the calendar beta branch but people have still been copying that pattern.
> 
> It's not a great pattern since views are closable, so you don't want to be holding a reference to the closed view and prevent it from being garbage collected.

Ah! Haha, good to know. I've updated things to use a getter instead: https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/main/src/main.ts#L52-L60

liamcain (approved):

lishid:
https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/main.ts#L77
Remember to unload this.

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/main.ts#L88
Not sure why the typecheck isn't working, but `this.view` should have given you errors here and below.

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/main.ts#L100
You need to run your check regardless of the `checking` variable - sometimes your function is called with `checking = false` without a prior "check" call, which would cause you to bypass the check and execute the function (which here isn't a problem because the function also runs the check, but you should ideally do the check in the `checkCallback`).

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/main.ts#L147
Consider using Obsidian's own event delegation system (`HTMLElement.on('event', '.selector', callback)`). Though if these are all elements you create from your code, you should probably hook them at the element level instead of doing event delegation. Event handling is rather efficient nowadays and event delegation isn't really that necessary anymore these days unless you have specific needs for it.

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/main.ts#L184
Why not just use `debounce` (from `obsidian`)?

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/view.ts#L98
Prefer to use async-await instead of `.then`.

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/tooltip.ts#L15
I think a better way to handle this is to hook the scroll event for each tooltip, and unhook the event when the tooltip is removed. If your tooltip has a proper lifecycle then I think that's better than having it globally managed. (Plus you won't have an event hooked globally when you have no tooltips shown)

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/emitter.ts#L13
This seems fairly similar to our event system: https://github.com/obsidianmd/obsidian-api/blob/38bd84fd66b6bf59d2eab243f27e33d3031ca51c/obsidian.d.ts#L1013

https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/tooltip.ts#L80
Discouraged use of innerHTML. Seems like the data source here originally came from an element, better to just store the element instead of its outerHTML. If you need copies you can clone the node.

Tip: https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml

mgmeyers:
Updates: https://github.com/mgmeyers/obsidian-pandoc-reference-list/commit/cbcb7a38c6cd3ea3faf47899ce15ab722e459f7a

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/main.ts#L77 Remember to unload this.

✅ 

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/main.ts#L88 Not sure why the typecheck isn't working, but `this.view` should have given you errors here and below.

Not sure what you mean. `this.view` refers to the getter above that returns either null or a `ReferenceListView`. What errors do you see here?

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/main.ts#L100 You need to run your check regardless of the `checking` variable - sometimes your function is called with `checking = false` without a prior "check" call, which would cause you to bypass the check and execute the function (which here isn't a problem because the function also runs the check, but you should ideally do the check in the `checkCallback`).

👍 In this case I'll leave the check in the `initLeaf` function because it's used elsewhere as well.

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/main.ts#L147 Consider using Obsidian's own event delegation system (`HTMLElement.on('event', '.selector', callback)`). Though if these are all elements you create from your code, you should probably hook them at the element level instead of doing event delegation. Event handling is rather efficient nowadays and event delegation isn't really that necessary anymore these days unless you have specific needs for it.

👍 Moved some of the listeners directly to the elements. Wasn't aware that modern browsers will correctly clean up event listeners when elements are removed!

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/main.ts#L184 Why not just use `debounce` (from `obsidian`)?

More of a stylistic preference, but I've replaced this one with `debounce`

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/view.ts#L98 Prefer to use async-await instead of `.then`.

✅ 

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/0f70e025a87c4d318d79f8902be6bc3d9cbbd570/src/tooltip.ts#L15 I think a better way to handle this is to hook the scroll event for each tooltip, and unhook the event when the tooltip is removed. If your tooltip has a proper lifecycle then I think that's better than having it globally managed. (Plus you won't have an event hooked globally when you have no tooltips shown)

✅ 

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/emitter.ts#L13 This seems fairly similar to our event system: https://github.com/obsidianmd/obsidian-api/blob/38bd84fd66b6bf59d2eab243f27e33d3031ca51c/obsidian.d.ts#L1013

Yeah, the main difference is that this one type checks the event payloads. I removed it, though in favor of `Events` since I don't actually have any payloads for these.

> https://github.com/mgmeyers/obsidian-pandoc-reference-list/blob/aabf423d57d0fe595c257f1e67119d3d8d2ee298/src/tooltip.ts#L80 Discouraged use of innerHTML. Seems like the data source here originally came from an element, better to just store the element instead of its outerHTML. If you need copies you can clone the node.
> 
> Tip: https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml

✅ Switched to `clone` for this.

lishid:
> Not sure what you mean. this.view refers to the getter above that returns either null or a ReferenceListView. What errors do you see here?

Oops, totally missed that, my bad!

PR #929, opened 2022-04-30
Add obsidian-attendance plugin.

Tiim:

## Repo URL

Link to my plugin: https://github.com/Tiim/obsidian-attendance

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
Really nice job with this plugin! I just found a few suggestions and nitpicky things but love seeing plugins with tests 😄 

---

https://github.com/Tiim/obsidian-attendance/blob/6c2e5bc8ef93613be2779c255f3425e71b872fb4/src/AttendanceRenderer.ts#L84 maybe a more verbose name for this arg?

https://github.com/Tiim/obsidian-attendance/blob/6c2e5bc8ef93613be2779c255f3425e71b872fb4/src/AttendanceRenderer.ts#L112 maybe a more verbose name for this arg?

https://github.com/Tiim/obsidian-attendance/blob/6c2e5bc8ef93613be2779c255f3425e71b872fb4/src/AttendanceRenderer.ts#L161 This is because the `<p>` tag adds a bunch of padding, right? You could just add a class to the container and target `.yourcontainer > p` to remove the padding instead of shuffling the html elements around

https://github.com/Tiim/obsidian-attendance/blob/35ec8d783e2df01ef0300ec5a442fd0872868506/src/AttendanceRenderer.ts#L85 you could use `this.containerEl.empty()` here instead

https://github.com/Tiim/obsidian-attendance/blob/35ec8d783e2df01ef0300ec5a442fd0872868506/src/parse/markdown-metadata-parser.ts#L22 getAllTags can return `null`

liamcain (approved):

lishid:
https://github.com/Tiim/obsidian-attendance/blob/b575c7aeda44348f91926034f00759e6771e3ebb/src/main.ts#L51

See this tip: https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-managing-references-to-custom-views

https://github.com/Tiim/obsidian-attendance/blob/b575c7aeda44348f91926034f00759e6771e3ebb/src/cache/cache.ts#L29
You registered these events with the component `SourceCache` but the `SourceCache` isn't a child component of your plugin, so its events are never unregistered when your plugin is disabled, causing memory leaks.

https://github.com/Tiim/obsidian-attendance/blob/b575c7aeda44348f91926034f00759e6771e3ebb/src/cache/cache.ts#L38
This will likely be a performance hog for users on initialization, especially if you're firing off asynchronous tasks for each file.

https://github.com/Tiim/obsidian-attendance/blob/b575c7aeda44348f91926034f00759e6771e3ebb/src/cache/cache.ts#L60
Prefer using `async-await` instead of `.then`.

https://github.com/Tiim/obsidian-attendance/blob/b575c7aeda44348f91926034f00759e6771e3ebb/src/parse/markdown-metadata-parser.ts#L27
I don't think you need to compute these yourself since we keep a very similar data structure in `MetadataCache.resolvedLinks`, which is also performantly computed with deferrals and worker threads. You should be able to query it directly from MetadataCache.

Realistically, I don't think you actually need to pre-compute/cache any of the things you're doing right now, because they can be read from the metadata cache data structures fairly quickly when they are needed (for your queries).

https://github.com/Tiim/obsidian-attendance/blob/b575c7aeda44348f91926034f00759e6771e3ebb/src/AttendanceData.ts#L6
Should use `moment` from `obsidian` instead of bringing your own copy of `moment`.

https://github.com/Tiim/obsidian-attendance/blob/b575c7aeda44348f91926034f00759e6771e3ebb/src/AttendanceData.ts#L239
Perhaps the caller should read the data of the file so you don't have to repeatedly re-read the file from disk?

https://github.com/Tiim/obsidian-attendance/blob/b575c7aeda44348f91926034f00759e6771e3ebb/src/ui/AttendanceCodeblockRenderer.ts#L78
Perhaps you should just create an event handler on your plugin (use the `Events` class from `obsidian`) instead of using Workspace as a message bridge?

https://github.com/Tiim/obsidian-attendance/blob/b575c7aeda44348f91926034f00759e6771e3ebb/src/ui/AttendanceCodeblockRenderer.ts#L105
Use `.empty()` instead.

https://github.com/Tiim/obsidian-attendance/blob/b575c7aeda44348f91926034f00759e6771e3ebb/src/util/link.ts#L16
There's a helper function in `app.fileManager.generateMarkdownLink` that will properly generate markdown links using the user's link style preferences (wikilink vs markdown link, relative path vs absolute path, etc).

Tiim:
Thanks @lishid for your great feedback! I have tried to resolve most issues you mentioned.

I have a few questions though:
> Realistically, I don't think you actually need to pre-compute/cache any of the things you're doing right now, because they can be read from the metadata cache data structures fairly quickly when they are needed (for your queries).

The plugin currently caches tags, links and attendance code blocks. There is a class "FolderCache" but I just realized it doesn't actually do any caching, I will rename it.
For the links and folders I agree, its easy to just read them directly, but for the tags I'm a bit stumped. I cant find an easy way to get all files with a certain tag, am I just missing it? The codeblocks I'm caching because I am working on making an overview over all attendance codeblocks in the vault, which probably should stay cached.

Another thing I'm wondering is, is there a reliable way to replace the content of the codeblock? Currently I'm manually trying to find the codeblock start and end index in the file and replace just this part. The problem is, this falls appart when for example inside a call out or quote.

lishid:
> For the links and folders I agree, its easy to just read them directly, but for the tags I'm a bit stumped. I cant find an easy way to get all files with a certain tag, am I just missing it?

You are right in that the data isn't nicely formatted for what you need, but it's already fairly accessible through MetadataCache on a per-file basis. I would say, if you query very frequently this might be necessary, but if the query only runs on user action, then you might as well just manually query the tags of each file to satisfy your filter.

> Another thing I'm wondering is, is there a reliable way to replace the content of the codeblock? Currently I'm manually trying to find the codeblock start and end index in the file and replace just this part. The problem is, this falls appart when for example inside a call out or quote.

This unfortunately I don't have a better solution to suggest. Trying to catch this case and throwing an error is likely best you can do, and instruct users to avoid wrapping it in other blocks.

Tiim:
Thank you again for all of the feedback, I think i have resolved all of your concerns.

Tiim:
Thanks again for the review :) I will make these changes when I have some free time next week. 

I do have some new features planned, but I don't have any changes that absolutely need to be in the plugin before it is added.

liamcain:
Hi @Tiim, just checking in, is this reading for re-review?

Tiim:
Hey @liamcain, I have not changed much except some performance improvements and the changes from the last review. So sure it is ready for a re-review :)

liamcain:
> Hey @liamcain, I have not changed much except some performance improvements and the changes from the last review. So sure it is ready for a re-review :)

Sorry, I totally missed this reply! Finally got around to re-reviewing. Thanks for your patience with waiting so long. Plugin looks great!

Tiim:
No worries! Thanks for merging :)

PR #932, opened 2022-05-02
Add `Obsidian Enhancing Export` plugin.

mokeyish:

## Repo URL

Link to my plugin:
https://github.com/mokeyish/obsidian-enhancing-export
## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
https://github.com/mokeyish/obsidian-enhancing-export/blob/cfb82d85dee44f059ef5781ed1eafb20ac3ab72f/manifest.json#L9 This is where clicking on the author's name in the plugin list will link to. If you have a personal domain you want to link to or you might want to link to `https://github.com/mokeyish`

https://github.com/mokeyish/obsidian-enhancing-export/blob/cfb82d85dee44f059ef5781ed1eafb20ac3ab72f/src/main.ts#L56 I recommend removing this or adding a filterable prefix such as `[obsidian-enhancing-export] saveSettings` so that it doesn't pollute the console as much

https://github.com/mokeyish/obsidian-enhancing-export/blob/cfb82d85dee44f059ef5781ed1eafb20ac3ab72f/src/main.ts#L40 Instead of downloading the files automatically, I suggest you bundle the lua files in your `main.js` file. You can either copy them into JS strings, or if you want to keep them as `.lua` files, you can just import them directly and specify an esbuild loader to handle them.


Adding the following to your esbuild config should work:
```
{
  loader: { ".lua": "text" }
}
```

A few advantages to this over automatically downloading the files onLoad:
- easier to manage between versions
- more secure
- fully offline

https://github.com/mokeyish/obsidian-enhancing-export/blob/cfb82d85dee44f059ef5781ed1eafb20ac3ab72f/src/polyfill.ts#L8 I recommend that you don't mutate the prototypes for any shared classes. This could be dangerous if multiple plugins mutated the same prototype and caused conflicts. You can just create helper functions that accept the Element as an argument.

https://github.com/mokeyish/obsidian-enhancing-export/blob/cfb82d85dee44f059ef5781ed1eafb20ac3ab72f/src/ui/setting_tab.ts#L78 remove this

https://github.com/mokeyish/obsidian-enhancing-export/blob/cfb82d85dee44f059ef5781ed1eafb20ac3ab72f/src/ui/rename_modal.ts#L26 should this be a value from `lang`?

mokeyish:
Hi [liamcain](https://github.com/liamcain): 
    Thanks for your reviewing. I have updated the code accord to your suggestion. 

For the third one, how to `automatically downloading the files onLoad`? I just write lua text string to file now.

https://github.com/mokeyish/obsidian-enhancing-export/blob/47bdad9f83d9cc6d05d8df0ea54a0b2b5120d817/src/main.ts#L65-L72

liamcain (approved):

mokeyish:
@liamcain Hi liamcain:
Is there any other suggestions about this PR? 
And when will this PR be merged?

lishid:
https://github.com/mokeyish/obsidian-enhancing-export/blob/47bdad9f83d9cc6d05d8df0ea54a0b2b5120d817/src/main.ts#L70
Here you can use `adapter.write()` instead of having to import `fs` from node, and computing your own file paths from the vault's basepath. I know your plugin requires running on the desktop anyway and you need node/electron API later on, but I still think it's best to stick with obsidian's API whenever possible.

community-plugins.json
```diff
@@ -4119,5 +4119,12 @@
         "author": "musug",
         "description": "Screenshot png to jpeg and compress and rename.",
         "repo": "musug/obsidian-paste-png-to-jpeg"
+    },
+    {
+        "id": "obsidian-enhancing-export",
+        "name": "Obsidian Enhancing Export",
+        "author": "YISH<moktyish@hotmail.com>",
```
liamcain, line 4126: Hi @mokeyish, just wanted to confirm: are you okay with including your email in the author section here? This will be visible from the list of community plugins, so I just wanted to confirm it's intentional

author-section

mokeyish: I have removed the email in it.

mokeyish:
> https://github.com/mokeyish/obsidian-enhancing-export/blob/47bdad9f83d9cc6d05d8df0ea54a0b2b5120d817/src/main.ts#L70 Here you can use `adapter.write()` instead of having to import `fs` from node, and computing your own file paths from the vault's basepath. I know your plugin requires running on the desktop anyway and you need node/electron API later on, but I still think it's best to stick with obsidian's API whenever possible.



> https://github.com/mokeyish/obsidian-enhancing-export/blob/47bdad9f83d9cc6d05d8df0ea54a0b2b5120d817/src/main.ts#L70 Here you can use `adapter.write()` instead of having to import `fs` from node, and computing your own file paths from the vault's basepath. I know your plugin requires running on the desktop anyway and you need node/electron API later on, but I still think it's best to stick with obsidian's API whenever possible.


This is better, I've updated it to use obsidian's API  instead.

liamcain:
https://github.com/mokeyish/obsidian-enhancing-export/blob/1b47102fad7daeb59cba5da5300505be0e507b18/src/main.ts#L71

Does this work? The lua files are not binary files, so this should be `write()` instead.

mokeyish:
> https://github.com/mokeyish/obsidian-enhancing-export/blob/1b47102fad7daeb59cba5da5300505be0e507b18/src/main.ts#L71
> 
> Does this work? The lua files are not binary files, so this should be `write()` instead.

Yes, I embed lua as binary. 
https://github.com/mokeyish/obsidian-enhancing-export/blob/1b47102fad7daeb59cba5da5300505be0e507b18/esbuild.config.mjs#L59
Text may diffrent on different platforms.

PR #934, opened 2022-05-02
Add "Meta title filename" plugin

Snezhig:

## Repo URL

Link to my plugin: https://github.com/Snezhig/obsidian-front-matter-title

## Release Checklist
  - [x]  Windows
  - [x]  Linux
  - [x]  Android
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
https://github.com/Snezhig/obsidian-meta-title-filename/blob/ba84acb8a17092c675e5e61af53c8ae3b106b442/main.ts#L7 this import looks unused, right?

https://github.com/Snezhig/obsidian-meta-title-filename/blob/ba84acb8a17092c675e5e61af53c8ae3b106b442/main.ts#L6 there's a debounce function available from the `"obsidian"` package if you want to make your bundle size a bit smaller

https://github.com/Snezhig/obsidian-meta-title-filename/blob/ba84acb8a17092c675e5e61af53c8ae3b106b442/main.ts#L64-L66 it's a bit confusing that you have 2 variables called `explorer` here but they are different types. Maybe `explorerView`?

Snezhig:
1. Yes, I have deleted this.
2. Changed.
3. Great proposal, thanks.
4. I saw this, but I decided to use my own implementation, because I need a little functionality.
5. Yes, the loop is excessive, I changed it.
6. I just want my plugin not to know what the title was and how it was made. So I just keep the original.

Snezhig:
Why action have tried to find manifest.json in release 1.1.0-beta, when the last release version is 1.0.1?

liamcain:
https://github.com/Snezhig/obsidian-meta-title-filename#introduction I think your README could be a little bit more clear as to what metadata it supports. You mention `my_title` and then your example shows a nested `additional.title`. I think you're better off being super clear and showing an example for both.

> Then set the path like additional.title and plugin will set Project ideas as filename

Similarly, this might give the impression that the file is getting renamed. Maybe instead of "set," just say "show"

https://github.com/Snezhig/obsidian-meta-title-filename/blob/588e1fa29f9c5535d85b9aa32175a45125816372/main.ts#L2 unused. And you can remove this from your package.json

https://github.com/Snezhig/obsidian-meta-title-filename/blob/588e1fa29f9c5535d85b9aa32175a45125816372/main.ts#L26 this might be a bit cleaner to just overwrite the `saveSettings` from within the `onLoad` function:

```ts
public async onload() {
  this.saveSettings = debounce(this.saveSettings.bind(this), 100, true);
}
```

https://github.com/Snezhig/obsidian-meta-title-filename/blob/588e1fa29f9c5535d85b9aa32175a45125816372/main.ts#L87 I recommend removing this to avoid polluting the console.  If you think it's important, change it to `console.debug` instead.

https://github.com/Snezhig/obsidian-meta-title-filename/blob/588e1fa29f9c5535d85b9aa32175a45125816372/main.ts#L131 I'm pretty sure the `bind` on this line is unnecessary since the closure is defined inline in the same context.

https://github.com/Snezhig/obsidian-meta-title-filename/blob/588e1fa29f9c5535d85b9aa32175a45125816372/main.ts#L134 `abstract` is a reserved word, I recommend renaming this to something else (maybe `fileOrFolder`)

https://github.com/Snezhig/obsidian-meta-title-filename/blob/588e1fa29f9c5535d85b9aa32175a45125816372/main.ts#L136 you could just use `.map()` here to map all the values to a promise. Could make this function a one-liner.

https://github.com/Snezhig/obsidian-meta-title-filename/blob/588e1fa29f9c5535d85b9aa32175a45125816372/src/Settings.ts#L56 👀 

https://github.com/Snezhig/obsidian-meta-title-filename/blob/588e1fa29f9c5535d85b9aa32175a45125816372/src/MetaTitleParser.ts#L6 Doesn't look like this needs to be async.

https://github.com/Snezhig/obsidian-meta-title-filename/blob/588e1fa29f9c5535d85b9aa32175a45125816372/src/Utils/Queue.ts#L7 looks like the typing here is wrong

https://github.com/Snezhig/obsidian-meta-title-filename/blob/588e1fa29f9c5535d85b9aa32175a45125816372/src/Utils/Queue.ts#L1 you can use Obsidian's debounce here

Snezhig:
@liamcain , Thanks for tips.
1. Updated readme with examples.
2. Removed, but I use it for queue, because this debounce can return promise. That lets test code easily.
3. I did not know, that I could do that. Thanks.
4. Forgot to delete. Done.
5. Right.
6. I agree.
7. I can not, because Map.prototype.values() returns Iterator. I can wrap it to Array, but it won't make the code more readable.
8. Wow. Changed.
9. Yea, I removed and and got a lot of problems, but solved them. Thank you, that was good remark.
10. Answered at second point.

liamcain (approved):
> I can not because Map.prototype.values() returns Iterator. I can wrap it to Array, but it won't make the code more readable.

Ah, missed that it was a map and not an object. Good call!

---

Looks great! Appreciate the README changes especially, looks super clear now.

community-plugins.json
```diff
@@ -4185,5 +4185,12 @@
         "description": "Convert Markdown files to BBCode",
         "repo": "salockhart/obsidian-bbcode",
         "branch": "master"
+    },
+    {
+        "id": "obsidian-meta-title-filename-plugin",
+        "name": "Meta title filenames",
```
liamcain, line 4191: The code looks good, but I think you might want to use the term "frontmatter" instead of "meta block." Maybe instead of "Meta title filenames," a more accurate plugin name would be "Frontmatter titles"

community-plugins.json
```diff
@@ -4185,5 +4185,12 @@
...
+        "author": "Snezhig",
+        "description": "Let you use data from meta block as a filename",
```
liamcain, line 4193:
```suggestion
        "description": "Lets you define a title in frontmatter to be displayed as the filename",
```

Maybe something like this is more descriptive?

Snezhig:
Nice, and new name of plugin is great, but I have to rename my repository, yes?

liamcain:
>  I have to rename my repository, yes?

That's up to you. The repo name isn't very visible from within Obsidian so I doubt anyone will be confused. But if you do want to rename the repo, I suggest doing it before we merge this so that we can point to the correct repo name here.

liamcain:
Sorry for the extra hoops to jump through, but for the rename, you'll also need to update the `id` field in your manifest file: https://github.com/Snezhig/obsidian-front-matter-title/blob/fd5ae1bacc5fb27a08b19f4d690ce685015255d3/manifest.json#L2.

Snezhig:
I remembered about it too and just did it in last beta and master releases.

liamcain:
Awesome, thanks! It's installing properly.

PR #936, opened 2022-05-03
Add Folder Focus Mode Plugin

grochowski:

## Repo URL

Link to my plugin: https://github.com/grochowski/obsidian-folder-focus-mode

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


aidenlx:
I think maybe your plugin shares something in common with [alx-folder-note's focus folder feature](https://github.com/aidenlx/alx-folder-note/releases/tag/0.12.2)?

liamcain:
https://github.com/grochowski/obsidian-folder-focus-mode/blob/699eaa00d1c8da88a884effc9e859e083aa36f62/main.ts#L51-L53 I don't think this logic is correct. You're looping over the list of registered file explorers for each iteration of the outer loop. The `registeredFileExplorers` var isn't doing anything here so I think you want to just get rid of inner loop and perform the operations on `fileExplorer.view`.

https://github.com/grochowski/obsidian-folder-focus-mode/blob/699eaa00d1c8da88a884effc9e859e083aa36f62/main.ts#L84-L86 same here.

https://github.com/grochowski/obsidian-folder-focus-mode/blob/699eaa00d1c8da88a884effc9e859e083aa36f62/main.ts#L139 getActiveFile can return null. You might want to use `checkCallback` here instead so that this command can only be run when there is an active file.

https://github.com/grochowski/obsidian-folder-focus-mode/blob/699eaa00d1c8da88a884effc9e859e083aa36f62/main.ts#L3 `path` is a node library so this might break on mobile, I'm not positive since I see you're also installing it from npm, but I think it's still an issue.

https://github.com/grochowski/obsidian-folder-focus-mode/blob/699eaa00d1c8da88a884effc9e859e083aa36f62/main.ts#L88 you should use `const` instead of `var` here

grochowski:
@liamcain thanks for the feedback! resolved mentioned issues in new release (0.9.2) - also, I've checked it on mobile and now it works just right :)

liamcain (approved):
Changes look good! This will get merged shortly.

PR #937, opened 2022-05-03
Add obsidian-database-plugin

tomaszkiewicz:

## Repo URL

Link to my plugin: https://github.com/tomaszkiewicz/obsidian-database-plugin

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4119,5 +4119,12 @@
         "author": "musug",
         "description": "Screenshot png to jpeg and compress and rename.",
         "repo": "musug/obsidian-paste-png-to-jpeg"
+    },
+        {
+        "id": "obsidian-database-plugin",
+        "name": "Obsidian database plugin",
+        "author": "Łukasz Tomaszkiewicz",
+        "description": "Enables creating databases based on files in Obsidian - like Dataview, but with editing!",
+        "repo": "tomaszkiewicz/obsidian-database-plugin"
     }
 ]


liamcain:
Hi! Really cool plugin. If you haven't already, I recommend you take a look at the DB Folder plugin: https://github.com/RafaelGB/obsidian-db-folder. It has similar goals, though it opens a custom view instead of being inline. Since I could see users wanting both, there's definitely room for both plugins, but there might be some collaboration opportunities

---

https://github.com/tomaszkiewicz/obsidian-database-plugin/blob/0bc5b83e454e4e000628a5ee7756bdb356c00fd2/src/main.ts#L88 if you're not using settings, I recommend just removing this and all the settings related things for now.

https://github.com/tomaszkiewicz/obsidian-database-plugin/blob/0bc5b83e454e4e000628a5ee7756bdb356c00fd2/manifest.json#L3 It's a bit redundant to include the word "plugin" in the name of the plugin.

https://github.com/tomaszkiewicz/obsidian-database-plugin/blob/0bc5b83e454e4e000628a5ee7756bdb356c00fd2/src/main.ts#L13 is this necessary? It looks like you are vuetifying it in the `Vuetify.ts` file already

https://github.com/tomaszkiewicz/obsidian-database-plugin/blob/0bc5b83e454e4e000628a5ee7756bdb356c00fd2/src/main.ts#L39 I think it's better to create your own subclass of MarkdownRenderChild and manage the lifecycle of the Vue component from there instead of attaching them to the plugin. On L68, you're calling destroy, but not removing the instance from `this.instances`. Your list of instances is going to get stale really quickly.

https://github.com/tomaszkiewicz/obsidian-database-plugin/blob/0bc5b83e454e4e000628a5ee7756bdb356c00fd2/src/source.ts#L55 I think you're better off using `cachedRead` here instead of read since you're going through all the files in the vault. `read` will read it from disc which can be a lot slower.

https://github.com/tomaszkiewicz/obsidian-database-plugin/blob/1d9d67c142f3180fa1d662eeacc4b97cf8e34c82/src/source.ts#L56 It's still a hack, but maybe this syntax for links is better?

```
%% db-links
column 1: [[link]]
column 2: [[link]]
%%
---

That way the Obsidian comment parser doesn't get as confused by seeing the triple %

---

Notes from playing with it myself:
- The plugin doesn't like it if you don't include `fields`. Nothing gets rendered and it throws the following error: `VM2182:17900 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'filter')
    at DatabasePlugin.eval (eval at <anonymous> (app.js:1:1), <anonymous>:17900:41)`
- If the yaml fails to parse, nothing is rendered. You should consider adding an error message inside the codeblock

ericaxu:

PR #939, opened 2022-05-04
add obsidian external embed

tnichols217:

## Repo URL

Link to my plugin:

https://github.com/tnichols217/obsidian-dynamic-import

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4119,5 +4119,13 @@
         "author": "musug",
         "description": "Screenshot png to jpeg and compress and rename.",
         "repo": "musug/obsidian-paste-png-to-jpeg"
+    },
+    {
+        "id": "obsidian-iframes",
+        "name": "Obsidian Iframes",
+        "author": "Trevor Nichols",
+        "description": "Allows you to create relative iframes and MD iframes in Obsidian",
+        "repo": "tnichols217/obsidian-iframes",
+        "branch": "main"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4121,11 +4121,11 @@
         "repo": "musug/obsidian-paste-png-to-jpeg"
     },
     {
-        "id": "obsidian-dynamic-import",
-        "name": "Obsidian Dynamic Import",
+        "id": "obsidian-external-embed",
+        "name": "Obsidian External Embed",
         "author": "Trevor Nichols",
         "description": "Allows you to Import other MD files into your notes, and improves iframes",
-        "repo": "tnichols217/obsidian-dynamic-import",
+        "repo": "tnichols217/obsidian-external-embed",
         "branch": "main"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4121,11 +4121,11 @@
         "repo": "musug/obsidian-paste-png-to-jpeg"
     },
     {
-        "id": "obsidian-iframes",
-        "name": "Obsidian Iframes",
+        "id": "obsidian-dynamic-import",
+        "name": "Obsidian Dynamic Import",
         "author": "Trevor Nichols",
-        "description": "Allows you to create relative iframes and MD iframes in Obsidian",
-        "repo": "tnichols217/obsidian-iframes",
+        "description": "Allows you to Import other MD files into your notes, and improves iframes",
+        "repo": "tnichols217/obsidian-dynamic-import",
         "branch": "main"
     }
 ]


tnichols217:
I also kinda need help naming my plugin, so suggestions are welcome

liamcain:
My plugin name suggestion would be `External Embed`. 

---

https://github.com/tnichols217/obsidian-dynamic-import/blob/ce63c92ed6cb78e7b30494961f4167ee3c0f7d7d/main.ts#L27 small code style nitpick - It's pretty standard for typescript types and interfaces to be capitalized

https://github.com/tnichols217/obsidian-dynamic-import/blob/ce63c92ed6cb78e7b30494961f4167ee3c0f7d7d/main.ts#L47 I recommend that your settings object only includes the `value`.

So you would have:

```
interface DynamicImportSettings {
	allowInet: boolean
	allowInline: boolean
	recursionDepth: number
	useCacheForFiles: boolean
	cacheRefreshTime: number
}
```

and separately you can have an object with the name and description for each property. There's not reason to save the name and description to JSON, especially because you might want to change those values in a later version of the plugin.

https://github.com/tnichols217/obsidian-dynamic-import/blob/ce63c92ed6cb78e7b30494961f4167ee3c0f7d7d/main.ts#L397 your settings are all text fields but your settings also include a lot of boolean values. They should be using `addToggle` instead.

https://github.com/tnichols217/obsidian-dynamic-import/blob/ce63c92ed6cb78e7b30494961f4167ee3c0f7d7d/main.ts#L20-L21 CACHE and SETTINGS should be instance variables on your plugin class.

https://github.com/tnichols217/obsidian-dynamic-import/blob/ce63c92ed6cb78e7b30494961f4167ee3c0f7d7d/main.ts#L142 this can be `return Promise.resolve()`

https://github.com/tnichols217/obsidian-dynamic-import/blob/ce63c92ed6cb78e7b30494961f4167ee3c0f7d7d/main.ts#L219 This overrides the behavior for all iframes (not just the designated ones). From what I can tell, iframes without the `iframe-md` class appear to be broken.

Why not use a custom codeblock instead so that users can want to use the regular iframe behavior can still use your plugin? By using a post-processor, this plugin doesn't work in Live Preview either.

tnichols217:
1. Fixed
2. I only save the actual info to the json now, without all the extra info
3. I didn't know addToggle existed! Thanks!
4. Fixed
5. Fixed
6. I disabled iframes for now, and will add the new codeblock in the future
    - However, the design of the !!!paste and !!!inline command require them to be done with the postProcessor, but the import and iframe commands I will add a codeblock alternative for. Would that be a good decision?

liamcain:
Hi, thanks for making those changes. I took another look and spotted a couple things that I missed the first time around

https://github.com/tnichols217/obsidian-external-embed/blob/358d8efbd91f9a9642aab6d85cee49692a161188/main.ts#L342 You have a lot of instances of `.innerHTML`. This is dangerous and can [pose a security risk for users](https://github.com/tnichols217/obsidian-external-embed/blob/358d8efbd91f9a9642aab6d85cee49692a161188/main.ts#L342). Please consider using alternative methods of manipulating/inserting DOM elements.

https://github.com/tnichols217/obsidian-external-embed/blob/358d8efbd91f9a9642aab6d85cee49692a161188/main.ts#L329 remove this

https://github.com/tnichols217/obsidian-external-embed/blob/358d8efbd91f9a9642aab6d85cee49692a161188/main.ts#L364-L367 what's going on here?

tnichols217:
1. Yes, it is a security risk, but by the design of this plugin, namely having importing arbitrary from remote sources as a feature, it is a must in the design of this plugin. I do have server safeguards in the settings of this plugin however, such as accessing the internet, or running scripts.
2. Removed
3. So since in obsidian, the onload function of an element does not actually get called when an element is loaded, I added the option to run this function when elements are loaded (also has its own safeguard setting which is off by default to prevent malicious code from executing unless explicitly turned on)

tnichols217:
However, I am willing to add more security settings to limit these risks if you think necessary. Currently there is no restriction on importing other local files only, while there are restrictions on the onload function, internet access, as well as the inline command. There are currently no restrictions on the paste command (which gets passed back into the obsidian renderMarkdown function), the inline commandblock (which only runs the HTML explicitly inside your notes from no external sources), and the import command (which is still limited by the internet setting)
@liamcain  Please let me know if you think there are other precautions I could take to prevent this security risk, since I can't think of a better way to import external HTML code without the use of innerHTML. I can think of using iframes with srcDocument, but that would not respect the style settings from obsidian.  Thanks.

lishid:
I've taken a look at the plugin code and unfortunately I think it would require a lot of changes to be in a state that I'd feel comfortable merging into the official plugin list.

The two major ones being:
1. Your plugin processes markdown block by removing the entire markdown block element and re-generating it from (a modified) source. This really breaks down the rendering pipeline because other markdown post-processors that has to interact with the element will have their elements thrown out. In practice, this means that anytime the user uses one of your directives, all other plugins that touches that section breaks.
2. Your plugin is explicitly designed to be insecure - not only things like `onload` but other issues like the lack of sanitization. If a user installs your plugin, and opens a markdown file with a malicious block like `<script>...</script>`, it would execute that code instantly due to the use of `innerHTML`. [(Read more here)](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml)

The combination of the two would likely to lead to many users getting frustrated with other plugins breaking, while at the same time opening themselves up to possible remote code execution exploits.

You are perfectly fine to work on the plugin for your own use case, possibly even distributing it to others, but unfortunately it would be irresponsible of us to merge your plugin with these flaws as it is right now.

PR #941, opened 2022-05-05
Add obsidian-translator plugin

luhaifeng666:

## Repo URL

Link to my plugin: https://github.com/luhaifeng666/obsidian-translator

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3897,13 +3897,6 @@
         "repo": "adifyr/obsidian-chat-view"
     },
     {
-        "id": "obsidian-link-keeper",
-        "name": "Link Keeper",
-        "author": "Haifeng Lu",
-        "description": "This is a plugin for Obsidian to manage your frequently used links. ",
-        "repo": "luhaifeng666/obsidian-link-keeper"
-    },
-    {
         "id": "obsidian-translator",
         "name": "Translator",
         "author": "Haifeng Lu",
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3902,5 +3902,12 @@
         "author": "Haifeng Lu",
         "description": "This is a plugin for Obsidian to manage your frequently used links. ",
         "repo": "luhaifeng666/obsidian-link-keeper"
+    },
+    {
+        "id": "obsidian-translator",
+        "name": "Translator",
+        "author": "Haifeng Lu",
+        "description": "This is a plugin for Obsidian to translate selected text.",
+        "repo": "luhaifeng666/obsidian-translator"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3895,5 +3895,12 @@
         "author": "Aditya Majethia",
         "description": "Chat View lets you quickly and easily create elegant Chat UIs in your Markdown Files.",
         "repo": "adifyr/obsidian-chat-view"
+    },
+    {
+        "id": "obsidian-link-keeper",
+        "name": "Link Keeper",
+        "author": "Haifeng Lu",
+        "description": "This is a plugin for Obsidian to manage your frequently used links. ",
+        "repo": "luhaifeng666/obsidian-link-keeper"
     }
 ]


liamcain:
Hi! 

https://github.com/luhaifeng666/obsidian-translator#install
I recommend you move the `Install` instructions in your README to the bottom. I think it will give users the wrong idea that they need to do anything special to install the plugin. It's best to lead with "Usage"

https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/settings.ts#L11 You shouldn't store secrets in the settings object. That gets serialized to JSON and it's a very secure place for a secret key. I would suggest you put the value in localStorage instead

https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/styles.css#L1 You should prefix these classnames with your plugin name, otherwise you risk impacting unrelated elements within Obsidian or other plugins

https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/main.ts#L35 If you use `editorCheckCallback` instead, you can make sure that the command is only available if the user has some text selected

https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/modals.ts#L28 you shouldn't make the parameter here as optional

https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/modals.ts#L33 Why are you fetching the plugin instance from the modal class? You can just pass in the plugin instance via the constructor instead.

https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/utils.ts#L4 You can use `request()` instead of axios for making api calls <https://github.com/obsidianmd/obsidian-api/blob/38bd84fd66b6bf59d2eab243f27e33d3031ca51c/obsidian.d.ts#L2674>

https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/utils.ts#L33 You can use `window.crypto.subtle.digest` instead rather than relying on the Node-JS only library.

https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/main.ts#L34 It's generally preferable to leave the default hotkey empty and let users configure this on their own.

luhaifeng666:
> > Hi!
> > https://github.com/luhaifeng666/obsidian-translator#install I recommend you move the `Install` instructions in your README to the bottom. I think it will give users the wrong idea that they need to do anything special to install the plugin. It's best to lead with "Usage"
> > https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/settings.ts#L11 You shouldn't store secrets in the settings object. That gets serialized to JSON and it's a very secure place for a secret key. I would suggest you put the value in localStorage instead
> > https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/styles.css#L1 You should prefix these classnames with your plugin name, otherwise you risk impacting unrelated elements within Obsidian or other plugins
> > https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/main.ts#L35 If you use `editorCheckCallback` instead, you can make sure that the command is only available if the user has some text selected
> > https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/modals.ts#L28 you shouldn't make the parameter here as optional
> > https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/modals.ts#L33 Why are you fetching the plugin instance from the modal class? You can just pass in the plugin instance via the constructor instead.
> > https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/utils.ts#L4 You can use `request()` instead of axios for making api calls https://github.com/obsidianmd/obsidian-api/blob/38bd84fd66b6bf59d2eab243f27e33d3031ca51c/obsidian.d.ts#L2674
> > https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/utils.ts#L33 You can use `window.crypto.subtle.digest` instead rather than relying on the Node-JS only library.
> > https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/main.ts#L34 It's generally preferable to leave the default hotkey empty and let users configure this on their own.
> 
> Hi there~ Thanks for your patient feedback~😁 I have created some issues for these issues and I'll resolve them as soon as possible. However, I have some different opinions on the following issues:
> 
> > https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/settings.ts#L11 You shouldn't store secrets in the settings object. That gets serialized to JSON and it's a very secure place for a secret key. I would suggest you put the value in localStorage instead
> 
> Before using this plugin, users need to register first and get their own `appId` and `secretKey`.I don't know what their contents are, but I need to request data through them. So I provide input boxes for the users to configure them in the setting interface.
> 
> > https://github.com/luhaifeng666/obsidian-translator/blob/8a764fec48323c8a2a39ef82db3d65f4410628c8/main.ts#L35 If you use `editorCheckCallback` instead, you can make sure that the command is only available if the user has some text selected
> 
> This plugin supports not only translating the selected text but also the custom phrases by the same command. So it's needless to do this.

I have resolved all issues that you raised except the above two.

liamcain:
> Before using this plugin, users need to register first and get their own appId and secretKey.I don't know what their contents are, but I need to request data through them. So I provide input boxes for the users to configure them in the setting interface

I understand that, but the settings interface just saves a file inside `<your_vault>/.obsidian/<plugin>/data.json`. I was just suggesting you store using `localStorage` instead to avoid making it easily read by other applications outside of Obsidian.

liamcain (approved):

luhaifeng666:
> > Before using this plugin, users need to register first and get their own appId and secretKey.I don't know what their contents are, but I need to request data through them. So I provide input boxes for the users to configure them in the setting interface
> 
> I understand that, but the settings interface just saves a file inside `<your_vault>/.obsidian/<plugin>/data.json`. I was just suggesting you store using `localStorage` instead to avoid making it easily read by other applications outside of Obsidian.

Got it thanks~ I'll optimize it asap.

liamcain:
@luhaifeng666 are you okay with this being merged now? Or would you like to make the above change first?

luhaifeng666:
> @luhaifeng666 are you okay with this being merged now? Or would you like to make the above change first?

Merge first.

PR #942, opened 2022-05-06
Add obsidian-user-plugins reference

mnowotnik:

## Repo URL

Link to my plugin: https://github.com/mnowotnik/obsidian-user-plugins

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
https://github.com/mnowotnik/obsidian-user-plugins/blob/04cec150455394fec3b80ae4a017ebc335ea9e0c/manifest.json#L8 I don't think this plugin will run on mobile

1. You're using `path` which is a node.js library
2. You're using `window.require` which won't run on mobile

Maybe just mark this as desktopOnly for now.

https://github.com/mnowotnik/obsidian-user-plugins/blob/04cec150455394fec3b80ae4a017ebc335ea9e0c/src/main.ts#L34 this function can throw. You should catch the error and continue

https://github.com/mnowotnik/obsidian-user-plugins/blob/731482b427b287cb53f069e604ac23cb0bfe8537/src/main.ts#L72 I recommend also just prefixing this with your plugin name (`[obsidian-user-plugins]`) to make it easier to filter the console

mnowotnik:
@liamcain 
I fixed it according to your specifications.
- removed dep on path and require
- tested it on mobile - works
- catching all the errors and logging
- prefixed logs with the plugin name

liamcain (approved):

liamcain:
Hi, sorry for the delay! Plugin looks good to me, should be available in the plugin gallery shortly.

holroy:
Is there any reason this plugin is not released until the public yet? If really like to have this functionality available.

mnowotnik:
> Is there any reason this plugin is not released until the public yet? If really like to have this functionality available.

@holroy I haven't looked into this plugin in a while, but last I checked it had been released to community plugins. Please check.

holroy:
My bad, I searched for "Users plugin" and it didn't show,  and when I continued my search and found it only "ready for review" in this list I  thought it wasn't ready. Neither did I notice the "Merged" tag on the top, nor that it is called "User plugins" with singular "User" and plural on "Plugins"...

PR #945, opened 2022-05-06
Add local-quotes plugin

ka1tzyu:

## Repo URL

Link to my plugin: https://github.com/ka1tzyu/local-quotes

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4126,5 +4126,12 @@
         "author": "turulix",
         "description": "Automatic MOC for folders",
         "repo": "turulix/obsidian-folder-index"
+    },
+    {
+	    "id": "local-quotes",
+	    "name": "Local Quotes",
+	    "author": "ka1tzyu",
+	    "description": "Allows you to collect your quotes from all over the repository and embed them in different locations with renew delays.",
+	    "repo": "ka1tzyu/local-quotes"
     }
 ]


ericaxu:
The indentation of the JSON file looks a bit off. Could you kindly fix that please?

ka1tzyu:
> The indentation of the JSON file looks a bit off. Could you kindly fix that please?

Ooops, sorry, fixed

liamcain:
https://github.com/ka1tzyu/local-quotes/blob/8ea85e696add6aa1c221b82246515b1365f6c9d4/src/main.ts#L21 You don't need to `await` this.

https://github.com/ka1tzyu/local-quotes/blob/8ea85e696add6aa1c221b82246515b1365f6c9d4/src/processors/code-block.ts#L21 Why are you saving from within the render function?

https://github.com/ka1tzyu/local-quotes/blob/8ea85e696add6aa1c221b82246515b1365f6c9d4/src/main.ts#L93 Why are you saving settings on unload?

https://github.com/ka1tzyu/local-quotes/blob/8ea85e696add6aa1c221b82246515b1365f6c9d4/src/processors/code-block.ts#L34 Do not use `innerHTML`. Even with the check for `<script>` tags, it's still a [security risk](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml). You should be able to avoid doing it by just appending the result of `renderMarkdown` to the element.

https://github.com/ka1tzyu/local-quotes/blob/8ea85e696add6aa1c221b82246515b1365f6c9d4/src/processors/code-block.ts#L55 Why are you calling save here?

https://github.com/ka1tzyu/local-quotes/blob/8ea85e696add6aa1c221b82246515b1365f6c9d4/src/utils/parser.ts#L56 you could just use Obsidian's [renderMarkdown](https://github.com/obsidianmd/obsidian-api/blob/38bd84fd66b6bf59d2eab243f27e33d3031ca51c/obsidian.d.ts#L1830) function instead

https://github.com/ka1tzyu/local-quotes/blob/8ea85e696add6aa1c221b82246515b1365f6c9d4/src/main.ts#L21 It seems to me like you're not getting much value out of the QuotesVault cache. Why not just parse the vault once and listen for changes when files with your quote tag are modified?

I think if you optimize the caching a bit, you shouldn't need the "rescan quotes" setting.

https://github.com/ka1tzyu/local-quotes/blob/6c31c984c47d8dc660cfcb487bce9d5d441a1307/src/settings.ts#L113 there are a few typos here.

https://github.com/ka1tzyu/local-quotes/blob/6c31c984c47d8dc660cfcb487bce9d5d441a1307/src/settings.ts#L126 Maybe "Quotes shorter than this length will not be included."

https://github.com/ka1tzyu/local-quotes/blob/6c31c984c47d8dc660cfcb487bce9d5d441a1307/src/settings.ts#L18 I don't think you need to save your quoteVault to disk

https://github.com/ka1tzyu/local-quotes/blob/6c31c984c47d8dc660cfcb487bce9d5d441a1307/src/settings.ts#L16 Same here. You are generating the metadata onload regardless, no reason to cache this to disk.

ka1tzyu:
@liamcain thank for your review. I will fix these issues except two last one's.

I'm a beginner developer so maybe I don't realise something. But below I described the purposes for saving a `blockMetadata` and a `quoteVault`.

1. I decided to save a `quoteVault` because on Obsidian startup, if a note with a quote block is first loaded one (an index note, for example) and it should refresh (get a new content) but there is no layout fully prepared (a note shows up on startup but layoutReady doesn't happen at that time it shows) a quote block cannot fetch quote content. This problem is more bright with large vault with more than 100 notes, on slow devices and mobile.
2. The reason of saving a `blockMetadata` are a `lastUpdate` field, that tracks last quote's update time to refresh with right timing. Also, if a quote content refresh is long term, we need to save current content (author and text) to load it on the next app launch.

I'm sorry for my English, it's not my native one.

ka1tzyu:
@liamcain I already fixed the most of the mistakes, except two described [here](https://github.com/obsidianmd/obsidian-releases/pull/945#issuecomment-1130564107) and your suggestion about updating a `quoteVault` when file modifies. I've a question about the last one. Maybe I should better use 'layout-change' rather than 'modify'. I come to this question when I was thinking about realization.

In my mind, using modify, I need to add an extra property to the `quoteVault` entries - file's path. The plugin allows a user to use more than one file to store one author. So, I can put some quotes of `Author1` in a file `Note1.md`, and then freely add some quotes of this author in another file, `Note2.md`, for example. But if I will to rescan file on it modify, I need to clear author's quotes, because user can do mistakes, or just not completed writing quote, but it'll be written to the `quoteVault`. So I need to save files' paths in `quoteVault` to save ability to put same author in some files.

But if I'll use `layout-change` it means that there are no mistakes in user input, because user end writing, so plugin can freely rescan file.

I'm sorry If I'm asking something stupid. But I really want to do best.

liamcain:
> I decided to save a quoteVault because on Obsidian startup, if a note with a quote block is first loaded one (an index note, for example) and it should refresh (get a new content) but there is no layout fully prepared (a note shows up on startup but layoutReady doesn't happen at that time it shows) a quote block cannot fetch quote content. This problem is more bright with large vault with more than 100 notes, on slow devices and mobile.

Okay, I see.

> The reason of saving a blockMetadata are a lastUpdate field, that tracks last quote's update time to refresh with right timing. Also, if a quote content refresh is long term, we need to save current content (author and text) to load it on the next app launch.

Makes sense. Thanks for clarifying!

> But if I'll use layout-change it means that there are no mistakes in user input, because user end writing, so plugin can freely rescan file.

I don't think you can be sure that there are no mistakes done when layout-changed is called. Layout changed is also problematic because it means you might be updating the vault way more often than you need to (if the user is just navigating around)

> In my mind, using modify, I need to add an extra property to the quoteVault entries - file's path. The plugin allows a user to use more than one file to store one author. So, I can put some quotes of Author1 in a file Note1.md, and then freely add some quotes of this author in another file, Note2.md, for example. But if I will to rescan file on it modify, I need to clear author's quotes, because user can do mistakes, or just not completed writing quote, but it'll be written to the quoteVault. So I need to save files' paths in quoteVault to save ability to put same author in some files.

If you go down this route, I would suggest caching things in a map where the key is the filename. That way, when a modify event happens, you can refresh the contents of just that key very quickly.
`Record<filename, BlockMetadata[]`.

If you're happy with the current performance, I think this can go in now. But let me know if you'll like to attempt refactoring the way you're caching things first.

ka1tzyu:
@liamcain Thank for your advice about modify event. I decided to refactor code to allow update cache on file modify. I'll make it in 2 days.

ka1tzyu:
@liamcain, I updated the plugin, so from [1.3.0](https://github.com/ka1tzyu/local-quotes/releases/tag/1.3.0) quote listing automatically updates when file modifies

liamcain:
Looking good! Just a few more things I noticed:

---

https://github.com/ka1tzyu/local-quotes/blob/c76913c123691f685703d067e7eec4bf4a137c5d/src/types/quote.ts#L21 the typing here doesn't really make sense. A TAbstractFile is either a file or a folder. So it's redundant to have the type `TFile | TAbstractFile`.

https://github.com/ka1tzyu/local-quotes/blob/c76913c123691f685703d067e7eec4bf4a137c5d/src/types/quote.ts#L22 I think you should remove the `convertTAbstractFileToTFile` function, it's not really doing much good for you. Instead of trying to cast this to `TFile | null`, you should add a check:

```
if (!file instanceof TFile) {
  return;
}
```

https://github.com/ka1tzyu/local-quotes/blob/c76913c123691f685703d067e7eec4bf4a137c5d/src/main.ts#L27 I think it might actually make more sense for you to listen to the metadataCache's [`changed`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L2112) event. The metadata cache gets updated shortly after the modify event gets triggered, so there's a chance that you're reading stale data from the metadata cache in your current approach. This will also decrease the chance that getting the file metadata will be `null`

ka1tzyu:
@liamcain, thank you for advices, I fixed all of these in [1.3.2](https://github.com/ka1tzyu/local-quotes/releases/tag/1.3.2)

PR #948, opened 2022-05-07
feat: Add plugin entry validation

liamcain:
This is an updated version of #608 which should hopefully fix the permission issues we've seen in testing.


ericaxu:
Great! Let's see how it goes.

PR #949, opened 2022-05-09
Add Upcoming plugin

charliecm:

## Repo URL

Link to my plugin: https://github.com/charliecm/obsidian-upcoming

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


ericaxu:
Please add it to the bottom, thanks!

liamcain:
Hi, sorry for the delay in getting to this review!

---

https://github.com/charliecm/obsidian-upcoming/blob/a90ec055d6ed31f7429246e049e99297f5ea1d64/main.ts#L35 `getActiveViewOfType` can return `null`. You should check before trying to access `.leaf`

https://github.com/charliecm/obsidian-upcoming/blob/a90ec055d6ed31f7429246e049e99297f5ea1d64/main.ts#L36-L37 Why are you accessing `id` here?

https://github.com/charliecm/obsidian-upcoming/blob/a90ec055d6ed31f7429246e049e99297f5ea1d64/main.ts#L43 to check if the active file is a daily note, you're better off using the [`daily-note-interface` package](https://github.com/liamcain/obsidian-daily-notes-interface) which automatically reads the format and folder path from the daily notes plugin and also handles additional edge cases, like users that have nested formats (`YYYY/MM/YYYY-MM-DD`).

https://github.com/charliecm/obsidian-upcoming/blob/a90ec055d6ed31f7429246e049e99297f5ea1d64/main.ts#L71 If you don't want to use the `checking` variable, you should use `callback` here instead of `checkCallback`.

https://github.com/charliecm/obsidian-upcoming/blob/a90ec055d6ed31f7429246e049e99297f5ea1d64/main.ts#L152 Please [avoid](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml) using `.innerHTML` whenever possible.

charliecm:
@liamcain Thanks for the review! I've addressed all of your suggested changes and am using the daily-note-interface for handling daily notes. I've also added a create notes option in the latest commit.

Re: accessing `id`, I'm trying to exclude the active leaf from being closed along with the leaves previously created by the command. Given that `id` isn't in the current type file, is there a recommended way of accessing it?

liamcain:
> Re: accessing id, I'm trying to exclude the active leaf from being closed along with the leaves previously created by the command. Given that id isn't in the current type file, is there a recommended way of accessing it?

Looking back through what you're trying to do, I think what you have is fine.

---

Code LGTM!

PR #950, opened 2022-05-11
Add Obsidian Weread Plugin

zhaohongxuan:

## Repo URL

Link to my plugin: [https://github.com/zhaohongxuan/obsidian-weread-plugin](https://github.com/zhaohongxuan/obsidian-weread-plugin)

## Release Checklist

- [x]    I have tested the plugin on
    - [ ]    Windows
    - [x]    macOS
    - [ ]    Linux
    - [ ]    Android _(if applicable)_
    - [ ]    iOS _(if applicable)_
-    My GitHub release contains all required files
    - [x]    `main.js`
    - [x]    `manifest.json`
    - [ ]    `styles.css` _(optional)_
- [x]    GitHub release name matches the exact version number specified in my manifest.json (_**Note:** Use the exact verion number, don't include a prefix `v`_)
- [x]    The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [x]    My README.md describes the plugin's purpose and provides clear usage instructions.
- [x]    I have read the tips in [https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md) and have self-reviewed my plugin to avoid these common pitfalls.
- [x]    I have added a license in the LICENSE file.
- [x]    My project respects and is compatible with the original license of any code from other plugins that I'm using.  
- [x] I have given proper attribution to these other projects in my `README.md`.


liamcain:
https://github.com/zhaohongxuan/obsidian-weread-plugin/blob/7a6fd5d0b04e67124710d835f8109052789214bf/main.ts#L60 you might want to include the name of the plugin here so users know what plugin is being unloaded

https://github.com/zhaohongxuan/obsidian-weread-plugin/blob/7a6fd5d0b04e67124710d835f8109052789214bf/styles.css#L1 you might want to remove the css file since you're not using it.

https://github.com/zhaohongxuan/obsidian-weread-plugin/blob/7a6fd5d0b04e67124710d835f8109052789214bf/main.ts#L43 [Prefer async/await over promise chaining](https://github.com/lishid/obsidian-releases/blob/master/plugin-review.md#prefer-asyncawait-over-promise). You're mixing `then/catch` with `await` which can be very confusing. 

https://github.com/zhaohongxuan/obsidian-weread-plugin/blob/7a6fd5d0b04e67124710d835f8109052789214bf/src/fileManager.ts#L46 this function isn't doing anything except wrapping an existing API function

https://github.com/zhaohongxuan/obsidian-weread-plugin/blob/7a6fd5d0b04e67124710d835f8109052789214bf/src/components/wereadLogoutModel.ts#L1 these filenames say "Model" instead of "Modal" which confused me at first

https://github.com/zhaohongxuan/obsidian-weread-plugin/blob/7a6fd5d0b04e67124710d835f8109052789214bf/src/parser/parseResponse.ts#L69 `===`

https://github.com/zhaohongxuan/obsidian-weread-plugin/blob/7a6fd5d0b04e67124710d835f8109052789214bf/src/parser/parseResponse.ts#L10 You should use `window.moment` instead of including your own version of `moment` in the bundle

https://github.com/zhaohongxuan/obsidian-weread-plugin/blob/7a6fd5d0b04e67124710d835f8109052789214bf/src/settings.ts#L7 it seems counterintuitive to me to save the Cookies into your settings object (which gets serialized to JSON). Is there any reason you don't just save them as cookies?

zhaohongxuan:
hi @liamcain , at first I store the cookies as string, but laterly, I found I need to refresh cookie because weread cookie's expire time is very short (1.5 hour), use Object is easy to manipulate single Cookie's key and value.

zhaohongxuan:
Thanks for your review, I have fixed the first 7 review point. and reason I'm not going to fix the last point in above.

liamcain:
Hi again, sorry for such a long delay before getting back to you. Everything looks good, just had one more comment for something I missed earlier.

[import { createProxyMiddleware } from 'http-proxy-middleware';](https://github.com/zhaohongxuan/obsidian-weread-plugin/blob/60bdd2a941f52bef7423e903e0e94ee07ab45927/src/networkManager.ts#L3-L4)  From what I can tell, you are creating the proxy middleware server for 2 reasons:
- to bypass CORS
- To automatically refresh cookie on expiration

Instead of setting up the proxy server, you can use [`requestUrl`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L2680), which will bypass the CORS restriction for you automatically. You'll still need to manually refresh the cookies, but it should be pretty straightforward.

zhaohongxuan:
Hi @liamcain , Tons of thanks for your advice, I have replaced the `http-proxy-middleware` server with `requestUrl` in PR: https://github.com/zhaohongxuan/obsidian-weread-plugin/pull/57,  I had tested and it works fine🥰.

PR #951, opened 2022-05-12
Update text-snippets-obsidian description

ArianaKhit:
Edit description for existing plugin Text Snippets (text-snippets-obsidian)


ericaxu:
@ArianaKhit Thanks for the edit! I added "and" before the last part so it reads a little better.

PR #953, opened 2022-05-13
Add nuke-orphans plugin

sandorex:

## Repo URL

Link to my plugin: https://github.com/sandorex/nuke-orphans-plugin

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
https://github.com/sandorex/nuke-orphans-plugin/blob/044be34c22651521181406337ad539b795e35830/src/main.ts#L18 I think this can be simplified by using `Object.values` instead of `entries` in both instances, no?

https://github.com/sandorex/nuke-orphans-plugin/blob/044be34c22651521181406337ad539b795e35830/src/main.ts#L38 you shouldn't need to pass `this` as the second argument to `.filter` here

https://github.com/sandorex/nuke-orphans-plugin/blob/044be34c22651521181406337ad539b795e35830/src/main.ts#L46 `===`

https://github.com/sandorex/nuke-orphans-plugin/blob/044be34c22651521181406337ad539b795e35830/src/settings.ts#L49 IMO, it's better and simpler to just always read this option instead of providing your own setting. 

https://github.com/sandorex/nuke-orphans-plugin/blob/044be34c22651521181406337ad539b795e35830/src/settings.ts#L53
If you want to remove the `ts-ignore` here, I suggest you add a TS module override like [this](https://github.com/kometenstaub/linked-data-helper/blob/f0704cfda75e8a98ac29d52a7490a77053a41f8d/src/interfaces.ts#L32).

https://github.com/sandorex/nuke-orphans-plugin/blob/044be34c22651521181406337ad539b795e35830/src/trash_modal.ts#L44 You should read the `this.app.vault.getConfig("trashOption")` to see if the user wants deleted files in their local trash or in the Obsidian trash. (Also not a public API)

sandorex:
> https://github.com/sandorex/nuke-orphans-plugin/blob/044be34c22651521181406337ad539b795e35830/src/main.ts#L18 I think this can be simplified by using Object.values instead of entries in both instances, no?

Improved the readability significantly, thanks

> https://github.com/sandorex/nuke-orphans-plugin/blob/044be34c22651521181406337ad539b795e35830/src/main.ts#L38 you shouldn't need to pass this as the second argument to .filter here

Fixed

> https://github.com/sandorex/nuke-orphans-plugin/blob/044be34c22651521181406337ad539b795e35830/src/main.ts#L46 ===

Fixed

> https://github.com/sandorex/nuke-orphans-plugin/blob/044be34c22651521181406337ad539b795e35830/src/settings.ts#L49 IMO, it's better and simpler to just always read this option instead of providing your own setting.

Now reading `attachmentFolderPath` from config by default and allows overriding it

> https://github.com/sandorex/nuke-orphans-plugin/blob/044be34c22651521181406337ad539b795e35830/src/settings.ts#L53
If you want to remove the ts-ignore here, I suggest you add a TS module override like [this](https://github.com/kometenstaub/linked-data-helper/blob/f0704cfda75e8a98ac29d52a7490a77053a41f8d/src/interfaces.ts#L32).

Added module override

> https://github.com/sandorex/nuke-orphans-plugin/blob/044be34c22651521181406337ad539b795e35830/src/trash_modal.ts#L44 You should read the this.app.vault.getConfig("trashOption") to see if the user wants deleted files in their local trash or in the Obsidian trash. (Also not a public API)

Fixed

PR #956, opened 2022-05-14
Add obsidian-sequence-hotkeys plugin

moolmanruan:

## Repo URL

Link to my plugin: https://github.com/moolmanruan/obsidian-sequence-hotkeys

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
Really cool plugin, nice job!

Just a few small nitpicks but I think it's good to go in.

---

https://github.com/moolmanruan/obsidian-sequence-hotkeys/blob/6e80f34878cebd910e227936c383306bf8963229/hotkey-manager.ts#L67 `!==`

https://github.com/moolmanruan/obsidian-sequence-hotkeys/blob/6e80f34878cebd910e227936c383306bf8963229/hotkey-manager.ts#L50 Minor optimization, but you already know that it's a partial match, so you don't need to compare the arrays again, just check to see if the currentSequence is the same length as any of the matching chordSequences

liamcain (approved):

moolmanruan:
Thanks for the review, @liamcain. Nice catch with the optimisation. I've made the requested changes.

PR #958, opened 2022-05-14
Add state switcher plugin

lijyze:

## Repo URL

Link to my plugin: https://github.com/lijyze/obsidian-state-switcher

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lijyze (commented):
Update `description`

liamcain:
https://github.com/lijyze/obsidian-state-switcher/blob/fadee9962a9d6cb324571c825020640ea7606def/src/setting.ts#L16 I suggest you don't include this in your DEFAULT_SETTINGS. The way your loading your settings means that users won't be able to remove this property from their settings (since it always gets merged in when the plugin loads). Instead, I suggest you just add a check inside your load plugin. If the `loadData` returns `null`, then add that setting to get the user started. This way, if the user already has some data saved, it won't get re-added.

https://github.com/lijyze/obsidian-state-switcher/blob/fadee9962a9d6cb324571c825020640ea7606def/src/main.ts#L23 this needs a try/catch block since the promise can also be rejected if the modal is canceled.

https://github.com/lijyze/obsidian-state-switcher/blob/fadee9962a9d6cb324571c825020640ea7606def/src/main.ts#L24 you should make sure that the result of `.find()` isn't `null` before trying to access `.key`

lijyze:
Thank you for your earnest work. I become get confused with some of them.

1. With structure of data
    ```ts
    {
      stateMaps: [
        { id: '0', key: "state", values: ["waiting", "ongoing", "completed"] }
      ],
    };
    ```
    I don't see why user won't be able to remove default map in DEFAULT_SETTINGS. If they remove it, data will be save as 

    ```ts
    {
      stateMaps: [],
    };
    ```
Which means default map has deleted actually.

2. I already finish this.

3. Datas of the first Suggester comes from `nonEmptyField`, I select on of them, this data considered as `id`, then I use `find` to get values which its `id` exactly equal to I just selected from the same set `nonEmptyField`. I don't why this find will get `null` in this case. 

    ```ts
    const nonEmptyField = this.settings.stateMaps.filter((map) => map.key);
    
    if (!nonEmptyField.length) {
	    new Notice('No map founded, please check your config');
	    return ;
    }
    
    const ids = nonEmptyField.map((field) => '' + field.id);
    const keys = nonEmptyField.map((field) => field.key);
    
    let selectedId: string;
    try {
	    selectedId = await Suggester.Suggest(this.app, keys, ids);
    } catch (error) {
	    console.log(error)
    }
    
    if (!selectedId) return ;

    const selectedKey = selectedKey = nonEmptyField.find((field) => field.id === selectedId).key;
    
    const values = nonEmptyField.find((field) => field.id === selectedId).values.filter((values) => values);
    ```

---

I think these might be blind area of my knowledge, please make me clear if it won't bother you. Thanks!

liamcain:
1. Good point! I was thinking it wasn't inside the `stateMaps` property for some reason. That's what I get for reviewing so late at night 🤦 
2. Looks good
3. Yeah, I was mainly concerned with the case where `selectedId` could be null, but you fixed that with the early exit so it seems fine to me now.

PR #959, opened 2022-05-15
Add Note Auto Creator plugin

SimonTC:

## Repo URL

Link to my plugin: https://github.com/SimonTC/obsidian-note-autocreation

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
https://github.com/SimonTC/obsidian-note-autocreation/blob/f8d574d1b6064fe06ebf7041db1f4158ade65f67/src/main.ts#L55 I'm pretty sure you want to register this to the EditorSuggest's scope from within its constructor instead (see: [nldates](https://github.com/argenos/nldates-obsidian/blob/e621a9609211174b964230f5bb890faa574fb7f1/src/suggest/date-suggest.ts#L28)

https://github.com/SimonTC/obsidian-note-autocreation/blob/f8d574d1b6064fe06ebf7041db1f4158ade65f67/src/main.ts#L53 what's the observer for?

https://github.com/SimonTC/obsidian-note-autocreation/blob/f8d574d1b6064fe06ebf7041db1f4158ade65f67/src/ObsidianInterop.ts#L40 You're passing in an empty string for the sourcePath here. I think you want to pass in the activeFile's path instead. This allows you to create links relative to the current path, instead of always relative to the vault root.

SimonTC:
Thanks for the review!

> https://github.com/SimonTC/obsidian-note-autocreation/blob/f8d574d1b6064fe06ebf7041db1f4158ade65f67/src/main.ts#L55 I'm pretty sure you want to register this to the EditorSuggest's scope from within its constructor instead (see: [nldates](https://github.com/argenos/nldates-obsidian/blob/e621a9609211174b964230f5bb890faa574fb7f1/src/suggest/date-suggest.ts#L28))

You are right. This will be fixed.

> https://github.com/SimonTC/obsidian-note-autocreation/blob/f8d574d1b6064fe06ebf7041db1f4158ade65f67/src/main.ts#L53 what's the observer for?

It was my clunky way to figure out which of the suggestions were selected when the user pressed tab. After looking at the code for nldates I see that I can access that through `this.suggestions`. I will do that instead. I wasn't aware of that field since it is not part of the public API.

> https://github.com/SimonTC/obsidian-note-autocreation/blob/f8d574d1b6064fe06ebf7041db1f4158ade65f67/src/ObsidianInterop.ts#L40 You're passing in an empty string for the sourcePath here. I think you want to pass in the activeFile's path instead. This allows you to create links relative to the current path, instead of always relative to the vault root.

I thought setting the sourcePath just acted as a helper here and would not have an influence on the path of the returned file.
I did it like this since my assumption was that the value of the linkPath always would be the path from the root. I am now working on a feature where I take the config value for default note location into account so I will change this to pass in the activeFile's path to make sure everything works correctly.

SimonTC:
I have released version 1.0.1 which fixes the issues you raised

PR #960, opened 2022-05-15
Add `Packrat plugin`

therden:

## Repo URL

Link to my plugin: [https://github.com/therden/packrat](https://github.com/therden/packrat)

## Release Checklist

    - [x]     Windows
    -    macOS
    - [x]     Linux
    -    Android _(if applicable)_
    -    iOS _(if applicable)_
- [x]     My GitHub release contains all required files
    - [x]     `main.js`
    - [x]     `manifest.json`
    - [x]     `styles.css` _(optional)_
- [x]     GitHub release name matches the exact version number specified in my manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
- [x]     The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [x]     My README.md describes the plugin's purpose and provides clear usage instructions.
- [x]    I have read the tips in [https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md) and have self-reviewed my plugin to avoid these common pitfalls.
- [x]    I have added a license in the LICENSE file.
- [-]  My project respects and is compatible with the original license of any code from other plugins that I'm using.  
- [-] I have given proper attribution to these other projects in my `README.md`.


liamcain:
https://github.com/therden/packrat/blob/be1b54105eabd11993c72d2389b4883a4c33fc88/main.ts#L30 under the hood, this ID gets prefixed with the plugin ID, so no need to include the plugin name here.

https://github.com/therden/packrat/blob/be1b54105eabd11993c72d2389b4883a4c33fc88/main.ts#L37-L38 It's better to invert this conditional instead of leaving an empty `if` statement. `if (activeFile && activeFile.extension === "md")`

https://github.com/therden/packrat/blob/be1b54105eabd11993c72d2389b4883a4c33fc88/main.ts#L43 take a look at [these docs on how to use `checkCallback`](https://marcus.se.net/obsidian-plugin-docs/guides/commands#conditional-commands). You'll want to make sure to always return a boolean value if `checking` is true. So I think you're just missing a return statement at the end of this function `return false;`

Given that your goal is to modify the active file, you might be better off using `editorCheckCallback ` instead of `checkCallback`. Then you could modify the active file using the Editor class. It would require refactoring the `ProcessCompletedRecurringTasks` function quite a bit, but it would have the advantage of making the command `undo-able`. Plus it wouldn't mess with the view's state, so folds will stay intact.

https://github.com/therden/packrat/blob/be1b54105eabd11993c72d2389b4883a4c33fc88/main.ts#L86 `===`

https://github.com/therden/packrat/blob/be1b54105eabd11993c72d2389b4883a4c33fc88/main.ts#L88 when checking for the existence of something, you shouldn't check if it's greater than `0` since it could appear at the beginning. So this should either be `!== -1` or you can use `.contains()` which is an Array polyfill included in the Obsidian API.

https://github.com/therden/packrat/blob/be1b54105eabd11993c72d2389b4883a4c33fc88/main.ts#L116 Since `archiveFilename` is stored in settings (and user configurable), you should use [`getFirstLinkpathDest`](https://github.com/obsidianmd/obsidian-api/blob/38bd84fd66b6bf59d2eab243f27e33d3031ca51c/obsidian.d.ts#L2075) to find the TFile instead of `getAbstractFileByPath`. It's a bit smarter and will normalize the path for you.

---

Looks good, nice job, and good job on the README, it's very clear and concise.

therden:
Thank you for the feedback Liam!  

After making the changes described below, I've uploaded a new release (1.1.0) to my Github repo.

> https://github.com/therden/packrat/blob/be1b54105eabd11993c72d2389b4883a4c33fc88/main.ts#L30 under the hood, this ID gets prefixed with the plugin ID, so no need to include the plugin name here.

Done!

> 
> https://github.com/therden/packrat/blob/be1b54105eabd11993c72d2389b4883a4c33fc88/main.ts#L37-L38 It's better to invert this conditional instead of leaving an empty `if` statement. `if (activeFile && activeFile.extension === "md")`

Done!

> 
> https://github.com/therden/packrat/blob/be1b54105eabd11993c72d2389b4883a4c33fc88/main.ts#L43 take a look at [these docs on how to use `checkCallback`](https://marcus.se.net/obsidian-plugin-docs/guides/commands#conditional-commands). You'll want to make sure to always return a boolean value if `checking` is true. So I think you're just missing a return statement at the end of this function `return false;`

Done!

> 
> Given that your goal is to modify the active file, you might be better off using `editorCheckCallback ` instead of `checkCallback`. Then you could modify the active file using the Editor class. It would require refactoring the `ProcessCompletedRecurringTasks` function quite a bit, but it would have the advantage of making the command `undo-able`. Plus it wouldn't mess with the view's state, so folds will stay intact.

I'd prefer to keep `checkCallback` for now -- but I'll definitely consider switching to `editorCheckCallback` in future.

> 
> https://github.com/therden/packrat/blob/be1b54105eabd11993c72d2389b4883a4c33fc88/main.ts#L86 `===`

"Doh!" -- and Done!

> 
> https://github.com/therden/packrat/blob/be1b54105eabd11993c72d2389b4883a4c33fc88/main.ts#L88 when checking for the existence of something, you shouldn't check if it's greater than `0` since it could appear at the beginning. So this should either be `!== -1` or you can use `.contains()` which is an Array polyfill included in the Obsidian API.

Done!  

A **Packrat** trigger actually _can't_ be at the beginning of the line, as the first (non-space) characters of a markdown task are always `- [` -- however, I agree that it is better to test for the value returned by `indexOf` when the search string is _not_ found.

> 
> https://github.com/therden/packrat/blob/be1b54105eabd11993c72d2389b4883a4c33fc88/main.ts#L116 Since `archiveFilename` is stored in settings (and user configurable), you should use [`getFirstLinkpathDest`](https://github.com/obsidianmd/obsidian-api/blob/38bd84fd66b6bf59d2eab243f27e33d3031ca51c/obsidian.d.ts#L2075) to find the TFile instead of `getAbstractFileByPath`. It's a bit smarter and will normalize the path for you.

I looked at the API documentation, tried to find examples of use 'in the wild', and experimented quite a bit, but I utterly failed in my efforts to use `getFirstLinkpathDest`.  

As an alternative, I've added `try/catch` to `ProcessCompletedRecurringTasks` so that errors are both displayed and logged, and updated README.md to clarify that the user will need to pre-create any non-root directory in which they want the archive file to be stored.

liamcain:
> I looked at the API documentation, tried to find examples of use 'in the wild', and experimented quite a bit, but I utterly failed in my efforts to use getFirstLinkpathDest

Sorry, the documentation for that could be a little better.

```ts
getFirstLinkpathDest(linkpath: string, sourcePath: string): TFile | null;
```

**linkpath:** This is the file path to the file you're looking for
**sourcePath:** This function is used for [[links]] so the sourcePath would either be the activeFile's path, or `""` if you want the path to be relative to the root of your vault.

Also, I was incorrect when I said it normalizes that path for you, all it really does is add `.md` to the lookup if it's missing for the user input.

I think what you have is fine, though it might not be the user's intent to have the file automatically created if they have the path wrong. You might want to just display a `new Notice("Unable to find archival file")` instead.

PR #961, opened 2022-05-17
Add `Obsidian Image Gallery` plugin.

lucaorio:

## Repo URL

Link to my plugin: https://github.com/lucaorio/obsidian-image-gallery

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
https://github.com/lucaorio/obsidian-image-gallery/blob/9f65d10cfdb0caee82f85a7b8e81998c7a7157e8/src/img-gallery-renderer.ts#L39 You might want to consider using a more familiar syntax for configuring your codeblock. Something like YAML or even JSON would be better to be consistent with other plugin codeblocks. Otherwise, it's tricky for users to remember the syntax they need for configuring each plugin.

https://github.com/lucaorio/obsidian-image-gallery/blob/9f65d10cfdb0caee82f85a7b8e81998c7a7157e8/src/img-gallery-renderer.ts#L45 you should surface this error to the user. Maybe show it in the codeblock as a message

https://github.com/lucaorio/obsidian-image-gallery/blob/9f65d10cfdb0caee82f85a7b8e81998c7a7157e8/src/img-gallery-renderer.ts#L64 You probably want to call [`normalizePath`](https://github.com/obsidianmd/obsidian-api/blob/38bd84fd66b6bf59d2eab243f27e33d3031ca51c/obsidian.d.ts#L2237) to sanitize the user input here.

https://github.com/lucaorio/obsidian-image-gallery/blob/9f65d10cfdb0caee82f85a7b8e81998c7a7157e8/src/img-gallery-renderer.ts#L64 I don't see anywhere in this function where you filter the folder to only include images. You probably want to filter out the list by file extension somewhere.

lucaorio:
Hi @liamcain thank you for the suggestions!

I addressed all of the above in the [`Plugin review`](https://github.com/lucaorio/obsidian-image-gallery/pull/1) pull request; please let me know if everything now looks good to you.

liamcain:
https://github.com/lucaorio/obsidian-image-gallery/blob/1cdba095d4ed00bf286817b418044b8fb5e6d3f0/src/img-gallery-renderer.ts#L37 No need to add `js-yaml` to your plugin bundle. The Obsidian API exports a [`parseYaml`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L2323) function that you can use.

https://github.com/lucaorio/obsidian-image-gallery/blob/1cdba095d4ed00bf286817b418044b8fb5e6d3f0/src/img-gallery-renderer.ts#L8 You might want to initialize this to `[]` just so the typing is correct

https://github.com/lucaorio/obsidian-image-gallery/blob/1cdba095d4ed00bf286817b418044b8fb5e6d3f0/src/img-gallery-renderer.ts#L56-L57 I'm a little concerned that you have some an issue with mixed types here. I think loading these values from YAML will give you a `number` but your fallback values are strings. If they are numbers, then you'll also need to use `??` instead of `||` when providing default values. Otherwise you'll run into bugs where you users can't pass `0` in their config.

lucaorio:
Hey @liamcain, I addressed all your great points [here](https://github.com/lucaorio/obsidian-image-gallery/pull/3).

I switched from `||` to `??`, but also ended up in defaulting to numbers as they go through template literals whenever necessary. 

Please let me know how it looks now, and thanks again for the code review; I'm not an engineer and definitely got out of my comfort zone with this one. :)

liamcain:
https://github.com/lucaorio/obsidian-image-gallery/blob/bf328d9f180727f77b6982b76b3e2d6eb0f529f8/src/img-gallery-renderer.ts#L82 you'll need to remove the `.` in `.tiff` and `.tif` for these to match

https://github.com/lucaorio/obsidian-image-gallery/blob/bf328d9f180727f77b6982b76b3e2d6eb0f529f8/src/img-gallery-renderer.ts#L163 instead of setting `.innerHTML` (which is [dangerous](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml)), you can set the text from within `createEl`:

```ts
wrapper.createEl('p', { text: `(Error) Image Gallery: ${error}` });
```

lucaorio:
Thank you, @liamcain ! I just [finished addressing those](https://github.com/lucaorio/obsidian-image-gallery/pull/4/commits/481b328e86e974f4bf3c5a67ad4180ebca675c0b) too 👍

PR #962, opened 2022-05-17
Add `Media DB Plugin` plugin

mProjectsCode:

## Repo URL

Link to my plugin: https://github.com/mProjectsCode/obsidian-media-db-plugin

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
https://github.com/mProjectsCode/obsidian-media-db-plugin/blob/344c1ebf6c7e20696063ca73b4a6701e226ec7ea/README.md?plain=1#L38 I think this could be a lot more clear if you turned this into a markdown table for people to read. Maybe include the bullets that are just "notes" below the table.

https://github.com/mProjectsCode/obsidian-media-db-plugin/blob/344c1ebf6c7e20696063ca73b4a6701e226ec7ea/src/main.ts#L60 remove

https://github.com/mProjectsCode/obsidian-media-db-plugin/blob/344c1ebf6c7e20696063ca73b4a6701e226ec7ea/src/main.ts#L62 You aren't doing anything with this result. I'm guessing this can be removed.

https://github.com/mProjectsCode/obsidian-media-db-plugin/blob/344c1ebf6c7e20696063ca73b4a6701e226ec7ea/src/main.ts#L81 you can use `cachedRead` here instead of `read`.

https://github.com/mProjectsCode/obsidian-media-db-plugin/blob/344c1ebf6c7e20696063ca73b4a6701e226ec7ea/src/main.ts#L94 you should use `getUnpinnedLeaf` here instead.

https://github.com/mProjectsCode/obsidian-media-db-plugin/blob/344c1ebf6c7e20696063ca73b4a6701e226ec7ea/src/main.ts#L96 maybe add a bit more context here, including the plugin name. e.g. `[media-db-plugin] No active leaf, not creating media db note`

https://github.com/mProjectsCode/obsidian-media-db-plugin/blob/344c1ebf6c7e20696063ca73b4a6701e226ec7ea/src/utils/Utils.ts#L7 unused

mProjectsCode:
I released version 0.1.8 which fixes the things you flagged @liamcain

mProjectsCode:
I resolved the merge conflict that occurred because other plugin PRs were merged before this one

liamcain:
All the fixes you made look great. I just did another pass of the changes you've made since I last reviewed.

https://github.com/mProjectsCode/obsidian-media-db-plugin/blob/8b4a50647db02a5278de8ab7520ba1136f096be2/src/utils/YAMLConverter.ts#L1 Why did you end up creating your own yaml converter class (instead of using obsidian's stringify function)?

https://github.com/mProjectsCode/obsidian-media-db-plugin/blob/8b4a50647db02a5278de8ab7520ba1136f096be2/src/main.ts#L91 You should make sure that the result of getAbstractFileByPath isn't null before calling `delete`

https://github.com/mProjectsCode/obsidian-media-db-plugin/blob/8b4a50647db02a5278de8ab7520ba1136f096be2/src/main.ts#L130 It's confusing that you're calling this a leaf, since this variable contains a `TFile` instance.

https://github.com/mProjectsCode/obsidian-media-db-plugin/blob/8b4a50647db02a5278de8ab7520ba1136f096be2/src/main.ts#L131 name isn't optional, however, there might not be an active file so you should check that `getActiveFile` didn't return null

https://github.com/mProjectsCode/obsidian-media-db-plugin/blob/8b4a50647db02a5278de8ab7520ba1136f096be2/src/main.ts#L142 This is bad casting. You shouldn't cast a plain object into instance of a class. Either find a way to create a real instance of the class that you need, or change the type definition of that function to accept some interface with the subset of properties that you need.

mProjectsCode:
@liamcain Thanks for the Review.

The entire update note thing is still in development and not yet on the release branch, so I did not optimize it yet.

I wrote my own YAML converter because I did not find an option in obsidians converter to always put quotation marks around strings.

That being said I will try to implement the changes later today.

mProjectsCode:
@liamcain 
All the things you flagged should be fixed in version 0.1.10 :)

liamcain:
No need to keep resolving the merge conflicts, I'll do that right before it gets merged!

liamcain:
https://github.com/mProjectsCode/obsidian-media-db-plugin/blob/6f5f198a5572245c7a01a97ddc33d9aa9892c980/src/api/apis/SteamAPI.ts#L84 You should use `===` here instead.

---

Latest fixes look great, so I think this is good to go in.

PR #964, opened 2022-05-19
Add Semantic Line Breaker Plugin

chrisgrieser:

## Repo URL

Link to my plugin: https://github.com/chrisgrieser/obsidian-sembr

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4155,5 +4155,12 @@
         "author": "YISH",
         "description": "A enhancing export plugin base on pandoc for Obsidian. It allows to export to formats like Html, DOCX, ePub and PDF or Hugo.",
         "repo": "mokeyish/obsidian-enhancing-export"
-  }
+    },
+    {
+        "id": "obsidian-sembr",
+        "name": "Semantic Line Breaker",
+        "author": "pseudometa",
+        "description": "Apply and remove Semantic Line Breaks (SemBr).",
+        "repo": "chrisgrieser/obsidian-sembr"
+    }
 ]


liamcain:
Hi, kudos on the really great README!

---

https://github.com/chrisgrieser/obsidian-sembr/blob/a4d9c2d63391ef2374c4a0e5ef8d8859111c344d/README.md?plain=1#L23 sorry for the pedantry, but I think this should be `complying with` rather than `complying to` (or even `abiding by`)

https://github.com/chrisgrieser/obsidian-sembr/blob/a4d9c2d63391ef2374c4a0e5ef8d8859111c344d/main.ts#L38 I really don't think this is the correct approach for properly filtering out content that shouldn't have line breaks. You'll also need to ignore code fences, indented code blocks, block quotes, tables, and maybe more. I think the approach you'll want to take would involve traversing the Abstract Syntax Tree generated by code mirror. Then you can ensure that you are only processing paragraph nodes. It will be quite a bit trickier and more complex than using regex, but you'll have a lot fewer issues to sort out.

```ts
import { ensureSyntaxTree } from '@codemirror/language';

const tree =  ensureSyntaxTree(editor.cm.state);
tree.iterate({

  // Check for paragraph notes...
});
```

I'm not sure of any plugins that currently do something exactly like this, but you might be able to get use nothingislost's plugins as guidance. (see: [cm6-attributes]( <https://github.com/nothingislost/obsidian-cm6-attributes/blob/3336ee508b5d999457b5c97818a16381000f4ed1/src/main.ts#L85>))

Then you can iterate over the tree, similar to this: As you iterate, you'll collect all the changed lines in a list that you can then write back to the file using an EditorTransaction.

chrisgrieser:
closing this for now, since I actually don't use it myself anymore, because I found a different setup that works for me.

PR #965, opened 2022-05-21
add obsidian-ics

muness:

## Repo URL

Link to my plugin: https://github.com/muness/obsidian-ics

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
https://github.com/muness/obsidian-ics/blob/c41564ccfc44bb771f2ddb29116daa29a55d9930/src/main.ts#L61 You can remove the console.log

https://github.com/muness/obsidian-ics/blob/c41564ccfc44bb771f2ddb29116daa29a55d9930/src/main.ts#L63 You should add some error handling for this `request()` function

https://github.com/muness/obsidian-ics/blob/c41564ccfc44bb771f2ddb29116daa29a55d9930/src/main.ts#L67 You can remove this

https://github.com/muness/obsidian-ics/blob/c41564ccfc44bb771f2ddb29116daa29a55d9930/src/main.ts#L51-L52 while there's no rule against it, I suggest removing the default hotkey and maybe mentioning a recommended hotkey in your README. Adding a hotkey for the user can end up causing issues of the user already has that bound to something else and they might be confused why their setup is broken all of a sudden.

https://github.com/muness/obsidian-ics/blob/c41564ccfc44bb771f2ddb29116daa29a55d9930/src/icalUtils.ts#L2 you should use `window.moment` instead to avoid including your own instance of `Moment.js`. Obsidian also has the `window.moment` instance localized based on the user settings, so it's best to avoid this.

https://github.com/muness/obsidian-ics/blob/c41564ccfc44bb771f2ddb29116daa29a55d9930/src/icalUtils.ts#L70 Isn't `dates` an array? You shouldn't use `for ... in` here. That is for iterating over objects. You should be using `for ... of`

ericaxu:
Closing this since development seems to have stopped: https://github.com/muness/obsidian-ics/commit/f3d796df3ee828517da33d9ef762880398e0ae98

Thanks for working on this, @muness !

muness:
@ericaxu - I've picked this up again and am working through the feedback above. I assume I should start a new PR when am ready, right?

ericaxu:
> @ericaxu - I've picked this up again and am working through the feedback above. I assume I should start a new PR when am ready, right?

I've reopened the PR for you!

muness:
Thank you, it's also at https://github.com/obsidianmd/obsidian-releases/pull/2208 :)

joethei:
closing this again as it's a duplicate

PR #966, opened 2022-05-22
Update community-plugins.json

osadasami:

## Repo URL

Link to my plugin: https://github.com/osadasami/obsidian-selected-search-highlight

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4155,5 +4155,13 @@
         "author": "YISH",
         "description": "A enhancing export plugin base on pandoc for Obsidian. It allows to export to formats like Html, DOCX, ePub and PDF or Hugo.",
         "repo": "mokeyish/obsidian-enhancing-export"
-  }
+    },
+    {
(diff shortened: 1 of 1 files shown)


liamcain:
https://github.com/osadasami/obsidian-selected-search-highlight/blob/e68455d2f8211bfd1eab301c5d22de278b6bb58e/main.ts#L5 Binding this to click won't work reliably since the user can focus and use the search panel from the keyboard. I'm not totally sure of a better way that you can hook into the search panel currently. Another approach would be adding a callback to [`layout-change`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3773), however, this won't trigger when the user triggers a search.

PR #967, opened 2022-05-22
Rename Zotero Desktop Connector plugin

mgmeyers:
Per the request of Zotero's devs, I'm renaming my plugin to create less confusion between my plugin and Zotero's native "Zotero Connector"


phibr0:
If you change the ID, every User's settings will be reset and the Plugin needs to be reinstalled. Are you sure, that changing the display name isn't enough?

mgmeyers:
Ah, good call @phibr0! I've kept the old ID to reduce friction

PR #968, opened 2022-05-22
added excalibrain

zsviczian:

## Repo URL

Link to my plugin: https://github.com/zsviczian/excalibrain

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


claremacrae:
Sounds intriguing! Any chance you could add the URL to the template above, I’m really curious to see it!

zsviczian:
https://github.com/zsviczian/excalibrain

liamcain:
https://github.com/zsviczian/excalibrain/blob/f7022f5576cc48ff5db88b9665af3d73921e791d/src/main.ts#L178 `app.internalPlugins.getPluginById` can return `null` if the core plugin is disabled

https://github.com/zsviczian/excalibrain/blob/f7022f5576cc48ff5db88b9665af3d73921e791d/src/main.ts#L228 maybe use checkCallback here instead so that the command isn't available if excalidraw isn't available

https://github.com/zsviczian/excalibrain/blob/f7022f5576cc48ff5db88b9665af3d73921e791d/src/main.ts#L252 you should coerce this to a boolean

https://github.com/zsviczian/excalibrain/blob/f7022f5576cc48ff5db88b9665af3d73921e791d/src/main.ts#L284 it's better to use `app.commands.executeCommandById()` here

https://github.com/zsviczian/excalibrain/blob/f7022f5576cc48ff5db88b9665af3d73921e791d/src/main.ts#L300 you could create a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) so that you won't have to ts-ignore here.  

https://github.com/zsviczian/excalibrain/blob/f7022f5576cc48ff5db88b9665af3d73921e791d/src/utils/fileUtils.ts#L11 you probably want to limit the `lastIndexOf` to only look at the last path segment. Otherwise you'll get false positives from `some/path/with.dots.in.the.path/like-this`

https://github.com/zsviczian/excalibrain/blob/f7022f5576cc48ff5db88b9665af3d73921e791d/src/utils/dataview.ts#L69 I don't think you need to use `any` for dataview objects. I think you can import the typing information from `obsidian-dataview`.

zsviczian:
Thanks! Awesome! I always learn so much from this process (and on OMG). I've implemented the changes.

https://github.com/zsviczian/excalibrain/releases/tag/0.0.18

PR #974, opened 2022-05-26
Add obsidian-markbase plugin

markbaseteam:

## Repo URL

Link to my plugin: https://github.com/markbaseteam/obsidian-markbase

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
https://github.com/markbaseteam/obsidian-markbase/blob/5364b05b2a620f9433f2786b9eadf37884621046/src/main.ts#L266-L274 I don't think these try/catches should be nested. This error looks to just be about zipping the files so I suggest just wrapping this around `zipDirectory`

https://github.com/markbaseteam/obsidian-markbase/blob/5364b05b2a620f9433f2786b9eadf37884621046/package.json#L28 please make sure that `electron` is not getting pulled into the output `main.js`. It is already marked as an external lib in the esbuild config, so just double-check since that would be bad (electron is a large lib)

https://github.com/markbaseteam/obsidian-markbase/blob/5364b05b2a620f9433f2786b9eadf37884621046/src/components/createProjectModal.ts#L43 since the slug needs to be unique, maybe you can do a debounced check for uniqueness while the user is choosing a slug, that way they get immediate feedback.

https://github.com/markbaseteam/obsidian-markbase/blob/5364b05b2a620f9433f2786b9eadf37884621046/src/helpers/modals.ts#L8 note, this isn't the correct keyboard shortcut on MacOS. It's `command+option+i`.

---

I don't see any showstoppers, so just let me know if the build output is fine (and doesn't have a version of `electron` bundled at all) and it can go in.

markbaseteam:
@liamcain Yep no electron in the build output - please merge whenever works best for you

PR #976, opened 2022-05-26
Add Zotero link integration

MunGell:

## Repo URL

Link to my plugin: https://github.com/MunGell/obsidian-zotero

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
https://github.com/MunGell/obsidian-zotero/blob/7f17036c513bbdc5ef9631d49e2cf0197a5a9c48/src/ZoteroConnector.ts#L13 you can turn this into a [`getter`](https://www.typescriptlang.org/docs/handbook/2/classes.html#getters--setters) instead of setting it in the constructor. That way the value can be derived from `this.settings` without needing to get recomputed.

https://github.com/MunGell/obsidian-zotero/blob/7f17036c513bbdc5ef9631d49e2cf0197a5a9c48/src/ZoteroPlugin.ts#L32 if you fix the above, then you don't need to call `init` from within saveSettings

https://github.com/MunGell/obsidian-zotero/blob/7f17036c513bbdc5ef9631d49e2cf0197a5a9c48/src/ZoteroItem.ts#L54 Before doing this, you should call [`sanitizeHTMLToDom`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L2726). That way you aren't leaving this method exposed to a security risk.

MunGell:
Thanks for review @liamcain 

Your suggestions were addressed in the code

community-plugins.json
```diff
@@ -4272,5 +4272,12 @@
         "author": "Zsolt Viczian",
         "description": "ExcaliBrain is inspired by TheBrain and Breadcrumbs. It is an interactive, structured mind-map of your Obsidian Vault generated based on the folders and files in your Vault by interpreting the links, dataview fields, tags and YAML front matter in your markdown files.",
         "repo": "zsviczian/excalibrain"
+    },
+    {
+        "id": "obsidian-zotero",
+        "name": "Zotero",
```
liamcain, line 4278: Sorry I didn't notice this earlier, but–

I think you might want to change the plugin name (something like Zotero Link is fine). Having it as just "Zotero" could imply it's an official product of the Zotero team and might cause a bit of confusion from users.

MunGell: Thank you for highlighting this, I've renamed it to ZoteroBridge here and in the repository now.

MunGell:
Unfortunately, I have to close this PR for the moment. I am working on a refactored version of the plugin that might have a different name (Zotero Link), so it is better to wait until it will be done to avoid user confusion. Will open new PR soon.

PR #978, opened 2022-05-27
Add File Hider plugin to Community Plugin List

Oliver-Akins:

## Repo URL

Link to my plugin: https://github.com/Oliver-Akins/file-hider

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
https://github.com/Oliver-Akins/file-hider/blob/716e12aa4051af0a2c481839166d75bfdb2501a4/src/utils.ts#L11 you can improve the typing here and make this `createStyleLine(type: "folder" | "file", path: string)`

https://github.com/Oliver-Akins/file-hider/blob/716e12aa4051af0a2c481839166d75bfdb2501a4/src/utils.ts#L24 don't use `var`. This should be `const` instead. Also, don't use `for ... in` to iterate over a list. That's for iterating over the properties of an object. You should use `for ... of` instead.

https://github.com/Oliver-Akins/file-hider/blob/716e12aa4051af0a2c481839166d75bfdb2501a4/src/utils.ts#L24 I think you can add an `id` to your stylesheet, then you don't have to traversal all the stylesheets looking at the contents to find yours. To do that, you can just create your own `<style id="file-hider-styles">` and attach the styles there, rather than relying on the `styles.css` file being loaded.

https://github.com/Oliver-Akins/file-hider/blob/716e12aa4051af0a2c481839166d75bfdb2501a4/src/main.ts#L82 you shouldn't need to save the settings on unload

https://github.com/Oliver-Akins/file-hider/blob/716e12aa4051af0a2c481839166d75bfdb2501a4/src/main.ts#L103 If there's an issue that causes the findStyleSheet to always return `null`, then users could end up in an infinite loop here. I suggest removing this redundancy and instead just initialize the styles inside `layoutReady`.

https://github.com/Oliver-Akins/file-hider/blob/716e12aa4051af0a2c481839166d75bfdb2501a4/src/main.ts#L125-L129 `this.style.disabled = this.settings.hidden;`

https://github.com/Oliver-Akins/file-hider/blob/716e12aa4051af0a2c481839166d75bfdb2501a4/src/main.ts#L126 you also need to assume that `this.style` could be null here.

Oliver-Akins:
@liamcain - I've made the requested changes (sorta), because I tweaked how the system was actually managing the hiding/showing as a whole.

It's now using inline styles on the parent element of the `data-path` element, changing the `display` property. So it's basically the same but it just changes the general behaviour.

If there's any other changes needed let me know.

liamcain:
Really nice job, works great!

Sorry for taking so long to get back to you on this.

POTPER:
excellent！

PR #987, opened 2022-05-31
Submitting Simple Dice Roller

yeeshue99:

## Repo URL

Link to my plugin: https://github.com/yeeshue99/SimpleDiceRoller

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


yeeshue99:
Finally got a successful workflow!

yeeshue99:
@liamcain I'm aware of dice roller, but I would still like to upload this as a simpler alternative. I found dice roller to be very feature rich but also very  intimidating to setup.

This serves as a simple, plug and play solution with zero setup required.

liamcain:
https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/main.ts#L98 I suggest you use `window.crypto.randomValues()` instead of `Math.random()` for a better, more _random_ generator.

https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/main.ts#L22 markdownView can be `null` here. You should add a guard to check.

https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/main.ts#L40 you should use `const` instead of `var` here. (and in the below function too)

https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/main.ts#L1 looks like you have a lot of unused imports here 🧹💨


https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/README.md?plain=1#L8 typo `cna` -> `can`

https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/README.md?plain=1#L3 Nothing in the README explains what the dice formula is. I think it would be helpful to include that information as well as maybe a screenshot or two.

https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/main.ts#L40 this is a bad way to store the formulas. It means that you can never have more than one of the same formula in a view. You're better off keeping the formulas in an array

https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/main.ts#L50 Maybe I just don't play enough TTRPGs but shouldn't `averageAllDiceOnScreen` return me a single value? It looks like it just averages each formula.

So let me know if this interpretation is correct:

```
Simulate 2d6 = 1d6 + 1d6
Average 2d6 = (1d6 + 1d6) / 2
```

Is that right?

yeeshue99:
https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/main.ts#L98 Is Window.crypto.randomValues() mobile safe? I know the review docs mention that using nodjs' crypto module isn't mobile safe, and when I looked into subtle crypto, I couldn't find a random number generator.

https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/main.ts#L22 Fixed.

https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/main.ts#L40 Fixed in both.

https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/README.md?plain=1#L8 Fixed.

https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/README.md?plain=1#L3 Noted and fixed.

https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/main.ts#L40 Noted. Refactored to remove it, since storing the calculations was found to not be necessary at all.

https://github.com/yeeshue99/SimpleDiceRoller/blob/09b108dfb4738b15e343f4ccb84ede307585c147/main.ts#L50 In this context, averaging a die doesn't use any type of randomness at all. The average value of any dice is calculated to be {diceSize + 1} / 2. This is a common formula for finding the average value of these simple series of numbers, I believe. For multiple dice, you don't round until the end.

```
Simulate 2d6 = 1d6 + 1d6 = {Some random value between 2 and 12}
Average 2d6 = {7/2} + {7/2} = 7
```

I hope that makes sense.

liamcain:
> Is Window.crypto.randomValues() mobile safe? I know the review docs mention that using nodjs' crypto module isn't mobile safe, and when I looked into subtle crypto, I couldn't find a random number generator

https://caniuse.com/getrandomvalues, yes it's mobile-safe.

>  The average value of any dice is calculated to be {diceSize + 1} / 2. This is a common formula for finding the average value of these simple series of numbers, I believe.

Thanks for the explanation, that makes sense!

---

the other fixes look good!

yeeshue99:
@liamcain the transition to crypto.getRandomValues has been finished.

PR #989, opened 2022-05-31
Add Linkify plugin.

matthewhchan:

## Repo URL

Link to my plugin: https://github.com/matthewhchan/linkify

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


matthewhchan:
Plugin validation failed because the tag in `manifest.json` did not match the tag in my repo (1.0.8 vs 1.08, a typo). I have fixed the tag to be 1.0.8.

liamcain:
Hey, really cool plugin. I think a lot of people could find this useful; however, they might be a bit deterred by how technical the README is. It's not until the bottom of the readme that I grasped the use case for the plugin. I think it's awesome that you provide a few default rules to get people started, but I would recommend trying to make this appeal to users that might be intimidated by regular expressions.

---

https://github.com/matthewhchan/linkify/blob/6dd0d6e75ffd308409ae1618bf522eeb36dd02bf/main.ts#L94 prefer `===` to `==`

https://github.com/matthewhchan/linkify/blob/6dd0d6e75ffd308409ae1618bf522eeb36dd02bf/main.ts#L86 I don't think you need to worry about unregistering the editor extensions yourself. I think you can just modify `this.viewPlugins` in place. Here's Licat's recommendation regarding modifying existing registered extensions: https://discord.com/channels/686053708261228577/840286264964022302/932271118943912017

---

Everything else looks good to me!

matthewhchan:
> Hey, really cool plugin. I think a lot of people could find this useful; however, they might be a bit deterred by how technical the README is. It's not until the bottom of the readme that I grasped the use case for the plugin. I think it's awesome that you provide a few default rules to get people started, but I would recommend trying to make this appeal to users that might be intimidated by regular expressions.
> 
> https://github.com/matthewhchan/linkify/blob/6dd0d6e75ffd308409ae1618bf522eeb36dd02bf/main.ts#L94 prefer `===` to `==`
> 
> https://github.com/matthewhchan/linkify/blob/6dd0d6e75ffd308409ae1618bf522eeb36dd02bf/main.ts#L86 I don't think you need to worry about unregistering the editor extensions yourself. I think you can just modify `this.viewPlugins` in place. Here's Licat's recommendation regarding modifying existing registered extensions: https://discord.com/channels/686053708261228577/840286264964022302/932271118943912017
> 
> Everything else looks good to me!

I will still need to unregister view plugins when they are deleted, or is there another way to do that?

liamcain:
> I will still need to unregister view plugins when they are deleted, or is there another way to do that?

So, `registerEditorExtension` can take in an array. You can just pass in `this.viewPlugins`. The plugins should get unregistered automatically if you just modify the `this.viewPlugins` in place instead of reassigning it.

matthewhchan:
> > I will still need to unregister view plugins when they are deleted, or is there another way to do that?
> 
> So, `registerEditorExtension` can take in an array. You can just pass in `this.viewPlugins`. The plugins should get unregistered automatically if you just modify the `this.viewPlugins` in place instead of reassigning it.

Thanks for explaining it, I didn't realize that's how it worked. I've made the requested changes.

I also added more details to the README to make it easier to understand what the plugin does.

liamcain:
So, I think it only works as long as you are modifying the same array instance.

https://github.com/matthewhchan/linkify/blob/cd27e483d7dfb79d23e9e35919c528f1b2a292b9/main.ts#L85 you're reassigning `this.viewPlugins` to a new array and registering that. I think you want to empty the existing array, and `this.viewPlugins.push(... this.settings.rules.map(createViewPlugin))`. Then you don't need to re-register the list. Let me know if that makes sense!

matthewhchan:
Ok, I think I've done what you proposed.

matthewhchan:
To provide more detail, https://github.com/matthewhchan/linkify/commit/c98f93af1d12d58eb9f6da45c5c3a23dfee0d082 registers the array `this.viewPlugins`, which I then modify in-place using `splice` whenever the view plugins need to be updated, i.e. when the user makes changes to the settings.

PR #990, opened 2022-05-31
Stack Overflow Answers plugin submission

bramses:

## Repo URL

Link to my plugin: https://github.com/bramses/obsidian-stack-overflow

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4207,5 +4207,12 @@
         "author": "Snezhig",
         "description": "Lets you define a title in frontmatter to be displayed as the filename",
         "repo": "Snezhig/obsidian-front-matter-title"
+    },
+    {
+        "id": "obsidian-stack-overflow",
+        "name": "Stack Overflow Answers",
+        "description": "Copy and Paste Stack Overflow answers directly into Obsidian.",
+        "author": "bramses",
+        "repo": "bramses/obsidian-stack-overflow"
     }
 ]


bramses:
I'm not sure why `bramses/obsidian-stack-overflow` is 404'ing. The repo link is here: https://github.com/bramses/obsidian-stack-overflow and I reflected that in `community-plugins.json`

lishid:
Might be something wrong with github's API.

bramses:
@lishid should i change anything from my end?

lishid:
Hmm actually I think there's a bug with the script that's outputting the wrong issue message, will double check.

liamcain:
I updated the validation script so that it spits out the correct error. Your plugin is missing a Github Release.

---

https://github.com/bramses/obsidian-stack-overflow/blob/cea62bb42d6accbf548df6ee6328a8206cbf68f8/main.ts#L192-L193 While there's no rule against it, I suggest that you don't provide a default hotkey for your command. You can add something to the readme to suggest a hotkey, but Mod+shift+v is already a default hotkey and it's likely one that could clash with an existing command. If a command clashes with another command, it can be a bad experience for the user, so it's better to just leave this omitted.

https://github.com/bramses/obsidian-stack-overflow/blob/cea62bb42d6accbf548df6ee6328a8206cbf68f8/main.ts#L101 This function should have a consistent return type. You are returning `Promise<string>` at the bottom, so it probably makes sense to return an empty string here and in the other return below.

https://github.com/bramses/obsidian-stack-overflow/blob/cea62bb42d6accbf548df6ee6328a8206cbf68f8/main.ts#L104 It might make sense to add additional validation here, such as making sure that `this.extractAnswerId` returns a nonempty string to ensure it's an answer page

liamcain:
@bramses, the release you added is empty. There's information here about what you need to include in the Github Release: <https://github.com/obsidianmd/obsidian-sample-plugin#releasing-new-releases>.

bramses:
Ok sorry @liamcain I'm dumb lol, I had assumed GitHub would take care of it for me

I added `main.js` and `manifest.json` is that cool? My `styles.css` is empty so I omitted it

Screen Shot 2022-06-02 at 3 18 42 PM

bramses:
Hi @liamcain just following up on this! Is the release good now?

PR #991, opened 2022-06-01
Add advanced codeblock plugin

lijyze:

## Repo URL

Link to my plugin: https://github.com/lijyze/obsidian-advanced-codeblock

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lijyze:
Thanks for you review. I have finished correction.

1. Rename plugin name to `ObsidianAdvancedCodeblockPlugin`

2. Delete anything concerns setting.

3. This README is copy from another plugin of mine, now already change this line to right description.

lijyze:
I was set tag name to `101` by mistake, I have already change id to `1.0.1`, but I don't know how to make checker rerun.

liamcain:
https://github.com/lijyze/obsidian-advanced-codeblock/blob/05e75a5f6c2616b5bb9b8a445c98256eedf96cdf/src/postProcessor.ts#L20 So I'm _pretty_ sure it's safe to assume this is always non-null. But I'd still suggest that you add another guard here since `getActiveViewOfType(MarkdownView)` can return `null` if there's no active file.

lijyze:
Thank you, that's how it looks now.

```ts
const origin = app.workspace
  .getActiveViewOfType(MarkdownView)
  ?.editor.getLine(codeBlock.lineStart)
  .slice(3);

// not specify anything or no active view
if (!origin) return null;
```

PR #993, opened 2022-06-02
migrating repo

yeeshue99:
Migrating repo name, since the given one seems to not allow downloading.


liamcain:
Hi @yeeshue99, it looks like the issue is that the manifest.json file inside your release is incorrect. It needs to match the version of the release itself. When I clicked install, I got version 1.0.2 instead of version 1.1, but it did install correctly.

yeeshue99:
I want to migrate to a new repo. I tried renaming this old one, but for some reason the renamed title doesn't register as a real one. That's all this pr is. The new repo with the correct name has the correct versioning in the release

PR #994, opened 2022-06-04
Add List Callouts plugin

mgmeyers:

## Repo URL

Link to my plugin: https://github.com/mgmeyers/obsidian-list-callouts

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


mgmeyers:
> [app.workspace.trigger("parse-style-settings");](https://github.com/mgmeyers/obsidian-list-callouts/blob/d90ce7bde65bc4461144b1b4f08bcd82a2dbbce5/src/main.ts#L60-L61) Maybe prefix this event with your plugin name?

@liamcain can you say more about this? The style settings plugin listens for the `parse-style-settings` event so that plugins can tell it when plugin CSS needs to be parsed. If I prefixed this, it would also need to be prefixed in style settings, right? That might be a tricky update since quite a few plugins use this event now, though it is doable.

> [255](https://github.com/mgmeyers/obsidian-list-callouts/blob/d90ce7bde65bc4461144b1b4f08bcd82a2dbbce5/src/main.ts#L14) This feels like a really strange way to store your settings? Why are they keyed off by rgb color values? This means you can't have 2 items share the same color, Can this just be a list with color + char instead? That would also make it easier if you wanted to extend this have separate light/dark color values.

Yeah, it is a bit awkward. I've updated it to use an array instead https://github.com/mgmeyers/obsidian-list-callouts/blob/main/src/main.ts#L12-L41


> I'm not sure but this might require some updates to support the latest version of CM6

Yeah, everything has been updated!

liamcain:
> The style settings plugin listens for the parse-style-settings event so that plugins can tell it when plugin CSS needs to be parsed. If I prefixed this, it would also need to be prefixed in style settings, right? That might be a tricky update since quite a few plugins use this event now, though it is doable.

Oh, I didn't realize this was coming from the Style Settings plugin. My bad.

PR #1004, opened 2022-06-08
Add plugin: Alias from heading

basham:

## Repo URL

Link to my plugin: https://github.com/basham/obsidian-alias-from-heading

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


basham:
@liamcain: Thanks for the review! I published [v1.0.8](https://github.com/basham/obsidian-alias-from-heading/releases/tag/1.0.8), and it is ready for you to review again. After some additional testing, I also fixed some bugs with how links are synchronized with headings. In doing so, I changed how the plugin is using `String.replace()`. Before this latest release, using `String.replaceAll()` would make more sense, as you suggested [1]. However, with the new changes, there is no longer an advantage in doing so. If curious, I added two comments ([first](https://github.com/basham/obsidian-alias-from-heading/blob/1.0.8/main.ts#L75-L81), [second](https://github.com/basham/obsidian-alias-from-heading/blob/1.0.8/main.ts#L94-L100)) that explains this new approach, as it is not obvious by reading the code. The [release notes](https://github.com/basham/obsidian-alias-from-heading/releases/tag/1.0.8) also provides a high-level description of the fixes.

[1]: The TypeScript compiler complained about using `String.replaceAll()` until I added `"ES2021"` to the `lib` compiler option in `tsconfig.json`. Perhaps the [Obsidian Sample Plugin](https://github.com/obsidianmd/obsidian-sample-plugin) should be similarly updated, if that configuration change is truly compatible.

liamcain:
- [metadataCache.metadataCache[hash] = updatedCache;](https://github.com/basham/obsidian-alias-from-heading/blob/9aa954d6fe4b0c44c51b320631a96d5f9f82d83d/main.ts#L203-L204) So, I don't love that you're modifying the metadataCache, that's not meant to be modified, especially since the cache gets persisted in IndexDB, so it's not really a good place for ephemeral data. An alternative approach (it's not much better) would be to just monkey patch the `getLinkSuggestions` function  on the metadataCache. That way you only need to insert your aliases when the autocomplete popup is shown.

basham:
I made updates according to your review, in [v1.0.12](https://github.com/basham/obsidian-alias-from-heading/releases/tag/1.0.12).

> Ideally you can just use registerEvent. If I had to guess, the issue might be because you're not calling super.unload()

It turns out, `Plugin.registerEvent()` does work as expected. I just needed to switch from `Plugin.unload()` to `Plugin.onunload()`.

> An alternative approach (it's not much better) would be to just monkey patch the getLinkSuggestions function on the metadataCache. That way you only need to insert your aliases when the autocomplete popup is shown.

That works out nicely! It cleans up a lot of the code. The patch is removed when the plugin unloads.

However, this change introduced one issue. `getLinkSuggestions()` does not affect the behavior of the [Quick Switcher](https://help.obsidian.md/Plugins/Quick+switcher). Do you have a recommendation about that? Is there a way to patch that core plugin, so these implicit heading aliases are included in that list of suggestions?

basham:
I'm just surprised the Quick Switcher (being a core plugin) does not use `getLinkSuggestions()`. Is that a bug or something that should be fixed in that plugin?

liamcain:
> I'm surprised the Quick Switcher (being a core plugin) does not use getLinkSuggestions(). Is that a bug or something that should be fixed in that plugin?

That's a good question, I'm not sure what the reason is. I'll make a note and see if we can use it.

Sorry for taking so long to get back to this. It got buried with some other pull requests and I lost track of it. The changes look great, very excited to see this plugin get included.

PR #1005, opened 2022-06-09
Update community-plugins.json

jon-heard:

## Repo URL

Link to my plugin:
https://github.com/jon-heard/obsidian-text-expander-js

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [X]  Linux
  - [X]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


jon-heard:
I just tested this plugin on Linux and It works fine.  Modifying original comment to reflect this.

jon-heard:
I just found a bug during shortcut-file initialization.  I fixed it and put up a new release: 0.16.3.

liamcain:
https://github.com/jon-heard/obsidian-text-expander-js/blob/8c3c92011959a837c986d73ffbb91fbb23c7d603/main.js#L60-L79 It looks like you like used the generated `main.js` file from sample-plugin releases as a starting point for your plugin then continued to write Javascript on top of that. If you don't want to use typescript for your plugin, you can still use ES5 or even ES6 Javascript for your plugin. That means you can use [`class`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes) instead of wrapping everything in an IIFE.

It will make it a lot easier for you in the long run and your code will be a lot more readable. If you want to see examples of plugins that don't use typescript at all, check out pjeby's plugins, e.g. <https://github.com/pjeby/tag-wrangler/blob/master/src/plugin.js>

jon-heard:
@liamcain, thanks for this input about classes.  I've successfully converted my plugin to use all classes, like the sample of pjeby's work.
I've committed the change, along with a new version number.  If you want I can make a new release for it.  Just let me know.

jon-heard:
I've decided to make a new release for the new version that uses class syntax.
So, if you want to view the latest, with class syntax, then it's **0.16.4**.  If you want to continue viewing the previous version, then it's revision **0.16.3**.

liamcain:
[IS_MOBILE = this.app.isMobile;](https://github.com/jon-heard/obsidian-text-expander-js/blob/c8b01d6cac0cb0e3487a70a524bcbfae6d8c9939/main.js#L71-L72) you can get the platform information from the [Platform](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L2326) object in the api 

[this.activeFilesName = "";](https://github.com/jon-heard/obsidian-text-expander-js/blob/c8b01d6cac0cb0e3487a70a524bcbfae6d8c9939/main.js#L1284) I recommend not saving your own version of the active file name in case it gets out of sync with the API's active view. It's simpler to just use `getActiveFile()` when you need the value instead.

jon-heard:
Thanks for this feedback!  I've created a new release based on it.

I've removed `IS_MOBILE` and `IS_WINDOWS`, replacing references to them with direct references to the API.  I'm pretty sure I still need to search the `navigator.appVersion` string to determine Windows platform, since it doesn't look like the Platform object has a flag for that.

I've refactored `this.activeFilesName` to `this.currentFilesName` and written comments to better reflect its purpose.  In short, I need to react to a change of active-file by referencing the previously active file.  I couldn't find a way to access this natively in the "active-leaf-change" event.  There's also no "active-leaf-about-to-change" event or anything similar.  I'd have named the variable `this.priorActiveFilesName`, except that it spends the majority of the time BEING the active file's name, though it's only actually useful in the moment between when it's not to when I make it so, again.

I'd be all in for a better solution with the active-file change event, as I know this one is not obvious (though the comments might be good enough now).

liamcain:
[this.app.isMobile](https://github.com/jon-heard/obsidian-text-expander-js/blob/e8b183965a3e222c520581fe3c729cfb47badd09/main.js#L484) is deprecated, so this should be `Platform.isMobile` instead.

> I'm pretty sure I still need to search the navigator.appVersion string to determine if the platform is Windows, since it doesn't look like the Platform object has a flag for this.

Yeah, you're right, it looks like there's nothing exposed in the API for this currently, so what you have is fine.

> I've refactored this.activeFilesName to this.currentFilesName and written comments to better reflect its purpose, which is not obvious. In short, I need to react to a change of active-file by referencing the previously-active-file. I couldn't find a way to access the previously-active-file as part of the "active-leaf-change" event. There's also no "active-leaf-about-to-change" event or anything similar. I'd considered naming the variable this.priorActiveFilesName, except that for the majority of the time it actually IS the active file's name. Though, it's only useful in the moment between when it goes out of sync with the active-file to when I sync it back up again.

I see. Internally, we have some utilities for helping with this but I'm not sure the best way for you to access that from within a plugin. For now, what you have is probably the best way, but I'll give some thought into if there's anything we can add to the API to make this easier for your use case.

jon-heard:
Hi @liamcain.  Thanks again for your work on this review.  I have a wrinkle to this process that I hope won't cause you trouble...

I have a plan for one adjustment to the plugin before it is released to the community.  This adjustment would involve some notable changes to the code and documentation.  I'm wondering about the best way to handle this:

- I could implement the changes after this plugin has passed the review, and before it goes out to the community.
- I could implement the changes after this plugin has passed a preliminary review, but with one last review focusing on JUST these final changes.
- I could implement the changes immediately, before the plugin has passed the review.

Do you see a best-plan here?  I apologize if this is inconvenient.  I recognize that, ideally, everything would have been wrapped in a nice bow before this review had started.

FYI, the planned change is to add a third string to each shortcut.  A shortcut has a "Test" regex string, an "Expansion" JS string and this adds an "About" string to provide help text.

jon-heard:
Release 0.16.12:
- Polish - Wrote DEFAULT_SETTING.shortcuts with backtick string for readability
- Polish - fixed typo
- Polish - changed "greet" shortcut to "hi"
- Polish - Moved Object.freeze into DEFAULT_SETTINGS assignment

jon-heard:
The latest release adds "test_preRelease.txt": a list of tests for ALL features of this plugin.  I thought I'd mention it, in case it helps you with the review.

Also, I'll add no further changes to the main branch unless you request it... or it's REALLY important.

jon-heard:
Also, I don't want you to think I'm impatient (I'm sure that you have plenty of tasks on your plate), but there's anything I could do to smooth the review process,, such as switching from Javascript to Typescript, then please let me know.

jon-heard:
I've decided to re-write the plugin in Typescript (instead of Javascript).  This is because the ability to split the code among multiple files allows for superior organization.  I was concerned that this would disrupt the current review, but given the length of time since the last comment, I'm guessing that a re-review is inevitable anyway.

Release 0.17.0:
- contains many updates, most notably the rewrite into Typescript.

liamcain:
- [const childProcess = require("child_process");](https://github.com/jon-heard/obsidian-text-expander-js/blob/f2c0d97c68e3b98d27faff8c66b1e0d49a22f286/src/_Plugin.ts#L18-L19) I'm pretty sure this will throw an error when running on mobile, right? If you want it to be conditionally available on desktop, I think the require needs to be scoped inside a function
- You use `!=` and `==` throughout the code when it's probably better/safer to be using `!==` and `===`

---

Otherwise everything else looks fine. You could let me know once you've confirmed that the `child_process` doesn't break on mobile and I can merge this?

jon-heard:
Thank you for all of your work on this review!
- I have tested this feature on both Android and IPhone and the plugin works ok and without any console errors.  My guess is that [const childProcess = require("child_process");](https://github.com/jon-heard/obsidian-text-expander-js/blob/f2c0d97c68e3b98d27faff8c66b1e0d49a22f286/src/_Plugin.ts#L18-L19) is just assigning null or undefined, though I haven't checked for this explicitly.  However, in the interest of more intentional code, I've just created release 0.17.1, which _does not_ call require("child_process") when the platform is mobile.
- I've also rewritten all of the equality and inequality operators as you've suggested.  That's part of release 0.17.1 as well.  I followed up this change with a run of the prerelease test (which tests all features) and everything is working as expected.

Release 0.17.1:
- polish - only run require("child_process") if on non-mobile platform.
- polish - "==" and "!=" to "===" and "!=="
- bug fix - Renaming the active note makes "_currentFilesName" inaccurate. If the active note is a shortcut-file, this causes a minor bug until the active note is changed.

jon-heard:
Release 0.17.2
- bug fix - shortcut-files with non-standard newlines (\r chars) cause bugs in the help system.

jon-heard:
Release 0.18.0
- bug fix - Import Library feature must fail more gracefully on unable to connect to repo.
- feature - Shortcut-files can be disabled, without removing them. This feature is implemented, but not yet exposed through UI.

PR #1008, opened 2022-06-10
Update community-plugins.json

jensmtg:

## Repo URL

Link to my plugin: [https://github.com/jensmtg/obsidian-diagrams-net](https://github.com/jensmtg/obsidian-diagrams-net)

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
[callback: () => this.attemptNewDiagram()](https://github.com/jensmtg/obsidian-diagrams-net/blob/568294e08542d538a01d84132434b25816050c91/src/main.tsx#L36) you should use [checkCallback](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L505) instead. That way you can check that there is an activeView before showing the command in the command palette.

[this.app.workspace.splitActiveLeaf('horizontal')](https://github.com/jensmtg/obsidian-diagrams-net/blob/568294e08542d538a01d84132434b25816050c91/src/main.tsx#L124) we've updated the API recently and now recommend that you use [`getLeaf(true)`](https://github.com/obsidianmd/obsidian-api/blob/86ad046945701047de9862384d1508dd49ca2bcb/obsidian.d.ts#L3761) instead of `splitActiveLeaf`

[vault.read(xmlFile)](https://github.com/jensmtg/obsidian-diagrams-net/blob/ebefc8dd69938b2187aad931b8b6ad8245d60b24/src/DiagramsApp.tsx#L27) this can be [cachedRead](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3373) instead

---

Otherwise everything looks pretty good to me!

jensmtg:
> [this.app.workspace.splitActiveLeaf('horizontal')](https://github.com/jensmtg/obsidian-diagrams-net/blob/568294e08542d538a01d84132434b25816050c91/src/main.tsx#L124) we've updated the API recently and now recommend that you use [`getLeaf(true)`](https://github.com/obsidianmd/obsidian-api/blob/86ad046945701047de9862384d1508dd49ca2bcb/obsidian.d.ts#L3761) instead of `splitActiveLeaf`

`workspace.splitActiveLeaf('horizontal')` does split the view horizontally, whereas `workspace.getLeaf(true, 'horizontal')` does not - it still splits the view vertically. Is this intended behaviour?

liamcain:
> workspace.splitActiveLeaf('horizontal') does split the view horizontally, whereas workspace.getLeaf(true, 'horizontal') does not - it still splits the view vertically. Is this intended behaviour?

Providing the splitDirection to `getLeaf` is only available in v0.15.X currently. On older versions of Obsidian, it will always split vertically, so that's expected. If you need the split to be vertical, I guess your best option is to keep it as `splitActiveLeaf` for now.

jensmtg:
> > workspace.splitActiveLeaf('horizontal') does split the view horizontally, whereas workspace.getLeaf(true, 'horizontal') does not - it still splits the view vertically. Is this intended behaviour?
> 
> Providing the splitDirection to `getLeaf` is only available in v0.15.X currently. On older versions of Obsidian, it will always split vertically, so that's expected. If you need the split to be vertical, I guess your best option is to keep it as `splitActiveLeaf` for now.

Ok. So:
- I don't replace splitActiveLeaf with getLeaf for now, but will do so when v0.15 becomes stable.
- cachedRead and checkCallback implemented, as suggested.
- I also changed the manifest to `isDesktopOnly: true`.

jensmtg:
> LGTM. One small thing you might want to fix:
> 
> The video in your README should use an absolute link (i.e. raw.githubusercontent.com/something/something), to be visible from within Obsidian. The relative link you have now won't get resolved since Obsidian doesn't download the whole repo for displaying the README

Done. Replaced the video with a gif in absolute link.

jensmtg:
@liamcain Does this pull request need any more action on my part?

PR #1009, opened 2022-06-11
Add Obsidian Wielder

victorb:

## Repo URL

Link to my plugin: https://github.com/victorb/obsidian-wielder

## Release Checklist
  - [X]  Windows
  - [X]  macOS
  - [X]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4312,5 +4312,12 @@
         "author": "mnowotnik",
         "description": "Use js files or snippets to code your own quick and dirty plugins",
         "repo": "mnowotnik/obsidian-user-plugins"
+    },
+    {
+        "id": "wielder",
+        "name": "Wielder",
+        "author": "victorb",
+        "description": "Clojure/ClojureScript inside your Obsidian documents!",
+        "repo": "victorb/obsidian-wielder"
     }
 ]


victorb:
Adding a note from @joethei (from Discord) as others might have the same question

> I did not spot anything problematic in your code, but is there any reason you are not using a MarkdownPostProcessor?

Answer:

>  Yeah, my initial implementation actually used MarkdownPostProcessor. But had issues with that as they don't load in the same order always, so sometimes the second block would be called before the first. I need to process things strictly from top to bottom.

> As a side-effect of building it like I built it, it was trivial to get it to work with Obsidian Publish websites as well, as they work/look mostly the same as the preview window, so adding support was like 20-30 lines or something.

lishid:
> Yeah, my initial implementation actually used MarkdownPostProcessor. But had issues with that as they don't load in the same order always, so sometimes the second block would be called before the first. I need to process things strictly from top to bottom.

As a side effect, your plugin won't work for any page that's much longer than the visible scroll area. The preview pane is optimized in a way that only ever displays a subset of the document if it's long.

Your plugin is also registering events without un-registering them on unload, and also attaching global handlers `window.setIntervalTracked` that's closure'd to your plugin causing a memory leak when your plugin unloads.

Your command should use `editorCallback` instead of being a callback that manages its own editor (it's missing null checks for things like `activeLeaf` and `activeLeaf.view instanceof MarkdownView`).

Also you should avoid using `innerHTML` to prevent DOM injection attacks. If you absolutely cannot live without raw HTML then you must [sanitize them](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L2726).

victorb:
Thanks a lot for the review @lishid!

> As a side effect, your plugin won't work for any page that's much longer than the visible scroll area. The preview pane is optimized in a way that only ever displays a subset of the document if it's long.

Yeah, that's one of the drawbacks of the approach unfortunately. I have some ideas of fixes to this issue but have yet to implement them. As it stands right now, it seems to render about 4 pages worth of elements in my vaults, and the issue has been documented in the documentation of the plugin here as well: https://wielder.victor.earth/Tutorials/06-Pitfalls#Obsidian+-+Long+documents

> Your plugin is also registering events without un-registering them on unload, and also attaching global handlers window.setIntervalTracked that's closure'd to your plugin causing a memory leak when your plugin unloads.

> Your command should use editorCallback instead of being a callback that manages its own editor (it's missing null checks for things like activeLeaf and activeLeaf.view instanceof MarkdownView).

Both of these seems to be true indeed, thanks for catching that. Will release new version with this fixed.

> Also you should avoid using innerHTML to prevent DOM injection attacks. If you absolutely cannot live without raw HTML then you must [sanitize them](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L2726).

Thanks for the reference, I'll start using this function together with innerHTML.

victorb:
I've just pushed a new release (1.0.4) that contains fixes to things mentioned here.

Changes since I made initial PR here:

- Some updates to the documentation which was using the old plugin name
- Fixing the plugin so it can be used on mobile as well (tested iOS and Android)
- Sanitizing anything passed into `*renderHTML` (although I'm not sure what the attack vector could be here, users are the ones passing their own input to the function. But, better sanitize too much than too little)
- The command for inserting a code block uses the right `editorCallback` attribute
- Cleaning up all event handlers properly when disabling/unloading the plugin
- Intervals also cleaned up via the internal `registerInterval` function
- Removing the closure for `setIntervalTracked` when plugin unloads

lishid:
> Sanitizing anything passed into *renderHTML (although I'm not sure what the attack vector could be here, users are the ones passing their own input to the function. But, better sanitize too much than too little)

Yeah you're right, I realized that since your plugin just lets the markdown file execute code directly there's really not much use to preventing attacks from the innerHTML side.

One possible attack vector I've seen most plugins that do code execution (like templater, dataview-js, etc) protect against is that someone could send a malicious `.md` file, and a user unknowingly importing such a file to their vault would cause code execution. I guess by the pure nature of this plugin this is allowed (and possibly even intended), but at the very least I think there should be a warning at the top of the README saying this plugin will execute code and should only be used with trusted vaults/notes.

victorb:
Version has been updated to 1.0.5 now, incorporating the changes mentioned above.

liamcain:
- [defaultTimeout: number;](https://github.com/victorb/obsidian-wielder/blob/0119b025ad1c63db1abbc2f8b73d2c5f38393d13/src/main.ts#L36-L37) this looks unused

PR #1011, opened 2022-06-12
Add PostgreSQL plugin

clouedoc:

## Repo URL

Link to my plugin: https://github.com/clouedoc/postgresql-obsidian

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
[export class SampleModal](https://github.com/clouedoc/postgresql-obsidian/blob/6c55b3b9937d9417348ec47fbd3d351728fe434e/src/classes/sample-modal.ts#L3) this can be removed

[const dv: DataviewApi = getAPI();](https://github.com/clouedoc/postgresql-obsidian/blob/38cfeba2ffba8fd82083ecba5932dd65618c3262/src/classes/postgresql-plugin.ts#L17) This will fail if your plugin loads before dataview. You should only call this after Obsidian has fully loaded. You can just move this the command callback

[this.app.workspace.getActiveFile()](https://github.com/clouedoc/postgresql-obsidian/blob/38cfeba2ffba8fd82083ecba5932dd65618c3262/src/classes/postgresql-plugin.ts#L25) this can be null, you should add a check for that.

---

Can you add more information to the readme about how a user should setup the database? I was assuming you were using a local DB while looking at the code but then I noticed your readme uses a db as a service.

clouedoc:
@liamcain `this.app.workspace.getActiveFile()` TypeScript definition should indicate that the file can be null.
Do you think that I could propose this change?

I updated the plugin to apply your suggestions. Would you mind reviewing my latest commit?

lishid:
> @liamcain `this.app.workspace.getActiveFile()` TypeScript definition should indicate that the file can be null.
> Do you think that I could propose this change?

It already does:
```ts
getActiveFile(): TFile | null;
```
https://github.com/obsidianmd/obsidian-api/blob/4583a6b8305b05139c6b1bf226ad15e50a9a5987/obsidian.d.ts#L3809

clouedoc:
@lishid you're right

I found out that the default `tsconfig.json` does not include the `strict` and `strictNullChecks` option, which led to this programming error.

Maybe there should be a disclaimer in the README.md of [obsidian-sample-plugin](https://github.com/obsidianmd/obsidian-sample-plugin)?

Or even better, these options should be enabled by default for new plugins.

lishid:
> I found out that the default tsconfig.json does not include the strict and strictNullChecks option, which led to this programming error.

I've added it!

PR #1016, opened 2022-06-13
add plugin - google contacts and calendar lookup

ntawileh:

## Repo URL

Link to my plugin:
https://github.com/ntawileh/obsidian-google-lookup

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
[firstName: names && names[0].givenName ? names[0].givenName : 'unknown',](https://github.com/ntawileh/obsidian-google-lookup/blob/48a8c44cc3e5a318042cf8a06be4c6bff9feb45e/src/api/google/people-search.ts#L94) An empty list is truthy, so you might want to do something like `names?.[0]?.givenName ?? 'unknown'` instead. That way you can handle an empty list.

[isViewInSourceMode](https://github.com/ntawileh/obsidian-google-lookup/blob/8da870f6c48b8cc750a3a412bfaab60cc80dee78/src/utils/files.ts#L3) why not pass in the `view` to this function?

[moment(selectedText)](https://github.com/ntawileh/obsidian-google-lookup/blob/8da870f6c48b8cc750a3a412bfaab60cc80dee78/src/ui/calendar-modal.ts#L91) creating moment instances is actually pretty expensive, so I would suggest saving the moment instance in a variable instead of regenerating later in the ternary.

---

Anyway, I think this is good to go in as is, feel free to fix the following at your convenience.

PR #1021, opened 2022-06-22
Added obsidian-paste-into-codeblock

inode-:

## Repo URL

Link to my plugin: https://github.com/inode-/obsidian-paste-into-codeblock

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [X]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4341,5 +4341,12 @@
         "author": "XSPGMike",
         "description": "Create braincache flashcards from Obsidian",
         "repo": "XSPGMike/braincache_obsidian"
+    },
+    {
+        "id": "obsidian-paste-into-codeblock",
+        "name": "Paste into codeblock",
+        "author": "inode",
+        "description": "Allow paste into codeblock",
+        "repo": "inode-/obsidian-paste-into-codeblock"
     }
 ]


liamcain:
Hi, this plugin seems pretty similar to this one: <https://github.com/paddan/code-block-plugin>. It might make sense for you to collaborate with them rather than having 2 separate plugins.

---

- [hotkeys: [{ modifiers: ["Ctrl", "Alt"], key: "v" }],](https://github.com/inode-/obsidian-paste-into-codeblock/blob/290b76ce387cade7dd47d672f1ca16e4f9f033a3/src/main.ts#L20) While there's no strict rule preventing it, we recommend not providing hotkeys by default to avoid them conflicting with user hotkeys. It's also tricky since hotkeys vary by OS, so on macOS this would make more sense as `command`.

inode-:
I spent an hour looking for the solution, but I didn't find this plugin. It definitely does the same thing as mine ...

I will close the pull request

liamcain:
> I spent an hour looking for the solution, but I didn't find this plugin. It definitely does the same thing as mine ...

Not your fault, that plugin only got merged into the plugin gallery yesterday 😄

PR #1015, opened 2022-06-13
Add Zotero Bridge and Zotero Link to the list of community plugins

MunGell:
# I am submitting a new Community Plugin

## Repo URL

Link to my plugin:
- Zotero Bridge: https://github.com/vanakat/zotero-bridge
- Zotero Link: https://github.com/vanakat/zotero-link

## Release Checklist
- [x] I have tested the plugin on
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  Android _(if applicable)_
  - [ ]  iOS _(if applicable)_
- [x] My GitHub release contains all required files
  - [x] `main.js`
  - [x] `manifest.json`
  - [ ] `styles.css` _(optional)_
- [x] GitHub release name matches the exact version number specified in my manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
- [x] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [x] My README.md describes the plugin's purpose and provides clear usage instructions.
- [x] I have read the tips in https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md and have self-reviewed my plugin to avoid these common pitfalls.
- [x] I have added a license in the LICENSE file.
- [x] My project respects and is compatible with the original license of any code from other plugins that I'm using.
      I have given proper attribution to these other projects in my `README.md`.

PR #1022, opened 2022-06-22
Add Zotero Bridge to the list of plugins

MunGell:

## Repo URL

Link to my plugin: https://github.com/vanakat/zotero-bridge

## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


MunGell:
Follow up on #1015

liamcain:
[this.zoteroAdapter](https://github.com/vanakat/zotero-bridge/blob/005cd521b710a864e58cb4d4f3d6babfbbab5cc7/src/ZoteroBridge.ts#L15) is there supposed to be a UI to configure the settings?

MunGell:
I've designed it with configurability in mind, but there is no need for it at the moment - Zotero will always be on `localhost` and on `23119` port - it is not possible to configure it in Zotero at the moment.

PR #1023, opened 2022-06-23
add obsidian-files-cooker plugin

ivaneye:

## Repo URL

Link to my plugin: https://github.com/ivaneye/obsidian-files-cooker

## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


ivaneye:
> #### Hello ivaneye! [](https://obsidian.md)
> **I found the following errors in your plugin, Files Cooker:**
> 
> ❌ Plugin ID mismatch, the ID in this repo is not the same as the one in your repo.
> 
> This check was done automatically.

has changed

joethei:
- You might want to update your description and readme, I needed to take a look at the code to understand what this plugin does.

- You should put a limit on how many entries this loads: <https://github.com/ivaneye/obsidian-files-cooker/blob/main/src/move/move-modal.ts>, just add a `this.limit = xyz` into the constructor.
If this gets called in a vault with many files this would cause quite a bit of lag.

- <https://github.com/ivaneye/obsidian-files-cooker/blob/53c45f627c8e0e53f140a4b25034557179dbb7d5/manifest.json#L9> I don't see any reason why this plugin would not work on mobile.

liamcain:
- [this.addCommand({](https://github.com/ivaneye/obsidian-files-cooker/blob/64133b2e4505e4954af924a917817cce49ebd245/main.ts#L60-L66) Maybe use [`editorCheckCallback`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L518) here and check if dataview is installed before making this command available.
- [onClose() {](https://github.com/ivaneye/obsidian-files-cooker/blob/64133b2e4505e4954af924a917817cce49ebd245/src/modal/delete-confirm-modal.ts#L58-L61) this shouldn't be necessary since the modal gets detached
- [this.app.vault.rename(info.sourceFile, info.targetDir + "/" + info.sourceFile.name);](https://github.com/ivaneye/obsidian-files-cooker/blob/64133b2e4505e4954af924a917817cce49ebd245/src/modal/confirm-modal.ts#L32-L33) You should use [`FileManager.renameFile`](https://github.com/obsidianmd/obsidian-api/blob/4583a6b8305b05139c6b1bf226ad15e50a9a5987/obsidian.d.ts#L1134) instead. It will safely update the links after moving the file, so no links in the vault will get broken.
- [this.moveInfos.forEach(info => {](https://github.com/ivaneye/obsidian-files-cooker/blob/64133b2e4505e4954af924a917817cce49ebd245/src/modal/confirm-modal.ts#L31) Because renaming is `async`, I suggest you await the rename function and change the `forEach` to be `for ... of` so that the bulk move happens synchronously. Otherwise, the `rename` function calls will be potentially modifying the same file at the same time and could cause issues.

ivaneye:
@joethei 
I have followed all the suggestions to modify the code。Have any other suggestions?

liamcain:
- https://github.com/ivaneye/obsidian-files-cooker/blob/f653c89ed0139c8c82abdd3a98311c4cc7186b5d/main.ts#L42 You can remove this to avoid spamming the console


Otherwise looks good.

PR #1024, opened 2022-06-23
Add Zotero Link plugin

MunGell:

## Repo URL

Link to my plugin: https://github.com/vanakat/zotero-link

## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


MunGell:
Follow up from #1015

joethei:
Any particular reason why you split this into two plugins?
You could just add this as a feature to your Bridge plugin, the other zotero/similar plugins also do it this way.

You can simplify this: <https://github.com/vanakat/zotero-link/blob/15d3459692a4f1d90559b22cba10964c1dd997b7/src/ZoteroLinkSettingTab.ts#L40-L48> to `keywordList.createEl(...`, they will automatically be children of `keywordList`

PR #1027, opened 2022-06-26
Adding the Calendar Week Habit Tracker

Narsail:

## Repo URL

Link to my plugin: [Habit Tracker](https://github.com/Narsail/habit-tracker-obsidian)
## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [X]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


joethei:
- You should enable Issues in the Repository settings.
They get disabled when forking, leaving users no way to create feature requests / bug reports.

- You can remove any references to the settings, as you don't use them.

- instead of doing this: <https://github.com/Narsail/habit-tracker-obsidian/blob/66832a1cb281e41aa90cace1b3dc72aa61d6ccba/main.ts#L76-L82> you should use `heatmapCalendarMonthsUl.creatEl...` Saving you that `parent: heatmapCalendarMonthsUl` part.

- You should use different CSS classes, these will create problems with the original plugin, if either plugin changes its style: <https://github.com/Narsail/habit-tracker-obsidian/blob/66832a1cb281e41aa90cace1b3dc72aa61d6ccba/main.ts#L61-L74>

- You can move this over to your `models.json` file:
<https://github.com/Narsail/habit-tracker-obsidian/blob/66832a1cb281e41aa90cace1b3dc72aa61d6ccba/main.ts#L52-L57>

---
The following is optional:

- Putting your code in a seperate directory(`src` usually) would increase the readability, as they are mixed with configuration files currently.

- You can replace this: <https://github.com/Narsail/habit-tracker-obsidian/blob/66832a1cb281e41aa90cace1b3dc72aa61d6ccba/common.ts#L22> with:
`const min = Math.min(...values)`
They are the same thing, you can decide which syntax you prefer.

- Using semicolons at the end of each statement is prefered, as omitting them can lead to rare, hard to debug problems.(I know, the original plugin also omits them)
Just be aware how JS interprets such code: https://flaviocopes.com/javascript-automatic-semicolon-insertion/>

liamcain:
https://github.com/Narsail/habit-tracker-obsidian/blob/1e26916d13f360d6e655c5909f0fd43b1b99c041/README.md?plain=1#L12-L13

These image URLs should point to github's CDN (raw.githubusercontent.com) for them to load properly within Obsidian. Otherwise they will show up like this:

image


https://github.com/Narsail/habit-tracker-obsidian/blob/1e26916d13f360d6e655c5909f0fd43b1b99c041/README.md?plain=1#L19 you can remove the `\` here

[var simple = new Date(y, 0, 1 + (w - 1) * 7);](https://github.com/Narsail/habit-tracker-obsidian/blob/1e26916d13f360d6e655c5909f0fd43b1b99c041/main.ts#L9-L12) I recommend avoiding `var` and switching all instances of `var` to `let` instead. `var` is function-scoped rather than block-scoped, which is often undesirable and can lead to bugs.

> instead of doing this: https://github.com/Narsail/habit-tracker-obsidian/blob/66832a1cb281e41aa90cace1b3dc72aa61d6ccba/main.ts#L76-L82 you should use heatmapCalendarMonthsUl.creatEl... Saving you that parent: heatmapCalendarMonthsUl part.

Looks like this wasn't addressed, you're still constructing your elements bottom-up, which makes it a bit harder to visualize the structure.

[let filteredEntries = habit.entries.filter(entry => entry.date == day);](https://github.com/Narsail/habit-tracker-obsidian/blob/1e26916d13f360d6e655c5909f0fd43b1b99c041/main.ts#L91-L92) Use `===` instead of `==` for equality checks.

---

In your README, you mention "It currently only supports up to 7 habits." Where is this limitation coming from?

Narsail:
@liamcain I should have addressed all of your points in the latest release. 

Regarding the limitation of 7 Habits. That currently comes from on how the CSS file has been set up. There is a fixed number of rows and columns in there. 

I plan to make it dynamic eventually, but haven't had the time to look at it yet. And my CSS skills are non-existent, so I need to experiment a bit first.

PR #1028, opened 2022-06-27
Added custom-modals-obsidian info

helloitsian:

## Repo URL

Link to my plugin: https://github.com/helloitsian/custom-modals-obsidian

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4456,5 +4456,12 @@
         "author": "Jens M Gleditsch",
         "description": "Enable diagrams.net (previously draw.io) type diagrams, with the diagrams.net embedded editor.",
         "repo": "jensmtg/obsidian-diagrams-net"
+    },
+    {
+      "id": "custom-modals-obsidian",
+      "name": "Custom Modals",
+      "author": "Ian Murawski (helloitsian)",
+      "description": "Create powerful Custom Modals with HTML and JavaScript from the comfort of your notes.",
+      "repo": "helloitsian/custom-modals-obsidian"
     }
 ]


joethei:
<https://github.com/helloitsian/custom-modals-obsidian/blob/1e763e0a3da9892946e3b1f8da52a088e9d56051/src/main.ts#L44-L58>
 Obsidian will remove listeners if you register them using: <https://github.com/obsidianmd/obsidian-api/blob/d4b79f95ccf3838a305915a137a844250766d023/obsidian.d.ts#L601>

<https://github.com/helloitsian/custom-modals-obsidian/blob/main/src/filesManager/FilesManager.ts>
Running this at startup can have quite the performance impact on large vaults,
Some data might not be available as Obsidian might be indexing it.

<https://github.com/helloitsian/custom-modals-obsidian/blob/1e763e0a3da9892946e3b1f8da52a088e9d56051/src/modal/CustomModal.ts#L30-L31>
Please use the `createEl` function instead.

<https://github.com/helloitsian/custom-modals-obsidian/blob/1e763e0a3da9892946e3b1f8da52a088e9d56051/src/modal/CustomModal.ts#L33-L47>
Put this in the `styles.css` file instead.

<https://github.com/helloitsian/custom-modals-obsidian/blob/1e763e0a3da9892946e3b1f8da52a088e9d56051/src/modal/CustomModal.ts#L74-L75>
You can save this query by dirrectly passing the button obj over from the render function.

<https://github.com/helloitsian/custom-modals-obsidian/blob/1e763e0a3da9892946e3b1f8da52a088e9d56051/src/modal/CustomModal.ts#L84-L86>
Obsidian will also unload them when you use the `registerEvent` method when closing the modal.

helloitsian:
@AnnaKornfeldSimpson WOW I didn't even know it was mentioned, thanks for letting me know! I've been dealing with a move and working on other things. Will resume work on this tomorrow.

helloitsian:
Went ahead and put the changes in per your request @joethei

joethei:
I'll do a re-review soon.
After that someone else will also take a look.

liamcain:
- [async handleCreateAndModify(file: TAbstractFile) {](https://github.com/helloitsian/custom-modals-obsidian/blob/99691387027af5a7897968efa022c35673ae8a28/src/main.ts#L19) You should check `file instanceOf TFile` since these callbacks are also called for TFolder.
- [if (file.path.startsWith("modals/")) {](https://github.com/helloitsian/custom-modals-obsidian/blob/99691387027af5a7897968efa022c35673ae8a28/src/main.ts#L29) You probably want to make the path configurable
- You're probably better off listening for 'changed' (<https://github.com/obsidianmd/obsidian-api/blob/ff121cd409350d3afe39b08420f468d8aaee925e/obsidian.d.ts#L2176>) which passes you the TFile and the file contents
- [const root = this.app.vault.getRoot();](https://github.com/helloitsian/custom-modals-obsidian/blob/99691387027af5a7897968efa022c35673ae8a28/src/filesManager/FilesManager.ts#L19) I think this would be simpler if you use `getMarkdownFiles().filter(file => file.startsWith('your-path')`.
-  [async getExtraFileData(file: any) {](https://github.com/helloitsian/custom-modals-obsidian/blob/99691387027af5a7897968efa022c35673ae8a28/src/filesManager/FilesManager.ts#L10) I suggest not using `any` here to avoid running into typing issues.
- [.innerHTML = this.customModalSettings.content;](https://github.com/helloitsian/custom-modals-obsidian/blob/99691387027af5a7897968efa022c35673ae8a28/src/modal/CustomModal.ts#L42) while I really don't like seeing `innerHTML` used, if there's no way to avoid it, can we at least sanitize the html using [`sanitizeHTMLToDom`](https://github.com/obsidianmd/obsidian-api/blob/ff121cd409350d3afe39b08420f468d8aaee925e/obsidian.d.ts#L2791)
- [const okButton = document.createElement("button");](https://github.com/helloitsian/custom-modals-obsidian/blob/99691387027af5a7897968efa022c35673ae8a28/src/modal/CustomModal.ts#L44) You can simplify these `document.create*` calls using Obsidian's helper functions (`createEl` can take a className and text)
- [onClose(): void {](https://github.com/helloitsian/custom-modals-obsidian/blob/99691387027af5a7897968efa022c35673ae8a28/src/modal/CustomModal.ts#L104) I don't think you need this since the whole modal should get detached.

helloitsian:
> @helloitsian Hi Ian! Do you still intend to submit this plugin? Thanks!

Unfortunately life has caught up with me and I no longer have time to invest in this. My apologies and thanks to everyone who spent time and reviewed it!

PR #1029, opened 2022-06-28
Add obsidian-trim-whitespace plugin

zlovatt:

## Repo URL

Link to my plugin: https://github.com/zlovatt/obsidian-trim-whitespace

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
<https://github.com/zlovatt/obsidian-trim-whitespace/blob/76a74fcb6739dcfb687ce14129811aff9f8ad022/src/main.ts#L128-L138>
You should use a `editorCallback´ instead(<https://marcus.se.net/obsidian-plugin-docs/guides/commands#editor-commands>), 
then you can get rid of your `_getEditor` function and pass that editor instance to your other functions.
This will also make sure that your commands are only available if there is an active editor instance.

zlovatt:
[Updated plugin](https://github.com/zlovatt/obsidian-trim-whitespace/releases/tag/0.2.0) to implement this change, as well as handled a few bugfixes & feature adds.

PR #1030, opened 2022-06-29
Add `obsidian-meta-bind-plugin` plugin

mProjectsCode:

## Repo URL

Link to my plugin: https://github.com/mProjectsCode/obsidian-meta-bind-plugin

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
<https://github.com/mProjectsCode/obsidian-meta-bind-plugin/blob/1ce297e664aeac05ab02343d20c8dd333ccc63d1/src/main.ts#L53>
You don't need to cast here, TypeScript will that for you as you already did a `instanceof` check.

<https://github.com/mProjectsCode/obsidian-meta-bind-plugin/blob/1ce297e664aeac05ab02343d20c8dd333ccc63d1/src/main.ts#L99-L101>
You can move this after the metadata check, as it will be used after this, saving you some unnecessary operations

<https://github.com/mProjectsCode/obsidian-meta-bind-plugin/blob/1ce297e664aeac05ab02343d20c8dd333ccc63d1/src/main.ts#L139-L166>
Obsidian already has the frontmatter indexed in the metadata cache: <https://github.com/obsidianmd/obsidian-api/blob/d4b79f95ccf3838a305915a137a844250766d023/obsidian.d.ts#L2151> `this.app.metadataCache.getFileCache(yourFile)?.frontmatter`

<https://github.com/mProjectsCode/obsidian-meta-bind-plugin/blob/1ce297e664aeac05ab02343d20c8dd333ccc63d1/src/main.ts#L121>
This will also include all non markdown files.
You might want to only use the markdown files, `this.app.vault.getMarkdownFiles()`.

<https://github.com/mProjectsCode/obsidian-meta-bind-plugin/blob/master/src/parsers/DateParser.ts>
Obsidian includes Moment(<https://momentjs.com/>), then you are also not limited to a few preset date formats.

mProjectsCode:
@joethei thanks for the review :)

> https://github.com/mProjectsCode/obsidian-meta-bind-plugin/blob/1ce297e664aeac05ab02343d20c8dd333ccc63d1/src/main.ts#L53
You don't need to cast here, TypeScript will that for you as you already did a instanceof check.

Fixed

> https://github.com/mProjectsCode/obsidian-meta-bind-plugin/blob/1ce297e664aeac05ab02343d20c8dd333ccc63d1/src/main.ts#L99-L101
You can move this after the metadata check, as it will be used after this, saving you some unnecessary operations

Fixed

> https://github.com/mProjectsCode/obsidian-meta-bind-plugin/blob/1ce297e664aeac05ab02343d20c8dd333ccc63d1/src/main.ts#L139-L166
Obsidian already has the frontmatter indexed in the metadata cache: https://github.com/obsidianmd/obsidian-api/blob/d4b79f95ccf3838a305915a137a844250766d023/obsidian.d.ts#L2151 this.app.metadataCache.getFileCache(yourFile)?.frontmatter

I found the metadata cache to be unreliable when calling it from the vault file modify event. I am guessing there are some race conditions involved.

> https://github.com/mProjectsCode/obsidian-meta-bind-plugin/blob/1ce297e664aeac05ab02343d20c8dd333ccc63d1/src/main.ts#L121
This will also include all non markdown files.
You might want to only use the markdown files, this.app.vault.getMarkdownFiles().

fixed

> https://github.com/mProjectsCode/obsidian-meta-bind-plugin/blob/master/src/parsers/DateParser.ts
Obsidian includes Moment(https://momentjs.com/), then you are also not limited to a few preset date formats.

I switched the parser

mProjectsCode:
I haven't issued a new release with those changes, I want to wait for further reviews

liamcain:
> I found the metadata cache to be unreliable when calling it from the vault file modify event. I am guessing there are some race conditions involved

Instead of listening for the vault modify event, you can listen for `"changed"` (https://github.com/obsidianmd/obsidian-api/blob/4583a6b8305b05139c6b1bf226ad15e50a9a5987/obsidian.d.ts#L2185). Then you will be sent the most up to date metadata from the cache.

---

Otherwise it looks good to me. Let me know when you're ready for this to be released.

mProjectsCode:
Thanks, I will look into that. 
But since my PC is currently broken, it will probably take me another week to do that and get the other stuff I have been working on release ready.

mProjectsCode:
I finally managed to come back to the plugin. It should be ready for release (the docs are still unfinished, but I will do that later).

> Instead of listening for the vault modify event, you can listen for "changed" (https://github.com/obsidianmd/obsidian-api/blob/4583a6b8305b05139c6b1bf226ad15e50a9a5987/obsidian.d.ts#L2185). Then you will be sent the most up to date metadata from the cache.

that is now implemented

liamcain:
Nice looks good! Something I noticed after merging this, you have the plugin name as "meta bind plugin" but it might be weird having the word 'plugin' in the plugin name itself.

PR #1032, opened 2022-07-02
add typing transformer plugin

aptend:

## Repo URL

Link to my plugin:
https://github.com/aptend/typing-transformer-obsidian

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
The plugin probably requires at least Obsidian 0.15 as the update to version 6.0 of the CodeMirror packages happened there.



Not using semicolons to end statements can lead to rare, hard to debug problems: <https://flaviocopes.com/javascript-automatic-semicolon-insertion/>

[ruleArea.settingEl.setAttribute("style", "display: grid; grid-template-columns: 1fr;")](https://github.com/aptend/typing-transformer-obsidian/blob/f8843aab346524d24999674c53c4c830c7a5a5ff/src/main.ts#L295-L296)
You should put those into your `styles.css` file and just add a class to the `settingEl`, making this available for themes to style.

aptend:
@joethei Thanks for the reivew!

> The plugin probably requires at least Obsidian 0.15 as the update to version 6.0 of the CodeMirror packages happened there.

Now I develop this plugin on Obsidian 0.14.15 , and it works well on Mac and Windows. And it seems that I have on access to Obsidian 0.15…… so what should I do, set min app version 0.14.15 would be enough?

> Not using semicolons to end statements can lead to rare, hard to debug problems: https://flaviocopes.com/javascript-automatic-semicolon-insertion/

Fixed

> [ruleArea.settingEl.setAttribute("style", "display: grid; grid-template-columns: 1fr;")](https://github.com/aptend/typing-transformer-obsidian/blob/f8843aab346524d24999674c53c4c830c7a5a5ff/src/main.ts#L295-L296) You should put those into your `styles.css` file and just add a class to the `settingEl`, making this available for themes to style.

Fixed [.setClass("rules-text-area")](https://github.com/aptend/typing-transformer-obsidian/blob/c6dd0838bce56c0dd164438391c50b643c34ee6a/src/settings.ts#L76)

joethei:
Okay if it works that is fine.
With version 0.15.0 there is a upgrade in the CM version Obsidian is using.
(0.19.x - > 6.0
A few API`s have changed as a result.
As I was not sure which one you tested on I left this warning.

aptend:
I have set min app version 0.14.15 and issued a new version contains all updates. Any other suggestions?

liamcain:
- [await init(wasmbin);](https://github.com/aptend/typing-transformer-obsidian/blob/9978096527b7b49e4f38739f1baa65c5a2d307a7/src/main.ts#L52) Plugin `onload` is a synchronous, so it should exit pretty quickly, is this call expensive? If it is, maybe we can defer the loading or wrap it in a setTimeout so it happens in it's own event loop

---

Otherwise, looks good to me!

aptend:
This call boils down to `WebAssembly.instantiate()`, it costs 2ms on my Mac, while the whole `onload` costs 4ms. I guess it is acceptable for now? Anyway I will keep an eye on this, maybe I can optimize it in next version, thank you~

PR #1035, opened 2022-07-03
Add obsidian-golinks

xavdid:

## Repo URL

Link to my plugin: https://github.com/xavdid/obsidian-golinks

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [goLinkRegex](https://github.com/xavdid/obsidian-golinks/blob/f2823ceae3257262ee15299a28cd0f9d11e6e378/src/utils.ts#L1) You might want to add some word boundaries to your regex to guard against false positives (i.e. `bongo/link`)
- [this.containerEl.replaceChildren(...results);](https://github.com/xavdid/obsidian-golinks/blob/f2823ceae3257262ee15299a28cd0f9d11e6e378/src/main.ts#L47-L48) Instead of `replaceChildren`, you can use [setChildrenInPlace](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L76) which will only replace items that need replacing, so it should be faster.


---

Just those 2 things, otherwise everything looks good!

xavdid:
Awesome, thank you! I was planning on improving the regex and adding some tests, so I'll get those in. Great tip on the `setChildrenInPlace`, too! Hard to know which method is the best way sometimes.

PR #1039, opened 2022-07-04
Add TimeDiff plugin to community-plugin.json

dominiczaq:

## Repo URL

Link to my plugin: https://github.com/dominiczaq/obsidian-plugin-time-diff

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
Please make sure that the `versions.json` file has the correct content(you don't need to every version, only those where the targeted Obsidian version changes)

[diffSpan.style.color = "var(--interactive-accent)";](https://github.com/dominiczaq/obsidian-plugin-time-diff/blob/f36b8546b2bb08e8d1266b6614f85b1df4b6b7c8/main.ts#L41)
Put these into the CSS file to make it easier for theme devs to change the styling.
I don't think anyone will do that with this small plugin, but keep it in mind for any potential future plugins.

Obsidian brings the MomentJS library by default, you might want to switch to it if possible.

dominiczaq:
Hey @joethei , thank you very much for the review!
I've pushed version 0.0.2 with all your suggested changes.

liamcain:
- https://github.com/dominiczaq/obsidian-plugin-time-diff/blob/3df8f802a491d0d15fa767b48be7bb5cd0febbcb/main.ts#L22 Maybe don't include the spaces in here and just split on "-" but use `trim()` on the resulting split elements.
- https://github.com/dominiczaq/obsidian-plugin-time-diff/blob/3df8f802a491d0d15fa767b48be7bb5cd0febbcb/main.ts#L59 You can also use Moment JS's `humanize` function for rendering diffs to text (https://momentjs.com/docs/#/durations/humanize/)

PR #1046, opened 2022-07-09
Add Path Finder Plugin

jerrywcy:

## Repo URL

Link to my plugin: https://github.com/jerrywcy/obsidian-path-finder

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[MyPlugin](https://github.com/jerrywcy/obsidian-path-finder/blob/f9a39af1348766d06bf50379a29e024dd9687c39/src/main.ts#L10)
please rename this class as it can lead to conflicts with other plugins.

[!adapter.exists(from)](https://github.com/jerrywcy/obsidian-path-finder/blob/f9a39af1348766d06bf50379a29e024dd9687c39/src/main.ts#L135)
use the [`app.vault.getAbstractFileByPath`](https://github.com/obsidianmd/obsidian-api/blob/b9bde9e32c007eb28c0fd632cf2c42b01dfd022a/obsidian.d.ts#L3411) function instead, it will return null if there is no such file.
The adapter is not guaranted to be available on all operating systems.

[findShortestPath](https://github.com/jerrywcy/obsidian-path-finder/blob/f9a39af1348766d06bf50379a29e024dd9687c39/src/main.ts#L54)[findAllPathsAsGraph](https://github.com/jerrywcy/obsidian-path-finder/blob/f9a39af1348766d06bf50379a29e024dd9687c39/src/main.ts#L90)[findAllPaths](https://github.com/jerrywcy/obsidian-path-finder/blob/f9a39af1348766d06bf50379a29e024dd9687c39/src/main.ts#L129)
These functions all do pretty much the same thing, with minor changes, I would recommend refactoring this to remove the duplication.

[modals.ts](https://github.com/jerrywcy/obsidian-path-finder/blob/f9a39af1348766d06bf50379a29e024dd9687c39/src/modals.ts)
You could split this up into seperate files to make it easier to read.


[g.getName(g.g[i].v)](https://github.com/jerrywcy/obsidian-path-finder/blob/f9a39af1348766d06bf50379a29e024dd9687c39/src/view.ts#L48)
You should rename these, I have no idea what any of this means.
Ss with any other variables/function that have these short names, rename them to have meaningful names.

[leftButtonContainer.style.setProperty("height", "100%");](https://github.com/jerrywcy/obsidian-path-finder/blob/f9a39af1348766d06bf50379a29e024dd9687c39/src/view.ts#L185-L189)
If you can set these in the CSS, this will make it easier for others to overwrite.

[link](https://github.com/jerrywcy/obsidian-path-finder/blob/f9a39af1348766d06bf50379a29e024dd9687c39/styles.css#L25)
You should use unique CSS classes as these are also used in Core Obsidian.

jerrywcy:
@joethei All problems should be fixed in [Release 1.0.10](https://github.com/jerrywcy/obsidian-path-finder/releases/tag/1.1.10)

jerrywcy:
@joethei Sorry for my mistake... The right link should be [Release 1.0.10](https://github.com/jerrywcy/obsidian-path-finder/releases/tag/1.0.10).

liamcain:
- [if (!adapter.exists(from)) {](https://github.com/jerrywcy/obsidian-path-finder/blob/b907da313b4f427bf1c1eec2e41be38b9158070f/src/main.ts#L69) `exists` is async, so this conditional will never actually be true because you're checking if `Promise<boolean>` is truthy. You should change these to getAbstractFileByPath, as joethei recommended above
- [await workspace.getLeaf(true).setViewState({](https://github.com/jerrywcy/obsidian-path-finder/blob/b907da313b4f427bf1c1eec2e41be38b9158070f/src/main.ts#L137-L138) You can simplify this a lot by assigning the result of `getLeaf` to a variable. Then you can do `leaf.setViewData()` followed by `leaf.view.setData()`. And you don't need to scan the whole workspace for the leaf you just created.
- [let pathViewLeaf = workspace.getLeavesOfType(VIEW_TYPE_PATHVIEW)[0];](https://github.com/jerrywcy/obsidian-path-finder/blob/b907da313b4f427bf1c1eec2e41be38b9158070f/src/main.ts#L182-L183) Same issue as above
- [const leftButtonContainer = container.createDiv();](https://github.com/jerrywcy/obsidian-path-finder/blob/b907da313b4f427bf1c1eec2e41be38b9158070f/src/view.ts#L207-L209) You can pass the classes into `createDiv` so that it doesn't have to be on 2 separate lines

jerrywcy:
@liamcain All problems mentioned should be fixed in [Release 1.0.12](https://github.com/jerrywcy/obsidian-path-finder/releases/tag/1.0.12)

PR #1049, opened 2022-07-12
Add table generator

Quorafind:

## Repo URL

Link to my plugin: https://github.com/Quorafind/Obsidian-Table-Generator

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4541,5 +4541,12 @@
         "author": "Grzegorz Dominiczak",
         "description": "Calculates and displays diff in hours and minutes between two dates in `timediff` markdown block",
         "repo": "dominiczaq/obsidian-plugin-time-diff"
+    },
+    {
+        "id": "obsidian-table-generator",
+        "name": "Table Generator",
+        "author": "Boninall",
+        "description": "A plugin for generate markdown table quickly like Typora.",
+        "repo": "quorafind/obsidian-table-generator"
     }
 ]


joethei:
[this.tableGeneratorEl.style.position = "absolute";](https://github.com/Quorafind/Obsidian-Table-Generator/blob/da99835da66e812030f0b45c18da95ae616eec79/src/tableGeneratorIndex.ts#L78-L80)
you should move these into the CSS file to make it easier to style, if possible.

[SampleSettingTab](https://github.com/Quorafind/Obsidian-Table-Generator/blob/da99835da66e812030f0b45c18da95ae616eec79/src/tableGeneratorIndex.ts#L126)
Rename this to avoid conflicts.

[this.applyDebounceTimer = window.setTimeout(() => {](https://github.com/Quorafind/Obsidian-Table-Generator/blob/da99835da66e812030f0b45c18da95ae616eec79/src/tableGeneratorIndex.ts#L138-L140)
You can use the [`.registerInterval`](https://github.com/obsidianmd/obsidian-api/blob/b9bde9e32c007eb28c0fd632cf2c42b01dfd022a/obsidian.d.ts#L623-L627) from the API instead, it will be automatically unloaded.
```ts
this.registerInterval(window.setTimeout(() => {
            plugin.saveSettings();
        }, 100)
);
```


---
The following are just style nitpicks, you don't have fix these.

[(evt.target as HTMLElement)](https://github.com/Quorafind/Obsidian-Table-Generator/blob/da99835da66e812030f0b45c18da95ae616eec79/src/tableGeneratorIndex.ts#L40)
You are doing this multiple times, at multiple places, you could just save this in a variable.

Quorafind:
@joethei Fixed in https://github.com/Quorafind/Obsidian-Table-Generator/commit/26476699f14e90a1d4ef4e01b4e577d46ea45dbd

liamcain:
- [target.classList](https://github.com/Quorafind/Obsidian-Table-Generator/blob/26476699f14e90a1d4ef4e01b4e577d46ea45dbd/src/tableGeneratorIndex.ts#L45) I think you need to check if target is null here too.
- [this.registerDomEvent(window, 'click', (evt: MouseEvent) => {](https://github.com/Quorafind/Obsidian-Table-Generator/blob/26476699f14e90a1d4ef4e01b4e577d46ea45dbd/src/tableGeneratorIndex.ts#L54) I think you don't need this event listener. You should just keep the event listener on `document` and the other `window-open` event listener. I think that's enough though.
- [if (this.tableGeneratorEl === undefined) return](https://github.com/Quorafind/Obsidian-Table-Generator/blob/26476699f14e90a1d4ef4e01b4e577d46ea45dbd/src/tableGeneratorIndex.ts#L111) this line is redundant, you can remove it.
- [class tableGeneratorSettingTab](https://github.com/Quorafind/Obsidian-Table-Generator/blob/26476699f14e90a1d4ef4e01b4e577d46ea45dbd/src/tableGeneratorIndex.ts#L131) capitalize the name of your class
- [el.style.minWidth = "2.3em";](https://github.com/Quorafind/Obsidian-Table-Generator/blob/26476699f14e90a1d4ef4e01b4e577d46ea45dbd/src/tableGeneratorIndex.ts#L161-L162) you should just add a class and do these tweaks with CSS so that themes/snippets can override if they need to.
- [defaultWindow?: boolean](https://github.com/Quorafind/Obsidian-Table-Generator/blob/26476699f14e90a1d4ef4e01b4e577d46ea45dbd/src/utils/modifiedTable.ts#L3) I don't think you need this flag. Also this function is really hard to read since you have so much on one line. Could you reformat this function? I'm also not sure why you are using both `body.contains` _and_ querying for the element. Querying for the element should be enough. That said, you have a reference to the table element, so there's no reason you should be querying for it in the document at all. Just use this.tableGeneratorEl.detach()
- [$: if (selectedTableEnd) {](https://github.com/Quorafind/Obsidian-Table-Generator/blob/26476699f14e90a1d4ef4e01b4e577d46ea45dbd/src/ui/TableGenerator.svelte#L15) this looks dangerous. Why are you marking these svelte blocks as reactive? I think you are better off keeping these inside the event callbacks to avoid these triggering too frequently
- [export const tableGeneratorMenu](https://github.com/Quorafind/Obsidian-Table-Generator/blob/26476699f14e90a1d4ef4e01b4e577d46ea45dbd/src/tableGeneratorMenu.ts#L10) I don't see much advantage to having this in a util function. Just put this code directly into your plugin
- [tableGeneratorBoard.style.left = `${ coords.left }px`;](https://github.com/Quorafind/Obsidian-Table-Generator/blob/26476699f14e90a1d4ef4e01b4e577d46ea45dbd/src/tableGeneratorMenu.ts#L25-L27) Put these styles in your css file instead

claremacrae:
I'm sorry not to see this added. I've found it useful via BRAT.

Quorafind:
Sorry that I didn't update this. And it closed because I pull a new request automatically. I will reopen this soon. @claremacrae

claremacrae:
Oh that's great news. Thanks!

Quorafind:
@liamcain hi, I fixed all these issues.

liamcain:
- [if (!this.tableGeneratorEl) return;](https://github.com/Quorafind/Obsidian-Table-Generator/blob/5bfeaffb612199004648e8174deac5bb4f0dc6db/src/tableGeneratorIndex.ts#L86-L87) Unnecessary check, `createElement` won't fail, so this will never run
- [if (this.tableGeneratorEl) this.tableGeneratorEl.detach();](https://github.com/Quorafind/Obsidian-Table-Generator/blob/5bfeaffb612199004648e8174deac5bb4f0dc6db/src/tableGeneratorIndex.ts#L80) No need to detach if you're just going to reattach. [`appendChild`](https://developer.mozilla.org/en-US/docs/Web/API/Node/appendChild) will automatically detach if the element is attached elsewhere. I feel like the logic you have here is a bit backwards. I think what you want to do is check if `tableGeneratorEl` doesn't exist, if so, create the element. Then unconditionally attach the element to the body.
- [registerTableGeneratorMenu](https://github.com/Quorafind/Obsidian-Table-Generator/blob/5bfeaffb612199004648e8174deac5bb4f0dc6db/src/tableGeneratorIndex.ts#L142) afaict, this function doesn't do anything. I think you can just remove it.
- [style](https://github.com/Quorafind/Obsidian-Table-Generator/blob/5bfeaffb612199004648e8174deac5bb4f0dc6db/src/ui/Table.svelte#L50) You can use svelte's [style directive](https://svelte.dev/tutorial/style-directive) instead. Another cool option would be passing in the row and column as css vars. That way you can keep the css inside your style tag.
- [hideTable](https://github.com/Quorafind/Obsidian-Table-Generator/blob/5bfeaffb612199004648e8174deac5bb4f0dc6db/src/ui/TableGenerator.svelte#L71) You already have a reference to your table instance in the plugin. I think it's better to just pass around the reference to the table rather than querying the DOM for it. Or just have `hideTable` be a function on the plugin instance, then call `plugin.hideTable` if you want to avoid passing the htmlelement instance around.

Quorafind:
@liamcain 

> [if (!this.tableGeneratorEl) return;](https://github.com/Quorafind/Obsidian-Table-Generator/blob/5bfeaffb612199004648e8174deac5bb4f0dc6db/src/tableGeneratorIndex.ts#L86-L87) Unnecessary check, createElement won't fail, so this will never run

Remove. And use `createEl` from Obsidian now.

> [if (this.tableGeneratorEl) this.tableGeneratorEl.detach();](https://github.com/Quorafind/Obsidian-Table-Generator/blob/5bfeaffb612199004648e8174deac5bb4f0dc6db/src/tableGeneratorIndex.ts#L80) No need to detach if you're just going to reattach. [appendChild](https://developer.mozilla.org/en-US/docs/Web/API/Node/appendChild) will automatically detach if the element is attached elsewhere. I feel like the logic you have here is a bit backwards. I think what you want to do is check if tableGeneratorEl doesn't exist, if so, create the element. Then unconditionally attach the element to the body.

But `appendChild` didn't work with popout windows, for example when I create a new table generator in the popout window, the old one would not auto detach. So if you know how to make this work well with popout windows , I need some help 😂 Or I have to leave the code here.

> [registerTableGeneratorMenu](https://github.com/Quorafind/Obsidian-Table-Generator/blob/5bfeaffb612199004648e8174deac5bb4f0dc6db/src/tableGeneratorIndex.ts#L142) afaict, this function doesn't do anything. I think you can just remove it.

This does same thing like above.

> [style](https://github.com/Quorafind/Obsidian-Table-Generator/blob/5bfeaffb612199004648e8174deac5bb4f0dc6db/src/ui/Table.svelte#L50) You can use svelte's [style directive](https://svelte.dev/tutorial/style-directive) instead. Another cool option would be passing in the row and column as css vars. That way you can keep the css inside your style tag.

Used style directive.

> [hideTable](https://github.com/Quorafind/Obsidian-Table-Generator/blob/5bfeaffb612199004648e8174deac5bb4f0dc6db/src/ui/TableGenerator.svelte#L71) You already have a reference to your table instance in the plugin. I think it's better to just pass around the reference to the table rather than querying the DOM for it. Or just have hideTable be a function on the plugin instance, then call plugin.hideTable if you want to avoid passing the htmlelement instance around.

Move hidetable to plugin class.

liamcain:
LGTM, do you mind rebasing this PR? Github has decided that I can't resolve the conflicts via the web UI for some reason.

Quorafind:
@liamcain https://github.com/obsidianmd/obsidian-releases/pull/1212

PR #1050, opened 2022-07-12
updating community-plugins with meeting notes plugin

TimHi:

## Repo URL

Link to my plugin: https://github.com/TimHi/obsidian-meeting-notes

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[await this.loadSettings();](https://github.com/TimHi/obsidian-meeting-notes/blob/6b788b646acaa21986e17fefdae06e07705a64a9/main.ts#L17)
You can just read from the settings object, as you are always updating it on settings change.
I also don't see you accessing the settings object anywhre in this function, so I don't think this is even necessary here.

You are using semicolons most of the time, but are missing them sometimes,
not having semicolons can lead to rare, hard to debug issues:
https://flaviocopes.com/javascript-automatic-semicolon-insertion/(this is not an issue in this plugin, but letting you know for future reference).

TimHi:
@joethei Added semicolons and removed the settings loading, tested without and it worked fine recognizing folder changes in the settings.

liamcain:
- https://github.com/TimHi/obsidian-meeting-notes/blob/763bdc3692fcc06ab0e11f498d9ca5179ad53914/src/filehandler/FileService.ts#L44 Should should probably also make sure that the file extension is `.md`
- https://github.com/TimHi/obsidian-meeting-notes/blob/763bdc3692fcc06ab0e11f498d9ca5179ad53914/src/filehandler/FileRenderer.ts#L7 Obsidian uses `{{date}}` for all of its core templating plugins (Daily Notes, Unique Note Creator, Templates), so it probably makes sense to stay consistent with that in case users want to use the same note as a template for the templates plugin down the line.
- https://github.com/TimHi/obsidian-meeting-notes/blob/763bdc3692fcc06ab0e11f498d9ca5179ad53914/src/filehandler/FileRenderer.ts#L3 I don't really see much reason for this to be a class, it could just be a function.

TimHi:
@liamcain Thanks for the suggestions, I added the check for the markdown extension and exchanged {Date} with {{Date}} as templating the generated file is something I plan to do as soon as I get the time.

When doing the template I plan to use the FileRenderer class for it so it will be used, thats just my preference to keep it seperated.

PR #1053, opened 2022-07-13
Added obsidian-ocr to community-plugins.json

MohrJonas:

## Repo URL

Link to my plugin:

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[MyPlugin](https://github.com/MohrJonas/obsidian-ocr/blob/93b2b7d724901b19950b53934118614f2db5b5f5/src/main.ts#L7)
Rename this to avoid conflicts with other plugins.

[vaultPathToAbs](https://github.com/MohrJonas/obsidian-ocr/blob/93b2b7d724901b19950b53934118614f2db5b5f5/src/utils.ts#L21)
you can use the [`vault.adapter.getFullPath´](https://github.com/obsidianmd/obsidian-api/blob/b9bde9e32c007eb28c0fd632cf2c42b01dfd022a/obsidian.d.ts#L1243) function instead.

[getFileEnding](https://github.com/MohrJonas/obsidian-ocr/blob/93b2b7d724901b19950b53934118614f2db5b5f5/src/utils.ts#L41)
A `TFile` has the `extension` property for this.

[file as TFile](https://github.com/MohrJonas/obsidian-ocr/blob/93b2b7d724901b19950b53934118614f2db5b5f5/src/utils.ts#L138)
This cast is useless.

MohrJonas:
Thanks, I'll change these problems ASAP 😁

MohrJonas:
Alright, I implemented the changes. Do I now have to create a new pull request?

joethei:
No, you don't need to create a new one.
There are multiple review rounds, this will be merged once all reviewers are satisfied.

liamcain:
- [(await listAllFiles(this.app.vault)).forEach(async (file) => { await processFile(this, file, this.app.vault); });](https://github.com/MohrJonas/obsidian-ocr/blob/29978415392ccf9daa4cc2c1d1ab7ca79fbd1ce0/src/main.ts#L14-L15) `onload` is blocking, so it's important for plugins to only do essential setup inside the `onload` function. For all other work, I suggest wrapping within an [`onLayoutReady`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3626) callback. This will avoid slowing down Obsidian on startup
- [//@ts-ignore](https://github.com/MohrJonas/obsidian-ocr/blob/29978415392ccf9daa4cc2c1d1ab7ca79fbd1ce0/src/main.ts#L24) You should avoid having so many tsignore's. It's hard to see if it's masking a bug and without any more context, I don't know what error you're surpressing
- [if (existsSync(vaultPathToAbs(vault, filePathToJsonPath(file.path))) || !(await isFileValid(vault, file))) return;](https://github.com/MohrJonas/obsidian-ocr/blob/29978415392ccf9daa4cc2c1d1ab7ca79fbd1ce0/src/utils.ts#L133-L134) You're doing a lot of work inside processFile before even checking if the file is an image file or PDF. I'd recommend adding an early exit to avoid adding too much extra processing after each file creation
- [static addIndexingFile(file: TAbstractFile) {](https://github.com/MohrJonas/obsidian-ocr/blob/29978415392ccf9daa4cc2c1d1ab7ca79fbd1ce0/src/status-bar.ts#L33-L34) type here should be TFile (not abstract file). Also, no need to process non-image or pdf files
- [export async function listAllFiles(vault: Vault): Promise<Array<TFile>> {](https://github.com/MohrJonas/obsidian-ocr/blob/29978415392ccf9daa4cc2c1d1ab7ca79fbd1ce0/src/utils.ts#L121-L122) You can use [`listAllFiles`](https://github.com/obsidianmd/obsidian-api/blob/4583a6b8305b05139c6b1bf226ad15e50a9a5987/obsidian.d.ts#L3505) instead. I also suggest pruning this list before returning it so that it only includes files with extensions that the plugin supports

MohrJonas:
The //@ts-ignore either ignores that vault.adapter.getFullPath or vault.adapter.basePath apparently do not exist or that an import doesn't work even though it does.

liamcain:
> The //@ts-ignore either ignores that vault.adapter.getFullPath or vault.adapter.basePath apparently do not exist or that an import doesn't work even though it does.

- The reason is you need to check the type of `adapter`.

```ts
if (vault.adapter instanceof FileSystemAdapter) {
  const absolutePath = vault.adapter.getFullPath(file);
}
```

The reason for this is that there is also a mobile file system adapter

---

- [console.log(jsonObject);](https://github.com/MohrJonas/obsidian-ocr/blob/eb7ccff7e90fe95788422a257a62c599ce371fc3/src/utils.ts#L165) you can remove this console.log to avoid spamming the console
- [this.app.workspace.getMostRecentLeaf(](https://github.com/MohrJonas/obsidian-ocr/blob/eb7ccff7e90fe95788422a257a62c599ce371fc3/src/modals/search-modal.ts#L67) This should be using `getLeaf` instead. `getMostRecentLeaf` can return null and it doesn't take into account if a leaf is pinned. This should all be taken care of for you with `getLeaf`.

- [new DependencyModal(this.app, await doesProgramExist("tesseract"), await doesProgramExist("gm"), await doesProgramExist("gs")).open();](https://github.com/MohrJonas/obsidian-ocr/blob/eb7ccff7e90fe95788422a257a62c599ce371fc3/src/main.ts#L10-L11) You shouldn't be opening a modal inside the `onload` function. Also, the multiple `doesProgramExist` function calls will run in series which will block Obsidian startup. You should move this down into `onLayoutReady`. I also suggest checking for all the dependencies and only calling `.open()` if any of the dendencies are missing.

PR #1054, opened 2022-07-14
add .mov support plugin

connectdotz:

## Repo URL

Link to my plugin: https://github.com/connectdotz/obsidian-mov-support

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4541,5 +4541,12 @@
         "author": "Grzegorz Dominiczak",
         "description": "Calculates and displays diff in hours and minutes between two dates in `timediff` markdown block",
         "repo": "dominiczaq/obsidian-plugin-time-diff"
+    },
+    {
+        "id": "obsidian-mov-support",
+        "name": "MOV Support",
+        "author": "connectdotz",
+        "description": "To view/play embedded .mov video inline",
+        "repo": "connectdotz/obsidian-mov-support"
     }
 ]


lishid:
Hey thanks for the PR. I wasn't aware that electron already works with .mov files!

It seems to have taken quite some effort for your plugin to do all that work to add mov support. I think it makes sense to add support for this directly in core, which should make your life easier.

Would it make sense to wait until this land in core and re-evaluate how this plugin can be adapted for enhancements?

aidenlx:
Something may worth noting, when the `.mov` format is whitelisted, an additional note in [obsidian help page around file format](https://help.obsidian.md/Advanced+topics/Accepted+file+formats) about codec support between safari and chromium ([vp9](https://caniuse.com/webm) and [hevc](https://caniuse.com/hevc) in particular) could be helpful since the mp4 and mov container can sometimes include hevc codec, which won't play well in electron apps, and audio recording from electron (using webm container) won't play on iOS because of missing Vorbis codec support in safari. 

btw, `.mkv` container format is also supported by chromium, and will play well if the codec is supported, but it won't work in safari. Considering that webm and ogg format are also not accepted by safari, maybe it can be whitelisted as well?

liamcain:
Hi @connectdotz, we just added `.mov` support to Obsidian core in v0.15.7–Inspired by your plugin. The build is currently only available for insiders; if you don't have an insider license yet, we can add one to your account (it's the least we can do after Sherlocking your plugin!). Just DM me on Discord (`liam#3676`)

PR #1055, opened 2022-07-14
Add New-Note-Plus plugin

ilucaslee:

## Repo URL

Link to my plugin:[obsidian-new-note-plus](https://github.com/ilucaslee/obsidian-new-note-plus)

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4541,5 +4541,13 @@
         "author": "Grzegorz Dominiczak",
         "description": "Calculates and displays diff in hours and minutes between two dates in `timediff` markdown block",
         "repo": "dominiczaq/obsidian-plugin-time-diff"
+    },
+    {
+        "id": "new-note-plus",
+        "name": "New Note Plus",
+        "author": "Lucas Lee",
+        "description": "A new URL Scheme to create new note using clipboard contents",
+        "repo": "ilucaslee/obsidian-new-note-plus",
+        "branch": "main"
     }
 ]


liamcain:
- [new-note-plus](https://github.com/ilucaslee/obsidian-new-note-plus/blob/c25dc89925a2ea80cccfb1371e3c1cffaff9cdea/main.ts#L6) It's not very clear from the parameter that this creates a file with the clipboard contents. Why not call the url action `new-file-from-clipboard` or similar?
- [await this.app.vault.modify(file, text);](https://github.com/ilucaslee/obsidian-new-note-plus/blob/c25dc89925a2ea80cccfb1371e3c1cffaff9cdea/main.ts#L35) this looks dangerous. Given the name of the plugin and the fact that the uri specifically says "new-note," I wouldn't expect it to modify an existing note. Maybe this should either create a new note with a unique number at the end or display a prompt asking if the user wants to replace the file contents.
- [} else {](https://github.com/ilucaslee/obsidian-new-note-plus/blob/c25dc89925a2ea80cccfb1371e3c1cffaff9cdea/main.ts#L36) it looks like it's possible for `file` to exist but be pointing to a TFolder and still get into this `else` block.
- [await editor.focus();](https://github.com/ilucaslee/obsidian-new-note-plus/blob/c25dc89925a2ea80cccfb1371e3c1cffaff9cdea/main.ts#L51-L52) I don't think `focus` is async, no need to `await` it
- [var newPane = false;](https://github.com/ilucaslee/obsidian-new-note-plus/blob/c25dc89925a2ea80cccfb1371e3c1cffaff9cdea/main.ts#L45-L46) this variable looks unnecessary, just pass in `false` directly into `openLinkText`
- [await navigator.clipboard.readText();](https://github.com/ilucaslee/obsidian-new-note-plus/blob/c25dc89925a2ea80cccfb1371e3c1cffaff9cdea/main.ts#L32) You should wrap `readText` in a try/catch since it can throw exceptions in a few cases, such as if the window isn't focused.

PR #1059, opened 2022-07-15
Plugin Submission: Hard Breaks Plugin

bkis:

## Repo URL

Link to my plugin: https://github.com/bkis/obsidian-hard-breaks

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [Whitespace](https://github.com/bkis/obsidian-hard-breaks/blob/a480b0ba40ec6458a30c62b395e6f638d9db1a9c/src/main.ts#L91) nitpick by you seem to switch between not using semi-colons and using semi-colons is this file. I suggest using semicolons throughout to avoid bugs.
- [replaceBreaks](https://github.com/bkis/obsidian-hard-breaks/blob/a480b0ba40ec6458a30c62b395e6f638d9db1a9c/src/main.ts#L43) My only concern is this might choke on some other custom MD syntaxes that obsidian uses, such as footnotes and LaTeX.

bkis:
> * [Whitespace](https://github.com/bkis/obsidian-hard-breaks/blob/a480b0ba40ec6458a30c62b395e6f638d9db1a9c/src/main.ts#L91) nitpick by you seem to switch between not using semi-colons and using semi-colons is this file. I suggest using semicolons throughout to avoid bugs.

Thank you for you feedback! I am not at home in JS. Coming from Java, I'd have used semi-colons everywhere by instinct. This was my first time working with a TS project and I see many folks leaving out the semi-colons in modern JS - so I thought, well, if that's how they're doing it today, I can adapt. But I guess my habits came through here and there. I'll fix this to use semi-colons. 

>     * [replaceBreaks](https://github.com/bkis/obsidian-hard-breaks/blob/a480b0ba40ec6458a30c62b395e6f638d9db1a9c/src/main.ts#L43) My only concern is this might choke on some other custom MD syntaxes that obsidian uses, such as footnotes and LaTeX.

Good point! I'll add functionality to the parser to detect those block types so they're not misinterpreted as paragraphs :+1:

bkis:
A quick follow-up: The plugin leaves footnotes and LaTeX blocks untouched. As it's GFM-aware, it seems to properly recognize said blocks. Thanks anyway for the hint - I didn't test this before.

PR #1060, opened 2022-07-17
Adds Tagged-Documents-Viewer Plugin

mgeduld:
obsidian-plugin-tagged-documents-viewer

## Repo URL

Link to my plugin: https://github.com/mgeduld/obsidian-tagged-documents-viewer

## Release Checklist
  - [ ]  Windows
  - [ x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
- [ x] My GitHub release contains all required files
  - [ x] `main.js`
  - [ x] `manifest.json`
  - [ x] `styles.css` _(optional)_


mgeduld:
I have updates the release to add main.js.

joethei:
[isMacOS](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/23515b906e98f189dde2bd7cfd96072f80a9844a/utils/links.ts#L8)
There is already [`Platform.isMacOS`](https://github.com/obsidianmd/obsidian-api/blob/b9bde9e32c007eb28c0fd632cf2c42b01dfd022a/obsidian.d.ts#L2426) build in.

[contents.push({file, text: await this.app.vault.cachedRead(file)})](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/23515b906e98f189dde2bd7cfd96072f80a9844a/main.ts#L225)
You can run into memory issues when doing this with many files.

[document.querySelector](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/23515b906e98f189dde2bd7cfd96072f80a9844a/main.ts#L249)
Don't do anything with `document` if you can avoid it,
With the recent Obsidian update to 0.15 `document` only references the main window.
Use the `contentEl` instead.

[// @ts-ignore](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/23515b906e98f189dde2bd7cfd96072f80a9844a/main.ts#L238)
just set the last parameter of `MarkdownRenderer.renderMarkdown` to `this`,
That will make sure that the lifecyle is properly managed,



---

For future reference:
You are nosemicolons most of the time,
not having semicolons can lead to rare, hard to debug issues:
https://flaviocopes.com/javascript-automatic-semicolon-insertion

mgeduld:
Thanks. I am working on corrections. Regarding "just set the last parameter of MarkdownRenderer.renderMarkdown to this," I get a typescript error when I try that. 

"Argument of type 'this' is not assignable to parameter of type 'Component'.
Type 'TaggedDocumentsModal' is missing the following properties from type 'Component': load, onload, unload, onunload, and 7 more"

TaggedDocumentsModal is a subclass of Modal. I'm trying to figure out what "this" needs. I tried adding 

load(): void {}
unload(): void {}
open(): void {}
close(): void {}

But it didn't work. This seems to be because renderMarkdown wants a Component for its last arg, but TaggedDocumentsModal -> Modal -> CloseableComponent isn't a Component.

joethei:
ooops, should be `null` instead

mgeduld:
Hi. I have addressed all the issues and also refactored the code to be better organized.

One thing: I still need that ts-ignore. Using null as the final arg still generates a typescript error.

joethei:
Ahh, sorry for the confussion, I should have checked beforehand, but its `this.plugin`.
Otherwise it looks good from my side

mgeduld:
Okay. Thanks. I made that (this.plugin) change.

liamcain:
- [isTagNode(target: HTMLElement): boolean](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/372230b717c65bd9cdcd55c40f2425758fbe246c/src/TaggedDocumentsViewer.ts#L75) this function doesn't need to live on the plugin class
- [this.plugin.onSettingChange](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/372230b717c65bd9cdcd55c40f2425758fbe246c/src/SettingsTab.ts#L61) you could probably simplify the code here a bit by calling saveSettings from `onSettingChange` and maybe even assigning the value from within the function as well.
- [this.registerDomEvent(document](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/372230b717c65bd9cdcd55c40f2425758fbe246c/src/TaggedDocumentsViewer.ts#L21) As of `0.15.0`, this should be `activeDocument`. This will point to the currently focused document. For this to also work within pop-out windows, you'll also need to listen for `window-open` and attach the event listener to those documents as well.  You might need to use [`requireAPIVersion`](https://github.com/obsidianmd/obsidian-api/blob/ff121cd409350d3afe39b08420f468d8aaee925e/obsidian.d.ts#L2783) to ensure that the version of Obsidian supports it. An example of a plugin trying to do something similar can be found here: <https://github.com/Quorafind/Obsidian-Table-Generator/blob/26476699f14e90a1d4ef4e01b4e577d46ea45dbd/src/tableGeneratorIndex.ts#L53>
- [openModalOnClick: boolean = true;](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/372230b717c65bd9cdcd55c40f2425758fbe246c/src/TaggedDocumentsViewer.ts#L11-L13) why are you trying to keep these variables in sync with `this.settings.openModalOnClick`? You can just access these values from `settings`, putting separate instance variables on the plugin object just makes this susceptible to bugs.
- [querySelector](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/372230b717c65bd9cdcd55c40f2425758fbe246c/src/PaginationControls.ts#L19) querying the DOM is expensive so it's better to avoid it when possible. You can just assign these elements to instance vars (or maybe an array of page buttons) and access them directly from instead of querying for them.
- [this.changePage((e.currentTarget as HTMLElement).getText());](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/372230b717c65bd9cdcd55c40f2425758fbe246c/src/PaginationControls.ts#L42) same feedback as the previous bullet, but this is more complicated than it needs to be. Just make your event listener into a closure and make sure that it calls `goToPage` with the correct page number.
- [previoiusButton](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/372230b717c65bd9cdcd55c40f2425758fbe246c/src/PaginationControls.ts#L75) Typo
- [Function](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/372230b717c65bd9cdcd55c40f2425758fbe246c/src/utils/documents.ts#L9) Small Typescript thing, but you really shouldn't use `Function` as a type. Instead this should probably be `(file: TFile) => string`. 
- [buildSiblingGetter](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/372230b717c65bd9cdcd55c40f2425758fbe246c/src/utils/dom.ts#L1) Looks like the code in this file is unused
- [destFile](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/372230b717c65bd9cdcd55c40f2425758fbe246c/src/utils/links.ts#L25) Looks like you're not handling `null` here at all.

mgeduld:
Hi. Can you help me understand this? "You'll also need to listen for window-open and attach the event listener to those documents as well." Which documents are you referring to? Do you mean window.addEventListener('open' ... or is there some sort of event called "window-open"?

mgeduld:
Thanks. I have made all the above changes.

liamcain:
- [return this.pageButtons.find(({ button }) =>](https://github.com/mgeduld/obsidian-tagged-documents-viewer/blob/93c4dd62610ad8fd5523eb4ce7a286145da7043f/src/PaginationControls.ts#L19-L22) can't use just use the list index instead of `find()` here?

mgeduld:
The reason I used find was because that array doesn't just contain the page-number buttons. It also contains the next and previous buttons. 

```
this.pageButtons.push({
			button: previousButton,
			listener: previousButtonListener,
		});
this.pageButtons.push({
			button: nextButton,
			listener: nextButtonListener,
		});
this.renderPageNumberButtons();
```

I guess I could store those in a different place or use an offset to find the index of the buttons. Do you want me to do that? Is the find going to slow things down to an unacceptable degree?

Probably storing the next/previous buttons in a separate structure would be better than an offset, because the push order might get changed in the future, and that could lead to bugs.

mgeduld:
Okay. I went ahead and simplified it: 

```
private findButton(page: number): HTMLElement | undefined {
		return this.pageButtons[page]?.button;
}
```

mgeduld:
Oh. I didn't notice it had been merged. Okay. Next update will have this simplification. Thanks.

PR #1061, opened 2022-07-17
Added Simple Note Review plugin.

dartungar:

## Repo URL

Link to my plugin: https://github.com/dartungar/obsidian-simple-note-review

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[this.app.workspace.on("file-menu", (menu, file) => {](https://github.com/dartungar/obsidian-simple-note-review/blob/1c0ba58861f564600e24a439e67280627a6efcb4/main.ts#L53)
before using it in your `reviewNote´ function, check that `file instanceof TFIle`, as it also can be a folder.

[note as TFile](https://github.com/dartungar/obsidian-simple-note-review/blob/1c0ba58861f564600e24a439e67280627a6efcb4/src/queue/queueService.ts#L26)
do check before casting, see above.


Your plugins requires Dataview to be installed, but you should still check if its available & show a notice if its not.

dartungar:
@joethei thanks for the review! I've made some amends:
- I check for Dataview in the beginning of onLoad() in main.ts (https://github.com/dartungar/obsidian-simple-note-review/blob/master/main.ts) and show notice (and stop loading plugin) if it is not found 
- I check the type of `note` parameter in `reviewNote` method (https://github.com/dartungar/obsidian-simple-note-review/blob/master/src/queue/queueService.ts#L25)

dartungar:
@joethei I've made further improvements - now Simple Note Review will load, but show notices every time some Dataview-related function is called. See https://github.com/dartungar/obsidian-simple-note-review/blob/master/src/dataview/dataviewFacade.ts, https://github.com/dartungar/obsidian-simple-note-review/blob/18605dc673270c11c8be42407a7719279f528df5/src/queue/queueService.ts#L34, and https://github.com/dartungar/obsidian-simple-note-review/blob/18605dc673270c11c8be42407a7719279f528df5/src/queue/selectQueueModal.ts#L36. 

I've tried it and it's quite smooth experience.

liamcain:
- [if (dv == null) throw new Error();](https://github.com/dartungar/obsidian-simple-note-review/blob/18605dc673270c11c8be42407a7719279f528df5/main.ts#L96) doesn't look like you need a try/catch here. Should be able to just return `!!getAPI()`
- [simple-note-review-open-modal](https://github.com/dartungar/obsidian-simple-note-review/blob/18605dc673270c11c8be42407a7719279f528df5/main.ts#L29) Not a big deal, but you don't need to prefix your command IDs with the plugin name. Under the hood, the commands get prepended with `<your-plugin-id:>` so it will be guaranteed to be unique anyway.
- [dataviewIsInstalled](https://github.com/dartungar/obsidian-simple-note-review/blob/18605dc673270c11c8be42407a7719279f528df5/main.ts#L16) You don't want to do this `onload` since Obsidian doesn't guarantee a plugin load order. Your plugin might load before `dataview`, in which case this will fire as a false positive. It's better to wrap this in [`onLayoutReady`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3626).
- [note as TFile](https://github.com/dartungar/obsidian-simple-note-review/blob/18605dc673270c11c8be42407a7719279f528df5/src/queue/queueService.ts#L32) I don't think you need `as Note` since you have the type check above.
- [const firstNoteIndex = 0;](https://github.com/dartungar/obsidian-simple-note-review/blob/18605dc673270c11c8be42407a7719279f528df5/src/queue/queueService.ts#L114) this variable looks unnecessary. Just pass in `0` or `1`
- [getNextFilePath](https://github.com/dartungar/obsidian-simple-note-review/blob/18605dc673270c11c8be42407a7719279f528df5/src/queue/queueService.ts#L110) Is this supposed to be removing items from the queue? Or is that all handled by re-running the dataview query?

dartungar:
@liamcain  Thanks for the thorough review! I've corrected these.

> * [getNextFilePath](https://github.com/dartungar/obsidian-simple-note-review/blob/18605dc673270c11c8be42407a7719279f528df5/src/queue/queueService.ts#L110) Is this supposed to be removing items from the queue? Or is that all handled by re-running the dataview query?

The idea is to not store notes in queue, only "search terms" aka queue parameters. This reduces complexity dramatically (no need to track anything, and the results are always accurate), and the speed of Dataview ensures no serious overhead (I tested it on vault with 10000 notes - opening next note takes ~1 second).

dartungar:
@liamcain hello there, 
I've corrected some minor bugs that became apparent during the last week in which I used the plugin myself. I've also tested it on mobile. It would be great if you've found time to get back to this review :)

liamcain:
Changes look good! One of the screenshots in your README is rotated though, you might want to fix that.

PR #1063, opened 2022-07-18
add Hide Sidebars on Window Resize plugin

NomarCub:

## Repo URL

Link to my plugin: https://github.com/NomarCub/obsidian-hide-sidebars-on-window-resize

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1949,10 +1949,10 @@
     },
     {
         "id": "obsidian-hide-sidebars-when-narrow",
-        "name": "Hide Sidebars When Narrow",
-        "author": "Michael Hanson",
-        "description": "Automatically hides the sidebars when your window is narrow.",
-        "repo": "mybuddymichael/obsidian-hide-sidebars-when-narrow"
+        "name": "Hide Sidebars on Window Resize",
+        "author": "NomarCub, Michael Hanson",
+        "description": "Automatically hides the sidebars when your window is narrow on window resize",
+        "repo": "NomarCub/obsidian-hide-sidebars-on-window-resize"
     },
     {
         "id": "obsidian-linter",


joethei:
Since this is a fork of a existing unmaintained plugin,
could you try and contact the author,
if they are maybe open to transfering the existing plugin entry over to your repo?

NomarCub:
I tried before [here](https://github.com/mybuddymichael/obsidian-hide-sidebars-when-narrow/pull/5), but he wasn't responsive since April.

I left a comment again, but if he doesn't respond in like a week I see no reason to try more.

NomarCub:
The week has passed. Should this be a new plugin then, or a continuation of the old entry?

joethei:
You are free to take over the old plugin entry, as long as you keep the credits to the original author in the author field.
You can just change the entry in this PR.

NomarCub:
Should the author stay the same, or could it be like `"author": "NomarCub, Michael Hanson"`?
Does the id need to be the same to keep continuity?
Should the new repo start with the next semver patch (1.0.3) for continuity?

joethei:
Use something like the second version for the author field.
Keep the ID, so users of the old plugin can update.
And do use a higher version number for the release,

NomarCub:
I think it's done.

liamcain:
- [this.registerDomEvent(window, 'resize', (_) => {](https://github.com/NomarCub/obsidian-hide-sidebars-on-window-resize/blob/076f52b963ed59fa0f1c15083a47a48a156c0bda/src/main.ts#L18-L20) there's a `"resize"` event in the API if you want to use that instead. It is already debounced so it could simplify your code a little bit. <https://github.com/obsidianmd/obsidian-api/blob/ff121cd409350d3afe39b08420f468d8aaee925e/obsidian.d.ts#L3867>

PR #1068, opened 2022-07-23
add open-related-url plugin

dpickett:

## Repo URL

Link to my plugin: https://github.com/dpickett/open-related-url

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4593,7 +4593,14 @@
         "name": "Hard Breaks",
         "author": "Börge Kiss",
         "description": "Turn soft line breaks in Markdown into hard line breaks",
-        "repo": "bkis/obsidian-hard-breaks",
+        "repo": "bkis/obsidian-hard-breaks"
+    },
+    {
+        "id": "open-related-url",
+        "name": "Open Related Url",
+        "author": "Dan Pickett",
+        "description": "Opens URLs found in a note's YAML frontmatter",
+        "repo": "dpickett/open-related-url",
         "branch": "master"
     }
 ]


joethei:
Just two tiny things, that you don't have to address:

1. Maybe add a link to the Obsidian documentation on frontmatter in your readmeas not everybody that might find your plugin knows what that is.(<https://help.obsidian.md/Advanced+topics/YAML+front+matter>) 

2. I see that you have "node-notifier" in your dependencies,
I don't see it being used anywhere, if you where it would mean that your plugin would not work on mobile.

dpickett:
great suggestions. I've made both changes and bumped the patch level. thanks!

liamcain:
- [setPlaceholder("separate by commas"](https://github.com/dpickett/open-related-url/blob/a284ec3493cdc5431fb5b82a822b786ed648dec3/src/open-related-url/SettingTab.ts#L40) Small UX nitpick, I suggest putting `(separated by commas)` at the end of your description. Then have the placeholder be something like `e.g. task,file`. I think that would be clearer.
- [Frontmatter Suffix](https://github.com/dpickett/open-related-url/blob/a284ec3493cdc5431fb5b82a822b786ed648dec3/src/open-related-url/SettingTab.ts#L19) Maybe file this under future improvements, but it might be nice to have a UI with rows where you can specify the key and the title separately. I'm guessing you introduced "frontmatter suffix" so that you can still call something "file" but it will look for `fileAddress`. But from a UX perspective, it feels like unnecessary misdirection.

liamcain:
@dpickett, do you mind rebasing this pull request from your end? looks like github is struggling to let me do it via their web ui

liamcain (approved):

dpickett:
I've adjusted the placeholder per your recommendations. I agree on the setting. I will file an issue for that as a future enhancement. Wow, I am very impressed with how thorough you all are with your plugin review. Thanks!

PR #1074, opened 2022-07-29
Add PodNotes plugin

chhoumann:

## Repo URL

Link to my plugin: [PodNotes repository](https://github.com/chhoumann/podnotes)

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[sleep](https://github.com/chhoumann/PodNotes/blob/d46703f537a64b7cce0842709da497686bce16d7/src/utility/sleep.ts#L1)
The Obsidian API already hase this exact function

You might want to make it more clear that QuickAdd and that macro/action is required for the "Podcast note" & "timestamp" features you showcase in the demo gif.
At first glance it looks like this plugin provides these feature directly.

liamcain:
- [this.addCommand](https://github.com/chhoumann/PodNotes/blob/d46703f537a64b7cce0842709da497686bce16d7/src/main.ts#L108) Is this command just for testing or did you mean to leave it in?
- [export const plugin = writable<PodNotes>();](https://github.com/chhoumann/PodNotes/blob/d46703f537a64b7cce0842709da497686bce16d7/src/store/index.ts#L9-L11) While I like the simplicity of declaring svelte stores in their own file like this and using them as singletons, I've found it a bit limiting if you ever want to support multiple views in the future. Or if you ever need to quickly reset your view state. You might want to wrap all of these stores into some type of state manager that belongs to the View instance.

chhoumann:
Thank you, @joethei and @liamcain for your reviews!
I really appreciate it.

I've removed the sleep utility function, as I don't use it.

The plugin will support native timestamps and note creation, which is currently WIP, but very close to release.

The command is there intentionally, although I'm considering removing it. It's just a neat way to do a refresh.

And thank you, @liamcain, for the tips on state management!

PR #1075, opened 2022-07-29
New plugin - Document Editing Shortcuts

grizzly-sh:

## Repo URL

Link to my plugin: https://github.com/happy-grizzly/obsidian-document-editing-shortcuts

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4543,5 +4543,12 @@
         "author": "Johnny✨ and phibr0",
         "description": "Customize your workspace by adding commands /everywhere/",
         "repo": "phibr0/obsidian-commander"
+    },
+    {
+        "id": "document-editing-shortcuts",
+        "name": "Document Editing Shortcuts",
+        "author": "Happy Grizzly",
+        "description": "remap document editing shortcuts, reduce the use of a mouse",
+        "repo": "happy-grizzly/obsidian-document-editing-shortcuts"
     }
 ]


happy-grizzly:
K I fixed that

joethei:
CodeMirror(the underlying editor engine) already provides some of these shortcuts: <https://codemirror.net/docs/ref/#commands>, they cannot be changed, and as such are not visible from the hotkey menu/command pallete.

[// Remember to rename these classes and interfaces!](https://github.com/happy-grizzly/obsidian-document-editing-shortcuts/blob/c81227ab38263556e8e22d0bc145b06f03b6fbc6/main.ts#L3)
Please actually do this, to avoid conflicts.

[repeatUntilKeyUp(code, this)](https://github.com/happy-grizzly/obsidian-document-editing-shortcuts/blob/c81227ab38263556e8e22d0bc145b06f03b6fbc6/main.ts#L28)
Instead of this you can set the [`repeatable`](https://github.com/obsidianmd/obsidian-api/blob/ff121cd409350d3afe39b08420f468d8aaee925e/obsidian.d.ts#L514) property of the command.

[// If the plugin hooks up any global DOM events (on parts of the app that doesn't belong to this plugin)](https://github.com/happy-grizzly/obsidian-document-editing-shortcuts/blob/c81227ab38263556e8e22d0bc145b06f03b6fbc6/main.ts#L368-L375)
You are not using any of this, you can just remove it.

[function isNormalChar(char: string): number](https://github.com/happy-grizzly/obsidian-document-editing-shortcuts/blob/c81227ab38263556e8e22d0bc145b06f03b6fbc6/main.ts#L393)
Any particular reason you return a number here and not a boolean?

ericaxu:
@grizzly-sh Hi! Do you still intend to submit this plugin? Thanks!

grizzly-sh:
Hi, I already tried to submit it.
I'll respond to joethei:

1. CodeMirror API is out of my control to edit, extension building wise and technical knowledge wise. Therefore the only way I saw to solve this problem is with some "hacky" javascript, the other way one can do it is to configure Vim shortcuts and use the insert mode only...not great, but works.
2. I'll rename them...sure.
3. The repeatable method does not give control over milliseconds for repeatability.
4. Sure I'll remove it you're right.
5. If you take a look where and how this method is implemented, you'll see I configured "modes" for character groups, I preferred to use numbers instead of booleans because I wasn't sure how many types of grouped characters the normal computer kernel differentiates for navigation. As an example press options/control right and left arrow to this word: dkfjgdjfgkh,,,?-?(
Notice how it stops after the normal letters and before the alternative ones. that's what I mean by character types. I replicated this with Javascript.

I'll make the rest of the small changes in the next few days. I need to test that it works on all the OSs before pushing it.

grizzly-sh:
Sorry I just can't find the time to do this.

Hard to find time to solve problems I don't have.
Might keep maintaining the project at a later date,having this itch is welcome to fork and take over.

Closing

PR #1076, opened 2022-07-29
Plugin name changed - "Text Expander JS" to "Inline Scripts"

jon-heard:
# I am submitting a rename of an existing Community Plugin

## Repo URL

Link to my plugin:
https://github.com/jon-heard/obsidian-inline-scripts

Link to the old plugin:
https://github.com/jon-heard/obsidian-text-expander-js

### DIFF
--- a/community-plugin-stats.json
+++ b/community-plugin-stats.json
@@ -10400,7 +10400,7 @@
     "1.0.4": 478,
     "1.0.5": 422
   },
-  "obsidian-inline-scripts": {
+  "obsidian-text-expander-js": {
     "0.18.0": 19,
     "0.16.2": 1,
     "0.15.1": 1,
--- a/community-plugin-stats.json
+++ b/community-plugin-stats.json
@@ -10400,7 +10400,7 @@
     "1.0.4": 478,
     "1.0.5": 422
   },
-  "obsidian-text-expander-js": {
+  "obsidian-inline-scripts": {
     "0.18.0": 19,
     "0.16.2": 1,
     "0.15.1": 1,
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4501,11 +4501,11 @@
         "branch": "main"
     },
     {
-        "id": "obsidian-text-expander-js",
-        "name": "Text Expander JS",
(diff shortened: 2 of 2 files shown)


joethei:
Please don't edit the `community-plugin-stats.json` file.

We would suggest against changing the plugin id, as that will break all existing installations.
Existing users will not be able to update the plugin or view plugin information from the Community plugins tab.

jon-heard:
Ok, so you are saying to keep everything connected to the old name, except for the displayed name itself?

The reason for this pull request is that this plugin's current name (__Text Expander JS__) is very similar to another plugin's name (__Text Expander__).  This was by intent, since this plugin was conceived of in relation to that other plugin (like C to C++).  It has since turned out to be a bad idea as it has confused at least a few users, and a number of other users have endorsed changing the plugin's name (though one rather influential user did think it unnecessary).

The current name, __Text Epander JS__, and proposed new name, __Inline Scripts__, are quite different from each other (for reasons stated above).  I'm concerned that if I change the displayed name, but keep the old plugin id and repo name then the situation might become _more_ confusing rather than less.

I really don't _want_ to disrupt the plugin's user-base.  I'm also concerned about the download count (thus changing `community-plugin-stats.json`).  I had planned to mitigate the effect of the rename by announcing it on forums, subreddits, discord channels and Facebook groups.  However, if you are telling me that changing the plugin id is not a viable option then I may need to abandon this plan to rename altogether (with some relief, I admit).  If you are telling me that changing the id is doable, but not a great idea, then I'm less certain on how to proceed.

Do you have any thoughts on the best course of action here?

joethei:
The plugin stats file will update automatically, your changes will be overwritten the next time the script runs(every 24h), so updating it is useless. 


You can move the repo, no problem there.
The only issue is with changing the ID,
as that causes problems for existing users.

The ID is internal mostly and is not be visible to most users, only to those digging through the files.
So it should not matter if you keep the old one.

jon-heard:
@joethei: thanks for your help with this.  I've updated this pull request to match what you've suggested.  Let me know if you need anything more from before accepting this PR

PR #1080, opened 2022-07-31
Add "No Dupe Leaves" plugin

scambier:

## Repo URL

Link to my plugin: https://github.com/scambier/obsidian-no-dupe-leaves

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [iterateAllLeaves](https://github.com/scambier/obsidian-no-dupe-leaves/blob/fe981b3592a535019b2f3994ec7664fc8d6c7f45/src/main.ts#L35) UX suggestion: I think you probably only want to iterate leaves in the current container. Otherwise you'll click a link and it might switch you to a pop-out window which could be unexpected for the user.
- [viewState.type === 'markdown' &&](https://github.com/scambier/obsidian-no-dupe-leaves/blob/fe981b3592a535019b2f3994ec7664fc8d6c7f45/src/main.ts#L38-L39) It's not public API but if you want to support more than just markdown files, you can check the default type of the linktext `this.app.viewRegistry.getTypeByExtension(file.extension)` then make sure that the leaf you're looking at as a view with that viewType.

Anyway, cool plugin and I think it can go in as is. Just pointing out some suggestions.

scambier:
Thanks for the suggestions, I'll implement them asap 👍

liamcain:
btw, this is what I meant by container: [`getContainer()`](https://github.com/obsidianmd/obsidian-api/blob/ff121cd409350d3afe39b08420f468d8aaee925e/obsidian.d.ts#L3977)

PR #1093, opened 2022-08-04
added Google Calendar Plugin

YukiGasai:

## Repo URL

[]()
Link to my plugin: https://github.com/YukiGasai/obsidian-google-calendar

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [X]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


joethei:
I would suggest a different name and ID for the plugin.
Having Obsidian at the start will throw of the sorting in the settings and the plugins folder.

[from 'http';](https://github.com/YukiGasai/obsidian-google-calendar/blob/034faf1b8d6a2229ec19bf656fda03c243d07d63/src/googleApi/GoogleAuth.ts#L10)
This a NodeJS API, the plugin will not work on mobile, please mark it as such in the manifest


[getET() == 0 ||](https://github.com/YukiGasai/obsidian-google-calendar/blob/034faf1b8d6a2229ec19bf656fda03c243d07d63/src/googleApi/GoogleAuth.ts#L45-L48)
This is not really readable, I would prefer longer names.

[await fetch](https://github.com/YukiGasai/obsidian-google-calendar/blob/034faf1b8d6a2229ec19bf656fda03c243d07d63/src/googleApi/GoogleAuth.ts#L58)
Use the build in [`requestUrl`](https://github.com/obsidianmd/obsidian-api/blob/3770989c2364ead5429eb7e4475e095588bbe914/obsidian.d.ts#L2745-L2749)
method instead, it works similar to fetch, but will make sure it works everywhere in regards to CORS.

[adapter.read](https://github.com/YukiGasai/obsidian-google-calendar/blob/034faf1b8d6a2229ec19bf656fda03c243d07d63/src/helper/AutoEventNoteCreator.ts#L179)
(Prefer the Vault API over the Adapter API)[https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#prefer-the-vault-api-over-the-adapter-api]


[await plugin.app.workspace.activeLeaf.detach();](https://github.com/YukiGasai/obsidian-google-calendar/blob/034faf1b8d6a2229ec19bf656fda03c243d07d63/src/helper/AutoEventNoteCreator.ts#L192)
[Avoid accessing `Workspace.activeLeaf` directly](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-accessing-workspaceactiveleaf-directly)

[GoogleCalander](https://github.com/YukiGasai/obsidian-google-calendar/blob/034faf1b8d6a2229ec19bf656fda03c243d07d63/src/helper/types.ts#L21)
Typo

YukiGasai:
Thank you for the review. 
I updated the code to fix the problems.

I would like to keep the name because I already created the [obsidian-google-tasks plugin](https://github.com/YukiGasai/obsidian-google-tasks) with a similar naming scheme and had to write a documentation for the plugin for university using the name all over the place.

The plugin is working on Mobile, the http libraryis only imported as type and only used if obsidian is running in [desktop mode](https://github.com/YukiGasai/obsidian-google-calendar/blob/034faf1b8d6a2229ec19bf656fda03c243d07d63/src/googleApi/GoogleAuth.ts#L89).
For mobile users, there is a workaround without using the http library.

grafik

YukiGasai:
Ready for review

YukiGasai:
Sorry @joethei, but could I get a second review to get the plugin published?

liamcain:
[.K36rb .QQYuzf](https://github.com/YukiGasai/obsidian-google-calendar/blob/3c45b0fb52ead01a014d989d84415bc51fe919a1/src/svelte/WebFrameComp.svelte#L20) this looks very fragile. These keys are probably autogenerated hashes based on the content, so if google calendar pushes an update, these keys will probably change and this will break.

[const File](https://github.com/YukiGasai/obsidian-google-calendar/blob/3c45b0fb52ead01a014d989d84415bc51fe919a1/src/helper/AutoEventNoteCreator.ts#L105) Nit: this should probably be lowercase

[const newLeafe](https://github.com/YukiGasai/obsidian-google-calendar/blob/3c45b0fb52ead01a014d989d84415bc51fe919a1/src/helper/AutoEventNoteCreator.ts#L124) typo: `leaf`?

[plugin.templaterPlugin && plugin.coreTemplatePlugin](https://github.com/YukiGasai/obsidian-google-calendar/blob/3c45b0fb52ead01a014d989d84415bc51fe919a1/src/helper/AutoEventNoteCreator.ts#L130) the logic here is more complicated than it needs to be. No need for this initial `if` condition. Just check if templater exists, if it doesn't, check if Templates exists.

[read](https://github.com/YukiGasai/obsidian-google-calendar/blob/3c45b0fb52ead01a014d989d84415bc51fe919a1/src/helper/AutoEventNoteCreator.ts#L208) I think this can safely be `cachedRead` instead of `read`.

[read](https://github.com/YukiGasai/obsidian-google-calendar/blob/3c45b0fb52ead01a014d989d84415bc51fe919a1/src/helper/AutoEventNoteCreator.ts#L241) Same here, this can just be `cachedRead`

[DateToPercent](https://github.com/YukiGasai/obsidian-google-calendar/blob/3c45b0fb52ead01a014d989d84415bc51fe919a1/src/helper/CanvasDrawHelper.ts#L8) this should probably be lowercase (`dateToPercent`)

[check4Word](https://github.com/YukiGasai/obsidian-google-calendar/blob/3c45b0fb52ead01a014d989d84415bc51fe919a1/src/helper/CheckEditorForAtDates.ts#L22) `checkForWord`?

[window.moment(blockDateString, "YYYY-MM-DD", true).isValid() ||](https://github.com/YukiGasai/obsidian-google-calendar/blob/3c45b0fb52ead01a014d989d84415bc51fe919a1/src/helper/CheckEditorForCodeBlocks.ts#L91-L99) you can pass in an array as the second argument here to look for multiple formats.

---

Anyway, I don't think anything here is a blocker. I do agree with joethei though, it might make sense for you to rename your plugins to exclude the name 'obsidian' in them. All plugins in the plugin gallery are connected to Obsidian, so including it in the plugin name doesn't add anything identifiable.

liamcain:
Let me know if you would like to change the plugin ID before this gets merged, it will be tough to change the ID later (you'll lose all your download stats as well).

YukiGasai:
Thank you for the fast review. 
If the name is such a problem we can remove the obsidian part from it. Is it possible to remove it from the obsidian google task plugin as well? I don't care about the loss of the download count.

liamcain:
> If the name is such a problem we can remove the obsidian part from it.

Okay, sounds good, thanks!

> Is it possible to remove it from the obsidian google task plugin as well?

Sure, feel free to do that in a separate PR. You can keep the `ID` the same and just change the name, then the download count won't be effected and it won't cause any issues with users trying to update the plugin.

YukiGasai:
I updated the name. I hope I didn't mess things up with the merge.

PR #1094, opened 2022-08-05
Add Symbols Prettifier plugin

FlorianWoelki:

## Repo URL

Link to my plugin: https://github.com/FlorianWoelki/obsidian-symbols-prettifier

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - []  Android _(if applicable)_
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[`callback`](https://github.com/FlorianWoelki/obsidian-symbols-prettifier/blob/374da8cee604ea8cdb71dc0a5f0d5ff3cbd3d6a7/src/main.ts#L38)
This would be better as an [`editorCallback´](https://marcus.se.net/obsidian-plugin-docs/user-interface/commands#editor-commands)
This will make sure that the command is only visible and can only be invoked if there is an active editor.

FlorianWoelki:
> [`callback`](https://github.com/FlorianWoelki/obsidian-symbols-prettifier/blob/374da8cee604ea8cdb71dc0a5f0d5ff3cbd3d6a7/src/main.ts#L38) This would be better as an [`editorCallback´](https://marcus.se.net/obsidian-plugin-docs/user-interface/commands#editor-commands) This will make sure that the command is only visible and can only be invoked if there is an active editor.

Thank you for the feedback.
I've updated the code with `editorCallback`.

liamcain:
Hi, sorry for the long wait! Everything looks good to me 🥳

PR #1096, opened 2022-08-05
Add Raindrop Highlights Plugin

kaiiiz:

## Repo URL

Link to my plugin: https://github.com/kaiiiz/obsidian-raindrop-highlights-plugin

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
Hi, Just a few small nits but I think this can go in. Is there any reason you opted to create a separate plugin instead of collaborating with the other Raindrop plugin author?

---

- [message += `${collection.title}: ${collection.lastSyncDate?.toLocaleString()}\n`](https://github.com/kaiiiz/obsidian-raindrop-highlights-plugin/blob/0d7062c341a4778eaecaca20002894d6ba179102/src/main.ts#L46-L47) Small nit: You might want to add each entry into a list, then you can do `collectionEntries.join('\n')`. That way you don't end up with a trailing newline. It's also more efficient that doing intermediate string concatenation
- [Link](https://github.com/kaiiiz/obsidian-raindrop-highlights-plugin/blob/0d7062c341a4778eaecaca20002894d6ba179102/src/main.ts#L55) Super small nitpick but obsidian commands are "Sentence-case" so it would make more sense not to capitalize each word in your commands.
- [this.localStorage = window.localStorage;](https://github.com/kaiiiz/obsidian-raindrop-highlights-plugin/blob/0d7062c341a4778eaecaca20002894d6ba179102/src/tokenManager.ts#L5-L6) Why are you storing a reference to localstorage?
- [const filecontentLine = content.split("\n");](https://github.com/kaiiiz/obsidian-raindrop-highlights-plugin/blob/0d7062c341a4778eaecaca20002894d6ba179102/src/sync.ts#L155-L156) on large files, this can be pretty inefficient. You might want to consider just searching for `\n---\n` and splitting at that index.

kaiiiz:
> Is there any reason you opted to create a separate plugin instead of collaborating with the other Raindrop plugin author?

To be honest, this is because I think the other Raindrop plugin is developed in a very slow progress and without active maintenance. There are lots of issues without response and the plugin was submitted to [obsidian-releasees at Mar 22 but still not be accepted](https://github.com/obsidianmd/obsidian-releases/pulls?q=is%3Apr+sort%3Aupdated-desc+Raindrop+is%3Aclosed).

In the meantime, I'm tired of repeating the copy-paste highlight procedure from Raindrop to Obsidian, so I decided to implement this feature according to my needs. The original Raindrop plugin is more focused on the dataview-like features, but Raindrop Highlights will mainly focus on the note creation features.

> [message += ${collection.title}: ${collection.lastSyncDate?.toLocaleString()}\n](https://github.com/kaiiiz/obsidian-raindrop-highlights-plugin/blob/0d7062c341a4778eaecaca20002894d6ba179102/src/main.ts#L46-L47) Small nit: You might want to add each entry into a list, then you can do collectionEntries.join('\n'). That way you don't end up with a trailing newline. It's also more efficient that doing intermediate string concatenation

I've refactored it to a more [functional programming way](https://github.com/kaiiiz/obsidian-raindrop-highlights-plugin/blob/0c384c77d333fbaea45454060358fbd9101ece17/src/main.ts#L42-L47).

> [Link](https://github.com/kaiiiz/obsidian-raindrop-highlights-plugin/blob/0d7062c341a4778eaecaca20002894d6ba179102/src/main.ts#L55) Super small nitpick but obsidian commands are "Sentence-case" so it would make more sense not to capitalize each word in your commands.

Thanks for the suggestion. [I've fixed it](https://github.com/kaiiiz/obsidian-raindrop-highlights-plugin/blob/0c384c77d333fbaea45454060358fbd9101ece17/src/main.ts#L54).

> [this.localStorage = window.localStorage;](https://github.com/kaiiiz/obsidian-raindrop-highlights-plugin/blob/0d7062c341a4778eaecaca20002894d6ba179102/src/tokenManager.ts#L5-L6) Why are you storing a reference to localstorage?

Yes, this is unnecessary. I've removed it.

> [const filecontentLine = content.split("\n");](https://github.com/kaiiiz/obsidian-raindrop-highlights-plugin/blob/0d7062c341a4778eaecaca20002894d6ba179102/src/sync.ts#L155-L156) on large files, this can be pretty inefficient. You might want to consider just searching for \n---\n and splitting at that index.

I've refactored this [using indexOf to find the split position of the frontmatter and the content](https://github.com/kaiiiz/obsidian-raindrop-highlights-plugin/blob/ec2c6f4cfe0e6d066191da7a39277f1965128ed2/src/sync.ts#L159-L176). I think this implementation is better than before, which doesn't require the linenumber-sized of array, splice and join process.

PR #1097, opened 2022-08-06
added Obsidian-PDF plugin

BluBloos:

## Repo URL

Link to my plugin: https://github.com/BluBloos/Obsidian-PDF

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4609,5 +4609,13 @@
         "author": "Christian B. B. Houmann",
         "description": "Write notes on podcasts with ease.",
         "repo": "chhoumann/podnotes"
+    },
+    {
+        "id": "obsidian-pdf",
+        "name": "Obsidian-PDF",
+        "author": "Noah Cabral",
+        "description": "Treats PDF files as first-class citizens, where each PDF is a pairing of a .pdf and .md file.",
+        "repo": "BluBloos/Obsidian-PDF",
+        "branch": "master"
     }
 ]


liamcain:
- [mdPretty += md.charAt(i);](https://github.com/BluBloos/Obsidian-PDF/blob/c235a749a32a751cfa3a4455021e530713fcf09d/src/main.ts#L49-L50) String concatenation is actually pretty expensive in javascript, so doing it for every character is a bit wasteful. Instead, you can just do `replaceAll('  ', ' ')`
- [getActiveViewOfType](https://github.com/BluBloos/Obsidian-PDF/blob/c235a749a32a751cfa3a4455021e530713fcf09d/src/main.ts#L79) you're not handling `getActiveViewOfType` can return `null`
- [name](https://github.com/BluBloos/Obsidian-PDF/blob/c235a749a32a751cfa3a4455021e530713fcf09d/src/main.ts#L69) Instead of casting to FileView, I'd recommend that that you still do an instanceof check in addition to checking the viewState's type. That will make the check a bit more future proof in case the assumption that 'pdf' !== FileView down the line.
- [newLeaf = false](https://github.com/BluBloos/Obsidian-PDF/blob/c235a749a32a751cfa3a4455021e530713fcf09d/src/main.ts#L73) I'm not sure I follow the logic here. If any markdown view is open anywhere, then you opt to replace the currently active file? Is that behavior intentional? Did you mean to look for a specific leaf?
- [workspace.off](https://github.com/BluBloos/Obsidian-PDF/blob/c235a749a32a751cfa3a4455021e530713fcf09d/src/main.ts#L138) This won't work. The callback you pass in needs to be the same instance as what's passed to `on`. You also don't need to call this since you registered the event so this can be removed.
- [pairOpen](https://github.com/BluBloos/Obsidian-PDF/blob/c235a749a32a751cfa3a4455021e530713fcf09d/src/main.ts#L30) I don't think this is the best way to keep track of leaf pairs. Honestly I think you could simplify things a lot by just keeping a map of leaf IDs to their corresponding PDF leaf ID. Then you would only need to do work on `file-open` instead of `active-leaf-change`.
- [this.pairOpen = true;](https://github.com/BluBloos/Obsidian-PDF/blob/c235a749a32a751cfa3a4455021e530713fcf09d/src/main.ts#L66-L67) You should set this at the end to avoid it being set before the new leaf is open.

BluBloos:
Thanks Liam for the feedback. Re: newLeaf = false - I think I've found sound bugs in testing, and this is likely the cause.

PR #1099, opened 2022-08-06
obsidian-keyboard-shortcuts plugin

aciq:

## Repo URL

Link to my plugin: https://github.com/aciq/obsidian-keyboard-shortcuts

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


pjeby:
May I suggest changing the plugin name to be more specific?  Keyboard Shortcuts sounds like it is a general purpose shortcut management tool, so perhaps something like, "Aciq's Block and Heading Commands" or something of that sort would be a good idea?  It's best to have something that people can search for.  (Also the plugin is providing commands, rather than shortcuts -- a shortcut or hotkey in Obsidian lingo is the actual key assignment.)

aciq:
@pjeby 
If i named it Aciq's Keyboard Shortcuts would that work?
I think it explicitly gives you commands that'd only be useful as keyboard shortcuts so i'd want to keep that.

pjeby:
Aren't all commands only useful as keyboard shortcuts?  It's just that some shortcuts are shorter than others.  :wink:

As for the name, I think the important question to ask is: who do you want to have using this?  Will they know from "keyboard shortcuts" what the plugin does?  Quite a lot of plugins do nothing but provide commands, so should they all be named "keyboard shortcuts", and if so, how would anyone ever find which plugin they want?

In general, a plugin's name should sound like the answer to a question or need, like, "How can I move between headings?", "Use Aciq's Block and Heading Commands".  While technically "Aciq's Keyboard Shortcuts" is an answer to a need as well, it sounds more like the answer to the question, "How can I set up keyboard shortcuts?", and ironically, the plugin does not actually include any keyboard shortcuts, only commands.

This is all just a suggestion, and you're not required to follow it.  However, the reason I'm commenting on this at all is because I saw this PR announced on the Discord updates channel and thought it was a plugin for managing keyboard shortcuts...  and so was disappointed to find out it wasn't relevant or useful to what I was looking for.

So, just as a data point, the current name is not just confusing but *actively misleading* as to the plugin's purpose and function.  It will also not be very memorable for someone who does look at it, and go, "oh I should remember if I ever need these type of block and heading commands", who will likely not remember it as being under the unrelated phrase "keyboard shortcuts".  :grin:

liamcain:
Hi, in general, this feels like a hodge podge of different features, and the functionality doesn't match the plugin name, as pjeby mentioned above. I would suggest splitting this into multiple plugins divided based on purpose. The navigating between headings seems especially useful, so maybe start with that?

--- 

Code-wise, things look good to me. Only noticed a couple small nits.

- [let [<Literal>] headingregex = """^(#+) """](https://github.com/aciq/obsidian-keyboard-shortcuts/blob/f08e42cead1ad43b0245c9928d1d239866ee76de/src/Commands.fs#L9) You could make this `#{1,6}` since Markdown only supports H1 - H6 headings
- [Admonition](https://github.com/aciq/obsidian-keyboard-shortcuts/blob/f08e42cead1ad43b0245c9928d1d239866ee76de/src/Commands.fs#L181) Have you considered adapting this to Obsidian's callout syntax instead? They are functionally the same but callouts have better interoperability with markdown and Obsidian's links. Obsidian also has a command to insert callouts already, so I wonder if this feature is necessary.

aciq:
@liamcain 

Ok so i updated the callout syntax and headings and i gave it a less vague name.

It's now named "Quick snippets and navigation"
I think that's more fitting and wont need a separate plugin in case i add something new.

the validation says the name hasn't changed yet - do i need to make a new PR or did i miss something?

liamcain:
Looks good! 

> the validation says the name hasn't changed yet - do i need to make a new PR or did i miss something?

That was just because the branch needed to be rebased. It passed validation after I did that.

PR #1101, opened 2022-08-07
Obsidian Raindrop Plugin Submission

mtopping:

## Repo URL

Link to my plugin: https://github.com/mtopping/obsidian-raindrop

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [<div />](https://github.com/mtopping/obsidian-raindrop/blob/aa9ba888f3ed89468e8799486ec167370d8391f6/src/components/WaitingIndicator.svelte#L5) I suggest adding a class to each of these. CSS matches right to left, so `> div` can be a lot more expensive than it needs to be.
- [new RaindropBlockQueryProvider({](https://github.com/mtopping/obsidian-raindrop/blob/aa9ba888f3ed89468e8799486ec167370d8391f6/main.ts#L124-L125) You'll need to add this to a [MarkdownRenderChild](https://github.com/obsidianmd/obsidian-api/blob/6b2138aa28fb6d6f58f78e6fabec02186fdc0f82/obsidian.d.ts#L1890) for the lifecycle to be properly managed. Otherwise, this is a memory leak. There is some info about it in the docs, but you might want to check other implementations for a usage example https://github.com/search?q=MarkdownRenderChild+obsidian&type=code

ericaxu:

PR #1103, opened 2022-08-08
Update community-plugins.json

shap-po:
Added shap-po/obsidian-party

## Repo URL

Link to my plugin:
https://github.com/shap-po/obsidian-party

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [this.registerDomEvent(document](https://github.com/shap-po/obsidian-party/blob/2217829909d61e4cb106f1882a44e6003686824c/main.ts#L10) This approach won't work if you want to support Obsidian's pop-out windows feature. You'll also need to add an eventListener to the new window when the [`window-open`](https://github.com/obsidianmd/obsidian-api/blob/6b2138aa28fb6d6f58f78e6fabec02186fdc0f82/obsidian.d.ts#L4092) event is called.

lishid:
Actually, window level clicks are bridged over, but you will need to use `.instanceOf(HTMLElement)` instead of `instanceof HTMLElement`.

liamcain:
@shap-po, you'll need to cast EventTarget to an `HTMLElement` or `Node` first.

The reason why `instanceof HTMLElement` won't work is that `HTMLElement` actually refers to `window.HTMLElement` and pop-out windows will have their own HTMLElement constructor. So `someWindow.HTMLElement !== someOtherWindow.HTMLElement`, meaning the instanceof checks will fail.

The `.instanceOf` function is a shim added internally by Obsidian to handle these multi-window edge cases.

shap-po:
@liamcain got it, thank you. Now I have another question: do I need to validate the type if I cast to it?
```ts
const target = evt.target as HTMLElement;
if (target.hasClass("confetti")) party.confetti(target);
if (target.hasClass("sparkles")) party.sparkles(target);
```
This should work, right?

liamcain:
It's possible that the [event target is not an HTMLElement](https://stackoverflow.com/questions/28900077/why-is-event-target-not-element-in-typescript). So it's safer for do some duck-typing.

```ts
if (target.hasOwnProperty('hasClass')) {
  const target = evt.target as HTMLElement;
  if (target.hasClass(...) {}
}
```

shap-po:
For some reason, `hasOwnProperty` keeps returning `false`. So I'll check if target is `HTMLElement` instead
```ts
const target = evt.target as HTMLElement;
if (target.instanceOf(HTMLElement)) {
    if (target.hasClass("confetti")) party.confetti(target);
    if (target.hasClass("sparkles")) party.sparkles(target);
}
```

liamcain:
Okay, works for me!

shap-po:
Slightly changed name & description of the plugin and pushed changes to repo

PR #1106, opened 2022-08-11
Add Scrybble plugin

Azeirah:

## Repo URL

Link to my plugin:

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


Azeirah:
Retry by commenting?

Azeirah:
.

joethei:
[MyPluginSettings](https://github.com/Azeirah/scrybble/blob/15ff2de8ac01dbf75c5d624fe5bad9692eac541c/main.ts#L6), [MyPlugin](https://github.com/Azeirah/scrybble/blob/15ff2de8ac01dbf75c5d624fe5bad9692eac541c/main.ts#L16), [SampleSettingTab](https://github.com/Azeirah/scrybble/blob/15ff2de8ac01dbf75c5d624fe5bad9692eac541c/main.ts#L101)
You probably want to rename these.

I don't know what one could do with the access token to your service,
saving it in the plugin settings object it will be saved in plain text in a text file inside the vault.
I would recommend saving it in [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) instead.

[await fetch](https://github.com/Azeirah/scrybble/blob/15ff2de8ac01dbf75c5d624fe5bad9692eac541c/main.ts#L56)
Obsidian provides the [requestUrl](https://github.com/obsidianmd/obsidian-api/blob/6b2138aa28fb6d6f58f78e6fabec02186fdc0f82/obsidian.d.ts#L2896) method, it will make sure that requests work on every plattform.

[new Setting(containerEl)](https://github.com/Azeirah/scrybble/blob/15ff2de8ac01dbf75c5d624fe5bad9692eac541c/main.ts#L118-L170)
I would hide the login settings if the user is logged in, having this here will probably confuse a few users.
Instead, show a logout button that removes the access key.

[console.log("error!", error);](https://github.com/Azeirah/scrybble/blob/15ff2de8ac01dbf75c5d624fe5bad9692eac541c/main.ts#L167)
I would also add a quick notice here to at least show to the user that the login was not successful.

[console.log('Failure: ', e);](https://github.com/Azeirah/scrybble/blob/15ff2de8ac01dbf75c5d624fe5bad9692eac541c/main.ts#L181)
Same here, let the user know that the synchronisation failed.

ericaxu:

Azeirah:
Hi, I updated the plugin! I was rushing to get it out initially.

liamcain:
- [const base_url = "https://scybble.ink";](https://github.com/Azeirah/scrybble/blob/a4c615f525ccaf0ba6f69a8df18bd531cc558299/src/sync.ts#L4-L5) Looks like you have a typo here, the url should be https://scrybble.ink, right?
- [if (token !== null) {](https://github.com/Azeirah/scrybble/blob/a4c615f525ccaf0ba6f69a8df18bd531cc558299/main.ts#L25-L26) this syncing shouldn't be performed directly in onload. onload is blocking, so any code executed here will slow down Obsidian's boot time. Instead you should call this from within [`onLayoutReady`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3626).
- [fetchSyncDelta](https://github.com/Azeirah/scrybble/blob/a4c615f525ccaf0ba6f69a8df18bd531cc558299/main.ts#L112) this code block looks identical to the code called in the onload function. Maybe it should be a function on the plugin? Like `this.sync()`

Azeirah:
> * [const base_url = "https://scybble.ink";](https://github.com/Azeirah/scrybble/blob/a4c615f525ccaf0ba6f69a8df18bd531cc558299/src/sync.ts#L4-L5) Looks like you have a typo here, the url should be https://scrybble.ink, right?
> 
>     * [if (token !== null) {](https://github.com/Azeirah/scrybble/blob/a4c615f525ccaf0ba6f69a8df18bd531cc558299/main.ts#L25-L26) this syncing shouldn't be performed directly in onload. onload is blocking, so any code executed here will slow down Obsidian's boot time. Instead you should call this from within [`onLayoutReady`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3626).
> 
>     * [fetchSyncDelta](https://github.com/Azeirah/scrybble/blob/a4c615f525ccaf0ba6f69a8df18bd531cc558299/main.ts#L112) this code block looks identical to the code called in the onload function. Maybe it should be a function on the plugin? Like `this.sync()`

Thanks again for the review! I was wondering why my Obsidian was so slow on startup, but only on my desktop...

I updated the plugin and fixed the merge conflict for the PR

liamcain:
Looks good! I just left one more small comment: https://github.com/Azeirah/scrybble/commit/54127eb422d1832bba3d9244e468491c1149ce66#r91911841

Azeirah:
> Looks good! I just left one more small comment: [Azeirah/scrybble@54127eb#r91911841](https://github.com/Azeirah/scrybble/commit/54127eb422d1832bba3d9244e468491c1149ce66#r91911841)

You're 100% right, I didn't test it :<

I should stop relying on my eyes for code. It only works if it runs

Azeirah:
@liamcain Updated the code, _did_ run it this time.

- Login works
- Syncing works

I found that initially the code was incorrect? I thought that Obsidian's `requestUrl` would be a drop-in replacement for fetch, but it's not. I had to change the `Content-Type` header from `x-www-form-urlencoded` to `application/json`

And more notably, the `json` property on `response` is _not_ a promise, it's a regular old object.

IE

Yes:

```js
function myFn() {
  const response = await requestUrl({...});
  return response.json;
```

No:

```js
async function myFn() {
  const response = await requestUrl({...});
  return await response.json();
```

liamcain:
> I was also somewhat surprised that requestUrl calls don't get recorded in the devtools network tab. I assume you're not using electron's fetch or XMLHttpRequest under the hood, but something external? (like node or native code?)

Yes, requestUrl is bridged to the main electron thread to bypass CORS issues.

PR #1108, opened 2022-08-12
obsidian-rpg-manager added to community-plugins.json

carlonicora:

## Repo URL

Link to my plugin: https://github.com/carlonicora/obsidian-rpg-manager

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
While it is allowed, having `obsdian` in the plugin ID is unnecessary.
Also isn't it usually written as RPG and not Rpg ?

You have quite a few instances of `//@ts-ignore` in your code, if you can try to eliminate as many as you can.

Since you use the Dataview API a lot I would add some sort of notification/error message when Dataview is not installed.

[declare module 'obsidian'](https://github.com/carlonicora/obsidian-rpg-manager/blob/285f81a8a6ec2701405ea6563a4530ce9aeac1bd/src/api.ts#L8)
Any reason why this is here and not with the other typing?

carlonicora:
Hi,
I will update the plugin updating the name, checking the ts-ignore and merging the obsidian in the other typings

carlonicora:
Updated the plugin with the following:
- ID Updated to `rpg-manager`
- name updated to `RPG Manager`
- updated minAppVersion to `0.15.9`
- many `//@ts-ignore` have been removed (with the exclusion of Factories

Question (it is my first submission): do I have to create a new pull request with the updated ID?

Carlo

joethei:
no, you don't need to resubmit

liamcain:
- [this](https://github.com/carlonicora/obsidian-rpg-manager/blob/0c25fb5c9d4c19e21ec10d822c68198e34634678/src/main.ts#L43) I think these event subscriptions would be simpler if you `bind(this)` first, then debounce that function. Something like:

```ts
		this.refreshViews = debounce(this.refreshViews.bind(this), 500, true);
		this.registerEvent(this.app.metadataCache.on('resolved', this.refreshViews);
```

- [private initialiseRoots() {](https://github.com/carlonicora/obsidian-rpg-manager/blob/0c25fb5c9d4c19e21ec10d822c68198e34634678/src/api.ts#L23) I think you can just use [`getBasePath`](https://github.com/obsidianmd/obsidian-api/blob/6b2138aa28fb6d6f58f78e6fabec02186fdc0f82/obsidian.d.ts#L1217). Or, if you are just after the full path of the attachmentFolderPath, then use `getFullPath`.
- [attachmentFolderPath](https://github.com/carlonicora/obsidian-rpg-manager/blob/0c25fb5c9d4c19e21ec10d822c68198e34634678/src/api.ts#L58) You should use `getAttachmentPathConfig()` instead since it takes into account the user's "attachment folder" preference.
- [nse = (abstractFile ? true : false);](https://github.com/carlonicora/obsidian-rpg-manager/blob/0c25fb5c9d4c19e21ec10d822c68198e34634678/src/api.ts#L67) This ternary doesn't do anything since abstractFile is always truthy

carlonicora:
Hi @liamcain 
Thanks for the feedback and the changes requested.

Version 0.3.16 updated.

**Changes applied**:
- debounce for subscriptions refactored in main.js
- useless ternary removed
- removed the attachmentFolderPath (was doing the same thing as calling `getAttachmentPathConfig()`

**HOWEVER...**
In terms of the `initialiseRoot`, I am using the function to provide the same functionalities on desktop AND mobile. By using `getBasePath` or `getFullPath`, the path I am getting is not able to display the images with the full path.
If I use a relative path for the images, the images simply don't work as they are prepended by `app://obsidian.md/`.
If I use the full path the images are still not displayed.
In the `initialiseRoot` I read the filename including the filesystem part, which allows Obsidian to read the image directly from the correct path.

The difference is something like this:
## getBasePath
DesktopUri: /Users/XXX/Library/Mobile Documents/iCloud~md~obsidian/Documents/VAULE/ATTACHMENTFOLDER/IMAGE
Image works in src in Desktop: **NO**
MobileUri: EMPTY
Image works in src in Mobile: **NO**

## getFullPath
DesktopUri: /Users/XXX/Library/Mobile Documents/iCloud~md~obsidian/Documents/VAULE/ATTACHMENTFOLDER/IMAGE
Image works in src in Desktop: **NO**
MobileUri: Documents/VAULT/ATTACHMENTFOLDER/IMAGE
Image works in src in Mobile: **NO**

## initialiseRoot
DesktopUri: app://local/Users/XXX/Library/Mobile%20Documents/iCloud~md~obsidian/Documents/VAULT/ATTACHMENTFOLDER/IMAGE
Image works in src in Desktop: **YES**
MobileUri: capacitor://localhost/capacitor_file/private/var/mobile/Library/Mobile%20Documents/iCloud~md~obsidian/Documents/VAULT/ATTACHMENTFOLDER/IMAGE
Image works in src in Mobile: **YES**

Maybe I am missing something, but in my digging in the Obsidian API I didn't manage to find a solution to this. If I am making a mistake somewhere without realising, please do let me know and I will update the code to simplify it.

Ciao
Carlo

liamcain:
Oh, okay. I'll double-check on the initialiseRoot issue. Maybe there's a helper function we can expose to make this easier for plugins. Otherwise, everything else looks good, nice job!

PR #1111, opened 2022-08-13
Add URL Namer plugin

zfei:

## Repo URL

Link to my plugin: https://github.com/zfei/obsidian-url-namer

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4987,5 +4987,13 @@
         "description": "This is a plugin for synchornizing Obsidian notes to other note-based softwares like Anki, following more strictly the principles of Zettelkasten and treating each Obsidian file as a note.",
         "repo": "tansongchen/obsidian-note-synchronizer",
         "branch": "master"
+    },
+    {
+        "id": "obsidian-url-namer",
+        "name": "URL Namer",
+        "author": "zfei",
+        "description": "This plugin retrieves the HTML titles to name the raw URL links.",
+        "repo": "https://github.com/zfei/obsidian-url-namer",
+        "branch": "master"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4993,7 +4993,7 @@
         "name": "URL Namer",
         "author": "zfei",
         "description": "This plugin retrieves the HTML titles to name the raw URL links.",
-        "repo": "https://github.com/zfei/obsidian-url-namer",
+        "repo": "zfei/obsidian-url-namer",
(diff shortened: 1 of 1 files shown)


joethei:
You can remove `obsidian` from your plugin id, its unnecesarry.

[static rawUrlPattern = /(?<!\]\(\s*)(?<=\s|\(|\[|^)(?:https?:\/\/)?[a-zA-Z0-9]+[a-zA-Z0-9\-_.]*\.[a-z]{2,6}[^\s]*\b/gim;](https://github.com/zfei/obsidian-url-namer/blob/b4605217e1b65c3a1f933da50bb53f195d1051c0/main.ts#L50)
As you are using lookbehinds here this plugin will not work on iOS

You should add a notice while the plugin is doing its thing.
Just set the duration to `0` and call the `hide()` method on the notice once its finished.

[const res = await requestUrl({ url: reqUrl });](https://github.com/zfei/obsidian-url-namer/blob/b4605217e1b65c3a1f933da50bb53f195d1051c0/main.ts#L105)
You should check `res.status` if the request was succesfull, it returns HTTP status codes.

zfei:
@joethei thanks for the kind comments.

As I'm looking to parametrize the url pattern regex into plugin settings, will just leave the compatibility flag as is.

Removed the `obsidian-` prefix in the plugin json. Addressed the comments on the loading indicator notice and response status code check in the latest commit: https://github.com/zfei/obsidian-url-namer/commit/6cc1474e3c8fd365e52a4287ddbedaa315c7085d.

community-plugins.json
```diff
@@ -4987,5 +4987,13 @@
         "description": "This is a plugin for synchornizing Obsidian notes to other note-based softwares like Anki, following more strictly the principles of Zettelkasten and treating each Obsidian file as a note.",
         "repo": "tansongchen/obsidian-note-synchronizer",
         "branch": "master"
+    },
+    {
+        "id": "url-namer",
+        "name": "URL Namer",
+        "author": "zfei",
+        "description": "This plugin retrieves the HTML titles to name the raw URL links.",
+        "repo": "https://github.com/zfei/obsidian-url-namer",
```
liamcain, line 4996:
```suggestion
        "repo": "zfei/obsidian-url-namer",
```

PR #1115, opened 2022-08-15
Obsidian Git Mobile

Vinzent03:

## Repo URL

Link to my plugin: https://github.com/Vinzent03/obsidian-git-mobile

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

The code of this plugin lies in [obsidian-git](https://github.com/denolehov/obsidian-git). Since 1,5 years I'm the maintainer of the plugin. So I'm updating the author field of Obsidian Git.
The codebase works with two different git implementations. I'm removing the native git dependency in the [build process](https://github.com/Vinzent03/obsidian-git-mobile/blob/main/.github/workflows/main.yml) of this plugin. I haven't found a way to package node dependencies in a mobile version, so I have to create a whole new plugin entry.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16,8 +16,8 @@
     {
         "id": "obsidian-git",
         "name": "Obsidian Git",
-        "author": "Denis Olehov",
-        "description": "Backup your vault with git.",
+        "author": "Vinzent, (Denis Olehov)",
+        "description": "Backup your vault with Git",
         "repo": "denolehov/obsidian-git"
     },
     {
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4671,10 +4671,10 @@
     },
     {
         "id": "obsidian-git-isomorphic",
-        "name": "Obsidian Git Isomorphic",
+        "name": "Obsidian Git Mobile",
         "author": "Vinzent",
-        "description": "Backup your vault with Git on all devices.",
-        "repo": "Vinzent03/obsidian-git-isomorphic",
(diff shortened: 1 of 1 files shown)


joethei:
I feel like I am missing something here.

In your GH action I can see that you are switching the simpleGit file from the main plugin to the version in this repo.
All I can see in this repo is that none of the methods in that file are implemented, leading me to think that the plugin will not work.

You should be able to just include this in the main plugin instead.
quoting Licat from when I did something similar in one of my plugins.

> You can use `require()` directly in your code selectively so that you don't need a second plugin.
>
>Instead of importing the NodeJS packages directly at the top of the file, you could instead import it inside a function like this:
>
>```ts
>function doSomeNodeStuff() {
>  let fs = require('fs');
>  ..
>}
>```
>
> Since the require will never be called unless you call the function, it will work on mobile as long as you don't call the function (test for `Platform.isDesktopApp`)

Vinzent03:
The methods for mobile are implemented in `isomorphicGit.ts` in the main repo. On mobile I ensure that I don't create an `SimpleGit` instance, but an `IsomorphicGit` instance instead.
Using `require` in a method may be the solution I've always searched for. Thanks.

Vinzent03:
@joethei Thank you very much for that tip. It's working now without an extra plugin. May you still merge this pr which the changes to the author field?

PR #1119, opened 2022-08-16
Auto Hide plugin

skelato1:

## Repo URL

Link to my plugin: https://github.com/skelato1/obsidian-auto-hide

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [getLeaf](https://github.com/skelato1/obsidian-auto-hide/blob/cf0a12bd77bef172af4d85d4b02b5b9d1d311398/main.ts#L54) You shouldn't be using `getLeaf` here. `getLeaf` doesn't guarantee you get the currently active leaf. Instead, you could use [activeLeaf](https://github.com/obsidianmd/obsidian-api/blob/6b2138aa28fb6d6f58f78e6fabec02186fdc0f82/obsidian.d.ts#L3880) which is deprecated but would better suit your needs. That being said, you're current logic only supports clicking to close. I think a better approach would be adding an event listener to [`active-leaf-change`](https://github.com/obsidianmd/obsidian-api/blob/6b2138aa28fb6d6f58f78e6fabec02186fdc0f82/obsidian.d.ts#L4079) and if the activeLeaf's container !== left or right split, then you can collapse the sidebars. This would allow the sidebars when you select a file from the file explorer, for example.

skelato1:
@liamcain

Thank you for your review and advice, Liam! And sorry for being late.
I tested your approach, But I didn't love its behavior. So I changed with another approach. Is it okay?

liamcain:
Sounds good, the new approach looks good to me as long as you're okay with it only working on mouse click and not based on focus.

PR #1137, opened 2022-08-25
Add 'Embed Code file' Plugin

almariah:

## Repo URL

Link to my plugin: https://github.com/almariah/embed-code-file

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- I think your readme could use a screenshot or 2 to show what this plugin does
- [onunload](https://github.com/almariah/embed-code-file/blob/f67c3a3e0c7f4b92bd97ec45d84571b770df830e/main.ts#L25) Since you're not putting any of the replaced content back, I don't think you need this. The code blocks will go away the next time the page is rendered
- [extractTitle](https://github.com/almariah/embed-code-file/blob/f67c3a3e0c7f4b92bd97ec45d84571b770df830e/utils.ts#L102) It looks like your codeblock syntax is the same as YAML. Instead of parsing this yourself, you can just use [`parseYaml`](https://github.com/obsidianmd/obsidian-api/blob/6b2138aa28fb6d6f58f78e6fabec02186fdc0f82/obsidian.d.ts#L2538) and get the title, path, and lines from there.
- [tFile](https://github.com/almariah/embed-code-file/blob/f67c3a3e0c7f4b92bd97ec45d84571b770df830e/main.ts#L44) instead of casting this to TFile, you should do `instanceof TFile` since an TAbstractFile can also be a folder

almariah:
@liamcain I have done the requested changes. Thanks for reviewing!

liamcain:
- [let matchTitle = codeBlockFirstLine.match(/TITLE:\s*"([^"]*)"/i)](https://github.com/almariah/embed-code-file/blob/f80ff755e41d55f4d4dc8dd7c548298831a37a44/main.ts#L107-L108) I think you should use parseYaml for this part too. What you have now looks like it would only work if `TITLE` is the first property in the codeblock, right? Regardless, I think the less manual parsing you do, the better.

almariah:
@liamcain 
There are two types of title:
1. when embed code file from vault. In that case I am using `parseYaml`, because the block content is same as YAML.
2. when you are not embedding any code file but write normal code block:
````
```python
# some code here
```
````
In the second case the only way to add the `TITLE` is first line (or it would be part of the code itself):
````
 ```python TITLE: "some title"
# some code here
```
````

or as part of python code
````
 ```python
TITLE: "some title" # not proper python code
# some code here
```
````


That is why I am using [let matchTitle](https://github.com/almariah/embed-code-file/blob/f80ff755e41d55f4d4dc8dd7c548298831a37a44/main.ts#L107-L108) in the second case.

Anyway I don't like the second case, I wanted to get rid of it, because there are other plugins provide it. But I kept it for consistency of title style in both cases; embed code file or write code block. Other plugins provide different style for title.

liamcain:
oh, interesting. I should have looked closer, I was mostly looking at the diff. Okay. That's fine then.

PR #1141, opened 2022-08-28
Add obsidian-day-and-night to community-plugins.json

CyberT17:

## Repo URL

Link to my plugin: https://github.com/CyberT17/obsidian-day-and-night

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4660,5 +4660,13 @@
         "description": "Create folder of snippets to activate them in one click !",
         "repo": "Mara-Li/obsidian-group-snippets",
         "branch": "master"
+    },
+    {
+        "id": "obsidian-day-and-night",
+        "name": "Day and Night",
+        "author": "Kevin Patel",
+        "description": "An Obsidian plugin to automatically toggle themes between day theme and night theme on a set time schedule.",
+        "repo": "CyberT17/obsidian-day-and-night",
+        "branch": "master"
     }
 ]


liamcain:
This functionality is already possible using "Adapt to system" and using your OS's native auto-dark mode.
image

For example, on macOS there's "auto" 
image

---

[addDropdown](https://github.com/CyberT17/obsidian-day-and-night/blob/02e91cf89826b7c8ab9fbc33d58f07ec01051fed/main.ts#L156) Wouldn't it be simpler to just assume night = dark mode and day = light mode? It's a bit confusing to refer to this as switching the "theme" since users might think it switches between different Community themes. Saying it switches between "light and dark mode" is probably better to avoid confusion.

[async updateTheme() {](https://github.com/CyberT17/obsidian-day-and-night/blob/02e91cf89826b7c8ab9fbc33d58f07ec01051fed/main.ts#L55-L56) I suggest checking if the currentTheme matches the themeToApply before calling setTheme. That way `css-change` doesn't get called every minute, which can have unexpected side-effects from other plugins that listen for that event.

CyberT17:
The functionality that the OS provides applies to all the apps. Changing it on the OS level would mean that it would change the theme for finder, system preferences, safari, etc, which is not something I want.
This plugin also allows specifying the time to switch themes outside of the normal sunrise/sunset.
Personally, I have the plugin set up to switch to the light theme at 10am and night theme at 3 pm.

As for changing it to Light/Dark from Day/Night. One of the things I would like to do in the future is to allow the user to choose between different community themes. I am still learning the ins and outs of what I can and I can't do in a plugin.
A user could choose 2 different Dark themes. By naming it Day/Night l was looking to keep the naming convention more flexible.

I have implemented your suggestion regarding checking the currentTheme before calling setTheme in release 1.0.3.

liamcain:
Sounds good! The changes look good, would you mind updating your pull request to add the entry to the bottom of the plugins list? (we use the order to sort plugins by "newest" in the plugin gallery)

liamcain:
Actually, nevermind. I was able to do it from Github.

PR #1144, opened 2022-08-29
Add Todoist Completed Tasks plugin

Ledaryy:

## Repo URL

Link to my plugin: https://github.com/Ledaryy/obsidian-todoist-completed-tasks

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [ribbonIconEl.addClass("my-plugin-ribbon-class")](https://github.com/Ledaryy/obsidian-todoist-completed-tasks/blob/e71f5404c27a206a75cd1f2c0533df5a5c991ac8/main.ts#L20) you can remove this, it's boilerplate
- [vault](https://github.com/Ledaryy/obsidian-todoist-completed-tasks/blob/e71f5404c27a206a75cd1f2c0533df5a5c991ac8/src/updateFileContent.ts#L10) Since your plugin only operates on open files, I suggest using the CodeMirror editor instead of using `read` and `modify`. There are a lot of benefits:
- You don't lose your scroll position
- The changes get added to the undo history
- Less risk of data loss in case the current view has unsaved changes
- For `keywordSegmentStart` and `keywordSegmentEnd`, I know they are customizable, but i think it would make a lot more sense using Obsidian's comments (`%% COMPLETED_TODOIST_TASKS %%`). That way the markers won't render in preview mode.

Ledaryy:
Thanks for the feedback!
I've removed all boilerplate code. And replaced modify/read with the editor.
About the markers, I've tested them, and they work fine with the current state of the code. I also added them as the default setting.
Also, I moved moments js from dependencies and took directly from Obsidian. And more other small changes/bugfixes.

liamcain:
Nice! The changes look great. One more small suggestion:

When using the Editor, you should try to just replace the ranges of text with changes. Currently, you're reconstructing the file as a strong and using `setValue`. This still has the same downsides as using `vault.modify` since Obsidian can't tell what text changed and so it will clear out any folded text, remove any selections, and move the cursor to the beginning of the file.

Instead, you'll want to do something like this (psuedo-code):

```ts
const rangeStart = fileContent.indexOf(settings.keywordSegmentStart);
const rangeEnd = fileContent.indexOf(settings.keywordSegmentEnd);

editor.replaceRange(
  formattedTodos,
  editor.offsetToPos(rangeStart),
  editor.offsetToPos(rangeEnd)
);
```

That will tell the editor to just replace text from the start to the end of your segment.

Ledaryy:
Thanks for the explanation!
I replaced the setValue with replaceRange, seems to be working great!

PR #1148, opened 2022-08-30
Remove obsidian-deep-work-theme

nikbrunner:
# I am ~~submitting~~ removing a Community Theme

## Repo URL

Link to my theme: [Obsidian Deep Work Theme](https://github.com/nikbrunner/obsidian-deep-work-theme)

## Explanation

Due to the new changes introduced for 0.16.0, I want to remove my Theme from the Community Themes List for the time being.
The new changes are very good, but I rather want to use my limited time for my [new original project](https://github.com/terra-theme), which will also include an Obsidian Theme, based on the new system.

I'd rather make a clean break and not let the project sit around untended.

### DIFF
--- a/community-css-themes.json
+++ b/community-css-themes.json
@@ -475,14 +475,6 @@
         "modes": ["dark"]
     },
     {
-        "name": "Deep Work",
-        "author": "nikbrunner",
-        "repo": "nikbrunner/obsidian-deep-work-theme",
-        "screenshot": "screenshot.png",
-        "modes": ["dark", "light"],
-        "legacy": true
-    },
-    {
         "name": "Blackbird",
         "author": "Ivan Chernov",
         "repo": "vanadium23/obsidian-blackbird-theme",


lishid:
We can keep it here with the legacy flag and Obsidian 0.16 will not list the them unless the user previously had it installed. Only versions before 0.16 can find it from the list.

lishid:
New project looks amazing by the way!

ericaxu:
@nikbrunner What do you think? You can submit the new theme without the legacy flag once it's finished.

nikbrunner:
Hi :)
Thank you for your response and kind words! If that is the case, just keeping it here would be also a very good option. I can update it when I am ready and finished with Terra. :)

Thank you very much again, and also for the kind words towards Terra. :) It is a lot of work and very ambitious, but I learned a lot through work and through my first theme here. I want to take my time and do it properly. First I want to finish the Theme for Vim and VSCode, and after that Obsidian.

Until then! I will close this for now. 👋

PR #1155, opened 2022-09-03
Add "Open In Other Editor" plugin

yekingyan:

## Repo URL
Link to my plugin: https://github.com/yekingyan/obsidian-open-in-other-editor

## Release Checklist
    - [x] Windows
    - [ ] macOS
    - [ ] Linux
    - [ ] Android (if applicable)
    - [ ] iOS (if applicable)
     - [x] main.js
     - [x] manifest.json
     - [ ] styles.css (optional)
 My README.md describes the plugin's purpose and provides clear usage instructions.
 - [x] My project respects and is compatible with the original license of any code from other plugins that I'm using. I have given proper attribution to these other projects in my README.md.


liamcain:
- [this.app.workspace.getActiveFile()?.path](https://github.com/yekingyan/obsidian-open-in-other-editor/blob/05363d3317e5e0744d40af037cbaa3ea5fb8dc16/main.ts#L46) if `getActiveFile` is null, you should just return. Otherwise, the call to `open` will fail since you're interpolating `undefined` with a string

yekingyan:
Thank you for the review.
I fixed these issues in version 0.0.3.
Please check it again when you have time.

PR #1156, opened 2022-09-04
Janitor Submission

Canna71:

## Repo URL

Link to my plugin:
https://github.com/Canna71/obsidian-janitor

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [updateNoteWithDate](https://github.com/Canna71/obsidian-janitor/blob/328df6683175cab8d9251a6ca1687d67119f377b/main.ts#L141) if possible, I would suggest changing this function to use the CodeMirror Editor instead of `modify`. From what i can tell, you are always modifying an open file, so it should be pssible and it has several advantages: doesn't lose scroll position, properly applies to undo history.

Otherwise everything else looks good, let me know when this is ready to go in.

Canna71:
Thank you @joethei  for your feedbacks, they are helping a lot. You are perfectly right, I should have thought about it!
I modified the code in order to use the Editor when possible. I would leave the possibility to set the expiration date also when not in edit mode (it doesn't even seem to lose scroll position) but of course now I am using the editor when in edit node and I only change the frontmatter part.
I already updated the release with updated code.

Canna71:
Ah, I also tested it on iOS.

PR #1157, opened 2022-09-05
Add "New Tab Default Page" to Community Plugins

chrisgrieser:
"New Tab Default Page" only works with Obsidian 0.16 or above. (Already correctly reflected in `manifest.json` and `versions.json` – just mentioning this here for anyone coming from Discord.)

## Repo URL

Link to my plugin: https://github.com/chrisgrieser/new-tab-default-page

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


pjeby:
Have you tested what happens if you switch to a tab with a non-file view?  The way I'm reading your code, it suggests that if somebody switches to a tab with a Custom Frame (for example), that it'd be replaced with the new tab view.  Also, if someone navigates backwards to a deleted file, say, it seems it would also trigger.

What I would suggest doing instead to avoid false positives is to create a `WeakSet<WorkspaceLeaf>` and populate it via `iterateAllLeaves()` upon `onLayoutReady()`, then watch for `layout-change` events.  When the layout-change event fires, iterate all leaves again, and any leaf that is not in the set and has `!leaf.view || leaf.view.getType() === "empty"`, that will let you spot *only* just-added leaves with no actual file or view of another kind, while ignoring leaves that already existed at startup or when the plugin opened.  The layout-change event fires even before the new tab command finishes running (let alone the file-open event), so you can get it started loading perhaps before the tab even starts rendering as well.

Another thing you may want to consider is using [`getFirstLinkpathDest()`](https://github.com/obsidianmd/obsidian-api/blob/master/obsidian.d.ts#L2276-L2281) to get the file, rather than getAbstractFileByPath, as the former supports any unique name and adds the `.md` for you as necessary.  In addition, using it would let you target non-markdown files (e.g. images), in the event somebody just wants a nice wallpaper image for their new tabs.  :)

chrisgrieser:
> Another thing you may want to consider is using [getFirstLinkpathDest()](https://github.com/obsidianmd/obsidian-api/blob/master/obsidian.d.ts#L2276-L2281) to get the file, rather than getAbstractFileByPath, as the former supports any unique name and adds the .md for you as necessary. In addition, using it would let you target non-markdown files (e.g. images), in the event somebody just wants a nice wallpaper image for their new tabs. :)

good idea, this even makes it work when the user forgot to enter the parent folder of an (otherwise unique) note. Thanks!

> What I would suggest doing instead to avoid false positives is to create a `WeakSet<WorkspaceLeaf>` and populate it via `iterateAllLeaves()` upon `onLayoutReady()`

Could you link me an example of how this works? (or make a PR, whatever you prefer)

pjeby:
So, in your plugin onload, you'd do something like:

```ts
app.workspace.onLayoutReady(() => {
    // Get existing leaves (tabs)
    const existingLeaves = new WeakSet<WorkspaceLeaf>();
    app.workspace.iterateAllLeaves(leaf => { existingLeaves.add(leaf); });
    // When layout changes, check for new ones
    this.registerEvent(app.workspace.on("layout-change", () => {
        app.workspace.iterateAllLeaves(leaf => {
            if (!existingLeaves.has(leaf)) {
                existingLeaves.add(leaf);  // only ever check a leaf once
                if (!leaf.view || leaf.view.getViewType() === "empty") {
                    // leaf is a new tab, do something with it here; it may not be
                    // active yet, so explicitly target leaf or set it active first, eg.:
                    app.workspace.setActiveLeaf(leaf, false, true)
                    app.workspace.openLinkText("linkpath", "", false);
                }
            }
        });
    }));
});
```

This creates a set of the open tabs when the plugin starts, as soon as the layout is ready.  Then, whenever the layout changes it checks for newly opened leaves that are empty.  (The reason to use a WeakSet instead of a regular set is that it won't hold references - the set will automatically remove leaves that are no longer in use.)

(It does occur to me that there is still a false positive possible, in that if you switch workspaces *all* the leaves in the new workspace will be considered "new" and checked for emptiness.  But that's a lot less likely, and a lot harder to do anything about.)

chrisgrieser:
Thank you, works great!

Just to check: you used `openLinkText` for the sake of brevity?

(I replaced that part with `getFirstLinkpathDest` and `activeLeaf.openFile`, since `openLinkText` seems to create a new file if the file at the path does not exist)

liamcain:
```ts
		app.workspace.setActiveLeaf(leaf, false, true); // ensure this is the active leaf, since `layout-change` can trigger before the the leaf becomes active
		app.workspace.activeLeaf.openFile(tFiletoOpen);
```

I think you should just do `leaf.openFile(tFileToOpen)` here and don't bother setting the leaf as active (which might cause issues)

pjeby:
The openLinkText was for brevity's sake, yes.  I concur with Liam that `leaf.openFile()` suffices; the only reason I used setActiveLeaf is because openLinkText only works on the active leaf.  Sorry for any confusion there.

chrisgrieser:
@pjeby no worries!

@liamcain done

liamcain:
@chrisgrieser I meant literally change the last line to `leaf.openFile(tFiletoOpen);` (instead of using `workspace.activeLeaf`) since you already have the `leaf`

chrisgrieser:
ah, I misunderstood. Yeah, done.

liamcain:
@chrisgrieser, one last thing: you also need to bump the minAppVersion to 0.16. https://github.com/Canna71/obsidian-janitor/blob/master/manifest.json#L5

PR #1159, opened 2022-09-06
Add "ob-table-enhancer" plugin

Stardusten:

## Repo URL

Link to my plugin: https://github.com/Stardusten/ob-table-enhancer/tree/master

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4691,5 +4691,13 @@
         "description": "Inserts the note file path above each note.",
         "repo": "b0o/obsidian-embedded-note-paths",
         "branch": "main"
+    },
+    {
+        "id": "ob-table-enhancer",
+        "name": "Table Enhancer",
+        "author": "stardust",
+        "description": "A plugin for improving the experience of using table in Obsidian.",
+        "repo": "Stardusten/ob-table-enhancer",
+        "branch": "master"
     }
 ]


joethei:
[MyPlugin](https://github.com/Stardusten/ob-table-enhancer/blob/3ee82f5381efdf6f6436b46b59cb07b63f9ceb26/main.ts#L23)
Please rename this class to something more meaningful.

[// 劫持滚动事件](https://github.com/Stardusten/ob-table-enhancer/blob/3ee82f5381efdf6f6436b46b59cb07b63f9ceb26/main.ts#L55-L67)
This will only work for the editor instance that was loaded at start and will completely block scrolling.

[export const insertBelowIcon](https://github.com/Stardusten/ob-table-enhancer/blob/3ee82f5381efdf6f6436b46b59cb07b63f9ceb26/src/tool-bar.ts#L7)
Obsidian includes all icons from <https://lucide.dev/>, you can use any of those with the `setIcon` function from the API out of the box.

In general, you do way to much DOM manipulation,
most things you do here can either be acomplished with the Obsidian API or in a CodeMirror editor extension.
<https://marcus.se.net/obsidian-plugin-docs/editor/extensions>

Stardusten:
> [MyPlugin](https://github.com/Stardusten/ob-table-enhancer/blob/3ee82f5381efdf6f6436b46b59cb07b63f9ceb26/main.ts#L23)
Please rename this class to something more meaningful.

Fixed!

> [// 劫持滚动事件](https://github.com/Stardusten/ob-table-enhancer/blob/3ee82f5381efdf6f6436b46b59cb07b63f9ceb26/main.ts#L55-L67)
This will only work for the editor instance that was loaded at start and will completely block scrolling.

My bad, I've deleted it. But I have tried to prevent abnormal scrolling by using `dispatch` in codemirror editor instead (see `editor-utils.ts`), but but it still sometimes goes wrong. Do you have any good idea about this?

> [export const insertBelowIcon](https://github.com/Stardusten/ob-table-enhancer/blob/3ee82f5381efdf6f6436b46b59cb07b63f9ceb26/src/tool-bar.ts#L7)
Obsidian includes all icons from https://lucide.dev/, you can use any of those with the setIcon function from the API out of the box.

There are a few icons I need that I can't find on this site, so I'd rather not change them.

> In general, you do way to much DOM manipulation,
most things you do here can either be acomplished with the Obsidian API or in a CodeMirror editor extension.
https://marcus.se.net/obsidian-plugin-docs/editor/extensions

Indeed, I didn't  reading this document carefully before, but looked directly at the ob api, so I ignored the `EditorExtension` which is really useful and much more elegant than DOM manipulation. But this is difficult to modify now, I will try this when adding new features or refactoring later.

joethei:
Hi, we currently don't feel comfortable merging this in its current state.

The DOM hacks you do are to unstable to have a stable plugin in the long term.

From playing with it in my test vault on 0.16.2(current insider) I also found some issues with data synchronisation.
Basically the version visible in edit mode with your plugin and the actual version in the file did not match at all.

Sorry about that.

ericaxu:

jwierzbo:
@Stardusten do you have any plans for code refactor to address that DOM manipulation change? Would be nice to have that plugin in the official plugin repo..

PR #1160, opened 2022-09-07
Add "Cmenu-Toolbar" plugin

cumany:

## Repo URL

Link to my plugin: https://github.com/cumany/Cmenu-Toolbar

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


tingmelvin:
I'm just thinking if 1) it might be worth asking for a merger into the original cMenu or 2) changing the name at least to drop "cMenu" or to call it something else. it's going to be quite confusing debugging and answering queries when there's two "cMenu" named plugins by different authors in the store

cumany:
> I'm just thinking if 1) it might be worth asking for a merger into the original cMenu or 2) changing the name at least to drop "cMenu" or to call it something else. it's going to be quite confusing debugging and answering queries when there's two "cMenu" named plugins by different authors in the store

Thanks for your pointers
1. I tried pr , but the cmenu plugin has not been updated for more than a year, I fixed many bugs and added some new features.
2. i have changed the name to "obsidian-editing-toolbar"

joethei:
[manifest.json](https://github.com/cumany/obsidian-editing-toolbar/blob/ad8bb0cdcdf0b5a10d17a8e3327b6b38a6346764/manifest.json)
You can remove `Obsidian` from your plugin name and id, it's unnecessary and confusing to users.
Did you test your plugin in Obsidian 0.9.12?
Please set the `minAppVersion` to the latest version you tested the plugin with.

[export function addFeatherIcons(appIcons: string[]) {](https://github.com/cumany/obsidian-editing-toolbar/blob/ad8bb0cdcdf0b5a10d17a8e3327b6b38a6346764/src/icons/customIcons.ts#L39)
You can remove all references to feather icons, Obsidian includes Lucide since a few version ago, its an up to date fork of Feather icons.

[window.newNotice =](https://github.com/cumany/obsidian-editing-toolbar/blob/ad8bb0cdcdf0b5a10d17a8e3327b6b38a6346764/src/plugin/main.ts#L187)
What is the reasoning behind this?
I feel like it is unnecessary, at least here.

[window.ISMORE = false;](https://github.com/cumany/obsidian-editing-toolbar/blob/ad8bb0cdcdf0b5a10d17a8e3327b6b38a6346764/src/modals/cMenuToolbarModal.ts#L10-L13)
I would recommend against saving state in the window, also these names are not really meaningful.

[Setfontcolor](https://github.com/cumany/obsidian-editing-toolbar/blob/ad8bb0cdcdf0b5a10d17a8e3327b6b38a6346764/src/modals/cMenuToolbarModal.ts#L235), [tabCell](https://github.com/cumany/obsidian-editing-toolbar/blob/ad8bb0cdcdf0b5a10d17a8e3327b6b38a6346764/src/modals/cMenuToolbarModal.ts#L190), [CreateDiv](https://github.com/cumany/obsidian-editing-toolbar/blob/ad8bb0cdcdf0b5a10d17a8e3327b6b38a6346764/src/modals/cMenuToolbarModal.ts#L184), [checksvg](https://github.com/cumany/obsidian-editing-toolbar/blob/ad8bb0cdcdf0b5a10d17a8e3327b6b38a6346764/src/modals/cMenuToolbarModal.ts#L179)
Your function naming is not consistent, making it awkward to read.
I would recommend on choosing one and only using that one.

cumany:
Hi，
Went through your comments and made some changes, would love to have you take another look, thanks :D

liamcain:
Hi, sorry for the long delay in reviewing. I went through your code and found a few of small issues. Just ping me when these get addressed and I think we can merge this, thanks!

---

- [let currentleaf = activeDocument.body](https://github.com/cumany/obsidian-editing-toolbar/blob/1adaa56673fc5188386d605ca7aeb460237c1ad4/src/plugin/main.ts#L697-L698) I know it's marked as deprecated, but you should use workspace.activeLeaf instead of querying the dom for this.
- [globalThis](https://github.com/cumany/obsidian-editing-toolbar/blob/1adaa56673fc5188386d605ca7aeb460237c1ad4/src/plugin/main.ts#L149) You should avoid using global variables if you don't need to. This pollutes the global namespace and could result in collisions with other plugins. I think these can just be private instance variables on the plugin class anyway.
- [hotkeys: [{ modifiers: ["Mod"], key: "1" }],](https://github.com/cumany/obsidian-editing-toolbar/blob/1adaa56673fc5188386d605ca7aeb460237c1ad4/src/plugin/main.ts#L319-L320) as of Obsidian 1.0, these hotkeys are being used for switching tabs. I suggest just unsetting these default hotkeys and allowing users to specify their own.
- [hotkeys](https://github.com/cumany/obsidian-editing-toolbar/blob/1adaa56673fc5188386d605ca7aeb460237c1ad4/src/settings/settingsData.ts#L50) why are you saving hotkeys in your settings file? These are automatically saved by Obsidian so no need to save them yourself.
- [querySelector(".workspace-leaf.mod-active")](https://github.com/cumany/obsidian-editing-toolbar/blob/1adaa56673fc5188386d605ca7aeb460237c1ad4/src/modals/cMenuToolbarModal.ts#L105) same here, you can use workspace.activeLeaf
- [activeDocument.body?.querySelector(".workspace-leaf.mod-active")](https://github.com/cumany/obsidian-editing-toolbar/blob/1adaa56673fc5188386d605ca7aeb460237c1ad4/src/modals/cMenuToolbarModal.ts#L624) same here

cumany:
@liamcain 
Thanks for the review, I've updated the code。

liamcain:
Hi @cumany, the changes look good and the plugin LGTM 🚀 . There is a small issue with the PR though since you have 2 plugins both submitted from the `master` branch of your fork, so it's not easy for me to rebase.

I manually added the entry to the store here https://github.com/obsidianmd/obsidian-releases/commit/2fa3ad5e20df61013066db5ae2f16e1a34f7317f so I'm closing this Pull Request (so don't be alarmed, the plugin is in the Plugin Gallery now!).

PR #1162, opened 2022-09-08
remove block reference counts from community plugins

shabegom:
Block Reference Counts is no longer maintained and has a bug that causes issues on link renaming.


joethei:
The diff you created is pretty big,
it seems like you touched every line with an autoformatter.

Can you try and undo that?

shabegom:
ugh, yeah sorry

shabegom:
wow, that was more painful than it should have been. @joethei should be fixed.

joethei:
thank you

PR #1163, opened 2022-09-08
Add "Copy Search URL" plugin

czottmann:

## Repo URL

Link to my plugin: https://github.com/czottmann/obsidian-copy-search-url

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [this.button.parentNode?.removeChild(this.button);](https://github.com/czottmann/obsidian-copy-search-url/blob/d18ff31efa4c599c970c338b8aefed1a7db22126/main.ts#L54) You might want to add a check to make sure that `this.button` isn't null. Also you can use [`.detach()`](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L38) which is available on all HTMLElements in Obsidian. So this would just become `this.button?.detach();`

- [return (this.getSearchLeaf().view as SearchLeafView)?.getQuery();](https://github.com/czottmann/obsidian-copy-search-url/blob/d18ff31efa4c599c970c338b8aefed1a7db22126/main.ts#L66-L67) You might want to return `""` if the search leaf can't be found, just so that your return type is always a string. It shouldn't make a difference, but you are checking for emptiness inside the click event listener so it might be safer.

czottmann:
Thanks, @liamcain, much appreciated!

PR #1164, opened 2022-09-10
Add Influx plugin to community-plugins.json

jensmtg:

## Repo URL

Link to my plugin: https://github.com/jensmtg/influx

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js` (Yes: `main.tsx`)
  - [x] `manifest.json`


liamcain:
- [const _file = file instanceof TFile ? file : null](https://github.com/jensmtg/influx/blob/7e1f90529aa74e67aed56109cd3a2bbc673dc563/src/main.tsx#L94) I think you only want to trigger the callbacks if file is an instance of `TFile`, right? I think you want to wrap this in an if statement instead of using the ternary.
- [Math.random().toString](https://github.com/jensmtg/influx/blob/7e1f90529aa74e67aed56109cd3a2bbc673dc563/src/InfluxFile.tsx#L22) Why not just use the file path or something deterministic for the id? Math.random isn't really a good fit for a UUID.
- [const callback](https://github.com/jensmtg/influx/blob/7e1f90529aa74e67aed56109cd3a2bbc673dc563/src/InfluxReactComponent.tsx#L119) I think you can move this into the useEffect.
- [const influxFile = new InfluxFile(leaf.view?.file.path, apiAdapter, this)](https://github.com/jensmtg/influx/blob/7e1f90529aa74e67aed56109cd3a2bbc673dc563/src/main.tsx#L150) You're handling the null case for `View` here, but I don't think you want to pass in `undefined` to the InfluxFile constructor. If you don't have a view or a file, just early exit out of this function.
- [rand={rand}](https://github.com/jensmtg/influx/blob/7e1f90529aa74e67aed56109cd3a2bbc673dc563/src/main.tsx#L166) Unused, you can remove this.

jensmtg:
Thanks! The suggested changes have been implemented now, so I think it is ready to be merged.

Re performance: I agree with you concerns. For the current version I've introduced selective/lazy re-rendering, as well as the abilty to disable live-updates completly - with a note about performance impact in the settings.

PR #1168, opened 2022-09-12
Add "System Theme" plugin

jgauth:

## Repo URL
Link to my plugin: https://github.com/jgauth/obsidian-system-theme

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


jgauth:
additional context for why this plugin exists https://forum.obsidian.md/t/color-scheme-adapt-to-system-not-working-on-linux-fedora-gnome/38743/13

liamcain:
Oh interesting! I wasn't aware of this bug. We should definitely fix this in the core app. Since the release cycle on our end takes a little while, I think it's fine to have this in the plugin gallery for now. I'll make a note that this is an issue for Linux users and we can take a look at fixing it for good.

PR #1170, opened 2022-09-13
Add Time Tracker plugin

daaru00:

## Repo URL

Link to my plugin: https://github.com/daaru00/obsidian-time-tracker

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
You can remove `obsidian` from the id, it's unnecessary.

[os.EOL](https://github.com/daaru00/obsidian-time-tracker/blob/a833ffe8a3b95198cf3f86b55263e528ec980734/src/lib/file-storage.ts#L52)
I would use something else instead that defines the EOL character,
this seems to be the only thing blocking this plugin from being mobile compatible.


Other than that, looks good.

liamcain:
- [TimerWidget](https://github.com/daaru00/obsidian-time-tracker/blob/0582f42f05c320d56ae424fe6de3af48fb32f673/src/main.ts#L212) You should wrap your widgets in [MarkdownPostProcessorContext](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L1874) to manage their lifecycle. That way you can properly detach the event listeners when the component is destroyed.
- [window.document.querySelectorAll](https://github.com/daaru00/obsidian-time-tracker/blob/0582f42f05c320d56ae424fe6de3af48fb32f673/src/main.ts#L130) I don't think this will work with pop-out windows, you'll either need to setup a separate interval with each window (there's a window-open event you can listen for) or change up the approach to have each widget responsible for its own interval.

ericaxu:
@daaru00 Hi! Do you still intend to submit this plugin? Thanks!

joethei:
Closing PR due to inactivity.

Feel free to reopen this PR if you still intend on submitting this plugin.

PR #1171, opened 2022-09-14
Add "3D Graph" plugin

AlexW00:

## Repo URL

Link to my plugin: https://github.com/AlexW00/obsidian-3d-graph

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5355,5 +5355,13 @@
         "author": "Sean Quinlan",
         "description": "This is a plugin for Obsidian that adds support for handlebars template blocks in notes.",
         "repo": "sbquinlan/obsidian-handlebars"
+    },
+    {
+        "id": "3d-graph",
+        "name": "3D Graph",
+        "author": "Alexander Weichart",
+        "description": "A 3D graph for Obsidian!",
+        "repo": "AlexW00/obsidian-3d-graph",
+        "branch": "master"
     }
 ]


joethei:
You can remove `obsidian` from the plugin id, it's unnecessary.

AlexW00:
> You can remove `obsidian` from the plugin id, it's unnecessary.

Done.

liamcain:
- [Graph3dPlugin.settingsState = new State<GraphSettings>(this.settings);](https://github.com/AlexW00/obsidian-3d-graph/blob/e0e5e34aedabadaad1acc1b9e0bb490996ec9dec/src/main.ts#L55-L57) Why are these variables static?
- [this.app.metadataCache.on](https://github.com/AlexW00/obsidian-3d-graph/blob/e0e5e34aedabadaad1acc1b9e0bb490996ec9dec/src/main.ts#L98) if you `bind` these function references, you don't need to make everything static. So this would become `this.app.metadataCache.on("resolved", this.onGraphCacheReady.bind(this));`
- I'm guessing you introduced [openFileState](https://github.com/AlexW00/obsidian-3d-graph/blob/e0e5e34aedabadaad1acc1b9e0bb490996ec9dec/src/main.ts#L20) as a workaround because all the variables are static? I think you can avoid storing this value. It shouldn't be needed. You can always access the active file via [Workspace.getActiveFIle](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L4033).
- [if (Graph3dPlugin.cacheIsReady.value) {](https://github.com/AlexW00/obsidian-3d-graph/blob/e0e5e34aedabadaad1acc1b9e0bb490996ec9dec/src/main.ts#L159-L160) I don't think you should delay opening the leaf since this could be a bad user experience. Instead I think the leaf can either be empty or show a load indicator while the cache isn't ready. This should also make the code a bit simpler.
- [(await this.loadData()) as GraphSettings](https://github.com/AlexW00/obsidian-3d-graph/blob/e0e5e34aedabadaad1acc1b9e0bb490996ec9dec/src/main.ts#L168) I really don't think you should be casting a plain javascript object to a class. I think you're better off just defining an interface for your settings and creating utility functions for `resetSettings` and `initializeSettings`.
- [Graph3dPlugin](https://github.com/AlexW00/obsidian-3d-graph/blob/e0e5e34aedabadaad1acc1b9e0bb490996ec9dec/src/views/graph/ForceGraph.ts#L65) I would suggest passing in a reference to App and to your view. Then you don't need all these static variables and you can just use access properties of your plugin via the view.
- [update_dimensions](https://github.com/AlexW00/obsidian-3d-graph/blob/e0e5e34aedabadaad1acc1b9e0bb490996ec9dec/src/views/graph/ForceGraph.ts#L93) Why did you switch to snake_case just for a few function names?

ericaxu:
@AlexW00 Hi! Do you still intend to submit this plugin? Thanks!

AlexW00:
> @AlexW00 Hi! Do you still intend to submit this plugin? Thanks!

Yes I do, there's just a lot going on right now.

ericaxu:
> > @AlexW00 Hi! Do you still intend to submit this plugin? Thanks!
> 
> Yes I do, there's just a lot going on right now.

Thanks for the reply! I'll keep it open for now. Keeps us updated!

AlexW00:
> * [Graph3dPlugin.settingsState = new State(this.settings);](https://github.com/AlexW00/obsidian-3d-graph/blob/e0e5e34aedabadaad1acc1b9e0bb490996ec9dec/src/main.ts#L55-L57) Why are these variables static?
> 
>     * [this.app.metadataCache.on](https://github.com/AlexW00/obsidian-3d-graph/blob/e0e5e34aedabadaad1acc1b9e0bb490996ec9dec/src/main.ts#L98) if you `bind` these function references, you don't need to make everything static. So this would become `this.app.metadataCache.on("resolved", this.onGraphCacheReady.bind(this));`
> 
>     * I'm guessing you introduced [openFileState](https://github.com/AlexW00/obsidian-3d-graph/blob/e0e5e34aedabadaad1acc1b9e0bb490996ec9dec/src/main.ts#L20) as a workaround because all the variables are static? I think you can avoid storing this value. It shouldn't be needed. You can always access the active file via [Workspace.getActiveFIle](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L4033).
> 
>     * [if (Graph3dPlugin.cacheIsReady.value) {](https://github.com/AlexW00/obsidian-3d-graph/blob/e0e5e34aedabadaad1acc1b9e0bb490996ec9dec/src/main.ts#L159-L160) I don't think you should delay opening the leaf since this could be a bad user experience. Instead I think the leaf can either be empty or show a load indicator while the cache isn't ready. This should also make the code a bit simpler.
> 
>     * [(await this.loadData()) as GraphSettings](https://github.com/AlexW00/obsidian-3d-graph/blob/e0e5e34aedabadaad1acc1b9e0bb490996ec9dec/src/main.ts#L168) I really don't think you should be casting a plain javascript object to a class. I think you're better off just defining an interface for your settings and creating utility functions for `resetSettings` and `initializeSettings`.
> 
>     * [Graph3dPlugin](https://github.com/AlexW00/obsidian-3d-graph/blob/e0e5e34aedabadaad1acc1b9e0bb490996ec9dec/src/views/graph/ForceGraph.ts#L65) I would suggest passing in a reference to App and to your view. Then you don't need all these static variables and you can just use access properties of your plugin via the view.
> 
>     * [update_dimensions](https://github.com/AlexW00/obsidian-3d-graph/blob/e0e5e34aedabadaad1acc1b9e0bb490996ec9dec/src/views/graph/ForceGraph.ts#L93) Why did you switch to snake_case just for a few function names?

1. Static variables: Fixed, as you suggested
2. Static functions: Fixed, as you suggested
3. I created the `openFileState` variable as an utility. The utility class `State` wraps any value and notifies listeners on change. The plugin needs to listen to state changes of the open file to change the state of the local graph view.
4. Delayed leaf: Fixed, leafs now open instantly and wait for cache to be ready
5. Casting object to class: Fixed, created utility functions
6. Static plugin: Fixed, as you suggested
7. Snake case: Fixed, renamed to camel case

PR #1172, opened 2022-09-14
Add "Image Inserter" community plugin

cloudy9101:

## Repo URL

Link to my plugin: https://github.com/cloudy9101/obsidian-image-inserter

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
You can remove `obsidian` from the plugin id, its unecessary.

[id: 'image-inserter-plugin',](https://github.com/cloudy9101/obsidian-image-inserter/blob/10f74b7aac144e4a46b98d121a4e8ecbb7f3aea1/main.ts#L24)
I would use a different id, like `insert-unsplash-image`, you don't need to add the plugin name to the command id, that is already handled by Obsidian.

[const url = new URL("/search/photos", "https://insert-unsplash-image.cloudy9101.com/")](https://github.com/cloudy9101/obsidian-image-inserter/blob/10f74b7aac144e4a46b98d121a4e8ecbb7f3aea1/fetcher.ts#L15)
I guess you are using your own server, so users don't have to have an API key for unsplash?
Please document this in the README.
I would also recommend adding some code to handle errors from your own API, since at the moment it would just fail without any notice.

[await fetch(url)](https://github.com/cloudy9101/obsidian-image-inserter/blob/10f74b7aac144e4a46b98d121a4e8ecbb7f3aea1/fetcher.ts#L21)
Use [`requestUrl`](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L2894-L2898) instead.

cloudy9101:
Thank you for the review. I updated the code.

1. Plugin id changed to `insert-unsplash-image`
2. Document the proxy in the plugin repo's README
3. Notice there's something went wrong when the API request failed
4. Replaced `fetch` to `requestUrl`

Please review again when you got time, thanks.

PR #1173, opened 2022-09-14
Add "Custom sort in File Explorer" plugin

SebastianMC:

## Repo URL

Link to my plugin: https://github.com/SebastianMC/obsidian-custom-sort

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
- [x] I have added a license in the LICENSE.md file.


joethei:
["authorUrl"](https://github.com/SebastianMC/obsidian-custom-sort/blob/57b1cd8ae2404c71e261826512c2cb1d3f67f7ad/manifest.json#L8)
Are you sure that you want to expose your email address like that?
It will be visible when browsing community plugins.

[const fileExplorer = this.app.viewRegistry.viewByType["file-explorer"](leaf) as FileExplorerView;](https://github.com/SebastianMC/obsidian-custom-sort/blob/57b1cd8ae2404c71e261826512c2cb1d3f67f7ad/src/main.ts#L239)
You should use the `getFileExplorer()` function you created below.

[_l1s5_processParsedSortGroupSpec](https://github.com/SebastianMC/obsidian-custom-sort/blob/57b1cd8ae2404c71e261826512c2cb1d3f67f7ad/src/custom-sort/sorting-spec-processor.ts#L695)
What is the reasoning for these function names?
They are not really descriptive.

SebastianMC:
Thank you @joethei for the valuable remarks!

Code of the plugin was updated and released, the updates include:
- removed email from the source code. Indeed I'm new to public projects and repositories and not (yet) so sensitive to the privacy aspects. Most of my prior projects were not public
- unified the way of accessing FileExplorerView: all of the access goes via `getFileExplorer()`
- removed the confusing prefixes like `_l1s3_....` from parser methods. They were a residue from early development stage naming. Originally it allowed clear ordering of these functions in the code. No longer needed

liamcain:
- [additionalSortspecFile: 'Inbox/Inbox.md',](https://github.com/SebastianMC/obsidian-custom-sort/blob/d672df6303b221d1907526b3d15171560b89ede2/src/main.ts#L37-L38) You might want to make the default value for this `''`. Then check if the setting is configured before reading anything. Inbox.md is pretty generic so it might be unexpected for this plugin to read it by default.

PR #1180, opened 2022-09-20
Add Week Planner plugin

rwirdemann:
## Repo URL
Link to my plugin: https://github.com/rwirdemann/obsidian-week-planner

## Release Checklist
    - [ ] Windows
    - [x] macOS
    - [ ] Linux
    - [ ] Android (if applicable)
    - [ ] iOS (if applicable)
  - [x] main.js
  - [x] manifest.json
  - [x] styles.css (optional)
- [x]  I have added a license in the LICENSE file.


liamcain:
- [folling](https://github.com/rwirdemann/obsidian-week-planner/blob/3557688da0b96acf0ac79652b67484c42aac93fe/README.md#L72) typo
- [console.log('Secret: ' + value);](https://github.com/rwirdemann/obsidian-week-planner/blob/3557688da0b96acf0ac79652b67484c42aac93fe/main.ts#L188-L189) remove this
- [let filecontent = await (await this.getFileContents())](https://github.com/rwirdemann/obsidian-week-planner/blob/3557688da0b96acf0ac79652b67484c42aac93fe/src/file.ts#L38-L39) No need to `await` this twice.
- [try](https://github.com/rwirdemann/obsidian-week-planner/blob/3557688da0b96acf0ac79652b67484c42aac93fe/src/file.ts#L48) I don't think you really need to wrap this in a try/catch. Just call `read` directly instead.
- [deleteLineAt](https://github.com/rwirdemann/obsidian-week-planner/blob/3557688da0b96acf0ac79652b67484c42aac93fe/src/file.ts#L19) All of your utility functions modify the file and use `vault.modify`. Since you have an instance of the Editor (passed in by the editor callback), I highly recommend changing these functions to use the editor to make file modifications when possible. For example, using [replaceRange](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L866). Modifying the file via the Editor instead of `modify` allows you to preserve the file state, so any folded content will stay in tact, and the change will get added to the undo stack, so the user can undo the change if they want.
- [let todo = await source.getLineAt(line)](https://github.com/rwirdemann/obsidian-week-planner/blob/3557688da0b96acf0ac79652b67484c42aac93fe/main.ts#L140-L141) same as the issue I just mentioned, but you can get the current line with [`getLine`](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L847) and it's a lot more efficient since you don't need to read the file from disk and split it into newlines. The lines are already buffered for you so it will be a lot faster
- [console.log('allowed days:' + allowedDays)](https://github.com/rwirdemann/obsidian-week-planner/blob/3557688da0b96acf0ac79652b67484c42aac93fe/src/file.ts#L194-L195) remove this
- [map](https://github.com/rwirdemann/obsidian-week-planner/blob/3557688da0b96acf0ac79652b67484c42aac93fe/src/file.ts#L208) Just a comment about coding best practices, but I would suggest you don't call this function `map`. It's not really descriptive enough and introduces confusion since it doesn't work like a `map()` function would.
- [getTomorrowDate](https://github.com/rwirdemann/obsidian-week-planner/blob/3557688da0b96acf0ac79652b67484c42aac93fe/src/file.ts#L170) Obsidian exports `moment` (momentJS) which has a lot of date utilities, so you might make this code a bit simpler.
- [return this.fullFileName.endsWith('Inbox.md')](https://github.com/rwirdemann/obsidian-week-planner/blob/3557688da0b96acf0ac79652b67484c42aac93fe/src/file.ts#L81-L82) This can have false positives if the user has a file like `TestingInbox.md`. You should instead just look at the TFile.basename.

rwirdemann:
Hi,

thanks for the review and your comments. How do I submit my changes after they've been fixed? Do I have to create a new release and a new pull request?

Kind regards,
Ralf

rwirdemann:
Hi @liamcain,

I did the changes. Could you take another look.

Many thanks,
Ralf

PR #1185, opened 2022-09-22
Add Micro.publish plugin to publish posts to Micro.blog

otaviocc:

Obsidian is the *source of truth* for most of my writings and I though it would be interesting to post to [Micro.blog](https://micro.blog/) directly from it. In the past I explored several applications such as Ulysses, iA Writer, and many others which allow publishing to Micro.blog, but they don't work for me, they don't reflect the way I think and take notes. Obsidian does.

Since the Obsidian Plugin architecture is OOP, I decided to go with OOP using MVVM, without third-party dependencies to avoid taking unnecessary complexity into Obsidian. I've added documentation to types and interface methods in the code, so that anyone can easily follow the technical decisions I made when writing the plugin.

Micro.publish has the features I need. It allows

- setting default tags
- setting the default post visibility
- setting a default blog (for those with more than one M.b blog)

And before publishing, it's possible to override these default settings for the post being published.

More features are planned, and will be coming soon.

It's worth mentioning it's getting [positive reactions](https://micro.blog/adammedcalf/13438127) on Micro.blog.

## Repo URL

Link to my plugin: https://github.com/otaviocc/obsidian-microblog

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


otaviocc:
Hi there! I've pushed another commit to this PR, changing the plugin name to `Micro.publish`. I wrote about it on [my blog](https://otavio.cc/2022/09/26/micropublish.html). Manton, Micro.blog's creator, [mentioned it here](https://www.manton.org/2022/09/26/if-you-use.html). Thanks!

liamcain:
- [editorCallback](https://github.com/otaviocc/obsidian-microblog/blob/2a04d3a32e30453e116a68bbc4d7e0a59930aad0/src/MicroPlugin.ts#L19) This command might actually be better suited as a regular `callback`. `editorCallback` commands will only be available if the file is in 'edit mode.'
- [console.log('Login in')](https://github.com/otaviocc/obsidian-microblog/blob/2a04d3a32e30453e116a68bbc4d7e0a59930aad0/src/views/MicroPluginSettingsViewModel.ts#L118) typo

PR #1186, opened 2022-09-23
Add "HTML Reader" plugin

nuthrash:

## Repo URL

Link to my plugin: https://github.com/nuthrash/obsidian-html-plugin

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [innerHTML](https://github.com/nuthrash/obsidian-html-plugin/blob/d918366da66288cd52c94b29540e8e19f7aa6edc/src/HtmlView.ts#L100) - don't do this. [innerHTML is a potential security risk](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml). I think you can get away without using innerHTML and just add the text. If you want some of the text to be bold, just add `noticeEl.createEl('strong', { text: 'Some bold text' })
- [import DOMPurify from 'dompurify';](https://github.com/nuthrash/obsidian-html-plugin/blob/d918366da66288cd52c94b29540e8e19f7aa6edc/src/HtmlView.ts#L2) Obsidian actually uses DOMpurify under the hood. To sanitize html, you can use [`sanitizeHTML`](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L2979) from the Obsidian API. It uses DOMpurify under the hood. If that doesn't work for you, you can also access DOMpurify from `window.DOMpurify`. That way you don't need to bundle your own version of the package with your plugin.
- [console.log](https://github.com/nuthrash/obsidian-html-plugin/blob/d918366da66288cd52c94b29540e8e19f7aa6edc/src/HtmlPlugin.ts#L23) You might want to show the user a Notice when this happens, otherwise they won't know why the plugin doesn't work.
- [this.addSettingTab(new HtmlSettingTab(this.app, this));](https://github.com/nuthrash/obsidian-html-plugin/blob/d918366da66288cd52c94b29540e8e19f7aa6edc/src/HtmlPlugin.ts#L26-L27) AFAICT, you don't use the settings for anything. I suggest not registering the settings page until you have something to include on it.

nuthrash:
> * [import DOMPurify from 'dompurify';](https://github.com/nuthrash/obsidian-html-plugin/blob/d918366da66288cd52c94b29540e8e19f7aa6edc/src/HtmlView.ts#L2) Obsidian actually uses DOMpurify under the hood. To sanitize html, you can use [`sanitizeHTML`](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L2979) from the Obsidian API. It uses DOMpurify under the hood. If that doesn't work for you, you can also access DOMpurify from `window.DOMpurify`. That way you don't need to bundle your own version of the package with your plugin.

I got a problem.  When I invoke `window.DOMpurify.sanitize()` in main.js, Obsidian would throw an exception `Cannot read properties of undefined(read 'sanitize')`. But `window.DOMPurify.addHook()` is worked. On the other hand, it is strange that I can invoke `window.DOMPurify.sanitize()` at Developer Console.
Is `window.DOMPurify.sanitize()` a private function?

By the way, the  [`sanitizeHTML`](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L2979) API sanitize out too much HTML elements, it is not very suitable for user experience. I need a customizable solution to adjust some visual effects for suitable user experience.

nuthrash:
> I got a problem. When I invoke `window.DOMpurify.sanitize()` in main.js, Obsidian would throw an exception `Cannot read properties of undefined(read 'sanitize')`. But `window.DOMPurify.addHook()` is worked. On the other hand, it is strange that I can invoke `window.DOMPurify.sanitize()` at Developer Console. Is `window.DOMPurify.sanitize()` a private function?
> 
> By the way, the [`sanitizeHTML`](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L2979) API sanitize out too much HTML elements, it is not very suitable for user experience. I need a customizable solution to adjust some visual effects for suitable user experience.

My workaround code is
```typescript
window.DOMPurify.clearConfig();
window.DOMPurify.setConfig( purifyConfig );
const cleanDom = sanitizeHTMLToDom( contents );
this.contentEl.appendChild( cleanDom );
window.DOMPurify.clearConfig();
window.DOMPurify.removeHook( 'afterSanitizeAttributes' );
```
It works. However, I found another problem.
Some internal links like `#cite_1` would be replaced as `app://obsidian.md/index.html#cite_1`. When clicking this modified link, the Obsidian would ask if user want to open it. It seems that is not correct procedure for internal links.
I think keeping origin internal links is a better solution, especially when there are many cites in HTML files.
Maybe there is a way to avoid this problem?

nuthrash:
Hello,
I have modify code to this commit [Modify code to satisfy Obsidian's review](https://github.com/nuthrash/obsidian-html-plugin/commit/90b8b4b0d6206f7081a7f0ec2afd1337abe685e9), please review it.

> * [innerHTML](https://github.com/nuthrash/obsidian-html-plugin/blob/d918366da66288cd52c94b29540e8e19f7aa6edc/src/HtmlView.ts#L100) - don't do this. [innerHTML is a potential security risk](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml). I think you can get away without using innerHTML and just add the text. If you want some of the text to be bold, just add `noticeEl.createEl('strong', { text: 'Some bold text' })

Done. Change `innerHTML` to `noticeEl.createEl()` of `showError()`.

> * [import DOMPurify from 'dompurify';](https://github.com/nuthrash/obsidian-html-plugin/blob/d918366da66288cd52c94b29540e8e19f7aa6edc/src/HtmlView.ts#L2) Obsidian actually uses DOMpurify under the hood. To sanitize html, you can use [`sanitizeHTML`](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L2979) from the Obsidian API. It uses DOMpurify under the hood. If that doesn't work for you, you can also access DOMpurify from `window.DOMpurify`. That way you don't need to bundle your own version of the package with your plugin.

No change. I have evaluated Obsidian's internal DOMPurify module, and I decided to keep original code.
I found a problem, which is described in [this reply](https://github.com/obsidianmd/obsidian-releases/pull/1186#issuecomment-1276457775), and I cannot find any solution of this problem. I prefer keeping origin internal links is a better solution, especially when there are many cites in HTML files.

> * [console.log](https://github.com/nuthrash/obsidian-html-plugin/blob/d918366da66288cd52c94b29540e8e19f7aa6edc/src/HtmlPlugin.ts#L23) You might want to show the user a Notice when this happens, otherwise they won't know why the plugin doesn't work.

Done. Change `console.log()` to `showError()`.

> * [this.addSettingTab(new HtmlSettingTab(this.app, this));](https://github.com/nuthrash/obsidian-html-plugin/blob/d918366da66288cd52c94b29540e8e19f7aa6edc/src/HtmlPlugin.ts#L26-L27) AFAICT, you don't use the settings for anything. I suggest not registering the settings page until you have something to include on it.

Done. Remove HtmlPluginSettings.ts and related code.

liamcain:
Sorry for not replying to the earlier question, we were a bit wrapped up with the 1.0 launch so I'm very behind on plugin review!

Yeah, we don't expose much customization with DOM sanitization, so if it's not configurable enough, I agree it's better to just provide your own instance of DOMpurify. I think what you have currently looks good.

PR #1190, opened 2022-09-25
add Onyx Boox Extractor plugin

akosbalasko:

## Repo URL

Link to my plugin: https://github.com/akosbalasko/Onyx-Boox-Annotation-Highlight-Extractor

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [file-menu](https://github.com/akosbalasko/Onyx-Boox-Annotation-Highlight-Extractor/blob/59d1a81e88c3ece7d039f6a0c7c1aab566133474/main.ts#L12) Not sure if it's better than the current workflow, but one idea you might want to consider: instead of having users save the exported file as `.md` you could have them save the files as .txt. Then only add this file item to `.txt` files. For users that don't have "Detect all file extensions" enabled, you could register the .txt file extension (see: [txt as md](https://github.com/deathau/txt-as-md-obsidian/blob/main/src/main.ts))

PR #1191, opened 2022-09-25
Add Theme Toggler plugin

larsmagnus:

## Repo URL

Link to my plugin: https://github.com/larsmagnus/obsidian-theme-toggler

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -4831,5 +4831,13 @@
         "description": "Adds a button to the search view for copying the Obsidian search URL.",
         "repo": "czottmann/obsidian-copy-search-url",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-theme-toggler",
+        "name": "Theme Toggler",
+        "author": "larsmagnus",
+        "description": "Toggle the theme in Obsidian's panels.",
+        "repo": "larsmagnus/obsidian-theme-toggler",
+        "branch": "main"
     }
 ]


liamcain:
- [buttonSettings: {](https://github.com/larsmagnus/obsidian-theme-toggler/blob/8fb14c29a90decac65f49bc84cf4eed126241ac3/src/constants.ts#L15-L19) this doesn't belong in the settings. Settings should only include data that's persisted across restarts.
- [await this.saveSettings(DEFAULT_SETTINGS)](https://github.com/larsmagnus/obsidian-theme-toggler/blob/8fb14c29a90decac65f49bc84cf4eed126241ac3/src/main.ts#L195)  I don't think you want to include this in onunload. If you don't want to persist the leaf configurations, I suggest you store them in an instance variable rather than settings.
- [getAppTheme](https://github.com/larsmagnus/obsidian-theme-toggler/blob/8fb14c29a90decac65f49bc84cf4eed126241ac3/src/main.ts#L167) This function call isn't doing anything
- [this.app.commands.executeCommandById(`${pluginId}:${id}`)](https://github.com/larsmagnus/obsidian-theme-toggler/blob/8fb14c29a90decac65f49bc84cf4eed126241ac3/src/main.ts#L41-L42) This doesn't look right. `id` here is coming from buttonSettings.id, right? This approach seems more complicated than it needs to be. Why not just have the function call `toggleTheme`? Having it call `executeCommandById` seems unnecessary.
- [mousedown](https://github.com/larsmagnus/obsidian-theme-toggler/blob/8fb14c29a90decac65f49bc84cf4eed126241ac3/src/main.ts#L38) I think this should be `'click'` instead of mousedown. You also don't need to use `registerDomEvent` here, you can just use `addEventListener('click')` here. When the plugin unloads, you are removing the button anyway.

larsmagnus:
Thanks for the thorough review @liamcain! 🙌

I've addressed all your points in the 1.0.1 release of the plugin. Let me know if there are any other steps I need to make as part of the review process, or if there's anything else that needs fixing.

PR #1203, opened 2022-10-02
add graph sync plugin

Xallt:

## Repo URL

Link to my plugin: https://github.com/Xallt/sync-graph-settings

## Release Checklist
- [x] have tested the plugin on
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
Please remove the `v` from the release name & tag, that will break Obsidian(and the script)

[path.join(configDir, 'graph.json');](https://github.com/Xallt/sync-graph-settings/blob/156bfd9dd030dcad1348f09d342f1b93bcd6918b/main.ts#L15)
I would recommend using the [`normalizePath`](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L2446-L2449) function to normalize a concatenated string.
That will also make the plugin compatible with mobile.

ericaxu:
@Xallt Hi! Do you still intend to submit this plugin? Thanks!

Xallt:
@ericaxu I fixed the issues with the plugin you pointed out. Can we resume the reviewing process?

liamcain:
- [app](https://github.com/Xallt/sync-graph-settings/blob/06375883b62800e1f2b59fb42440d548b3d1f125/main.ts#L15) this is referencing app as a global variable. You should be using `this.app` instead.
- [var viewState](https://github.com/Xallt/sync-graph-settings/blob/06375883b62800e1f2b59fb42440d548b3d1f125/main.ts#L42) - I recommend you change all instances of `var` to either `const` or `let`. `var` has function-level scope, so it can easily lead to bugs if you're not careful. Here's a pretty good guide with examples of why not to use `var`: [why `var` is obsolete](https://javascript.plainenglish.io/4-reasons-why-var-is-considered-obsolete-in-modern-javascript-a30296b5f08f)
- [import {readFile} from "fs/promises";](https://github.com/Xallt/sync-graph-settings/blob/06375883b62800e1f2b59fb42440d548b3d1f125/main.ts#L3-L4) you should remove this line. This is a nodeJS package that's not available on mobile. So even if you're not using this import, it will prevent your plugin from working on mobile.

PR #1204, opened 2022-10-02
adding Obsidian MtG plugin

omardelarosa:

## Repo URL

Link to my plugin: [omardelarosa/obsidian-mtg](https://github.com/omardelarosa/obsidian-mtg)

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
Please remove Obsidian from the plugin id and name, its unnecessary.

["csv": "^6.2.0"](https://github.com/omardelarosa/obsidian-mtg/blob/6103bf0991d14e3e14dfdf7bc67a96ae36e8428f/package.json#L39)
Did you test that this works on mobile?
From what I can see this package will crash Obsidian on mobile as it depends on NodeJS API's.

[errorNode.innerHTML = error;](https://github.com/omardelarosa/obsidian-mtg/blob/6103bf0991d14e3e14dfdf7bc67a96ae36e8428f/main.ts#L60)
Avoid `innerHTML´: <https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml>
In general, I would move the error handling inside ofthe catch block.

[.mtg.collection.csv](https://github.com/omardelarosa/obsidian-mtg/blob/6103bf0991d14e3e14dfdf7bc67a96ae36e8428f/main.ts#L101)
I would recommend a different placeholder, as files starting with `.` are always hidden.

[file.name.endsWith(settings.collection.fileExtension);](https://github.com/omardelarosa/obsidian-mtg/blob/6103bf0991d14e3e14dfdf7bc67a96ae36e8428f/src/collection.ts#L29)
A `TFile` already has a `extension` parameter.

[const counts = document.createElement('span');](https://github.com/omardelarosa/obsidian-mtg/blob/6103bf0991d14e3e14dfdf7bc67a96ae36e8428f/src/renderer.ts#L314-L323)
Just as a little trick, no need to implement it, you can also do:
```ts
const xyz = otherEl.createSpan({cls: 'mtg....', text: 'hello world'});
```
This will automatically append it as a child to `otherEl`.

[httpsRequest(options, (response: any) => {](https://github.com/omardelarosa/obsidian-mtg/blob/6103bf0991d14e3e14dfdf7bc67a96ae36e8428f/src/scryfall.ts#L177)
I would recommend using `requestUrl` instead:
<https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L2894-L2898>

[hostname: 'api.scryfall.com',](https://github.com/omardelarosa/obsidian-mtg/blob/6103bf0991d14e3e14dfdf7bc67a96ae36e8428f/src/scryfall.ts#L218)
Any reason you are not using the variable you defined at the start of this file?

I also think that the `obsidian-plugin-` prefix for your CSS classes is unnecessary.

omardelarosa:
@joethei thanks for the review and comments!  It would be ideal for this to work on mobile so I'm gonna try to just replace the CSV parsing logic and drop the dependency on `csv` altogether.  I'll push up a new version shortly with that and some other fixes based on your feedback (like using that `createSpan()` method)

omardelarosa:
@joethei I made all the changes request in the [`1.0.2` release](https://github.com/omardelarosa/obsidian-mtg/releases/tag/1.0.2).

omardelarosa:
@joethei forgot to update the http request library to use `obsidian.requestUrl` in that last version, but I just got it in [the `1.0.3` release](https://github.com/omardelarosa/obsidian-mtg/releases/tag/1.0.3)

liamcain:
- [syncCounts](https://github.com/omardelarosa/obsidian-mtg/blob/ed3ee5a98abf3c2c57fa37782b35065a58e0e071/main.ts#L41) This shouldn't be in the onload function. The vault is not guaranteed to be initialized before your plugin. Instead, you should wrap this in [`onLayoutReady`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3626), which be called once the workspace has finished initializing.
- [syncCounts](https://github.com/omardelarosa/obsidian-mtg/blob/ed3ee5a98abf3c2c57fa37782b35065a58e0e071/main.ts#L55) `syncCounts` seems pretty expensive, does it need to get called every time _each_ markdown block gets rendered?

omardelarosa:
@liamcain it only needs to be called if the collection CSV changes.  Is there a way to attack a change callback to a specific file (or files) in the vault?  I think I only need to call `syncCounts` then (i.e. when the collection CSVs change) and once on load.

liamcain:
>  Is there a way to attack a change callback to a specific file (or files) in the vault? 

You can listen for the Vault's [`modify`](https://github.com/obsidianmd/obsidian-api/blob/009a10f196df551256a5894ea60fdf9ddaab19a1/obsidian.d.ts#L3706) event and just look at the file passed into the callback, maybe just check if the `file.name` matches your csv extension criteria.

omardelarosa:
@liamcain I just opened a PR on the repo using the `modify` event to reduce the number of times syncCounts is called:

https://github.com/omardelarosa/obsidian-mtg/pull/16

There's [a one-time call to `syncCounts`](https://github.com/omardelarosa/obsidian-mtg/pull/16/files#diff-564c860307257bfac9176f554a35aa858c350a59d75c0dcd0a5d63a480805bd0R63) on the first render when `this.cardCounts` is undefined, since it seems that even with the `onLayoutReady` callback if I open Obsidian directly on a decklist that first render has `undefined` for `this.cardCounts`. 

Do you know if there's another event I can observe that would get called after the vault is ready, but before the `registerMarkdownCodeBlockProcessor`'s first callback invocation?

liamcain:
> Do you know if there's another event I can observe that would get called after the vault is ready, but before the registerMarkdownCodeBlockProcessor's first callback invocation?

Sorry, I don't think an event for this exists.

liamcain:
- [csv](https://github.com/omardelarosa/obsidian-mtg/blob/01b4b82f95b59d913ec731a61766df65db3de056/main.ts#L42) I think you should include the `.` here. So `endsWith('.csv')` to avoid false positives

Otherwise, I think what you have works fine. Nice job!

PR #1206, opened 2022-10-03
Add Readavocado Sync Plugin

innneang:

## Repo URL

Link to my plugin: [https://github.com/innneang/obsidian-readavocado-sync](https://github.com/innneang/obsidian-readavocado-sync)

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [if(!((Date.now() - this.settings.lastSyncTime) / 1000 / 60 > this.settings.syncInterval)) {](https://github.com/innneang/obsidian-readavocado-sync/blob/d248c2475bf08ec8e799079c506d13ee54171373/main.ts#L64-L65) You probably want to store the value of this expression so that when you print it out, you can print out the same value that you used in your comparison.
- [^a-zA-Z ]](https://github.com/innneang/obsidian-readavocado-sync/blob/d248c2475bf08ec8e799079c506d13ee54171373/main.ts#L97) you probably also want to support numbers, right?
- [includes("http")](https://github.com/innneang/obsidian-readavocado-sync/blob/d248c2475bf08ec8e799079c506d13ee54171373/main.ts#L100) this should be `startsWith`. And even better, you probably want to use a regex so that you can make sure it's `http://` or `https://` and not just a but called `http`.

PR #1210, opened 2022-10-09
Add Projects plugin

marcusolsson:

## Repo URL

Link to my plugin: https://github.com/marcusolsson/obsidian-projects

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [ignore](https://github.com/marcusolsson/obsidian-projects/blob/38e09baa96456bf685199b84ca357218a282b9ce/src/view.ts#L12) What's the tsignore here for?
- [onRegisterProjectView](https://github.com/marcusolsson/obsidian-projects/blob/38e09baa96456bf685199b84ca357218a282b9ce/src/view.ts#L77) Instead of doing it this way, why not emit an event for plugins to listen to and register their views (similar to how Obsidian's file-menu works)?

marcusolsson:
> What's the tsignore here for?

`Property 'component' has no initializer and is not definitely assigned in the constructor.ts(2564)`

I have pretty strict linter settings, so it might not show up for others. Though I just realized I could write this as:

```ts
component: App | undefined;
```

That way I'll at least get rid of the ignores.

> why not emit an event for plugins to listen

The API for registering custom views is still wildly experimental, and not something I'm advertising. I've been prioritizing a stable 1.0, but this will be the next major feature. More information in https://github.com/marcusolsson/obsidian-projects/issues/27. 

Thanks for reviewing this! 🙏

PR #1211, opened 2022-10-09
add obsidian-ava

louis030195:

## Repo URL

Link to my plugin: https://github.com/louis030195/obsidian-ava

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


community-plugins.json
```diff
@@ -4854,5 +4854,13 @@
         "description": "Plugin to automatically update to system theme.",
         "repo": "jgauth/obsidian-system-theme",
         "branch": "master"
+    },
+    {
+        "id": "obsidian-ava",
```
liamcain, line 4859:
```suggestion
        "id": "ava",
```

liamcain:
- ["id": "obsidian-ava",](https://github.com/louis030195/obsidian-ava/blob/5481c211b646f34f0b333673f7483ed183d0fb6d/manifest.json#L2-L3) No need to include `obsidian` in the plugin ID
- [const suggest = new AvaSuggest(this.app, this, 1000, 3);](https://github.com/louis030195/obsidian-ava/blob/5481c211b646f34f0b333673f7483ed183d0fb6d/main.tsx#L36-L37) the `onload` function is called synchronously and is therefore blocking. So anything expensive should be deferred in either an IIFE or (if it relies on the workspace being loaded) inside a [`onLayoutReady`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3626) callback.
- [hotkeys: [{modifiers: ["Shift"], key: "tab"}],](https://github.com/louis030195/obsidian-ava/blob/5481c211b646f34f0b333673f7483ed183d0fb6d/main.tsx#L46-L47) we usually recommend not adding a default hotkey to avoid it interfering with other plugins or user preferences. Instead you can just leave instructions in the README of a "preferred" hotkey for users to configure themselves.
- ["isDesktopOnly": false](https://github.com/louis030195/obsidian-ava/blob/5481c211b646f34f0b333673f7483ed183d0fb6d/manifest.json#L9-L10) I think you need to set "isDesktopOnly" to true. You are importing `node` libs that are only available in the desktop (Electron) environment.

PR #1221, opened 2022-10-14
Add Checkbox 3 states plugin 

hrenaud:

## Repo URL

Link to my plugin:
https://github.com/hrenaud/obsidian-checkbox3states-plugin

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [SampleSettingTab](https://github.com/hrenaud/obsidian-checkbox3states-plugin/blob/083467dff488a90c3b54686067995a7c3d6c799b/main.ts#L189) nit: you should rename your class here.
- [Object](https://github.com/hrenaud/obsidian-checkbox3states-plugin/blob/083467dff488a90c3b54686067995a7c3d6c799b/main.ts#L49) Instead of casting to an object like this, it's preferable to cast to the type that you need, in this case `HTMLElement`.
- [touchstart](https://github.com/hrenaud/obsidian-checkbox3states-plugin/blob/083467dff488a90c3b54686067995a7c3d6c799b/main.ts#L62) Instead of configuring your own `longtouch` for mobile, I think you can listen for the "contextmenu" event on mobile. Obsidian fires that event after the user does a longpress.
- [loadSettings](https://github.com/hrenaud/obsidian-checkbox3states-plugin/blob/083467dff488a90c3b54686067995a7c3d6c799b/main.ts#L281) Instead of calling loadSettings here, you can just assign the new value to your settings object. loadSettings requires reading from disk, so it can be a bit more expensive

PR #1241, opened 2022-10-19
Add obsidian-things3-sync

royxue:

## Repo URL

Link to my plugin: https://github.com/royxue/obsidian-things3-sync

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [Statue](https://github.com/royxue/obsidian-things3-sync/blob/d37ddaa0c52ac68a937903ed2ffcbd3dc48a9ecf/main.ts#L226) I think this is supposed to say "status"
- [false](https://github.com/royxue/obsidian-things3-sync/blob/d37ddaa0c52ac68a937903ed2ffcbd3dc48a9ecf/manifest.json#L9) I'm not sure this works on mobile, I think you might need to confirm it. If it doesn't work, I suggest changing this flag to false and creating a new Github Release.

royxue:
> * [Statue](https://github.com/royxue/obsidian-things3-sync/blob/d37ddaa0c52ac68a937903ed2ffcbd3dc48a9ecf/main.ts#L226) I think this is supposed to say "status"
> * [false](https://github.com/royxue/obsidian-things3-sync/blob/d37ddaa0c52ac68a937903ed2ffcbd3dc48a9ecf/manifest.json#L9) I'm not sure this works on mobile, I think you might need to confirm it. If it doesn't work, I suggest changing this flag to false and creating a new Github Release.

* I will fix the typo.
* I tested on iPhone, it could work well, so I will keep the current flag

PR #1246, opened 2022-10-20
Add relation pane plugin

mottox2:

## Repo URL

Link to my plugin: https://github.com/mottox2/obsidian-relation-pane

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [const container = this.containerEl.children[1]](https://github.com/mottox2/obsidian-relation-pane/blob/22afaccde4d4c523619c28e7462937258fc4d5e1/src/RelationView.tsx#L91-L92) you should be using `this.contentEl` instead. It should be the same element.
- [backLinks](https://github.com/mottox2/obsidian-relation-pane/blob/22afaccde4d4c523619c28e7462937258fc4d5e1/src/RelationView.tsx#L105) you're not using your own getBacklinks function here
- Instead of passing in `metaKey` here, it's highly recommended that you use [`Keymap.isModEvent`](https://github.com/obsidianmd/obsidian-api/blob/009a10f196df551256a5894ea60fdf9ddaab19a1/obsidian.d.ts#L1639) and pass the result into `getLeaf`. That will keep the behavior consistent with clicking links, and it will support opening in new tabs, new splits, or new windows, based on the modifiers pressed.
- [<div className="search-empty-state">No backlinks found.</div>](https://github.com/mottox2/obsidian-relation-pane/blob/22afaccde4d4c523619c28e7462937258fc4d5e1/src/RelationView.tsx#L179-L180) You might want to add a similar empty state if there is no active file

mottox2:
@liamcain 
Thanks for the review. Sorry for the delay.
I have corrected the points you pointed out.

PR #1252, opened 2022-10-23
Add Daily Note Outline plugin

iiz00:

## Repo URL

Link to my plugin: https://github.com/iiz00/obsidian-daily-note-outline

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [debouncer](https://github.com/iiz00/obsidian-daily-note-outline/blob/4291ce9b7ec214337cb8a4322f01914ffa309279/src/view.ts#L65) I would suggest giving this a more meaningful name, such as `requestRefresh` ("request" implying that the function is debounced)
- [while ( i < arrayAll.length){](https://github.com/iiz00/obsidian-daily-note-outline/blob/4291ce9b7ec214337cb8a4322f01914ffa309279/src/view.ts#L178-L179) this can be a lot more efficient. I suggest you instead just create a new date equal to the earlier, then iterate up to latest. Then just check for

```ts
allDailyNotes[`day-${date.format()`]
```

iiz00:
Thank you for your review!
I see, it seems that if I can create keys for allDailyNotes from the date, arrayAll itself would be unnecessary. I will try to fix it.

PR #1255, opened 2022-10-24
Adding Custom File Extensions Plugin

MeepTech:
## Repo URL
Link to my plugin: https://github.com/MeepTech/obsidian-custom-file-extensions-plugin

## Release Checklist
    * [x]   Windows
    * [x]   macOS
    * [x]   Linux
    * [ ]   Android _(if applicable)_
    * [ ]   iOS _(if applicable)_
  * [x]  My GitHub release contains all required files
    * [x]  `main.js`
    * [x]  `manifest.json`
    * [ ]  `styles.css` _(optional)_
  * [x]  I have read the tips in https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md and have self-reviewed my plugin to avoid these common pitfalls.
  * [x]  My project respects and is compatible with the original license of any code from other plugins that I'm using.  I have given proper attribution to these other projects in my `README.md`.


liamcain:
- [parse](https://github.com/MeepTech/obsidian-custom-file-extensions-plugin/blob/6d1f627770f770b106c1ed495670f7232346be07/main.ts#L41) I think you should wrap this in a try/catch here as well.
- [currentValueIsInvalidJson](https://github.com/MeepTech/obsidian-custom-file-extensions-plugin/blob/6d1f627770f770b106c1ed495670f7232346be07/main.ts#L12) this is unused
- [revert](https://github.com/MeepTech/obsidian-custom-file-extensions-plugin/blob/6d1f627770f770b106c1ed495670f7232346be07/main.ts#L49) I think revert is misnamed and maybe doing the incorrect thing here. Reverting would be unregistering the current view types and registering the previous ones. But this appears to only _unregister_ the previous ones. I think instead of the current approach, you're better off looping through the currently saved extensions within your onChange, and if any extensions aren't in `value` then unregister them. I don't think you need the `revert` function then. If you want to unregister your custom views in onload, then you should be unregistering the `additionalFileTypes` instead of `previousAdditionalFileTypes`.

liamcain:
Hi @MeepTech, sorry for taking so long to re-review! Your reply got buried in my notification feed. Everything looks good. There's just a small validation error.

You changed the ID in the community-plugins.json file but not in your manifest.json file here: <https://github.com/MeepTech/obsidian-custom-file-extensions-plugin/blob/544f9262c6559dc5ca964d92a9b51dbaa4d8057a/manifest.json#L2> you'll need to update that locally in your repo and push a new version of the plugin to Github Releases. Once that's done, just @ me here and I'll merge it in!

MeepTech:
> Hi @MeepTech, sorry for taking so long to re-review! Your reply got buried in my notification feed 😢. Everything looks good. There's just a small validation error.
> 
> You changed the ID in the community-plugins.json file but not in your manifest.json file here: https://github.com/MeepTech/obsidian-custom-file-extensions-plugin/blob/544f9262c6559dc5ca964d92a9b51dbaa4d8057a/manifest.json#L2 you'll need to update that locally in your repo and push a new version of the plugin to Github Releases. Once that's done, just @ me here and I'll merge it in!

@liamcain I think i fixed everything you've asked:
https://github.com/MeepTech/obsidian-custom-file-extensions-plugin/releases/tag/1.0.1

Thanks so much!

drventure:
Downloaded and giving it a spin. So far, this is exactly what I was looking for in terms of handling plain text files.
In particular, any kind of script (.js, .ps, .bat, etc) that would result in +EXECUTING+ the file when you clicked on it in Obsidian (which was +really+ bad).

Thanks for this plugin!

r-erd:
Great work! One question:
Is there any way to add functionality so that e.g. yaml files are displayed as raw text? 
I know that there is source mode, but that still results in huge headings when I have comments in my code.

PR #1256, opened 2022-10-26
Sort search results by number of matches

BradleyYeo:
Screenshot 2022-10-26 at 3 39 22 PM
Can Search results be sorted by the number of matches it has?


joethei:
Please open feature requests on our forum: <https://forum.obsidian.md/>

PR #1267, opened 2022-10-29
update community-plugins.json

kokoromati:

## Repo URL

Link to my plugin:
https://github.com/kokoromati/obsidian-gt-quote

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5040,5 +5040,12 @@
         "author": "Cyrus Zhang",
         "description": "Sync your Readavocado highlights with Obsidian",
         "repo": "innneang/obsidian-readavocado-sync"
+    },
+    {
+        "id": "obsidian-gt-quote",
+        "name": "Obsidian GT Quote",
+        "description": "This is a plugin to help with quote(>) operations.",
+        "author": "kokoromati",
+        "repo": "kokoromati/obsidian-gt-quote"
     }
 ]


liamcain:
- [clipboard.readText](https://github.com/kokoromati/obsidian-gt-quote/blob/d365c2bdd692daaace0dbe492883c843281e89f4/main.ts#L24) No need to access electron remote for this, you can just use navigator.clipboard.readText(). On desktop, we patch that function to use electron's readText anyway, so it will do the same thing (and your plugin will have the advantage of working on mobile)
- [var](https://github.com/kokoromati/obsidian-gt-quote/blob/d365c2bdd692daaace0dbe492883c843281e89f4/main.ts#L27) avoid using var if you can. You can switch this to `let`.
- I think you could probably simplify this logic a bit by using `.join('\n')`. Maybe something like:

```ts
text.trim().split(/\r\n|\n/).map(line => '> ' + line).join('\n');
```
- [quoteString](https://github.com/kokoromati/obsidian-gt-quote/blob/d365c2bdd692daaace0dbe492883c843281e89f4/main.ts#L8) I think more commonly, blockquotes start with `> ` on each line (note the space). I don't know if it's always correctly parsed if you omit the space.
- [Setting #1](https://github.com/kokoromati/obsidian-gt-quote/blob/d365c2bdd692daaace0dbe492883c843281e89f4/main.ts#L104) Calling the setting "setting 1" is not very useful or descriptive. I would suggest making the name "quote format" and the description should be something more useful. I also think you are fine to exclude the Settings tab entire and keep it simple. Just hard coding the quote character to "> " seems fine to me.

joethei:
Hi @kokoromati ,

liamcain:
Hi @kokoromati, it's been a while since we've heard from you. I'm going to close this pull request to make room for other plugins in the queue. But feel free to resubmit the plugin whenever you are ready for us to take another look. Thanks! 🙏

PR #1270, opened 2022-10-30
Update community-plugins.json

TfTHacker:

## Repo URL

Link to my plugin: https://github.com/TfTHacker/obsidian42-strange-new-worlds/

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5057,5 +5057,12 @@
         "author": "Marcus Olsson",
         "repo": "marcusolsson/obsidian-projects",
         "branch": "main"
-    }
+    }, 
+    {
+        "id": "obsidian42-strange-new-worlds",
+        "name": "Obsidian42 - Strange New Worlds",
+        "description": "Revealing networked thought and the strange new worlds created by your vault.",
+        "author": "TfTHacker",
+        "repo": "TfTHacker/obsidian42-strange-new-worlds"
+    }	
 ]


joethei:
Just a few nitpicks for the start:

[appName = "Obsidian42 - Strange New Worlds";](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/e0b86e7d4e0e6763ca0bd904d2c536b6c2feae22/src/main.ts#L17-L18)
You can use `this.manifest.name´ and `this.manifest.id` here instead.

[if(state===true)](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/e0b86e7d4e0e6763ca0bd904d2c536b6c2feae22/src/main.ts#L128)
This can be shortened to `if(state)´

[getDisplayText() { return "Strange New Worlds"}](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/e0b86e7d4e0e6763ca0bd904d2c536b6c2feae22/src/ui/sidebar-pane.ts#L20)
You already have the same string as the `VIEW_TYPE_SNW` variable.

[snwTitleRefCountDisplayCountEl = document.createElement("div");](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/e0b86e7d4e0e6763ca0bd904d2c536b6c2feae22/src/ui/headerRefCount.ts#L83-L85)
This can be shortened to `wrapper.createDiv({cls: 'snw-header-count'});`
This is possible in a few places.

TfTHacker:
I am very thankful for this feedback. Learned some subtle but useful things. Also, it's a project with a lot of "goo," so thank you for the time you put into it.

I have made all the changes as suggested, and is now a part of the 1.0.6 release in the repo.

liamcain:
## Nitpicks

- [ThePlugin](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/4b14a10daf70e43938fa4f73af69f1d84b04a8e1/src/main.ts#L16) I recommend renaming the plugin name to something more unique (maybe just SNWPlugin)
- [snwPluginActivelyShowingCounts](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/4b14a10daf70e43938fa4f73af69f1d84b04a8e1/src/main.ts#L20) maybe just rename this to something simpler like `showCounts`
- [markdownPostProcessorSNW](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/4b14a10daf70e43938fa4f73af69f1d84b04a8e1/src/main.ts#L26) seems like this could just be `markdownPostProcessor`
- [(file) => indexDebounce()](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/4b14a10daf70e43938fa4f73af69f1d84b04a8e1/src/main.ts#L64) There's no need to add another thunk here. This can just be `this.app.metadataCache.on("resolve", indexDebounce)`
- [this.app.workspace.getRightLeaf(false).setViewState({type: VIEW_TYPE_SNW, active: false});](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/4b14a10daf70e43938fa4f73af69f1d84b04a8e1/src/main.ts#L86) this can just get called inside the onLayoutReady function, I don't think it needs to be in the resolved callback
- [this.app.workspace.rightSplit.expand();](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/4b14a10daf70e43938fa4f73af69f1d84b04a8e1/src/main.ts#L110) I suggest not doing this. revealLeaf will already handle this, and you can't assume the user didn't drag the view into the leftSplit
- [setTimeout](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/4b14a10daf70e43938fa4f73af69f1d84b04a8e1/src/main.ts#L114) you shouldn't need a timeout here
- [if(state===true)](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/4b14a10daf70e43938fa4f73af69f1d84b04a8e1/src/main.ts#L130) I think this function could be more readable if you got rid of `state`. Instead I think you could just say `if (this.settings.displayIncomingFilesheader && this.showingCounts) { on() } else { off() }`
- [state===true](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/4b14a10daf70e43938fa4f73af69f1d84b04a8e1/src/main.ts#L144) same thing as my last comment. I think just combining the checks into the conditional of the `if` statement will make this a lot more readable.
- [e.ctrlKey || e.metaKey) && e.altKey](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/4b14a10daf70e43938fa4f73af69f1d84b04a8e1/src/ui/components/uic-ref--parent.ts#L73) please don't try to configure this yourself. Instead you can use [`Keymap.isModEvent`](https://github.com/obsidianmd/obsidian-api/blob/009a10f196df551256a5894ea60fdf9ddaab19a1/obsidian.d.ts#L1639) and pass the result to `getLeaf`. It will return the correct paneType so you don't need to figure out the modifier keys (Since they also very by OS)
- [400](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/4b14a10daf70e43938fa4f73af69f1d84b04a8e1/src/ui/components/uic-ref--parent.ts#L108) what is this setTimeout for?
- [e.ctrlKey || e.metaKey](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/4b14a10daf70e43938fa4f73af69f1d84b04a8e1/src/ui/components/uic-ref--parent.ts#L118) Instead of checking for ctrl or meta, you can use `isModifier(evt, 'Mod')` which will check the correct modifier key for the OS.
- [obsidian](https://github.com/TfTHacker/obsidian42-strange-new-worlds/blob/4b14a10daf70e43938fa4f73af69f1d84b04a8e1/src/types.ts#L3) since you already have the obsidian type override here, I suggest moving move things into this, so you don't have to keep casting to `any`. It will help in the long run when you're debugging code you wrote months ago and don't remember what it does 😆

TfTHacker:
@liamcain thank you for the great feedback and all the time you put into this review. Not a small effort, and I appreciate it.

I have applied all the changes your recommend. And I learned some useful things along the way. I really appreciate the utility functions provided by the API for common tasks (like how handle mouse events).

You asked about: "400 what is this setTimeout for?" When I call the following function:
`thePlugin.app.workspace.getActiveViewOfType(MarkdownView).setEphemeralState({line: lineNu });` on a large file, the large file takes a bit more time to load in the UI, and thus if I try to scroll right away it doesn't scroll. This brief timeout gives obsidian a little time to get the file loaded before I try to reposition the focus.

Please let me know if there is a better way to handle this.

liamcain:
> You asked about: "400 what is this setTimeout for?" When I call the following function:
thePlugin.app.workspace.getActiveViewOfType(MarkdownView).setEphemeralState({line: lineNu }); on a large file, the large file takes a bit more time to load in the UI, and thus if I try to scroll right away it doesn't scroll. This brief timeout gives obsidian a little time to get the file loaded before I try to reposition the focus

Interesting, okay. I don't think that's the best way to handle it, but I can't think of anything better right now. We might have something that handles that already internally, I'll take a look.

PR #1277, opened 2022-10-30
Nerd of iron add betterfrontmatter plugin

Nerd-of-Iron:

## Repo URL

Link to my plugin:

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5063,7 +5063,6 @@
         "name": "Better Front Matter",
         "description": "JavaScript API for full YAML front matter editing.",
         "author": "Rob Pitt",
-        "repo": "Nerd-of-Iron/obsidian-betterfrontmatter",
-        "branch": "master"
+        "repo": "Nerd-of-Iron/obsidian-betterfrontmatter"
     }	
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5057,5 +5057,13 @@
         "author": "Marcus Olsson",
         "repo": "marcusolsson/obsidian-projects",
         "branch": "main"
-    }
+    },
+    {
+        "id": "obsidian-betterfrontmatter",
+        "name": "Better Front Matter",
+        "description": "JavaScript API for full YAML front matter editing.",
+        "author": "Rob Pitt",
+        "repo": "Nerd-of-Iron/obsidian-betterfrontmatter",
+        "branch": "master"
+    }	
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5059,7 +5059,7 @@
         "branch": "main"
     },
     {
-        "id": "obsidian-betterfrontmatter",
+        "id": "betterfrontmatter",
         "name": "Better Front Matter",
         "description": "JavaScript API for full YAML front matter editing.",
         "author": "Rob Pitt",


joethei:
You can remove `branch` from this pull request, its no longer used.

[delete frontMatterCopy.position;](https://github.com/Nerd-of-Iron/obsidian-betterfrontmatter/blob/aaaacc68901caf706f88c456b977676dd66fa090/main.ts#L22)
Why is this deleted?
I don't see any other refences of it.

[function getTFile(file: TFile | string | undefined): TFile {](https://github.com/Nerd-of-Iron/obsidian-betterfrontmatter/blob/aaaacc68901caf706f88c456b977676dd66fa090/utility.ts#L6)
Why is it possible to pass `undefined` into this function?
I would recommend making the `file´ parameter optional, like you are doing in the `save`/`load `functions

[const fileContent = await app.vault.cachedRead(view.file);](https://github.com/Nerd-of-Iron/obsidian-betterfrontmatter/blob/aaaacc68901caf706f88c456b977676dd66fa090/utility.ts#L52)
If you are writing to a file it is recommended to use `.read` instead.
https://marcus.se.net/obsidian-plugin-docs/vault#read-files

Nerd-of-Iron:
Hi joethei,

Forgive my mistakes, I am new at contributing to open source and this was also my first Typescript project.

Can you explain what you mean by remove branch? Remove the branch I used as a pull request?

I delete the position property because it exists in the frontmatter returned by getFileCache in the latest version of Obsidian and is not a YAML property.

I will correct the other mistakes now. What is the correct thing to do once I have? I will push it to GitHub but then what do I do?

community-plugins.json
```diff
@@ -5057,5 +5057,13 @@
...
         "branch": "main"
-    }
+    },
+    {
+        "id": "betterfrontmatter",
+        "name": "Better Front Matter",
+        "description": "JavaScript API for full YAML front matter editing.",
+        "author": "Rob Pitt",
+        "repo": "Nerd-of-Iron/obsidian-betterfrontmatter",
+        "branch": "master"
```
joethei, line 5067: I mean to remove this line.

joethei:
Just let us know once you made the change, @liamcain will review the plugin once he has the time.

Nerd-of-Iron:
@joethei I have used the correct read function (thank you), released a new version and removed the unused branch property.

liamcain:
Hi, just a heads up, we already have some metadata improvements planned for the Obsidian API so this plugin might be less relevant in the near future.

- [const frontMatter =](https://github.com/Nerd-of-Iron/obsidian-betterfrontmatter/blob/ad831b41077936ab1ef4a000edd452673a9c7ea2/main.ts#L19-L20) Because you are reading the frontmatter with the intent to write back to disk, I don't think you should be fetching the frontmatter out of the cache. This data could be stale compared to what's in the file currently. I think you're better off doing a `read` of the file, parsing the YAML from the file, then writing back to disk.
- [app.metadataCache.getFileCache(view.file)?.frontmatter](https://github.com/Nerd-of-Iron/obsidian-betterfrontmatter/blob/ad831b41077936ab1ef4a000edd452673a9c7ea2/utility.ts#L54) same thing here, you're reading the latest version of the file contents, but then using stale version of the frontmatter.

Nerd-of-Iron:
@liamcain Hello, I did not actually use the plugin myself in the end as I found my database needs more suitably met by a payloadcms server than Obsidian and my task management needs better met by TickTick (though I do still use and value Obsidian for it's note taking). If you are going to implement this sort of thing yourself in the near future is there much point in my fixing it? I'll fix it if you really would like but it seems like it won't be required soon.

liamcain:
>  If you are going to implement this sort of thing yourself in the near future is there much point in my fixing it? I'll fix it if you really would like but it seems like it won't be required soon.

It's up to you. I was warning you just to see if you still want this plugin included in the plugin gallery. If you do, I think it's worth fixing up, given that the current implementation has risks of data loss for users (since it's writing stale data back to disk)

Nerd-of-Iron:
Nah, there's no need if you are going to implement it yourselves. I'll
delete the project.

PR #1278, opened 2022-10-31
Add Copy document as HTML plugin

mvdkwast:

## Repo URL

Link to my plugin: https://github.com/mvdkwast/obsidian-copy-as-html

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[path.dirname(file.path)](https://github.com/mvdkwast/obsidian-copy-as-html/blob/a3496cd47751d6931affc4b61524784ba22300c8/main.ts#L171)
[path.extname(filePath)](https://github.com/mvdkwast/obsidian-copy-as-html/blob/a3496cd47751d6931affc4b61524784ba22300c8/main.ts#L349)
Instead of using the `path` package, you can get this information from the `TFile`.
You are already retrieving it here: <https://github.com/mvdkwast/obsidian-copy-as-html/blob/a3496cd47751d6931affc4b61524784ba22300c8/main.ts#L166>
Just make sure that the result is `instanceof TFile` as it can also be a folder.
Removing the dependency on the `path` package means that your plugin will work on mobile.

<https://github.com/mvdkwast/obsidian-copy-as-html/blob/a3496cd47751d6931affc4b61524784ba22300c8/main.ts#L328-L339>
This can be replaced with a call to [`arrayBufferToBase64`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L330)

<https://github.com/mvdkwast/obsidian-copy-as-html/blob/a3496cd47751d6931affc4b61524784ba22300c8/main.ts#L437-L446>
I recommend using a `checkCallback` here instead, this means that the command will not be visible in these circumstances, meaning users will not be confused when they run the command and nothing happens.
<https://marcus.se.net/obsidian-plugin-docs/user-interface/commands#conditional-commands>

mvdkwast:
Hi @joethei , thanks for reviewing !

I'm not sure I agree about the [arrayBufferToBase64](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L330) : 
- it makes the conversion to b64 synchronous, which I was hoping to avoid here
- I was hoping that the native FileReader would be more efficient than looping with `btoa` calls which the `arrayBufferToBase64` probably does.
- I would still have to add the `data:image/XXX;base64,` header

Agreed on the other points, thanks for the pointers !

mvdkwast:
Addressed all in [release 0.1.2](https://github.com/mvdkwast/obsidian-copy-as-html/releases/tag/0.1.2)

PR #1280, opened 2022-10-31
Add Plugin: Chorded Hotkeys 

ConnorMeyers:

## Repo URL

Link to my plugin: https://github.com/ConnorMeyers/obsidian-chorded-hotkeys

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
You can remove `obsidian` from your plugin id, its unecessary.

[var content = await app.vault.read(file);](https://github.com/ConnorMeyers/obsidian-chorded-hotkeys/blob/8c216450a83974ce3fb600caf0b7df2dcb4e856b/src/chords/template-chord.ts#L19)
You can use `cachedRead` here instead.
<https://marcus.se.net/obsidian-plugin-docs/vault#read-files>

[const command = app.commands.listCommands().find(c => c.name == value);](https://github.com/ConnorMeyers/obsidian-chorded-hotkeys/blob/8c216450a83974ce3fb600caf0b7df2dcb4e856b/src/chords/command-chord.ts#L14)
I would recommend saving the command id in the settings instead.

liamcain:
- [document](https://github.com/ConnorMeyers/obsidian-chorded-hotkeys/blob/8c216450a83974ce3fb600caf0b7df2dcb4e856b/src/main.ts#L32) I think you should register this event to `window` instead of document. Then it should work in pop-out windows (since obsidian copies window events from the main window into pop-out windows automatically).
- [bind(this)](https://github.com/ConnorMeyers/obsidian-chorded-hotkeys/blob/8c216450a83974ce3fb600caf0b7df2dcb4e856b/src/main.ts#L74) Nit: You can just use an es6 `() => {}` arrow function then you won't need to call `bind`.
- [app](https://github.com/ConnorMeyers/obsidian-chorded-hotkeys/blob/8c216450a83974ce3fb600caf0b7df2dcb4e856b/src/chords/command-chord.ts#L18) Don't rely on the global `app` instance. This is only accessible in the app for debugging purposes and isn't intended to be used by plugins. Instead you should pass the `app` instance into your util functions.
- [templaterPlugin](https://github.com/ConnorMeyers/obsidian-chorded-hotkeys/blob/8c216450a83974ce3fb600caf0b7df2dcb4e856b/src/chords/template-chord.ts#L16) you should use `plugins.getPlugin` instead of accessing the plugin map directly
- [sort](https://github.com/ConnorMeyers/obsidian-chorded-hotkeys/blob/8c216450a83974ce3fb600caf0b7df2dcb4e856b/src/utils/string.extensions.ts#L7) IMO, this can just be a `sortStr` util and doesn't need to be attached to the string prototype.

PR #1284, opened 2022-11-01
Add Auto_Link Plugin

Zenchyi1:

## Repo URL

Link to my plugin: https://github.com/Zenchyi1/Auto-Link

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


Zenchyi1:
Thanks for reviewing my submission,
Command is removed now, it was only meant for testing.
"3" is the Number of lines it checks for links, a.e if the cursor is on line 70, this will be set to 67. Later on in the for loop it will go through every line from 67 to 70 and check for links. 
I wanted to **avoid** searching the entire document, i also renamed the variable names, hoping that this makes it a bit clearer.

liamcain:
Hi, sorry for the delay in reviewing! I took a look at the plugin and found a few small issues. If you don't mind addressing them, I can get this merged in.

---

- [fileHandler](https://github.com/Zenchyi1/Auto-Link/blob/b7e33c5c3872d3cce29fada7c3b57d73feadb623/src/file_handler.ts#L3) classes are usually capitalized in javascript/typescript
- [any](https://github.com/Zenchyi1/Auto-Link/blob/b7e33c5c3872d3cce29fada7c3b57d73feadb623/src/file_handler.ts#L6) this should be TFile
- [any](https://github.com/Zenchyi1/Auto-Link/blob/b7e33c5c3872d3cce29fada7c3b57d73feadb623/src/file_handler.ts#L12) this should be the type of your plugin. Also it doesn't look like you need to pass it in at all.
- [view](https://github.com/Zenchyi1/Auto-Link/blob/b7e33c5c3872d3cce29fada7c3b57d73feadb623/src/insertLinks.ts#L7) you need to check that view isn't null. Also don't cast view to "any"
- [cursorLine: number](https://github.com/Zenchyi1/Auto-Link/blob/b7e33c5c3872d3cce29fada7c3b57d73feadb623/src/insertLinks.ts#L9) you are casting from `number | null` to number here. You have an optional unwrap around view here but not on the line above.
- [const view = app.workspace.getActiveViewOfType(MarkdownView);](https://github.com/Zenchyi1/Auto-Link/blob/b7e33c5c3872d3cce29fada7c3b57d73feadb623/src/insertLinks.ts#L26) you're not checking if view is null at all
- [plugin: any](https://github.com/Zenchyi1/Auto-Link/blob/b7e33c5c3872d3cce29fada7c3b57d73feadb623/src/event_manager.ts#L6) don't cast this to any, just use the type of your plugin
- Looks like you're not tracking new files that get created at all

PR #1288, opened 2022-11-01
Add Writing plugin

johackim:

## Repo URL

Link to my plugin: https://github.com/johackim/obsidian-writing

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
There's no need to include 'obsidian' in the plugin ID. The ID is used for your plugin's settings folder so it should closely match your plugin name for user convenience.

<https://github.com/johackim/obsidian-writing/blob/64a5755afde3cb4821011f66b826db73ea0bbdbc/src/main.js#L17-L21>
Using the callback approach is prefered here:
```ts
    this.initLeaf();
});
```

<https://github.com/johackim/obsidian-writing/blob/64a5755afde3cb4821011f66b826db73ea0bbdbc/src/main.js#L91>
This is a async function, I would recommend adding a `await`.

<https://github.com/johackim/obsidian-writing/blob/64a5755afde3cb4821011f66b826db73ea0bbdbc/src/components/previewer.js#L23>
I would recommend encoding this image with base64 and including it that way directly.
The current approach will fail if the user has no internet/blocked imgur.

<https://github.com/johackim/obsidian-writing/blob/64a5755afde3cb4821011f66b826db73ea0bbdbc/src/utils.js#L5-L10>
The Obsidian API already includes the 
[`MarkdownRenderer.renderMarkdown`]
(https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L2001-L2010)
function for this, it will render any Obsidian specific markdown.

liamcain:
- [if (this.app.workspace.layoutReady) {](https://github.com/johackim/obsidian-writing/blob/64a5755afde3cb4821011f66b826db73ea0bbdbc/src/main.js#L17-L21) Instead of approach, you can use [`onLayoutReady`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3626) which will automatically call the plugin if/when the layout is ready. So it should reduce the boilerplate here a lot
- [attr: { style: 'border-left: none; padding-left: 0px;' },](https://github.com/johackim/obsidian-writing/blob/64a5755afde3cb4821011f66b826db73ea0bbdbc/src/views/chapters.js#L49) I suggest moving this to CSS if possible. It's easier for themes to override.
- [getMarkdownFiles().find((f) => f.path === `${chapter}.md`)](https://github.com/johackim/obsidian-writing/blob/64a5755afde3cb4821011f66b826db73ea0bbdbc/src/views/chapters.js#L53) You can simplify this code by using [getAbstractFileByPath](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L3612) instead. (note: it will return a TAbstractFile, which is either a TFile or a TFolder)
- [getMostRecentLeaf](https://github.com/johackim/obsidian-writing/blob/64a5755afde3cb4821011f66b826db73ea0bbdbc/src/views/chapters.js#L55) you shouldn't be using `getMostRecentLeaf` for this. This won't respect if the leaf is pinned. Instead, you should use [`getLeaf`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L3964). Also, when using getLeaf, I recommend calling it like this: `getLeaf(Keymap.isModEvent(evt))` (where `evt` is the click event). This will allow users to `command/control`-click on the button to open in a new tab, for example.

johackim:
Thanks for your review @joethei and @liamcain!

I made the changes, but I don't know how to integrate `MarkdownRenderer.renderMarkdown` with my React components : 

https://github.com/johackim/obsidian-writing/blob/markdownrenderer/src/views/goal.js#L59-L71

Can you help?

Thanks again!

liamcain:
I think you just need to use dangerouslySetInnerHTML: https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml

liamcain:
@johackim just bumping this to in case you missed my last message. I think this is fine to go in as is, but I do recommend you switch to using Obsidian's renderMarkdown function for better compatibility.

PR #1291, opened 2022-11-02
Add Rapid Notes plugin

valteriomon:

## Repo URL

Link to my plugin: https://github.com/valteriomon/obsidian-rapid-notes

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


liamcain:
- [noteName = originalNoteName.substring(1);](https://github.com/valteriomon/obsidian-rapid-notes/blob/843012e20939d8a48f569bbc9f09fa01d0176f15/src/main.ts#L58) Nitpick: you might want to use an autoformatter for your code so that the indentation lines up correctly.
- [const folderPaths = Object.keys(folders);](https://github.com/valteriomon/obsidian-rapid-notes/blob/843012e20939d8a48f569bbc9f09fa01d0176f15/src/main.ts#L72-L73) this is a super roundabout way of getting the folder paths in the vault. Why not just call `getFolders` and map those to the corresponding tfolder.path? No need to use reduce the folders into an object, then get the keys of the object. The other option is to use [`getAllLoadedFiles`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L3684). Then just loop through the results and find which are `instanceof TFolder`. That gets you all the folders in the vault.

PR #1297, opened 2022-11-02
Added Quran Lookup plugin

abuibrahim2:

## Repo URL

Link to my plugin: https://github.com/abuibrahim2/quranlookup

## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


liamcain:
- [surahJson](https://github.com/abuibrahim2/quranlookup/blob/f0ed902409794ef40599ec2566b27d689647c7d7/main.ts#L56) I recommend loading the JSON file and initializing Fuse inside [`onLayoutReady`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3626) just to avoid slowing down the plugin load time. Since onload is called synchronously, this could slow down the user's startup time. onLayoutReady is called asynchronously and will be effectively the same experience.
- [editor.replaceSelection(totalT);](https://github.com/abuibrahim2/quranlookup/blob/f0ed902409794ef40599ec2566b27d689647c7d7/main.ts#L89-L90) it looks like you might want to add some handling here for if there is no match found
- [[!TIP]](https://github.com/abuibrahim2/quranlookup/blob/f0ed902409794ef40599ec2566b27d689647c7d7/main.ts#L198) A future improvement might be adding a way to customize the format here.

PR #1306, opened 2022-11-03
Add Auto Glossary plugin

ennioitaliano:

## Repo URL

Link to my plugin: https://github.com/ennioitaliano/obsidian-auto-glossary

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-css-themes.json
+++ b/community-css-themes.json
@@ -1018,5 +1018,12 @@
         "repo": "ashwinjadhav818/obsidian-aura",
         "screenshot": "assets/screenshot-preview.png",
         "modes": ["dark"]
+    },
+    {
+        "name": "ion",
+        "author": "zamsyt",
+        "repo": "zamsyt/obsidian-ion",
+        "screenshot": "thumbnail.png",
+        "modes": ["dark"]
     }
 ]


joethei:
You can remove `obsidian` from the plugin id, its uncesarry.

[if (folder.path.contains(".md")) {](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/1131a2409f7e7389015188cb34bdeb2c06eeb968/main.ts#L42)
Test for `folder instanceof TFolder` instead, your method does not work detecting images for example.
I would also recommend putting this before you call `menu.addItem`, that way users will only see it when clicking on folders.

[if (folder.path.contains(".md")) {](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/1131a2409f7e7389015188cb34bdeb2c06eeb968/main.ts#L81)
Same here.

[if (folder.path.contains(".md")) {](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/1131a2409f7e7389015188cb34bdeb2c06eeb968/main.ts#L120)
And here.

[// If the plugin hooks up any global DOM events (on parts of the app that doesn't belong to this plugin)](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/1131a2409f7e7389015188cb34bdeb2c06eeb968/main.ts#L177-L186)
You can remove this

[let notes: string[] = [];](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/1131a2409f7e7389015188cb34bdeb2c06eeb968/utils.ts#L32-L48)
You are itterating over the same set of data multiple times here, it's not really efficient.
This can be replaced by:
```ts
return notesTFiles.some(file => file.name.contains(fileName));
```

[const fileContents: string[] = await Promise.all(](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/1131a2409f7e7389015188cb34bdeb2c06eeb968/utils.ts#L55-L70)
Same here, can be replaced with, for example:
```ts
vault.getMarkdownFiles().forEach(file => {
		const fileConents = vault.cachedRead(file);
		if(!fileContents[i].toString().contains("---\ntags: obsidian-auto-glossary\n---\n")) {
			cleanedNotes.push(file);
		}
	});
```


I would generally recommend moving all code files(with the `.ts` extension) into a seperate `src` folder, making it easier to navigate.
The only config you need to update for it, is this line: [entryPoints: ['main.ts'],](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/1131a2409f7e7389015188cb34bdeb2c06eeb968/esbuild.config.mjs#L18)

PR #1307, opened 2022-11-04
Update community-plugins.json

Quorafind:

## Repo URL

Link to my plugin: https://github.com/Quorafind/Obsidian-Bionic-Reading

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
You can remove `obsidian` from the plugin id, its unecessary.

Please use a different name for the plugin, "Bionic Reading" is already copyrighted/trademarked.
I would recommend removing every reference of that name in code.

You mention in the readme that the plugin works in reading mode, where is this implemented?

Quorafind:
@joethei I think `A plugin to enable bionic reading mode in Live preview mode of Obsidian.` is equal to a plugin that enable `Bionic reading mode` in `live preview mode` of Obsidian.

I will rename bionic reading to another name, but I don't think I have to remove every reference of that name (or maybe just in code?). Someone wants to search bionic reading in Obsidian so that I have to leave at least one in README or plugin's description.

Quorafind:
Finished @joethei

liamcain:
- [linePos](https://github.com/Quorafind/Obsidian-Better-Reading-Mode/blob/61922663e82c10f0cf49a51831750202c2a8b922/src/betterReadingWidget.ts#L81) I don't think you need the `?` optional unwrap here. lineAt doesn't return null

community-plugins.json
```diff
@@ -5409,5 +5409,12 @@
         "description": "A plugin to enable using obsidian notes as calculation sheets.",
         "author": "monesga",
         "repo": "monesga/obsidian-sigma"
+    },
+    {
+        "id": "obsidian-bionic-reading-plugin",
```
liamcain, line 5414:
```suggestion
        "id": "better-reading-mode",
```

PR #1312, opened 2022-11-05
Add obsidian-achievements plugin

Zachatoo:

## Repo URL

Link to my plugin: https://github.com/Zachatoo/obsidian-achievements

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- Instead of creating the uninstallCommands array, you could just use [`Plugin.register`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L638) which will automatically get called at unload

PR #1314, opened 2022-11-08
add plugin: asciimath

widcardw:

## Repo URL

Link to my plugin:

https://github.com/widcardw/obsidian-asciimath

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


widcardw:
Please wait for a moment. I forgot that I had changed the source code of some dependency, which is not available on npmjs. This PR is pending for some change...

widcardw:
Now it's ok to use this plugin.

liamcain:
- [this.app.vault.modify(view.file, content)](https://github.com/widcardw/obsidian-asciimath/blob/21cf0930c9694dddb25f365ce0005b142d636757/main.ts#L115) Since you have an Editor instance, it's better to make the modifications via Editor transactions. This let's codemirror keep track of the actual changes to the file, meaning undo history will be preserved and the view state (cursor location, folded regions, etc) won't get wiped. Take a look at the api for [`Editor.transaction`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L901).
- [.filter(Boolean)](https://github.com/widcardw/obsidian-asciimath/blob/21cf0930c9694dddb25f365ce0005b142d636757/main.ts#L265-L266) this looks redundant

widcardw:
image

image

I have refactored the command with [`Editor.transaction`](https://github.com/widcardw/obsidian-asciimath/blob/main/main.ts#L171), but I am not sure whether I am using it properly. Please review the code and give me some instructions, thanks!

liamcain:
Looks good! There are probably ways to make it slightly more efficient, but this works. Why did you switch from the old method of regex matching across the entire file contents? That would avoid splitting the file by line and search each line, which can be more expensive and memory taxing for longer files.

widcardw:
> Looks good! There are probably ways to make it slightly more efficient, but this works. Why did you switch from the old method of regex matching across the entire file contents? That would avoid splitting the file by line and search each line, which can be more expensive and memory taxing for longer files.

Since the method `Editor.transaction` receives the param `changes`, which looks like this:

```ts
interface EditorChange {
  text: string
  from: EditorPosition
  to: EditorPosition
}
```

However, the `EditorPosition` consists of `{ line: number, ch: number }` and I must get the **line number and char number** of the target formulas.

If I use the global regex, the result only contains the index of the target formula, and it **does not contain the line number and char number**. Maybe it's better to develop an API that can match the content and find out the line number and ch number 👀 ?

liamcain:
Oh, I didn't realized that was the point of confusion. If you are iterating over the whole file contents, you should be able to get the start index of each match from the regex result. Then you can use [`offsetToPos`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L907) to convert the "offset" (offset just means the character offset of the string within the file) into the EditorPosition.

PR #1324, opened 2022-11-11
Add obsidian-webpage-export to plugins

KosmosisDire:

## Repo URL

Link to my plugin: https://github.com/KosmosisDire/obsidian-webpage-export

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [innerHTML](https://github.com/KosmosisDire/obsidian-webpage-export/blob/14b4f7e899b0426f0f92d88da385950a19c8f796/scripts/main.ts#L20) instead of setting the HTML as a string, it's much safer to just construct the elements. You don't risk creating malformed html and it's more efficient. You could make darkModeToggle into a function called `addDarkModeToggle(el: HTMLElement)` and have it construct the same thing you have in the html string.
- [async](https://github.com/KosmosisDire/obsidian-webpage-export/blob/14b4f7e899b0426f0f92d88da385950a19c8f796/scripts/main.ts#L55) don't need to mark this as async
- [getFileNameFromFilePath](https://github.com/KosmosisDire/obsidian-webpage-export/blob/14b4f7e899b0426f0f92d88da385950a19c8f796/scripts/main.ts#L57) you can just use file.basename instead of creating this function to parse the full path.
- [.contains(".")](https://github.com/KosmosisDire/obsidian-webpage-export/blob/14b4f7e899b0426f0f92d88da385950a19c8f796/scripts/main.ts#L57) Don't do this to check if something is a file instead of a folder. Just use `if (file instanceof TFile) {}`. You can have files without file extensions so this assumption isn't safe.
- [var](https://github.com/KosmosisDire/obsidian-webpage-export/blob/14b4f7e899b0426f0f92d88da385950a19c8f796/scripts/main.ts#L137) don't use `var`. Use `let` or `const` instead.
- [switchToLeafWithFile](https://github.com/KosmosisDire/obsidian-webpage-export/blob/14b4f7e899b0426f0f92d88da385950a19c8f796/scripts/leaf-handler.ts#L95) this is unused
- [getLeafByFile](https://github.com/KosmosisDire/obsidian-webpage-export/blob/14b4f7e899b0426f0f92d88da385950a19c8f796/scripts/leaf-handler.ts#L78) this is unused
- [var](https://github.com/KosmosisDire/obsidian-webpage-export/blob/14b4f7e899b0426f0f92d88da385950a19c8f796/scripts/html-gen.ts#L134) You have a bunch of `var`s in this file as well
- [const $ = jQuery;](https://github.com/KosmosisDire/obsidian-webpage-export/blob/14b4f7e899b0426f0f92d88da385950a19c8f796/scripts/html-gen.ts#L7-L8) I don't see any jquery being used. Can this be removed?
- [script src='https://code.jquery.com/jquery-3.6.0.js](https://github.com/KosmosisDire/obsidian-webpage-export/blob/14b4f7e899b0426f0f92d88da385950a19c8f796/scripts/html-gen.ts#L321) Does a copy of jquery need to be included in the output HTML?

KosmosisDire:
Those changes have now been implemented!

Here are some notes on the requested changes.
- I have now rewritten the html generation to procedurally generate html rather than using strings.
- This function now does need async after some changes.
- I have fixed the places where I could use `file.basename`, but I still need the function for some other uses.
- I am now using `TFile` rather than `.contains(".")`
- I have removed all extraneous `vars`.
- Removed `switchToLeafByFile` and `getLeafByFile`
- jquery is definitely being used here, in the `fixLinks`, `inlineImages`, and `outlineImages` functions.
- Yes, jquery is used by the final html page for the color change toggle, the outline, the collapsing headers, and other general behaviors and animations.

Thank you for looking over this :)

liamcain:
The name is fine as is. I usually mention to devs that they don't need to include "obsidian" in the plugin ID. A lot of people think the plugin ID needs to match the name of the repo, but it's really just used for the folder name inside `.obsidian/plugins`. So ideally the ID should be similar to the plugin name so that it's easy for users to find if they are looking for their plugin data. So if you're open to changing it, I'd suggest removing 'obsidian' from the ID. You'll need to update it in the manifest.json file too.

KosmosisDire:
Would I remove it from the ID and the plugin name? Or just the ID?

liamcain:
image

The word "obsidian" doesn't appear in the name at all. It doesn't matter if the name obsidian appears in the repo name (in fact, I would suggest keeping it in the repo name just so it's more clear that it's an obsidian plugin).

KosmosisDire:
Yeah, I just remembered that the actual displayed name there is different than the repo name haha. I guess I will make it match the displayed name exactly as `webpage-html-export`, since I am changing it anyways. I'll let you know when I have finished updating that. Thanks!

KosmosisDire:
@liamcain Okay, I have changed the plugin ID, as far as I know everything should be done!

PR #1325, opened 2022-11-11
Add Auto Glossary Plugin

ennioitaliano:

## Repo URL

Link to my plugin: https://github.com/ennioitaliano/obsidian-auto-glossary

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [this.app](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/b1aa050721d17ff659f1c04e91f1b1c896820740/src/glossaryIndex.ts#L138) you're not in a class scope so I'm guessing `this` here is a bug. You should just pass in a reference to the App instead.
- [this.app](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/b1aa050721d17ff659f1c04e91f1b1c896820740/src/glossaryIndex.ts#L138) Don't access the App from the `global` it's only exposed here for testing purposes but isn't meant to be used by plugins. We may remove it in the future.
- [global.app](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/b1aa050721d17ff659f1c04e91f1b1c896820740/src/utils.ts#L31) don't access app from `global`
- [fileExists](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/b1aa050721d17ff659f1c04e91f1b1c896820740/src/utils.ts#L30) a better approach is to use [`getAbstractFileByPath`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L3612) to find out if a file exists. Or [`DataAdaptor.exists`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L688)
- [contains("---\ntags: obsidian-auto-glossary\n---\n"](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/b1aa050721d17ff659f1c04e91f1b1c896820740/src/utils.ts#L65) this is going to be pretty limiting since it will only check if the file starts with `tags:` in the frontmatter. You might want to check the metadataCache instead to find out if the file has that tag.

ennioitaliano:
@liamcain  Thank you for the review, I fixed some things and have doubts about some others:

>   * [this.app](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/b1aa050721d17ff659f1c04e91f1b1c896820740/src/glossaryIndex.ts#L138) you're not in a class scope so I'm guessing `this` here is a bug. You should just pass in a reference to the App instead.
> 
>   * [this.app](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/b1aa050721d17ff659f1c04e91f1b1c896820740/src/glossaryIndex.ts#L138) Don't access the App from the `global` it's only exposed here for testing purposes but isn't meant to be used by plugins. We may remove it in the future.
> 
>   * [global.app](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/b1aa050721d17ff659f1c04e91f1b1c896820740/src/utils.ts#L31) don't access app from `global`

The use of `this` as reported in the first point is definitely a bug, but I don't get what should I do instead of accessing the App from `global`: it's enough to remove the `global` or there is something else?
***
>   * [fileExists](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/b1aa050721d17ff659f1c04e91f1b1c896820740/src/utils.ts#L30) a better approach is to use [`getAbstractFileByPath`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L3612) to find out if a file exists. Or [`DataAdaptor.exists`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L688)
 
Corrected using `DataAdapter.exists`.
***

>   * [contains("---\ntags: obsidian-auto-glossary\n---\n"](https://github.com/ennioitaliano/obsidian-auto-glossary/blob/b1aa050721d17ff659f1c04e91f1b1c896820740/src/utils.ts#L65) this is going to be pretty limiting since it will only check if the file starts with `tags:` in the frontmatter. You might want to check the metadataCache instead to find out if the file has that tag.

This is the intended behavior since I write that tag **in the frontmatter** everytime a file is created with this plugin.
***
[Here you can find the commit.](https://github.com/ennioitaliano/obsidian-auto-glossary/commit/133254d8a2c278ca481d83acab281d89e56385ef)

liamcain:
> The use of this as reported in the first point is definitely a bug, but I don't get what should I do instead of accessing the App from global: it's enough to remove the global or there is something else?

So the one option is to just pass the reference `this.app` from your plugin into `createFile`, then just keep passing it to each other function that needs it. Since that can be pretty annoying, another option is to just move all your utils into the plugin class itself. Then you can just reference the app via `this.app`.

ennioitaliano:
> So the one option is to just pass the reference `this.app` from your plugin into `createFile`, then just keep passing it to each other function that needs it. Since that can be pretty annoying, another option is to just move all your utils into the plugin class itself. Then you can just reference the app via `this.app`.

Ok, [I fixed it](https://github.com/ennioitaliano/obsidian-auto-glossary/commit/9be9e9df067f72155ec36b5f1104926e4eaab874) using the first option you suggested.
Thanks again

PR #1337, opened 2022-11-14
added obsidian-audio-player community plugin

noonesimg:

## Repo URL

Link to my plugin:
https://github.com/noonesimg/obsidian-audio-player

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [const DEFAULT_SETTINGS: AudioPlayerSettings = {](https://github.com/noonesimg/obsidian-audio-player/blob/2cdade74157648a9524a663248c90f71898498ef/src/main.ts#L12-L14) you can remove this
- [addCommand](https://github.com/noonesimg/obsidian-audio-player/blob/2cdade74157648a9524a663248c90f71898498ef/src/main.ts#L25) just a nitpick, but I recommend you use a code formatter (something like [`Prettier`](https://prettier.io)) so that the code is a bit easier to read/review.
- [container.style.width = '100%';](https://github.com/noonesimg/obsidian-audio-player/blob/2cdade74157648a9524a663248c90f71898498ef/src/main.ts#L60-L61) Instead of adding style attributes, I suggest you move these styles into a styles.css file in your plugin. It will make the plugin easier for Themes to make changes if they want.
- [var](https://github.com/noonesimg/obsidian-audio-player/blob/2cdade74157648a9524a663248c90f71898498ef/src/utils.ts#L3) I recommend you change all instances of `var` to either `const` or `let`. `var` has function-level scope, so it can easily lead to bugs if you're not careful. Here's a pretty good guide with examples of why not to use `var`: [why `var` is obsolete](https://javascript.plainenglish.io/4-reasons-why-var-is-considered-obsolete-in-modern-javascript-a30296b5f08f)
- [createApp(VueApp, { filepath: link.path, ctx: ctx, mdElement: el, audio: player }).mount(el)](https://github.com/noonesimg/obsidian-audio-player/blob/2cdade74157648a9524a663248c90f71898498ef/src/main.ts#L67-L68) Since you're mounting a Vue component in your markdown postProcessor, you'll need to worry about lifecycle management. Basically, the issue is that you need to clean up and properly unmount the Vue component when the tab is closed. Otherwise, it can be a memory leak. To do this, you'll need to move the code here into a class that extends MarkdownRenderChild (something like `class AudioPlayer extends MarkdownRenderChild {}`. Then you'll want to call `ctx.addChild(new AudioPlayer())`. Your AudioPlayer class will then need an onunload() function that calls the $destroy(). (That seems to be the way to unmount Vue components AFAICT)

noonesimg:
> * [const DEFAULT_SETTINGS: AudioPlayerSettings = {](https://github.com/noonesimg/obsidian-audio-player/blob/2cdade74157648a9524a663248c90f71898498ef/src/main.ts#L12-L14) you can remove this
> * [addCommand](https://github.com/noonesimg/obsidian-audio-player/blob/2cdade74157648a9524a663248c90f71898498ef/src/main.ts#L25) just a nitpick, but I recommend you use a code formatter (something like [`Prettier`](https://prettier.io)) so that the code is a bit easier to read/review.
> * [container.style.width = '100%';](https://github.com/noonesimg/obsidian-audio-player/blob/2cdade74157648a9524a663248c90f71898498ef/src/main.ts#L60-L61) Instead of adding style attributes, I suggest you move these styles into a styles.css file in your plugin. It will make the plugin easier for Themes to make changes if they want.
> * [var](https://github.com/noonesimg/obsidian-audio-player/blob/2cdade74157648a9524a663248c90f71898498ef/src/utils.ts#L3) I recommend you change all instances of `var` to either `const` or `let`. `var` has function-level scope, so it can easily lead to bugs if you're not careful. Here's a pretty good guide with examples of why not to use `var`: [why `var` is obsolete](https://javascript.plainenglish.io/4-reasons-why-var-is-considered-obsolete-in-modern-javascript-a30296b5f08f)
> * [createApp(VueApp, { filepath: link.path, ctx: ctx, mdElement: el, audio: player }).mount(el)](https://github.com/noonesimg/obsidian-audio-player/blob/2cdade74157648a9524a663248c90f71898498ef/src/main.ts#L67-L68) Since you're mounting a Vue component in your markdown postProcessor, you'll need to worry about lifecycle management. Basically, the issue is that you need to clean up and properly unmount the Vue component when the tab is closed. Otherwise, it can be a memory leak. To do this, you'll need to move the code here into a class that extends MarkdownRenderChild (something like `class AudioPlayer extends MarkdownRenderChild {}`. Then you'll want to call `ctx.addChild(new AudioPlayer())`. Your AudioPlayer class will then need an onunload() function that calls the $destroy(). (That seems to be the way to unmount Vue components AFAICT)

Thank you for the review! 
I addressed all the issues you've mentioned. 

Just a little sidenote here: since vue3 was introduced, the $destroy() method [was removed](https://v3-migration.vuejs.org/breaking-changes/#removed-apis) as 
> Users should no longer manually manage the lifecycle of individual Vue components

I've moved the initialization of a vue app to a separate class that extends MarkdownRenderChild with explicit lifecycle managment

liamcain:
Sorry for the long wait. The plugin looks good, happy to have it included in the list of community plugins!

noonesimg:
Thank you! Glad I've finally made it through!

PR #1339, opened 2022-11-14
Add new Plugin "ObsiDOOM"

twibiral:

## Repo URL

Link to my plugin: https://github.com/twibiral/ObsiDOOM

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [workspace: Workspace](https://github.com/twibiral/ObsiDOOM/blob/6685560fee358c9e778a47891b4f35a8e4a3f21f/src/SidePanel.ts#L87) no need to pass in the workspace instance, you already have access to it from this.leaf.view.app.workspace (or you could just pass in the app in the constructor to simplify that chaining a bit
- [gameIFrame.style.position = "relative";](https://github.com/twibiral/ObsiDOOM/blob/6685560fee358c9e778a47891b4f35a8e4a3f21f/src/SidePanel.ts#L154-L157) I recommend moving all the styles here into a CSS file, just add a className to your iframe and style it via CSS. It will be easier for themes to override.
- [for (let child of Array.from(this.contentEl.children)) {](https://github.com/twibiral/ObsiDOOM/blob/6685560fee358c9e778a47891b4f35a8e4a3f21f/src/SidePanel.ts#L165-L169) I don't think you need any of this code, the elements will automatically get removed from the container (and garbage collected) when the view is closed.
- ["DOOM": "https://dos.zone/player/?bundleUrl=https%3A%2F%2Fcdn.dos.zone%2Fcustom%2Fdos%2Fdoom.jsdos?anonymous=1",](https://github.com/twibiral/ObsiDOOM/blob/6685560fee358c9e778a47891b4f35a8e4a3f21f/src/GAME_URLS#L2-L3) Looks like your GAME_URLS file is unused.
- [const GAME_URLS = JSON.parse(](https://github.com/twibiral/ObsiDOOM/blob/6685560fee358c9e778a47891b4f35a8e4a3f21f/src/SidePanel.ts#L6-L7) why are you making this a string, then parsing it? It can just be an object.

twibiral:
> * [workspace: Workspace](https://github.com/twibiral/ObsiDOOM/blob/6685560fee358c9e778a47891b4f35a8e4a3f21f/src/SidePanel.ts#L87) no need to pass in the workspace instance, you already have access to it from this.leaf.view.app.workspace (or you could just pass in the app in the constructor to simplify that chaining a bit

This changes would complicate things because `activate` couldn't be static anymore. So yes, it would save the passing of `workspace` but would also introduce other boilerplate code. I kept it.

> * [gameIFrame.style.position = "relative";](https://github.com/twibiral/ObsiDOOM/blob/6685560fee358c9e778a47891b4f35a8e4a3f21f/src/SidePanel.ts#L154-L157) I recommend moving all the styles here into a CSS file, just add a className to your iframe and style it via CSS. It will be easier for themes to override.

Changes according to your suggestion.

> * [for (let child of Array.from(this.contentEl.children)) {](https://github.com/twibiral/ObsiDOOM/blob/6685560fee358c9e778a47891b4f35a8e4a3f21f/src/SidePanel.ts#L165-L169) I don't think you need any of this code, the elements will automatically get removed from the container (and garbage collected) when the view is closed.

You are completely right, I removed it.

> * ["DOOM": "https://dos.zone/player/?bundleUrl=https%3A%2F%2Fcdn.dos.zone%2Fcustom%2Fdos%2Fdoom.jsdos?anonymous=1",](https://github.com/twibiral/ObsiDOOM/blob/6685560fee358c9e778a47891b4f35a8e4a3f21f/src/GAME_URLS#L2-L3) Looks like your GAME_URLS file is unused.

Not anymore, as I now load it instead of parsing a JSON string (see below).

> * [const GAME_URLS = JSON.parse(](https://github.com/twibiral/ObsiDOOM/blob/6685560fee358c9e778a47891b4f35a8e4a3f21f/src/SidePanel.ts#L6-L7) why are you making this a string, then parsing it? It can just be an object.

You are right, I was just lazy. Changed it to read in the GAMES_URL file.


Thank you for your suggestions! I hope the changes made sure sufficient.

PR #1342, opened 2022-11-16
Publish SamePage Extension

dvargas92495:

## Repo URL

Link to my plugin: https://github.com/samepage-network/obsidian-samepage

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


dvargas92495:
Hey `samepage-network` is my organization by which I'm submitting this plugin - can there be a check in this validation that sees whether the submitter is part of the organization?

dvargas92495:
Another question - is there a hosted version of Obsidian?

I'd like to start running end to end tests for the extension, and it would be easier if the test had a URL to go to. Otherwise, I need to figure out a solution for running the product locally, which may not be possible if Obsidian is closed source

dvargas92495:
Created a PR for this: https://github.com/obsidianmd/obsidian-releases/pull/1376/files

liamcain:
Sorry the the long wait for a review! I just found a couple of things I wanted to ask about to get a better picture of what's happening before we merge this.

---

- [label.toLowerCase().replace(/ /g, "-")](https://github.com/samepage-network/obsidian-samepage/blob/53c114fd925ad504bca64fa852d682a1bab12510/src/main.ts#L108) Even though command IDs are normally kebab case, they don't need to be, so can omit the slugifying here
- [document.body](https://github.com/samepage-network/obsidian-samepage/blob/53c114fd925ad504bca64fa852d682a1bab12510/src/protocols/sharePageWithNotebook.ts#L164) I think you'll want to register these on the `window` instead since Obsidian will automatically copy over events on the newly spawned windows. That said, you will manually need to register "paste" on popout windows since we don't bring that yet (will be changed in the next release though). To register the "paste" event per window, you can listen for [`window-open`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L4102), then attach an event listener to Window from the callback.
- [plugin.app.vault.on](https://github.com/samepage-network/obsidian-samepage/blob/53c114fd925ad504bca64fa852d682a1bab12510/src/protocols/sharePageWithNotebook.ts#L131) I don't quite understand why you need to keep registering and unregistering this event handler. Is there stale data trapped in the closure? Ideally you can just use `plugin.registerEvent` and only register this event handler once.
- [plugin.registerDomEvent(document.body, "paste", bodyPasteListener);](https://github.com/samepage-network/obsidian-samepage/blob/53c114fd925ad504bca64fa852d682a1bab12510/src/protocols/sharePageWithNotebook.ts#L178-L179) I'm actually note quite sure why you need to capture all keystrokes + paste? Isn't the `modify` event enough?
- [app.workspace.getLeaf("split", "vertical");](https://github.com/samepage-network/obsidian-samepage/blob/53c114fd925ad504bca64fa852d682a1bab12510/src/protocols/sharePageWithNotebook.ts#L79) The recommended way to handle this now is to call: `app.workspace.getLeaf(Keymap.isModEvent(e))`. That should properly handle where to open the link: new tab, new pane, or new window.
- [getNotebookPageId](https://github.com/samepage-network/samepage.network/blob/8d9512f1484f0b137bda2aa1e069e6af42017a47/package/protocols/sharePageWithNotebook.ts#L55) is there any reason that this function definition can only use an HTMLElement for the lookup? I feel like querying for the leaf title isn't the best way to handle this. One workaround that would be slightly safer is creating a WeakMap<HTMLElement, string> where you listen for the leaf open event, then store the file name by the corresponding view element. That would at least be slightly more futureproof.

dvargas92495:
> so can omit the slugifying here

you're right! removed.

>  think you'll want to register these on the window

switched to `window` and added a listener for `window-open`

> I don't quite understand why you need to keep registering and unregistering this event handler.

This is one of the hairier issues that I'll try explaining and then will propose another solution.

So the problem of just registering a global `"modify"` handler is that this callback captures BOTH user events and events from extensions. This causes an infinite loop:
1. User key presses in Obsidian 1
1. Obsidian 1 `modify` handler captures event, sends update to Obsidian 2
1. Obsidian 2 `modify` handler captures event, sends update to Obsidian 1
1. Repeat steps 2 and 3 repeatedly

So, we focus on _user events_ so that we could differentiate between updates from _users_ and updates from the extension itself.

I would love to do a global `modify` handler. In order to do that I need to differentiate updates from users vs. updates from samepage. Here's one solution that will enable me to do this:
- I submit updates with `plugin.app.vault.modify(abstractFile, expectedText, {id: "samepage"})`
- In the modify handler, I can do: 
```typescript
plugin.app.vault.on("modify", async (file, options) => {
  if (options.id === "samepage") return;
  // continue with callback
});
```

> I'm actually note quite sure why you need to capture all keystrokes + paste? Isn't the modify event enough?

See above.

> app.workspace.getLeaf(Keymap.isModEvent(e))

Thank you!

> any reason that this function definition can only use an HTMLElement for the lookup

Here's where it gets used: https://github.com/samepage-network/samepage.network/blob/main/package/protocols/sharePageWithNotebook.ts#L209-L229

To give some color on what's going on: We use mutation observers to detect when to insert the SamePage status bar into the DOM. 

This is from the common `samepage` npm library, where other apps don't necessarily have an equivalent to "on leaf open". Additionally, other apps render pages in embedded contexts that we want to show the status bar in which is another reason for mutation observers. Because we use mutation observers, we have to infer the notebook page id from the dom element to determine whether or not to show the status bar.

Now for Obsidian, I agree that it is far more future proof to use on leaf open/on leaf close instead of this. Based on this feedback, I'm going to change both the underlying library and the obsidian extension to allow for using these callbacks instead of the mutation observer. This will take some amount of work (but should be doable within a day). Do you consider this to be a review-blocking issue? This will help me prioritize accordingly

---

Thanks for the thoughtful review! I just pushed version 2.1.9 incorporating this feedback. I had one final question. I'd like to run end-to-end tests on the obsidian samepage extension, similar to what I'm doing with Roam:
- [example test](https://github.com/samepage-network/roam-samepage/blob/main/tests/extension.test.ts)
- [example demo video](https://samepage.network/data/tests/roam/data/159fc20b12da4825d8f86bc2b6e1287179ec42cf.webm)

The easiest way to do this is if there was a hosted version of Obsidian. Does one exist or is that something your team would consider? otherwise, I'll have to download a copy of Obsidian on each test run, I think playwright might support running tests against electron instances, but it would slow down testing considerably (some caching solutions might exist with github actions).

liamcain:
> User key presses in Obsidian 1
Obsidian 1 modify handler captures event, sends update to Obsidian 2
Obsidian 2 modify handler captures event, sends update to Obsidian 1
Repeat steps 2 and 3 repeatedly

I think you're overcomplicating it, you should be able to just check if the file actually changed. How Obsidian Sync handles this is by saving its own hash, then you can just compare the incoming hash to the current hash.

It also seems like you're not using the mtime to figure out what changes to accept. You'll probably also want to introduce some logic so that newer local changes aren't overridden by stale server changes.

> Do you consider this to be a review-blocking issue?

I don't think it's blocking, just pointing out that the current approach is quite fragile so it will lead to some potential issues down the line.

> Does one exist or is that something your team would consider?

There was an old guide for how to do e2e testing via spectron. Since that's deprecated, this is the latest info I've come across related to e2e testing: <https://forum.obsidian.md/t/standard-approach-for-writing-automated-end-to-end-tests-for-plugins/31535>. That's might be your best bet for the time being.

dvargas92495:
>  think you're overcomplicating it, you should be able to just check if the file actually changed. How Obsidian Sync handles this is by saving its own hash, then you can just compare the incoming hash to the current hash.

So this solves the infinite loop, but it still causes step 3, where Obsidian 2 received the change from Obsidian 1, and then sends another message out claiming it made an update. I would need to avoid this for history management and messages stored

> It also seems like you're not using the mtime

Not sure I'm familiar, is this just the time the modification was made? I'm already using a CRDT library under the hood that I expect to handle conflict resolution

> I don't think it's blocking, just pointing out that the current approach is quite fragile so it will lead to some potential issues down the line.

agreed - I'm adding it to my list for monday

> this is the latest info I've come across related to e2e testing

this doesn't solve the issue I rose since it's still running the tests on a local install of Obsidian. To clarify, I'm looking to run tests locally as well as in CI via a Github action. For this approach, I would need to continually install Obsidian in CI (or cache a copy somehow) to run tests. But if that's the best bet for the time being I think I see the path to making it work, just wanted to see if a hosted one was available or on the horizon

lishid:
> User key presses in Obsidian 1
Obsidian 1 modify handler captures event, sends update to Obsidian 2
Obsidian 2 modify handler captures event, sends update to Obsidian 1
Repeat steps 2 and 3 repeatedly

Here the way to solve this is to have your plugin save the file with an mtime option (in the options parameter when you write the file). When you receive a file change event, you compare the file's mtime and size with the last known server version of that file. If those are the same, you can safely ignore the change event. If they are different, then you read the file from disk and hash it to compare with the last known hash.

Using this algorithm prevents the situation in your case where Obsidian 2 modified handler reacts to the event, and don't send any update to the server.

dvargas92495:
> When you receive a file change event, you compare the file's mtime and size with the last known server version of that file.

Ah so the issue is _there is no centralized server version_. Each TFT has their own version of the data. So Obsidian 2 would receive the update not knowing which peer to compare the mtime & hash with. If it compares with its own, it's no different than a user event change so then another update gets sent.

Decentralization is a core philosophical tenant for SamePage, so we're looking to avoid a centralized version of the data as much as possible.

lishid:
I still think that's perfectly doable with what I described without storing anything on the server - you can compute a hash and generate an mtime (using `Date.now()`) before calling `vault.write()`, and then store that in memory on your plugin such that if there was any `modified` event matching the same mtime and hash gets ignored. If any real change happens locally, then it would end up with a different mtime and hash which you can then decide to upload. 

Since everything is stored in memory while your plugin is running, you will lose it all upon reboot, but that's not a big deal since you will have to do a re-sync on boot anyway and you can keep a copy of mtime/hash then.

dvargas92495:
🔥 🔥 🔥 

YES - this is great. Version 2.2.1 should now be using file hashes and mtime. I realized that your solution was actually very similar to my proposal, thanks for the suggestion! Now to get other apps to enable something like this...

liamcain:
Looks great! One more thing though: you still want to register your 'modify' event listener with the plugin so that it gets removed when the plugin is unloaded.

```ts
plugin.registerEvent(plugin.app.vault.on("modify", ...))
```

dvargas92495:
ah right of course - just published version 2.2.3 with this change!

dvargas92495:
Also a minor point of feedback - on each update, I've had to resolve conflicts bc another extension got merged. This could be avoided if each json object were its own file within a directory instead of continually adding to the bottom of one file

liamcain:
Oh, sorry, I should have mentioned before, you don't need to rebase each time, I will do the rebase on my end before I merge. But agreed we could improve on the system a bit. It still needs to be a single JSON file at the end, since we fetch it from the app and rely on the ordering, but we could probably add a CI step to generate a clean JSON file 🤔

PR #1343, opened 2022-11-17
Add Mind Map+ plugin

jing-yen:

## Repo URL

Link to my plugin: https://github.com/jing-yen/obsidian-mind-map-plus

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5333,5 +5333,13 @@
         "author": "mvdkwast",
         "repo": "mvdkwast/obsidian-copy-as-html",
         "branch": "master"
+    },
+    {
+        "id": "obsidian-mind-map-plus",
+        "name": "Mind Map+",
+        "description": "Displays markdown notes as mind maps using Markmap, with easy SVG export and color options.",
+        "author": "Tong Jing Yen",
+        "repo": "jing-yen/obsidian-mind-map-plus",
+        "branch": "main"
     }
 ]


liamcain:
- [this.mindmapView](https://github.com/jing-yen/obsidian-mind-map-plus/blob/ad556c4cb9b39f21d3b1ff6b297dbea767034e4f/src/main.ts#L35) Don't store a reference to the view. This prevents the view from being unloaded from memory when it's closed. It also looks like you're not using the reference for anything, so it should be safe to remove this assignment
- [{path:this.activeLeafPath(this.workspace), basename: this.activeLeafName(this.workspace)}](https://github.com/jing-yen/obsidian-mind-map-plus/blob/ad556c4cb9b39f21d3b1ff6b297dbea767034e4f/src/main.ts#L35) I don't think you need to bother passing these values in from the Plugin since you can retrieve the activeLeaf from your view class.
- [activeLeaf](https://github.com/jing-yen/obsidian-mind-map-plus/blob/ad556c4cb9b39f21d3b1ff6b297dbea767034e4f/src/mindmap-view.ts#L166) Accessing activeLeaf is discouraged because it's often not the correct API to use based on what you're trying to accomplish. The active leaf will change when the user interacts with the sidebar views or non-file views. And when that happens, you don't actually want to update your view. Instead you should listen for the [`file-open`](https://github.com/obsidianmd/obsidian-api/blob/ec589e9762a1d7e2faad01f894cb34c41b10ecaf/obsidian.d.ts#L4189) event. That event will only fire when a new markdown view is focused, so you should be able to simplify a lot of your code here.
- [currentMd: string](https://github.com/jing-yen/obsidian-mind-map-plus/blob/ad556c4cb9b39f21d3b1ff6b297dbea767034e4f/src/mindmap-view.ts#L16) If you make the change I suggested above, you shouldn't need to store the currentMd on your view. Instead, you can just immediate parse the file contents and update the view accordingly.
- [vault: Vault;](https://github.com/jing-yen/obsidian-mind-map-plus/blob/ad556c4cb9b39f21d3b1ff6b297dbea767034e4f/src/mindmap-view.ts#L17-L18) I suggest you only store the app instance on your view, instead of storing vault and workspace. Vault and workspace are also singletons, but it's not a great assumption for your plugin to make. If (for example) Obsidian reloaded the workspace and the instance got created, your plugin would be holding a stale reference.
- [this.app.vault.adapter.read](https://github.com/jing-yen/obsidian-mind-map-plus/blob/ad556c4cb9b39f21d3b1ff6b297dbea767034e4f/src/mindmap-view.ts#L191) You should be using [cachedRead](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3373) here instead. It's a lot faster than reading directly from the disk.

joethei:
Hi @jing-yen,

ericaxu:

PR #1346, opened 2022-11-18
Improved plugin validation & added theme validation.

joethei:
- Moved plugin validation into a separate script
- added a few new validation steps:
   - only allow plugin ID's Obsidian can parse.
   - only allow version numbers Obsidian can parse
   - plugin with id/name/repo already in the list
   - discourage including `Obsidian` in name/ID
   - issues need to be enabled
   - encourage including a license in repo
   -  discourage email address included in author field
   - Warn when including the `branch` param in the PR.
- Added validation steps for themes
   - screenshot has to exist
   - screenshots should all be a similar size


.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,28 @@
...
+      - uses: actions/checkout@v3
+        with:
+          ref: "refs/pull/${{ github.event.number }}/merge"
+      - uses: actions/setup-node@v2
+      - run: npm install probe-image-size
+      - uses: actions/github-script@v6
+        with:
+          script: |
+            const script = require('./validation-theme.js');
+            const probe = require('probe-image-size');
```
liamcain, line 23: I'm not too familiar with github scoping. Is there any reason this import needs to be done here and passed in? Can we just import it from the validation-theme.js script?

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  theme-validation:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: "refs/pull/${{ github.event.number }}/merge"
+      - uses: actions/setup-node@v2
+      - run: npm install probe-image-size
```
liamcain, line 18: Is there a simple image size script we can embed directly into the workflow so that we don't need to npm install here? This means that it will run npm install for each execution which I suspect will slow down the script a lot. Obviously it's not going to be hugely detrimental (current execution is ~5s E2E), but still would be nice to avoid.

validation-plugin.js
```diff
@@ -0,0 +1,160 @@
...
+            await github.rest.issues.addLabels({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                labels: ['Ready for review']
+            });
+        }
+    }
+
+    return context.payload.client_payload.value
```
liamcain, line 159:
```suggestion
    return context.payload.client_payload.value;
```

validation-plugin.js
```diff
@@ -0,0 +1,160 @@
+module.exports = async ({ github, context, core, probe }) => {
```
liamcain, line 1:
```suggestion
module.exports = async ({ github, context, core }) => {
```

validation-theme.js
```diff
@@ -0,0 +1,180 @@
...
+    let manifest = {};
+    console.log(`Repo info: ${owner}/${repo}`);
+
+    if (owner.toLowerCase() !== author.toLowerCase()) {
+        addError(`The newly added entry is not at the end, or you are submitting on someone else's behalf. Last theme in the list is: ${theme.repo}`);
+    }
+    try {
+        const repository = await github.rest.repos.get({ owner, repo });
+        if(!repository.data.has_issues) {
+            addError('Your repository does not have issues enabled. Users will not be able to report bugs and request features.');
```
liamcain, line 46: Maybe:

```suggestion
            addWarning('Your repository does not have issues enabled. Users will not be able to report bugs and request features.');
```

We have never stated that this is a requirement, so I'd rather not show this as a hard error.

validation-theme.js
```diff
@@ -0,0 +1,180 @@
...
+            await github.rest.issues.addLabels({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                labels: ['Ready for review']
+            });
+        }
+    }
+
+    return context.payload.client_payload.value
```
liamcain, line 179:
```suggestion
    return context.payload.client_payload.value;
```

.github/workflows/validate-theme-entry.yml
joethei: `require` here is a wrapper around the regular require.
We could also also pass in that wrapper.
<https://github.com/actions/github-script#use-npm-packages>

validation-plugin.js
joethei: We need this here, related to: <https://github.com/obsidianmd/obsidian-releases/pull/1346/files#r1026772896>

joleaf:
Hey, one question: If the script prints „warnings" and „errors", is the author required to fix the the found "warnings"? If yes: maybe add a comment to the bot output, that these warnings should be fixed as well. If no: the bot should print the warnings but also would say, that a fix is not required but optimal?

joethei:
No, warnings are never required to be fixed.
The wording is clear enough for me, warnings are always just that warnings.

validation-plugin.js
```diff
@@ -0,0 +1,168 @@
...
+            }
+        }catch(e) {
+            addError(`The newly added entry is not at the end, or you are submitting on someone else's behalf. Last plugin in the list is: ${plugin.repo}`);
+        }
+        
+    }
+    try {
+        const repository = await github.rest.repos.get({ owner, repo });
+        if(!repository.data.has_issues) {
+            addError('Your repository does not have issues enabled. Users will not be able to report bugs and request features.');
```
liamcain, line 54: this should just be a warning IMO

liamcain (approved):

liamcain (approved):

validation-theme.js
```diff
@@ -18,15 +21,66 @@ module.exports = async ({ github, context, core, probe }) => {
         console.log('Found issue: ' + warning);
     }
 
-    if (context.payload.pull_request.additions <= context.payload.pull_request.deletions) {
-        // Don't run any validation checks if the user is just modifying existing theme config
-        return;
+    const createMessage = async (theme = undefined) => {
+        if (errors.length > 0 || warnings.length > 0) {
+            let message = `#### Hello ${author}!<a href="https://obsidian.md"><img align="right" height="30px" src="https://user-images.githubusercontent.com/59741989/139557624-63e6e31f-e617-4041-89ae-78b534a8de5c.png"/></a>\n`;
```
liamcain, line 26: doesn't look like author is defined yet

PR #1348, opened 2022-11-19
Submit my Obsidian Mindmap Nextgen plugin

MasterTuto:

## Repo URL

Link to my plugin: https://github.com/AdrianSimionov/obsidian-mindmap-nextgen

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5333,5 +5333,13 @@
         "author": "mvdkwast",
         "repo": "mvdkwast/obsidian-copy-as-html",
         "branch": "master"
+    },
+    {
+	"id": "obsidian-mindmap-nextgen"
+    "name": "Obsidian Mindmap Nextgen",
+    "author": "AdrianSimionov",
+	"description": "A plugin to preview notes as Markmap mind maps",
+    "repo": "AdrianSimionov/obsidian-mindmap-nextgen",
+	"branch": "main"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5339,7 +5339,7 @@
     	"name": "Obsidian Mindmap Nextgen",
     	"author": "AdrianSimionov",
 	"description": "A plugin to preview notes as Markmap mind maps",
-    	"repo": "AdrianSimionov/obsidian-mindmap-nextgen",
+    	"repo": "https://github.com/MasterTuto/obsidian-mindmap-vb",
 	"branch": "main"
     }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5335,11 +5335,11 @@
         "branch": "master"
     },
     {
-	"id": "obsidian-mindmap-nextgen"
-    "name": "Obsidian Mindmap Nextgen",
-    "author": "AdrianSimionov",
+	"id": "obsidian-mindmap-nextgen",
+    	"name": "Obsidian Mindmap Nextgen",
+    	"author": "AdrianSimionov",
 	"description": "A plugin to preview notes as Markmap mind maps",
-    "repo": "AdrianSimionov/obsidian-mindmap-nextgen",
+    	"repo": "AdrianSimionov/obsidian-mindmap-nextgen",
 	"branch": "main"
     }
 ]


liamcain:
Hi, because this plugin is a fork of an existing plugin in the community plugins list, we kindly ask that you include the original author's name in the "author" field.

Something like "author": "Your Name (original by Their Name)".

---

- [this.mindmapView](https://github.com/AdrianSimionov/obsidian-mindmap-nextgen/blob/b2cf7183f696ec51bd56b606c608450523bf6139/src/main.ts#L54) Don't store a reference to the view. This prevents the view from being unloaded from memory when it's closed. It also looks like you're not using the reference for anything, so it should be safe to remove this assignment
- [activeLeafPath](https://github.com/AdrianSimionov/obsidian-mindmap-nextgen/blob/b2cf7183f696ec51bd56b606c608450523bf6139/src/main.ts#L105) don't think you need to bother passing these values in from the Plugin since you can retrieve the activeLeaf from your view class.
- [checkActiveLeaf](https://github.com/AdrianSimionov/obsidian-mindmap-nextgen/blob/b2cf7183f696ec51bd56b606c608450523bf6139/src/mindmap-view.ts#L331) Accessing activeLeaf is discouraged because it's often not the correct API to use based on what you're trying to accomplish. The active leaf will change when the user interacts with the sidebar views or non-file views. And when that happens, you don't actually want to update your view. Instead you should listen for the [file-open](https://github.com/obsidianmd/obsidian-api/blob/ec589e9762a1d7e2faad01f894cb34c41b10ecaf/obsidian.d.ts#L4189) event. That event will only fire when a new markdown view is focused, so you should be able to simplify a lot of your code here.
- [currentMd](https://github.com/AdrianSimionov/obsidian-mindmap-nextgen/blob/b2cf7183f696ec51bd56b606c608450523bf6139/src/mindmap-view.ts#L28) If you make the change I suggested above, you shouldn't need to store the currentMd on your view. Instead, you can just immediate parse the file contents and update the view accordingly.
- [vault: Vault;](https://github.com/AdrianSimionov/obsidian-mindmap-nextgen/blob/b2cf7183f696ec51bd56b606c608450523bf6139/src/mindmap-view.ts#L29-L30) I suggest you only store the app instance on your view, instead of storing vault and workspace. Vault and workspace are also singletons, but it's not a great assumption for your plugin to make. If (for example) Obsidian reloaded the workspace and the instance got created, your plugin would be holding a stale reference.
- [this.app.vault.adapter.r](https://github.com/AdrianSimionov/obsidian-mindmap-nextgen/blob/b2cf7183f696ec51bd56b606c608450523bf6139/src/mindmap-view.ts#L361) You should be using [cachedRead](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3373) here instead. It's a lot faster than reading directly from the disk.
- [overrideStopPropagation](https://github.com/AdrianSimionov/obsidian-mindmap-nextgen/blob/b2cf7183f696ec51bd56b606c608450523bf6139/src/mindmap-view.ts#L229) Please don't do this. This will impact more than just your plugin. What are you trying to accomplish with this change?
- [this.workspace.on("css-change", async () => await this.checkAndUpdate()),](https://github.com/AdrianSimionov/obsidian-mindmap-nextgen/blob/b2cf7183f696ec51bd56b606c608450523bf6139/src/mindmap-view.ts#L213-L215) I don't think you want any of these events. You should only need to track the [`file-open`](https://github.com/obsidianmd/obsidian-api/blob/ec589e9762a1d7e2faad01f894cb34c41b10ecaf/obsidian.d.ts#L4189) event
- [quick-preview](https://github.com/AdrianSimionov/obsidian-mindmap-nextgen/blob/b2cf7183f696ec51bd56b606c608450523bf6139/src/mindmap-view.ts#L195) There's no point to debouncing quick-preview. That defeats the purpose of quick-preview, which is to access the file contents for quickly updating a view without an additional read. If you don't want to update your view that often, just remove the quick-preview listener
- Related to the last point, there's no need to create your own debounce logic. You can just the [debounce](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L772) function provided by Obsidian
- Avoid writing your own style tag like this. You should be able to move all this CSS into the styles.css file that's bundled with your plugin.
- [querySelector](https://github.com/AdrianSimionov/obsidian-mindmap-nextgen/blob/b2cf7183f696ec51bd56b606c608450523bf6139/src/mindmap-view.ts#L134) Instead of querying the document (which won't work if the plugin view is opened in a popout window) you should just store a reference to the toolbar HTMLElement on the view.
-

MasterTuto:
Changes have been done!

PR #1353, opened 2022-11-21
Submit a new plugin MAKE.md

jp-cen:

## Repo URL

Link to my plugin: http://github.com/make-md/makemd

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


jp-cen:
Is this error due to mismatch between pull request account and repo account?

ericaxu:
> Is this error due to mismatch between pull request account and repo account?

Yes, that's correct. Although it's not a deal-breaker if you're obviously the author. Are you the author of MAKE.md?

jp-cen:
> > Is this error due to mismatch between pull request account and repo account?
> 
> Yes, that's correct. Although it's not a deal-breaker if you're obviously the author. Are you the author of MAKE.md?

Thanks for the reply. Yep, I'm the author.

community-plugins.json
```diff
@@ -5333,5 +5333,13 @@
         "author": "mvdkwast",
         "repo": "mvdkwast/obsidian-copy-as-html",
         "branch": "master"
+    },
+    {
+        "id": "make-md",
+        "name": "MAKE.md",
+        "description": "The ultimate experience for makers on Obsidian",
```
liamcain, line 5340: The description is a searchable field for users browsing the plugin gallery, I think you might want to provide a better description so that users know what the plugin does and have it more easily discoverable.

jp-cen: Thanks, is there a character limit? I have one that's 264 characters long.

jp-cen: Gotcha, I lowered it to under 256 for now and committed the change.

liamcain:
Hi, I'm really excited about this project and initiative. I just wanted to see if you had considered splitting "Make.md" into multiple plugins? It's up to you, but I think more modular plugins fit better into the Obsidian ethos. They could still stay in the "make.md" plugin family, but users would have more control over what features they think fit with their workflow and vault. Obviously, there are some downsides to it too, so it's just something to keep in mind!

---

- [this.settings.sidebarRibbon](https://github.com/Make-md/makemd/blob/ac7c59bdf897f5c26fd94c074130bf0af94654fa/src/main.ts#L78) starting in Obsidian 1.1.0, ribbon items are toggleable at the app level. Once that version becomes public, we'll start asking plugins to remove their custom ribbon configuration settings so that users can manage their ribbons all in one location
- [FileTreeView](https://github.com/Make-md/makemd/blob/ac7c59bdf897f5c26fd94c074130bf0af94654fa/src/main.ts#L221) why not just use an instanceof check here, then you don't need to look for the .destroy method
- [this.app.vault.off('create', this.onCreate);](https://github.com/Make-md/makemd/blob/ac7c59bdf897f5c26fd94c074130bf0af94654fa/src/main.ts#L253-L254) you can just register these using [`registerEvent`](https://github.com/obsidianmd/obsidian-api/blob/7fa6f1ebb103c99a8bec930233499bb7e79b6e6d/obsidian.d.ts#L666) in onload, then you don't need to manually call .off() in onunload
- [InlineMenuComponent](https://github.com/Make-md/makemd/blob/ac7c59bdf897f5c26fd94c074130bf0af94654fa/src/cm-extensions/inlineStylerView/InlineMenu.tsx#L25) You should consider using an autoformatter like Prettier on these files, I think it could be a whole lot easier to read with some proper indentation.
- [function nosuper](https://github.com/Make-md/makemd/blob/ac7c59bdf897f5c26fd94c074130bf0af94654fa/src/components/FlowEditor/FlowEditor.tsx#L45) I see a lot of code in this file borrowed from the Hover Editor plugin, but I don't see any attribution. When reusing code from another plugin (or open source project), please make sure to follow the terms of the software license.
- [window](https://github.com/Make-md/makemd/blob/ac7c59bdf897f5c26fd94c074130bf0af94654fa/src/dispatch/flowDispatch.ts#L7) you might need to change all instances of "window" here into "activeWindow" to support Obsidian's popout windows.
- [recursive(node, target](https://github.com/Make-md/makemd/blob/ac7c59bdf897f5c26fd94c074130bf0af94654fa/src/utils/utils.ts#L46) chromium doesn't support tail call optimization, so recursive functions like this are better written as a while loop.
- [window.make = this;](https://github.com/Make-md/makemd/blob/ac7c59bdf897f5c26fd94c074130bf0af94654fa/src/main.ts#L168-L169) Ideally you don't put this on the global window object. This is a potential memory leak because the plugin will never get removed from memory on unload, you also risk colliding with other plugins if they are trying to use the same variable name.

jp-cen:
Thanks for the review, @liamcain and for the detailed comments! I have pushed an update that addresses all of the concerns you've raised here. 

We're looking at certain parts to become more modular, however, there are some non-obvious integrations and dependencies between the features (flow editor and the inline styler for example), so we'll have circle back to this in the near future.

Changes made:
- Sidebar ribbon has been addressed and that feature has been removed.
- Fixed the filetree, event listener, code formatting, attribution, activewindow points and some recursive function improvements and removed global variable.

liamcain:
Okay, looking good! Do you mind rebasing this pull request and moving your plugin entry to the bottom?

liamcain:
Hmm, I'm not quite sure what you did, but I think you merged all the commits on top of your branch instead of rebasing.

Try something like:

git fetch origin master
git rebase -i origin/master

Then in the interactive rebase, make sure to only keep your own commits. Delete any of the other commits that got pulled in.

jp-cen:
> Hmm, I'm not quite sure what you did, but I think you merged all the commits on top of your branch instead of rebasing.
> 
> Try something like:
> 
> git fetch origin master git rebase -i origin/master
> 
> Then in the interactive rebase, make sure to only keep your own commits. Delete any of the other commits that got pulled in. Sorry! Normally I just rebase for people, but it seems that your obsidian-releases repo has some weird permissions so I can't push to it.

Thanks, fixed!

PR #1354, opened 2022-11-21
Add External Link Opener Plugin

zorazrr:

## Repo URL

Link to my plugin: https://github.com/zorazrr/obsidian-link-opener

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [&& href](https://github.com/zorazrr/obsidian-link-opener/blob/eaa2425a140ff06931490e0ec10c0bd367bb5fbd/src/main.ts#L27) code style nitpick: you might want to just check for `href` first and if it doesn't exist, just return early. You're checking for `href` inside each conditional which is unnecessary. Then you could even use a switch statement if you preferred.
- [this.app.workspace.revealLeaf](https://github.com/zorazrr/obsidian-link-opener/blob/eaa2425a140ff06931490e0ec10c0bd367bb5fbd/src/main.ts#L49) this might not be the same leaf that you just created. Instead of using `getLeafOfType`, you should store a reference to your newly created leaf, then pass that into `revealLeaf`:

```ts
let leaf = workspace.getLeaf("tab");
leaf.setViewState({ ... });
workspace.revealLeaf(leaf);
```

PR #1356, opened 2022-11-21
Mathjax WikiLinks Plugin

aaron-jack-manning:

## Repo URL

Link to my plugin: https://github.com/aaron-jack-manning/obsidian-mathjax-wikilinks

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [x]  Android
  - [ ]  iOS
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [isMath = !isMath;](https://github.com/aaron-jack-manning/obsidian-mathjax-wikilinks/blob/467ab74ed19d471fc2b6ad789b6ede1f9e242a33/main.ts#L34) Just a heads up, I don't think this is the best way to check if there's a Math expression in the wikilink. For example, it will misunderstand links that start with a $ but don't end with one. It also doesn't take into account `\$` which should escape the Math expression.

PR #1359, opened 2022-11-22
Add "Todo sort" plugin

ryangomba:

## Repo URL

Link to my plugin: https://github.com/ryangomba/obsidian-todo-sort

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5772,5 +5772,13 @@
         "description": "File Forgetting Curve",
         "repo": "ptrsvltns/file-forgetting-curve-obsidian",
         "branch": "main"
+    },
+    {
+        "id": "todo-sort",
+        "name": "Todo sort",
+        "description": "Sort todos by completion status",
+        "author": "Ryan Gomba",
+        "repo": "ryangomba/obsidian-todo-sort",
+        "branch": "main"
     }
 ]


liamcain:
- [loadSettings](https://github.com/ryangomba/obsidian-todo-sort/blob/ef1e2c15c89bd20ea17d53494c2e058d570294c3/src/main.ts#L43) Don't do this. This will mean that each file change will cause the plugin to read the settings from disk. While the file might be small, unnecessary disk reads should always be avoided since they can be expensive (especially if it's a harddrive read and not solid state) plus it puts more burden on the disk than it needs.
- [value.indexOf](https://github.com/ryangomba/obsidian-todo-sort/blob/ef1e2c15c89bd20ea17d53494c2e058d570294c3/src/sort.ts#L24) you could end up with false positive matches by looking for todos like this. Instead you should use regular expressions to test if the line is a TODO. Something like `^ */(-|\*) \[.\]/`
- [const completed](https://github.com/ryangomba/obsidian-todo-sort/blob/ef1e2c15c89bd20ea17d53494c2e058d570294c3/src/sort.ts#L25) A completed task doesn't always use "x" as the completion character. Any character that's not a space will be treated as "completed"

ryangomba:
Thanks @liamcain for the help. I've made the changes you requested and also added a few tests for the failure cases you raised. After experimenting with a debounce, I chose to make the delay a customizable setting. A delay makes the feature feel a little sluggish and sometimes surprising, so I'd like to leave it up to the user to decide what works best for them. The default is 250ms.

PR #1362, opened 2022-11-22
Add ProZen plugin

cmoskvitin:

## Repo URL

Link to my plugin: https://github.com/cmoskvitin/obsidian-prozen

## Release Checklist
  - [x]  Windows
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [animate](https://github.com/cmoskvitin/obsidian-prozen/blob/eef64fbe70d0e12ae3e2789e66b40e9be82d1e59/styles.css#L30) I suggest you give your classNames a prefix. So this would be something like `.prozen-animate {` instead. The reason being that these class names could collide with obsidian classes or other plugins.
- [fullscreenMode](https://github.com/cmoskvitin/obsidian-prozen/blob/eef64fbe70d0e12ae3e2789e66b40e9be82d1e59/main.ts#L48) maybe this command should toggle zen mode so that you can leave zenmode by calling the command again?

PR #1365, opened 2022-11-23
Add Tray plugin

dragonwocky:

## Repo URL

Link to my plugin: https://github.com/dragonwocky/obsidian-tray

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
  - [ ] styles.css _(optional)_


dragonwocky:
Oh, I assumed it would just pull the files from the repo matching the release's commit hash. I've added the `main.js` and `manifest.json` files to the release now.

liamcain:
- [querySelector(".mod-close");](https://github.com/dragonwocky/obsidian-tray/blob/5a81b44df12ada6811bc6cc2304c9cbf1b1bce6a/main.js#L40) this is not a good way to predict if the user is closing the window. It is very susceptible to a lot of false positives. There are several buttons in the UI with the class "mod-close" on them so this could lead to some really unexpected bugs where users click and the whole window disappears. You are also not handling cases where the user closes a windows via hotkey. Popout windows also close when the user clicks the 'tab close' button on the last open tab in the windows. I think this plugin is otherwise safe but it's important that this is addressed before merging.

dragonwocky:
> You are also not handling cases where the user closes a windows via hotkey.

This is intentional: system-level hotkeys like `Alt+F4` aren't handled so that they can act as force-quits, if ever the user happens to need to actually close the app and not just hide it (which can also be done through the tray icon's menu, but that can be disabled). The plugin options do only say the will be hidden and run in the background "...instead of quitting it when pressing the **window close button** or **toggle focus hotkey**".

> There are several buttons in the UI with the class "mod-close" on them so this could lead to some really unexpected bugs where users click and the whole window disappears.

Gotcha, didn't know that class was reused elsewhere. Would it be sufficient to change that selector to the more specific `.titlebar-button-container > .titlebar-button.mod-close`?

I did originally attempt to handle `BrowserWindow.on("close")`, but for whatever reason it didn't seem to work. The other option might be to attempt to override `window.onbeforeunload`, but that could make the behaviour inconsistent with what I described above.

> Popout windows also close when the user clicks the 'tab close' button on the last open tab in the windows.

As in, they should but don't, or do but shouldn't? They seem to be behaving the same for me with/without the plugin, other than the expected minimisation to the tray of the main window?

---

There are apparently also some macOS compatibility issues as well that I need to look into sorting out. That may take me a little while since I don't have a mac, but I'll let you know when I've fixed what I can.

Thanks!

liamcain:
> As in, they should but don't, or do but shouldn't? They seem to be behaving the same for me with/without the plugin, other than the expected minimisation to the tray of the main window? Not sure I understand what needs changing on that point, sorry!

oh, good point. I forgot this plugin doesn't really need to worry about that part. In that case, what you have is fine.

> Gotcha, didn't know that class was reused elsewhere. Would it be sufficient to change that selector to the more specific .titlebar-button-container > .titlebar-button.mod-close?

Yes, more specific would be better. However, the close button on macOS (and I think Linux too) use native elements and you won't be able to attach event listeners to them. So this approach will only work on Windows.

> I did originally attempt to get the current window's  Electron.BrowserWindow instance and handle .on("close"), but for whatever reason it didn't seem to work. The other option might be to attempt to override window.onbeforeunload, but that could make the behaviour inconsistent with what I described above.

`window.onbeforeunload = () => {}` should work but you won't be able to tell if it was closed via the mouse or keyboard.

dragonwocky:
@liamcain thanks, having a look at this now, I've been focusing on some other stuff for a little while.

> window.onbeforeunload = () => {} should work but you won't be able to tell if it was closed via the mouse or keyboard.

I attempted to override `window.onbeforeunload` with `event.preventDefault(); return false;` to cancel the window close event, and minimise it to the tray instead, but it didn't seem to have any effect, I'm not sure why. I'll see if I can work out another way tomorrow, I'm open to suggestions though if you have any.

dragonwocky:
Sorry for the wait @liamcain @joethei, this should be ready for review/merge now. Summary of changes (more detail is listed on the releases page):
- various bugfixes and improvements to macOS/Linux compatibility.
- better option defaults and more specific control of window management.
- window close is intercepted at electron level instead of by overriding the dom click handler.
- multi-vault support (open, hide & quit actions target vault windows only).
- customisable global shortcut to create a new "quick note" from anywhere.
- customisable tray/menubar icon images.

liamcain:
- Run in background doesn't take affect until you restart Obsidian

Okay, I tried this out on macOS and the experience there is very buggy unfortunately. It's likely that the behavior varies a lot from platform to platform for something like this. On macOS, the "tray" is the Menu bar. Clicking on the menu bar icon causes Obsidian to refocus/unminimize which is unexpected. You might need to find someone with a mac to collaborate with and fix some of these glitches.

dragonwocky:
Thanks for merging!

> On macOS, the "tray" is the Menu bar. Clicking on the menu bar icon causes Obsidian to refocus/unminimize which is unexpected.

Hm, a left-click should act to toggle window visibility and a right-click should open the menu. That's the normal behaviour on Windows at least, so that's what I've implemented- does a left-click bring up the menu on macOS?

I've fixed most of what's been reported in the plugin repo and users seem happy so far, I'm sure if there are more platform-specific complaints they'll find their way there for me to fix too.

liamcain:
> That's the normal behaviour on Windows at least, so that's what I've implemented- does a left-click bring up the menu on macOS?

Yes, on macOS, there is no right-click for menu items. Left click opens the menu.

PR #1366, opened 2022-11-23
External Link Helper

nakalsio:
## Repo URL
Link to my plugin: [https://github.com/nakalsio/obsidian-danpung](https://github.com/nakalsio/obsidian-danpung)

## Release Checklist

  - [x] Windows
  - [x] macOS
  - [x] Linux
  - [x] main.js
  - [x] manifest.json
  - [x] styles.css
- [x] GitHub release name matches the exact version number specified in my manifest.json (Note: Use the exact version number, don't include a prefix v


nakalsio:
Changed the tag from 'v0.8.0' to '0.8.0'

liamcain:
- [ribbonIconEl.addClass('my-plugin-ribbon-class');](https://github.com/nakalsio/obsidian-danpung/blob/5b881ed49e1ed223fcb32034d13d2e56f4eb43c8/src/main.ts#L37) you can remove this line.
- [modifiers: ["Shift", "Alt"], key: ";"](https://github.com/nakalsio/obsidian-danpung/blob/5b881ed49e1ed223fcb32034d13d2e56f4eb43c8/src/main.ts#L47) While there's no hard rule against it, we recommend against providing a default hotkey when possible. The reason being that it's easy to pick a hotkey that a user already has configured and they could get confused when that key doesn't do what they expect. Also, it's hard choosing a safe default hotkey that's available for all operating systems. For example, Shift+Alt hotkeys won't work for macOS since that is reserved for inputting unicode characters.
- [getCache(file.path)](https://github.com/nakalsio/obsidian-danpung/blob/5b881ed49e1ed223fcb32034d13d2e56f4eb43c8/src/main.ts#L96) assign this to a variable so you don't need to call getCache multiple times in a row.

nakalsio:
1. Removed `ribbonIconEl.addClass('my-plugin-ribbon-class');`
2. Removed hotkeys
3. Assigned to `theCache` : https://github.com/nakalsio/obsidian-danpung/blob/8b929e5aadfb1dbfd7b970f7602f40f9dfca1708/src/main.ts#L92
4. gifs are looping

PR #1376, opened 2022-11-28
Update validate plugin entry to account for publishing on behalf of orgs

dvargas92495:
This change should allow plugin authors to publish plugins on behalf of their GitHub organization

### DIFF
--- a/.github/workflows/validate-plugin-entry.yml
+++ b/.github/workflows/validate-plugin-entry.yml
@@ -50,7 +50,16 @@ jobs:
             console.log(`Repo info: ${owner}/${repo}`);
 
             if (owner.toLowerCase() !== author.toLowerCase()) {
-              addError(`The newly added entry is not at the end, or you are submitting on someone else's behalf. Last plugin in the list is: ${plugin.repo}`);
+              const onBehalfOfOrg = await octokit.rest.orgs
+                .checkMembershipForUser({
+                  org: owner.toLowerCase(),
+                  username: author.toLowerCase(),
+                })
+                .then(() => true)
+                .catch(() => false);
+              if (!onBehalfOfOrg) {
+                addError(`The newly added entry is not at the end, or you are submitting on someone else's behalf. Last plugin in the list is: ${plugin.repo}`);
+              }
             } else {
               try {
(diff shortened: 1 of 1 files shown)


.github/workflows/validate-plugin-entry.yml
```diff
@@ -50,7 +50,16 @@ jobs:
             console.log(`Repo info: ${owner}/${repo}`);
 
             if (owner.toLowerCase() !== author.toLowerCase()) {
-              addError(`The newly added entry is not at the end, or you are submitting on someone else's behalf. Last plugin in the list is: ${plugin.repo}`);
+              const onBehalfOfOrg = await octokit.rest.orgs
```
joethei, line 53: This will break, on GH actions it would be `github.rest...`, will work the same though.

joethei (changes requested):

.github/workflows/validate-plugin-entry.yml
dvargas92495: gah sorry forgot to swap it out from my internal testing

lishid:
Hmm according to the github docs, it seems to care whether the requester is the org member:
https://docs.github.com/en/rest/orgs/members?apiVersion=2022-11-28#check-organization-membership-for-a-user

Seems like it returns 302 if the requester isn't and 204 if both the requester and the username passed in is a member. Does anyone know how octokit handles the differentiation between 302 and 204?

dvargas92495:
302s seem to just return the data, which aligns with the current `then => true, catch => false` logic

https://github.com/octokit/request.js/blob/master/src/fetch-wrapper.ts#L69-L118

dvargas92495:
I just did

```
.checkMembershipForUser({
  org: "obsidian-community",
  username: "joethei",
})
```

and it actually returned a 204 locally, despite me not being in the obsidian-community org. In any case, I believe the Promise should resolve

lishid:
Ok! Let's merge and see. Thank you for the contribution ❤️

dvargas92495:
Seems to have worked!

https://github.com/obsidianmd/obsidian-releases/pull/1342

PR #1378, opened 2022-11-30
Zk4 evergreen plugin

zk4:
EverGreen workflow with stack tabs in Obsidian

## Repo URL

 
Link to my plugin: https://github.com/zk4/obsidian-evergreen

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [console.log(viewState.type)](https://github.com/zk4/obsidian-evergreen/blob/c451ed0a9172b99688c0d233298a2cfcc95174ff/src/main.ts#L27) remove this
- [getLeavesOfType](https://github.com/zk4/obsidian-evergreen/blob/c451ed0a9172b99688c0d233298a2cfcc95174ff/src/main.ts#L23) I don't think you want to iterate over all leaves. I think you actually want to get the current tab group instead. const tabs = workspace.activeLeaf.parent.children should do the trick
- [if (viewState.state?.file?.endsWith(sourcePath)) {](https://github.com/zk4/obsidian-evergreen/blob/c451ed0a9172b99688c0d233298a2cfcc95174ff/src/main.ts#L35) I think it makes sense to stop searching for the source leaf after you've found the target leaf. The behavior should only work when the target is to the right of the source.

joethei:
Hi @zk4 do you still intend on submitting this plugin?

ericaxu:

PR #1391, opened 2022-12-02
Add 'Translate' plugin

Fevol:

## Repo URL

Link to my plugin: https://github.com/Fevol/obsidian-translate

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [X]  Linux
  - [X]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


Fevol:
Let me know if I need to change the name or ID, I've seen some comments that adding `obsidian-` is unnecessary.
Currently, there are three other translation plugins:
- `language-translator`: [Language Translator](https://github.com/twentytwokhz/language-translator)
- `obsidian-translator`: [Translator](https://github.com/luhaifeng666/obsidian-translator)
- `deepl`: [DeepL](https://github.com/friebetill/obsidian-deepl)

liamcain:
> Let me know if I need to change the name or ID, I've seen some comments that adding obsidian- is unnecessary.

Yeah, we usually point out that you don't need the 'obsidian-' prefix since all it does it make it harder for users to find the plugin folder inside .obsidian/plugins/. Otherwise it's not a big deal.

> I could change the name to Rosetta Translate to differentiate from the other plugins, though that name might also be confusing for people searching the plugin.

It's up to you if you'd prefer a more unique name. It usually makes it more helpful for users to refer to your plugin if the name is more unique. But I can understand wanting something short and sweet too.

Code LGTM, this can go in whenever. Let me know if you'd like to fix the ID first though. You'll need to change it here and in your repo manifest.json. And you'll need update the latest release with the updated manifest.json (or alternatively cut a new release).

Fevol:
Thank you for the review @liamcain!

For the name: I haven't been able to come up with a name that is both more descriptive, and is not easily confusable with already existing services, so I think I'm going to keep it as is.

As for the ID, am I allowed to shorten it to just `translate`?

liamcain:
> As for the ID, am I allowed to shorten it to just translate?

Sure! As long as it's not already taken, it's fair game.

PR #1392, opened 2022-12-02
add Obsidian Mermaid plugin

dartungar:
Add Obsidian Mermaid plugin: improved Mermaid.js experience for Obsidian: visual toolbar with common elements (with more functions to come)

## Repo URL
Link to my plugin:
https://github.com/dartungar/obsidian-mermaid

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [showNotice](https://github.com/dartungar/obsidian-mermaid/blob/4ae94e7f37ef1b244fea869d4e22d39eb7467fb0/main.ts#L76) there's no benefit to having this function on the plugin
- [svg](https://github.com/dartungar/obsidian-mermaid/blob/4ae94e7f37ef1b244fea869d4e22d39eb7467fb0/src/ui/toolbarView/viewHelpers.ts#L86) I think you should call [sanitizeHTMLToDom](https://github.com/obsidianmd/obsidian-api/blob/6b2138aa28fb6d6f58f78e6fabec02186fdc0f82/obsidian.d.ts#L2977) just in case there's the possibility of DOM injection from the mermaid svg value.
- [click](https://github.com/dartungar/obsidian-mermaid/blob/4ae94e7f37ef1b244fea869d4e22d39eb7467fb0/main.ts#L26) this is not a reliable way to get keep track of the current editor. If you're targeting Obsidian v1.1 or newer, you can use the [`activeEditor`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4009). That should stay up to date automatically so you can remove the logic around managing your own active editor reference.

dartungar:
@liamcain thanks for the review! I've dealt with first 2 points. As for using `activeEditor` - I've switched to it, but still have to rely on a separate variable for tracking it, since clicking on custom view makes `activeEditor` null (since there is no active editor in custom view).

liamcain:
> As for using activeEditor - I've switched to it, but still have to rely on a separate variable for tracking it, since clicking on custom view makes activeEditor null (since there is no active editor in custom view).

In that case, you should add a listener on the [`active-leaf-change`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4202) event. Then check if the new leaf has an editor associated with it and update your activeEditor inside there. Using document click event won't work if the user switches views via hotkey or another another interaction.

dartungar:
@liamcain that makes sense, thanks! I've changed the code accordingly: https://github.com/dartungar/obsidian-mermaid/blob/b0d3cecf004190d3b866cdb70134e47fc5c8e837/main.ts#L30

liamcain:
Looks good. One more small thing. We're asking plugins to avoid using the word "Obsidian" in their plugin name to avoid over-saturating the brand name and reserve the naming for first-party plugins (see: Obsidian Publish, Obsidian Sync). Would you mind renaming this plugin, could just be something like "Mermaid Improved"

dartungar:
@liamcain sure, done it!

liamcain:
Thanks!

PR #1393, opened 2022-12-02
add 'Audio Notes' plugin

jjmaldonis:

## Repo URL

Link to my plugin: https://github.com/jjmaldonis/obsidian-audio-notes

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


jjmaldonis:
I created a release with the tag "0.1.0".

liamcain:
- [const view = this.app.workspace.getActiveViewOfType(MarkdownView);](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/main.ts#L610) no need to fetch the current view if you already have access to the leaf. Just use `leaf.view`
- [Note: `.catch` is required](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/main.ts#L731) if you'd rather use `async/await`, you could just move this into its own async function and call it without `await`. Since you don't need the return value, it will work fine.
- [monkeyPatchConsole](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/main.ts#L924) I don't recommend that you keep this in your plugin since this will impact the entire app. If you want to keep this code around, I suggest putting it behind some "debug" flag.
- [new Notice(`Must be in editor mode.`), 10000](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/main.ts#L1592) the time is outside of the parens
- [sourceView](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/main.ts#L1589) you don't need to query for this, just check the mode with getMode()
- [var repeat](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/main.ts#L1217) I recommend you change all instances of `var` to either `const` or `let`. `var` has function-level scope, so it can easily lead to bugs if you're not careful.
- [input[type="range"](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/styles.css#L184) these styles are global. You should make sure to wrap all the styles in a class that you control. Otherwise, this will impact all the sliders throughout the app.
- [knownAudioPlayers](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/main.ts#L626) I don't think you should keep a map of all the audioplayer elements. Obsidian has a lot of optimizations under the hood that will cause elements to get removed or recreated, so creating a map will mean that removed audio elements won't get cleaned up from memory and will become a memory leak. You could use a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) to keep a reference to elements by some ID, but one disadvantage to WeakMaps is you can't iterate over them. So you'll need a different way to handle that.

jjmaldonis:
Thanks for the fantastic feedback!

I have fixed everything you mentioned. See below for details:

* [const view = this.app.workspace.getActiveViewOfType(MarkdownView);](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/main.ts#L610) no need to fetch the current view if you already have access to the leaf. Just use leaf.view
    * Fixed. See [here](https://github.com/jjmaldonis/obsidian-audio-notes/blob/master/src/CreateNewAudioNoteInNewFileModal.ts#L72).
* [Note: .catch is required](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/main.ts#L731) if you'd rather use async/await, you could just move this into its own async function and call it without await. Since you don't need the return value, it will work fine.
    * I decided to continue to use `.then().catch()` (rather than async/await) so the code that runs the command is right on the screen if someone is reading the code. I changed the comment so it makes more sense.
* [monkeyPatchConsole](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/main.ts#L924) I don't recommend that you keep this in your plugin since this will impact the entire app. If you want to keep this code around, I suggest putting it behind some "debug" flag.
    * Fixed. I put the code behind a "debug" flag and it only runs on mobile. I wanted to keep this because I can't access the developer tools on mobile, so it has been very helpful for me to get the error messages in a log.txt file. Plus other users can send me their log.txt file if they are having issues. See [here](https://github.com/jjmaldonis/obsidian-audio-notes/blob/master/src/main.ts#L122).
* [new Notice(Must be in editor mode.), 10000](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/main.ts#L1592) the time is outside of the parens
    * Thanks! These are now fixed.
* [sourceView](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/main.ts#L1589) you don't need to query for this, just check the mode with getMode()
    * I was able to use `getMode()`, and then I found out I can remove the code entirely. So it's gone now.
* [var repeat](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/main.ts#L1217) I recommend you change all instances of var to either const or let. var has function-level scope, so it can easily lead to bugs if you're not careful.
    * Fixed, thanks! See [here](https://github.com/jjmaldonis/obsidian-audio-notes/blob/master/src/main.ts#L615).
* [input[type="range"](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/styles.css#L184) these styles are global. You should make sure to wrap all the styles in a class that you control. Otherwise, this will impact all the sliders throughout the app.
    * Fixed. I now use css class inheritance to ensure the styles are applied appropriately. See [here](https://github.com/jjmaldonis/obsidian-audio-notes/blob/master/styles.css#L184).
* [knownAudioPlayers](https://github.com/jjmaldonis/obsidian-audio-notes/blob/2e299ecbce26c2ef862cc52acdf78c8fcc2c69b4/main.ts#L626) I don't think you should keep a map of all the audioplayer elements. Obsidian has a lot of optimizations under the hood that will cause elements to get removed or recreated, so creating a map will mean that removed audio elements won't get cleaned up from memory and will become a memory leak. You could use a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) to keep a reference to elements by some ID, but one disadvantage to WeakMaps is you can't iterate over them. So you'll need a different way to handle that.
    * This was a tough one. I couldn't get `WeakMaps` to work because the keys need to be objects in a WeakMap. I ended up going with a solution that limits the number of audioplayer elements that are cached/saved. I set the max to 30, which is a pretty low number and should take up negligible memory. See [here](https://github.com/jjmaldonis/obsidian-audio-notes/blob/master/src/AudioNotes.ts#L167) and the reference to the `AudioElementCache` class in `main.ts`.
    * The functionality that uses the cache is pretty important because it keeps the audio players synced in the Live Preview and Reading views. Technically there is still a small memory leak, but it is limited to a very small size (max 30 HTMLElements) and is now a negligible amount of the memory usage of this plugin.
    * I also tried for a few hours to hook into an event that gets triggered when the HTMLElements are unloaded/destroyed, but I wasn't able to get anything to work. I had never heard of on-destroy events for HTMLElements until I googled today, but the functionality for that doesn't seem to exist/work.

jjmaldonis:
@liamcain I updated my plugin based on your requests, please see above. Thanks!

liamcain:
> This was a tough one. I couldn't get WeakMaps to work because the keys need to be objects in a WeakMap. I ended up going with a solution that limits the number of audioplayer elements that are cached/saved. I set the max to 30, which is a pretty low number and should take up negligible memory. See [here](https://github.com/jjmaldonis/obsidian-audio-notes/blob/master/src/AudioNotes.ts#L167) and the reference to the AudioElementCache class in main.ts.

Hmm, I still think you could simplify things a bit. I'm picturing just keeping track of 2 variables:
- activePlayer
- audioPlayers: WeakMap<HTMLElement, HTMLElement>

Then you could map Audio htmlelements to AudioPlayers. `updateCurrentTimeOfAudio` would be able to look up audio players using the weakmap.

Anyway, I think it looks good! Feel free to continue down that path for future improvements.

jjmaldonis:
Thanks @liamcain!

I'll implement that.

PR #1398, opened 2022-12-05
Add tab switcher

Quorafind:

## Repo URL

Link to my plugin: https://github.com/Quorafind/Obsidian-Tab-Switcher

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5525,5 +5525,14 @@
         "author": "SamePage",
         "repo": "samepage-network/obsidian-samepage",
         "branch": "main"
+    },
+    {
+    
+        "id": "tab-switcher",
+        "name": "Tab Switcher",
+        "description": "A plugin to help you switch tabs that works similiar to CTRL+SHIFT+A in Edge.",
+        "author": "Boninall",
+        "repo": "quorafind/obsidian-tab-switcher",
+        "branch": "master"
     }
 ]


liamcain:
- [parentSplit](https://github.com/Quorafind/Obsidian-Tab-Switcher/blob/04b6c99801bd57da708288d02b09a9efd0fe5901/tabSwitcherIndex.ts#L10) use parent instead of parentSplit. parentSplit will get removed from Obsidian soon
- [getCurrentLeavesInParent](https://github.com/Quorafind/Obsidian-Tab-Switcher/blob/04b6c99801bd57da708288d02b09a9efd0fe5901/tabSwitcherIndex.ts#L49) this should return `leaf.parent.children` instead of iterating over all the leaves
- [console.log(key);](https://github.com/Quorafind/Obsidian-Tab-Switcher/blob/04b6c99801bd57da708288d02b09a9efd0fe5901/tabSwitcherIndex.ts#L100) remove this
- [registerDomEvent](https://github.com/Quorafind/Obsidian-Tab-Switcher/blob/04b6c99801bd57da708288d02b09a9efd0fe5901/tabSwitcherIndex.ts#L147) isn't leafEl just an element in the modal? you don't need to register the dom event on the plugin if that's the case.
- [TabSwitcherModal extends Modal](https://github.com/Quorafind/Obsidian-Tab-Switcher/blob/04b6c99801bd57da708288d02b09a9efd0fe5901/tabSwitcherIndex.ts#L36) why not just use the FuzzySuggestModal instead?
- [Fuse](https://github.com/Quorafind/Obsidian-Tab-Switcher/blob/04b6c99801bd57da708288d02b09a9efd0fe5901/tabSwitcherIndex.ts#L118) I'm pretty sure creating the fuse instance is expensive. I think you should just keep a reference to the fuse instance then only update the fuse search when updating the filter.

liamcain:
@Quorafind you have too many plugins that I think you forgot about this one!

ericaxu:

PR #1408, opened 2022-12-09
Add set-in-obsidian time planner

sandorex:

## Repo URL

Link to my plugin: https://github.com/sandorex/set-in-obsidian-plugin

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [var leaf](https://github.com/sandorex/set-in-obsidian-plugin/blob/e08d8cb437372a3ab9bdf1bfe26f7f18ba1eef9a/src/main.ts#L62) I recommend you change all instances of `var` to either `const` or `let`. `var` has function-level scope, so it can easily lead to bugs if you're not careful. Here's a pretty good guide with examples of why not to use `var`: [why `var` is obsolete](https://javascript.plainenglish.io/4-reasons-why-var-is-considered-obsolete-in-modern-javascript-a30296b5f08f)
- [View](https://github.com/sandorex/set-in-obsidian-plugin/blob/e08d8cb437372a3ab9bdf1bfe26f7f18ba1eef9a/src/view.ts#L25) You should extend ItemView instead of view. It will give you better file options in the options menu.
- [Show Timeline Ribbon Icon](https://github.com/sandorex/set-in-obsidian-plugin/blob/e08d8cb437372a3ab9bdf1bfe26f7f18ba1eef9a/src/settings.ts#L79) 
- [Show Timeline Ribbon Icon](https://github.com/sandorex/set-in-obsidian-plugin/blob/e08d8cb437372a3ab9bdf1bfe26f7f18ba1eef9a/src/settings.ts#L79) As of v1.1.0 of Obsidian, the ribbon can be customized within Obsidian. So users can choose what items to include and in what order all in one place. Because if this, we're recommending that new plugins avoid introducing their own toggles for ribbon items.
- [await](https://github.com/sandorex/set-in-obsidian-plugin/blob/e08d8cb437372a3ab9bdf1bfe26f7f18ba1eef9a/src/view.ts#L197) no need to await here. This is an event callback so it's nonblocking

sandorex:
I have applied the requested changes, thanks for the help

sandorex:
I have applied the changes, feel free to take a look @liamcain

PR #1414, opened 2022-12-11
Add Toggle Case Plugin

MatthewAlner:

## Repo URL

Link to my plugin:
[obsidian-toggle-case](https://github.com/MatthewAlner/obsidian-toggle-case)

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [Sync case multi-cursor](https://github.com/MatthewAlner/obsidian-toggle-case/blob/e6a213f401a6f86cf5e9c74c450c11b82fafdae7/src/main.ts#L135) why make this an option? Seems like it would make the most sense to just always apply the transformations to every cursor.
- [wordRangeAtPos](https://github.com/MatthewAlner/obsidian-toggle-case/blob/e6a213f401a6f86cf5e9c74c450c11b82fafdae7/src/utils.ts#L105) can't you use `wordAt` for this? (<https://codemirror.net/docs/ref/#state.EditorState.wordAt>)

PR #1419, opened 2022-12-13
Add File Color plugin

ecustic:

## Repo URL

Link to my plugin: [File Color](https://github.com/ecustic/obsidian-file-color)

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [fileColor.path.replace(oldPath, newFile.path)](https://github.com/ecustic/obsidian-file-color/blob/80b71ee673126207d98b2d39d7fa6f7c81b7d1e4/src/FileColorPlugin.ts#L42) I don't think you need to call .replace() here. You should be able to just reassign path. fileColor.path = newFile.path
- [await this.saveSettings()](https://github.com/ecustic/obsidian-file-color/blob/80b71ee673126207d98b2d39d7fa6f7c81b7d1e4/src/FileColorPlugin.ts#L44) Small performance suggestion: Maybe create a new function called requestSave that wraps your saveSettings function in [debounce](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L772). Then in your event callbacks, I would suggest calling requestSave and not awaiting it. That should improve performance when a user renames an entire folder (you don't want to bother writing the settings to disk each time. There's also no need to wait for the settings to save before applying the colors to the file explorer, so it should feel snappier
- [applyColorStyles](https://github.com/ecustic/obsidian-file-color/blob/80b71ee673126207d98b2d39d7fa6f7c81b7d1e4/src/FileColorPlugin.ts#L55) I don't think you need to call this on delete
- [onunload](https://github.com/ecustic/obsidian-file-color/blob/80b71ee673126207d98b2d39d7fa6f7c81b7d1e4/src/FileColorPlugin.ts#L62) You should delete the style tag on unload

ecustic:
@liamcain I've applied all of your suggestions. I've also added a short debounce to `applyColorStyles` as it was also being called multiple times in succession when moving/renaming any folder with children.

I was originally under the impression that the `rename` event was only called on the file that was actually being renamed (hence the `path.replace(oldPath, newFile.path)` calls), and not for all of the files that are affected, which actually makes things much simpler.

You can see all of the changes in the [1.0.2 release](https://github.com/ecustic/obsidian-file-color/releases/tag/1.0.2). 😃

PR #1420, opened 2022-12-13
add obsidian plugin paste as html

maotong06:
 ## Repo URL
 Link to my plugin: https://github.com/maotong06/obsidian-paste-as-html-plugin
 
 ## Release Checklist
   
   * [x]   Windows
   * [x]   macOS
   * [x]   Linux
 * [x]  My GitHub release contains all required files
   
   * [x]  `main.js`
   * [x]  `manifest.json`
 * [x]  I have read the tips in https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md and have self-reviewed my plugin to avoid these common pitfalls.


maotong06:
i am sorry, the last merge was deleted by mistake

liamcain:
- [hotkeys](https://github.com/maotong06/obsidian-paste-as-html-plugin/blob/aa30b2528e097be6c465fd0fe3fb83fd0437665f/main.ts#L8) While there's no hard rule against it, we recommend against providing a default hotkey when possible. The reason being that it's easy to pick a hotkey that a user already has configured and they could get confused when that key doesn't do what they expect. Also, it's hard choosing a safe default hotkey that's available for all operating systems. Also, alt-shift-v won't work on macOS since that inserts a unicode character instead.

PR #1424, opened 2022-12-14
Add File Forgetting Curve plugin

ptrsvltns:

## Repo URL

Link to my plugin: https://github.com/ptrsvltns/file-forgetting-curve-obsidian

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [dayjs](https://github.com/ptrsvltns/file-forgetting-curve-obsidian/blob/0193e7274c148689988c8935f36e1673530c533d/src/view.ts#L28) you should assign dayjs() to a variable outside of the for loop so that you don't need to reconstruct it for each file in the list
- [width](https://github.com/ptrsvltns/file-forgetting-curve-obsidian/blob/0193e7274c148689988c8935f36e1673530c533d/src/view.ts#L65) you should move all the styles into CSS so that they are more easily adaptable by themes
- [var](https://github.com/ptrsvltns/file-forgetting-curve-obsidian/blob/0193e7274c148689988c8935f36e1673530c533d/src/view.ts#L156) I recommend you change all instances of `var` to either `const` or `let`. `var` has function-level scope, so it can easily lead to bugs if you're not careful. Here's a pretty good guide with examples of why not to use `var`: [why `var` is obsolete](https://javascript.plainenglish.io/4-reasons-why-var-is-considered-obsolete-in-modern-javascript-a30296b5f08f)
- [if (count % 5 != 0) return](https://github.com/ptrsvltns/file-forgetting-curve-obsidian/blob/0193e7274c148689988c8935f36e1673530c533d/src/plugin.ts#L73) What's the reasoning behind this? Why are you showing it every 5th file-open?

PR #1425, opened 2022-12-14
Add Table Code Block plugin

ptrsvltns:

## Repo URL

Link to my plugin: https://github.com/ptrsvltns/table-code-block-obsidian

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5548,5 +5548,13 @@
         "description": "Hidden Folder",
         "repo": "ptrsvltns/hidden-folder-obsidian",
         "branch": "main"
+    },
+    {
+        "id": "table-code-block-obsidian",
+        "name": "Table Code Block",
+        "author": "ptrsvltns",
+        "description": "Table Code Block",
+        "repo": "ptrsvltns/table-code-block-obsidian",
+        "branch": "main"
     }
 ]


liamcain:
- [e.ctrlKey && (e.key == "v" || e.key == "V")](https://github.com/ptrsvltns/table-code-block-obsidian/blob/4bddaeda6d8042bd8c112b5c5d6c65c773893b13/src/plugin.ts#L136) this doesn't cross-platform. macOS uses command-V and command-C for copy/paste
- [clipboard](https://github.com/ptrsvltns/table-code-block-obsidian/blob/4bddaeda6d8042bd8c112b5c5d6c65c773893b13/src/plugin.ts#L133) no need to use electron's clipboard.write. You can just use navigator.clipboard.writeText
- [marginRight](https://github.com/ptrsvltns/table-code-block-obsidian/blob/4bddaeda6d8042bd8c112b5c5d6c65c773893b13/src/plugin.ts#L501) You should avoid assigning styles via javascript and instead move all these styles into CSS so that they are more easily adaptable by themes.
- [await](https://github.com/ptrsvltns/table-code-block-obsidian/blob/4bddaeda6d8042bd8c112b5c5d6c65c773893b13/src/plugin.ts#L744) This shouldn't need to be wrapped in an [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE).
- [await](https://github.com/ptrsvltns/table-code-block-obsidian/blob/4bddaeda6d8042bd8c112b5c5d6c65c773893b13/src/plugin.ts#L762) same here, the [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) looks unnecessary

joethei:
Hi @ptrsvltns do you still intend on submitting this plugin?

ericaxu:

PR #1427, opened 2022-12-14
Propose new plugin obsidian-console

dellermann:
Propose a new Obsidian plugin which renders console commands and their output.

## Repo URL

Link to my plugin: https://github.com/dellermann/obsidian-console

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


community-plugins.json
```diff
@@ -5788,5 +5788,13 @@
         "author": "Ryan Gomba",
         "repo": "ryangomba/obsidian-todo-sort",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-console",
+        "name": "Console Markdown Plugin",
```
liamcain, line 5794: You might want to choose a simpler name, it's a bit redundant to include the word "Plugin" in the plugin name.

community-plugins.json
```diff
@@ -5788,5 +5788,13 @@
         "author": "Ryan Gomba",
         "repo": "ryangomba/obsidian-todo-sort",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-console",
```
liamcain, line 5793: While there are a lot of plugins that include "obsidian" in the ID, it's not necessary to include it.

liamcain (approved):
The plugin looks good to me. Let me know if you'd like to change the name or ID before this gets merged.

community-plugins.json
dellermann: I'm much too late with my response. Yes, if possible, change the ID to a shorter one. Thank you!

liamcain: You'll need to change the name/ID on your end since they need to be changed here and in your plugin repo to match.

community-plugins.json
```diff
@@ -5924,5 +5924,13 @@
...
         "description": "Monthly Habit Calendar for DataviewJS. This plugin helps you render a calendar inside DataviewJS code block, showing your habit status within a month.",
         "repo": "hedonihilist/obsidian-habit-calendar"
+    },
+    {
+        "id": "console",
+        "name": "Console Markdown Plugin",
+        "author": "Daniel Ellermann",
+        "description": "Renders console commands and their output.",
+        "repo": "dellermann/obsidian-console",
+        "branch": "master"
```
liamcain, line 5934:
```suggestion
        "repo": "dellermann/obsidian-console"
```

PR #1431, opened 2022-12-16
Add Checklist Reset plugin

lhansford:

## Repo URL

Link to my plugin: https://github.com/lhansford/obsidian-checklist-reset

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5866,6 +5866,14 @@
         "description": "Create notes for audio files.",
         "repo": "jjmaldonis/obsidian-audio-notes",
         "branch": "master"
+    },
+    {
+        "id": "obsidian-checklist-reset",
+        "name": "Checklist Reset",
+        "author": "Luke Hansford",
+        "description": "Adds a command to reset the state of any checklists in a document",
+        "repo": "lhansford/obsidian-checklist-reset",
+        "branch": "main"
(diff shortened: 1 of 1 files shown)


liamcain:
- [replaceAll](https://github.com/lhansford/obsidian-checklist-reset/blob/71982e3d3d09726f9b6ab5ebc51330d48ba33e01/main.ts#L10) technically Obsidian checklists can have any non-space character to mark it as completed (e.g. `- [c]`, `- [/]`, etc) so you might want to handle this more generically.
- The current approach is susceptible to a lot of false positives. It's possible that the user has "[x]" in the middle of a line of text and it's not a task. You might want to use a more sophisticated regex to make sure that you're only finding tasks.

lhansford:
@liamcain Thanks for the feedback. I've made those suggested changes.

liamcain:
Looks good. You might still want to ensure that the list item is captured at the start of a new line, to avoid capturing things like `foo - [x] - bar`

lhansford:
Thanks for the suggestion, I'll see if I can patch that.

PR #1432, opened 2022-12-16
add plugin obsidian-disable-plugin-temporarily

nguyenvanduocit:

## Repo URL

Link to my plugin: [repo](https://github.com/nguyenvanduocit/obsidian-disable-plugin-temporarily)

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5556,5 +5556,13 @@
         "author": "MAKE.md",
         "repo": "Make-md/makemd",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-disable-plugin-temporarily",
+        "name": "Disable Plugin Temporarily",
+        "description": "Disable all plugin temporarily",
+        "author": "DuocNV",
+        "repo": "nguyenvanduocit/obsidian-disable-plugin-temporarily",
+        "branch": "main"
     }
 ]


liamcain:
- [containerEl.createEl](https://github.com/nguyenvanduocit/obsidian-disable-plugin-temporarily/blob/9294023377750558da10668a261acc4a5a85d970/src/SetingTab.ts#L20) You should avoid assigning styles via javascript and instead move all these styles into CSS so that they are more easily adaptable by themes.
- [saveSettings](https://github.com/nguyenvanduocit/obsidian-disable-plugin-temporarily/blob/9294023377750558da10668a261acc4a5a85d970/src/main.ts#L99) why are you calling saveSettings inside loadSettings? This seems unnecessary
- [splice](https://github.com/nguyenvanduocit/obsidian-disable-plugin-temporarily/blob/9294023377750558da10668a261acc4a5a85d970/src/main.ts#L69) don't mutate the list while you're iterating over it. I think you're better off just iterating the whole list first, then clearing the affectedPlugins.

joethei:
Hi @nguyenvanduocit,

nguyenvanduocit:
yes, I still intend, just some distract so that I can not see the notification, I will update it

nguyenvanduocit:
I don't think I'm suitable to continue submitting this plugin, I don't have time on it.

PR #1434, opened 2022-12-17
Add WakaTime plugin

alanhamlett:

## Repo URL

Link to my plugin: https://github.com/wakatime/obsidian-wakatime

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


alanhamlett:
@WhiteNoiseDev ready for review

alanhamlett:
Ping, this PR is ready.

liamcain:
- [this.logger](https://github.com/wakatime/obsidian-wakatime/blob/462513e31ee2ac29bb9e31c99f7b535dbc272b53/src/main.ts#L25) looks like this is not defined yet. This will throw an error.
- [promptForApiKey](https://github.com/wakatime/obsidian-wakatime/blob/462513e31ee2ac29bb9e31c99f7b535dbc272b53/src/main.ts#L90) your api key prompt isn't a singleton so it's possible for multiple prompts to get stacked on top of each other. This can be a bad user experience. Likewise, I suggest avoiding showing a prompt on plugin load. I think you're better off providing a Setting Tab with a place to set the API key. Currently, for example, if I don't have the API key ready to go, I might dismiss the modal, then the only way to proactively get the API key modal to appear is to wait for the heartbeat check.

alanhamlett:
Fixed with 80dae7b71a31010387297b74eea7c59a25cc3b08.

> * I think you're better off providing a Setting Tab

When considering this, I wanted to make it similar to the other WakaTime plugins where a menu item just re-opens the api key prompt. Is there any way to do that to re-use code instead of making a new settings tab?

liamcain:
I think a settings tab would be the easiest, most familiar UX for users, but if you're really trying to avoid it, I would suggest just fixing the following behaviors:
- don't prompt more than once (currently, if a user is missing an API key, the prompt will appear on each heartbeat check. The only way to avoid this is to disable the plugin)
- There should be a predictable way to get the api key prompt to appear _when desired_. This could be a command that you register, or maybe allow clicking the statusbar item to open the modal, or maybe both.

alanhamlett:
> * There should be a predictable way to get the api key prompt to appear _when desired_. This could be a command that you register, or maybe allow clicking the statusbar item to open the modal, or maybe both.

There's already a command to open the api key prompt. I'll stop it from auto-prompting more than once and re-release version 1.0.0.

alanhamlett:
Fixed in https://github.com/wakatime/obsidian-wakatime

alanhamlett:
Thank you!

PR #1437, opened 2022-12-18
Added obsidian-file-chucker to the list

kenlim:
Added obsidian-file-chucker to the list

## Repo URL

Link to my plugin: https://github.com/kenlim/file-chucker-plugin

## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


liamcain:
- [await app.vault.getAbstractFileByPath](https://github.com/kenlim/obsidian-file-chucker/blob/a0495cb7106f60ee658d43ba9d82b7ad67381fe9/src/main.ts#L157) this function isn't async. You shouldn't need the await
- [editorCallback](https://github.com/kenlim/obsidian-file-chucker/blob/a0495cb7106f60ee658d43ba9d82b7ad67381fe9/src/main.ts#L37) editorCallback is only called if the file is in edit mode. If a user prefers preview mode, this command won't be available. I suggest using [checkCallback](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L505) and checking for `this.app.workspace.getActiveViewOfType(MarkdownView)`

community-plugins.json
```diff
@@ -5788,5 +5788,12 @@
         "author": "Ryan Gomba",
         "repo": "ryangomba/obsidian-todo-sort",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-file-chucker",
+        "name": "Obsidian file chucker",
```
liamcain, line 5794: We're asking plugins to avoid using the word "Obsidian" in their plugin name to avoid desaturating the brand name and reserve the naming for first-party plugins (see: Obsidian Publish, Obsidian Sync).

community-plugins.json
```diff
@@ -5788,5 +5788,12 @@
         "author": "Ryan Gomba",
         "repo": "ryangomba/obsidian-todo-sort",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-file-chucker",
```
liamcain, line 5793: Similarly, you don't need to include the word "obsidian" in your ID.

community-plugins.json
kenlim: Have changed this to just "file-chucker"

community-plugins.json
kenlim: it's just "file-chucker" now.

kenlim:
Have removed await to getAbstractFileByPath.
Have also changed to the editorCallback, so this plugin now works in Preview mode!
Also changed the plugin name and repo name to match

PR #1438, opened 2022-12-18
fix: update ava description

louis030195:
Update of `ava` plugin name and description

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5274,9 +5274,9 @@
     },
     {
         "id": "ava",
-        "name": "Obsidian AI",
+        "name": "Ava",
         "author": "louis030195",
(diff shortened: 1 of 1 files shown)


liamcain:
Hi @louis030195, I discussed it with the team, and we have some concerns over the rebranding of this plugin. Calling it "Obsidian AI" is too easily mistakable for a first-party Obsidian product (following the same scheme as Obsidian Publish, Obsidian Sync). It also bears resemblance to "Notion AI" so it could be easily misconstrued that this is an Obsidian product in direct competition with Notion.

What's wrong with the name "Ava"? I thought it was a recognizable name already.

louis030195:
@liamcain No problem. Updated :)

PR #1439, opened 2022-12-18
Add "Boost Link Suggestions" plugin

jglev:

## Repo URL

Link to my plugin: https://github.com/jglev/obsidian-boost-link-suggestions

_(This plugin follows [this thread](https://discord.com/channels/686053708261228577/716028884885307432/1053860523646656563) in Discord.)_

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [X]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5881,6 +5881,13 @@
         "author": "dartungar",
         "description": "Improved Mermaid.js experience for Obsidian: visual toolbar with common elements & more",
         "repo": "dartungar/obsidian-mermaid"
+    },
+    {
+        "id": "boost-link-suggestions",
+        "name": "Boost Link Suggestions",
+        "author": "Jacob Levernier",
+        "description": "Alternative inline link suggester that orders results by link count and manual boosts.",
+        "repo": "jglev/obsidian-boost-link-suggestions"
     }
 ]
 


liamcain:
- [this.plugin.app.workspace.getActiveFile](https://github.com/jglev/obsidian-boost-link-suggestions/blob/2b5d5b6fd9439380b9fdbce32fa267eca307d38b/main.ts#L274) there are cases where this can be null, such as inside a Canvas card (where there is an editor but no active file). You should make sure to handle null here. Actually, you should be able to access the file from the EditorSuggestContext instead
- [queryWords.every](https://github.com/jglev/obsidian-boost-link-suggestions/blob/2b5d5b6fd9439380b9fdbce32fa267eca307d38b/main.ts#L101) up to you, but you can also use [prepareFuzzySearch](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L2910) to get the same matching algorithm that [[ already uses.

community-plugins.json
```diff
@@ -5788,5 +5788,12 @@
         "author": "Ryan Gomba",
         "repo": "ryangomba/obsidian-todo-sort",
         "branch": "main"
+    },
+    {
+        "id": "obsidian-boost-link-suggestions",
```
liamcain, line 5793: You don't need to include the word 'obsidian' in your ID.

jglev:
@liamcain, thank you! I appreciate your review!

I've incorporated your suggestions in a new branch [here](https://github.com/jglev/obsidian-boost-link-suggestions/compare/main...1.0.0). Are those in line with what you had in mind? If so, I'll merge and release them.

> Your plugin describes its behavior as "altering" the default link suggest, but it's not really altering it. It's creating a different suggest. One that's configurable. I'm not sure if anyone cares about the distinction there, but I was surprised to not see any monkeypatching at least.

I take your point -- I meant to convey that the plugin allows altering the order of inline suggestions from the plugin, rather than the default suggest function. I can see how it could be read either way. Would "Suggest inline links ordered by link count and manual boosts" read clearer to you?

> I was surprised to not see any monkeypatching at least.

This exposes the limits of my knowledge -- is monkeypatching officially supported / sanctioned in the API spec? (And do you mean using something like https://github.com/pjeby/monkey-around?) If not, I'm happy to use the current approach of implementing a new suggest trigger; but if so, I'm happy for the tip!

> You don't need to include the word 'obsidian' in your ID. Let me know if you'd like to change the ID before this gets merged (it's difficult to change later)

Thank you; I'd like to take you up on this, and drop the `obsidian-` prefix.

Again, thank you for your time and expertise!

liamcain:
> I've incorporated your suggestions in a new branch [here](https://github.com/jglev/obsidian-boost-link-suggestions/compare/main...1.0.0). Are those in line with what you had in mind? If so, I'll merge and release them.

Changes look good to me!

> I take your point -- I meant to convey that the plugin allows altering the order of inline suggestions from the plugin, rather than the default suggest function. I can see how it could be read either way. Would "Suggest inline links ordered by link count and manual boosts" read clearer to you?

Yeah, I think that reads better. Maybe "Alternative inline link suggester that orders results by link count and manual boosts" (maybe a bit wordy though)

> This exposes the limits of my knowledge -- is monkeypatching officially supported / sanctioned in the API spec? (And do you mean using something like https://github.com/pjeby/monkey-around?) If not, I'm happy to use the current approach of implementing a new suggest trigger; but if so, I'm happy for the tip!

Oh, it's not really sanctioned, when I see it I just usually look the other way. I'm not proposing that you switch to this approach, I think what you have is better and safer. I was just saying it was surprising given the description.

> Thank you; I'd like to take you up on this, and drop the obsidian- prefix.

Cool, sounds good. Just remember to remove it in all places. Needs to be dropped from this PR, your manifest.json, and the manifest.json of your latest Github Release.

jglev:
> Changes look good to me!

Great! Thank you for your suggestions! I can see how the plugin is stronger having incorporated them.

> Yeah, I think that reads better. Maybe "Alternative inline link suggester that orders results by link count and manual boosts" (maybe a bit wordy though)

That wording reads clearly to me -- I've gone ahead and incorporated it both in this PR and in the plugin manifest and `package.json` file. Thank you!

> Oh, it's not really sanctioned, when I see it I just usually look the other way. I'm not proposing that you switch to this approach, I think what you have is better and safer. I was just saying it was surprising given the description.

Ha, good to know : ) Thank you for explaining more!

> Cool, sounds good. Just remember to remove it in all places. Needs to be dropped from this PR, your manifest.json, and the manifest.json of your latest Github Release.

Done (in all places)! I've created a new [release](https://github.com/jglev/obsidian-boost-link-suggestions/releases/tag/1.0.0) reflecting all of these changes just now, as well.

jglev:
(I also just rebased on current `master` and force-pushed to my branch for this PR, hoping it would make everything easier for you to merge. Is that typically the advised flow here? If not, I'm sorry if that makes things harder!)

jglev:
Thank you, @liamcain!

PR #1451, opened 2022-12-20
Added Visually Numbered Headings plugin

platon-ivanov:

## Repo URL

Link to my plugin: [https://github.com/platon-ivanov/obsidian-releases-visually-numbered-headings](https://github.com/platon-ivanov/obsidian-releases-visually-numbered-headings)

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5563,5 +5563,13 @@
         "description": "This is a markdown export plugin for Obsidian.",
         "author": "bingryan",
         "repo": "bingryan/obsidian-markdown-export-plugin"
+    },
+    {
+        "id": "obsidian-visually-numbered-headings",
+        "name": "Visually Numbered Headings",
+        "description": "Heading numbering without modifying your files.",
+        "author": "Platon Ivanov",
+        "repo": "platon-ivanov/obsidian-visually-numbered-headings",
+	"branch": "master"
     }
 ]


liamcain:
- [editor-change](https://github.com/platon-ivanov/obsidian-visually-numbered-headings/blob/397172e34ec62605d17aef46cb526e7904ccf80e/src/main.tsx#L33) what is this for? It shouldn't be necessary to trigger the rerender on change.
- [mdNumGenCache](https://github.com/platon-ivanov/obsidian-visually-numbered-headings/blob/397172e34ec62605d17aef46cb526e7904ccf80e/src/main.tsx#L50) I think this approach is clever, unfortunately, I think there are some cases were it won't work reliably. It won't work if the file is in a heading embed `![[some file#heading]]` and I think it might have issues on longer documents since Obsidian does some optimizations under the hood to prevent long documents from being rendered all at once. I think you're better off using getSectionInfo to get the lineStart and lineEnd, then using that data to calculate the heading level from the metadataCache. That should work more reliably.

joethei:
Hi @platon-ivanov do you still intend on submitting this plugin?

platon-ivanov:
Hello @joethei! Thanks for looking into the plugin! Yes, I still intend on submitting it.

liamcain:
@platon-ivanov have you seen my comment above? https://github.com/obsidianmd/obsidian-releases/pull/1451#issuecomment-1368081368 I'm reviewing the code before this plugin gets merged so I pointed out a few things I found in the review process.

ericaxu:
Hi @platon-ivanov , do you still intend to submit this plugin?

PR #1452, opened 2022-12-21
Add Link Nodes in Canvas

Quorafind:

## Repo URL

Link to my plugin: https://github.com/Quorafind/Obsidian-Link-Nodes-In-Canvas

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [tempEdge.attach();](https://github.com/Quorafind/Obsidian-Link-Nodes-In-Canvas/blob/bad276b3223e9a97ddb95f192b8426db53830bfa/linkNodesInCanvasIndex.ts#L59-L60) this shouldn't be necessary. addEdge should be enough.
- [canvas.requestSave();](https://github.com/Quorafind/Obsidian-Link-Nodes-In-Canvas/blob/bad276b3223e9a97ddb95f192b8426db53830bfa/linkNodesInCanvasIndex.ts#L32) don't call requestSave for each iteration in this loop, just call it once after the forEach.

Quorafind:
I move `canvas.requestSave()` outside the loop.

But I think if I want to see the edge render real time , I need to leave `.attach()` there, else I cannot see the edge is renderred real-time. What do you think? @liamcain

liamcain:
That's surprising. It should be rendered automatically in the next frame. But either way, attach isn't super expensive so it's fine.

PR #1453, opened 2022-12-21
Add obsidian-local-images-plus plugin

Sergei-Korneev:
This repo is a reincarnation of obsidian-local-images plugin which main aim was downloading images in md notes to local storage.

## Repo URL

Link to my plugin:

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


Sergei-Korneev (commented):
Sorry, I have made some mistypes.
	"description": "Local Images plus plugin is a reincarnation of obsidian-local-images which main aim was downloading images in md notes to local storage."

liamcain:
- [proccessPage](https://github.com/Sergei-Korneev/obsidian-local-images-plus/blob/0a7c4d9e047684877bd0e6fca7b8ff208f971a2a/src/main.ts#L30) typo: process
- [I could not find the way to access newly pasted content immediatily, after pasting, Plugin's API returns old text for a while.](https://github.com/Sergei-Korneev/obsidian-local-images-plus/blob/0a7c4d9e047684877bd0e6fca7b8ff208f971a2a/src/main.ts#L305) this is probably too much information for a user to need to know. Maybe just gave the relevant info.
- [require](https://github.com/Sergei-Korneev/obsidian-local-images-plus/blob/0a7c4d9e047684877bd0e6fca7b8ff208f971a2a/src/utils.ts#L12) you should use the vault's [`adapter.read`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L732) which will read the file properly across all operating systems (including mobile)

Sergei-Korneev:
Thank you so much. Will be added in 0.14.5

liamcain:
You're referring to this event in the API? https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4247. Sorry, mind pointing me at the code that's returning the wrong thing?

Sergei-Korneev:
Yeah, maybe I expressed myself badly, I mean when you try immediately get a note content in the "editor-paste" event e.g. 


this.app.workspace.on(
      "editor-paste",
        ( evt: ClipboardEvent, editor: Editor, info: MarkdownView) => {


const data = this.app.vault.read(this.app.workspace.getActiveFile());
//or
const data = this.app.workspace.activeLeaf.view.data;
}

it returns the previous content, despite the fact, that in editor window we of course can see the actual content. But the event called on timer a second later 
return the actual content with any of those methods.
I presume, it takes some time to write on disk maybe and  expose actual content to the plugin. Am I right?

liamcain:
Oh, yeah. The paste event is fired before the content gets written to the file so that you can intercept it if you want. Maybe it's too much to change from the current implementation, but you do have access to the clipboardData from the PasteEvent. So instead of trying to read the whole file contents, you could just modify the clipboardData before it gets inserted.

Sergei-Korneev:
I thought about that too. Thanks.

PR #1455, opened 2022-12-21
Add 'google-photos' plugin

alangrainger:

## Repo URL

Link to my plugin: https://github.com/alangrainger/obsidian-google-photos

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [maxHeight](https://github.com/alangrainger/obsidian-google-photos/blob/2a68abf11b59431504b8721d506b3a51bf3a1a1b/main.ts#L50) You should avoid assigning styles via javascript. Instead, add a class to the containerEl and move all these styles into CSS so that they are more easily adaptable by themes.
- [setDesc](https://github.com/alangrainger/obsidian-google-photos/blob/2a68abf11b59431504b8721d506b3a51bf3a1a1b/main.ts#L99) maybe use the description to add some information about where the user can find this info.
- [Image filename format](https://github.com/alangrainger/obsidian-google-photos/blob/2a68abf11b59431504b8721d506b3a51bf3a1a1b/main.ts#L143) I suggest adding the default values as placeholder text here
- [Thumbnail height](https://github.com/alangrainger/obsidian-google-photos/blob/2a68abf11b59431504b8721d506b3a51bf3a1a1b/main.ts#L134) I suggest adding the default value as placeholder text here too
- [setDesc](https://github.com/alangrainger/obsidian-google-photos/blob/2a68abf11b59431504b8721d506b3a51bf3a1a1b/main.ts#L135) maybe give some info about what the format is for this value. Is it pixels? A percentage? Either?
- [close](https://github.com/alangrainger/obsidian-google-photos/blob/2a68abf11b59431504b8721d506b3a51bf3a1a1b/main.ts#L199) maybe wait until the request is done loading before closing the modal. That way users on slow internet will know that it's still processing
- [width](https://github.com/alangrainger/obsidian-google-photos/blob/2a68abf11b59431504b8721d506b3a51bf3a1a1b/main.ts#L213) same here, I suggest moving this to CSS

alangrainger:
All suggestions fixed in v1.1.5 - thanks for the review @liamcain !

PR #1456, opened 2022-12-21
7596ff: obsidian-hyphenation

7596ff:

## Repo URL

Link to my plugin: https://github.com/7596ff/obsidian-hyphenation

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [this.registerEvent](https://github.com/7596ff/obsidian-hyphenation/blob/1b6ebe70fa51518ec4334bcb1d515ba58c58f017/src/main.ts#L38) looks like you're registering your callback for file-open twice. I think what you meant to do was the following:

```ts
this.app.workspcae.onLayoutReady(() => {
  this.setLang(navigator.language);
  this.registerEvent(this.app.workspace.on("file-open", this.fileOpen));
});
```

- [document.getElementsByClassName](https://github.com/7596ff/obsidian-hyphenation/blob/1b6ebe70fa51518ec4334bcb1d515ba58c58f017/src/main.ts#L11) this approach won't work correctly. You're forgetting that multiple files can be open at once. What you really want here is the container element for the activeLeaf. I would recommend switching your file-open event to [`active-leaf-change`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4202). Then you'll be given the new active leaf. From there, you can get the containerEl: `leaf.containerEl`. That should work a bit more reliably.

7596ff:
Thank you so much for the detailed feedback. I'll make the changes soon!

PR #1461, opened 2022-12-22
Update community-plugins.json

willasm:

## Repo URL

Link to my plugin: https://github.com/willasm/obsidian-open-weather

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


willasm:
Can someone explain to me how to fix this? First it asked me to correct the entry, which I did. The entry after mine didn't add the first brace { and now it is saying I submitted that entry. A little help would be appreciated.

liamcain:
Hi, no need to worry about that validation error. I'll fix the pull request once it's ready to be merged.

- [view.setViewData(editor,false);](https://github.com/willasm/obsidian-open-weather/blob/e66bda521490853dd31f23716850bd65fb12b56f/main.ts#L724) Just to be clear: if the user is in Preview mode, this will only affect the output display and the change won't be saved to the file. I don't think that's what you want to do. I think if the user is in Preview mode, you instead want to [modify](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L3760) the file. Optionally, just add a check here to make sure the user is in edit mode.
- [catch](https://github.com/willasm/obsidian-open-weather/blob/e66bda521490853dd31f23716850bd65fb12b56f/main.ts#L182) you should just `return` here. There's no point to executing the rest of this function if the API failed.
- [weatherString](https://github.com/willasm/obsidian-open-weather/blob/e66bda521490853dd31f23716850bd65fb12b56f/main.ts#L230) returning this error message means that the error message will be what gets inserted into the note. I don't think that's what you want to do
- [ribbonIconEl.addClass('my-plugin-ribbon-class');](https://github.com/willasm/obsidian-open-weather/blob/e66bda521490853dd31f23716850bd65fb12b56f/main.ts#L292) remove this
- [this.newFileNamePath](https://github.com/willasm/obsidian-open-weather/blob/e66bda521490853dd31f23716850bd65fb12b56f/main.ts#L438) it looks like `this.newFileNamePath` is unused. I recommend removing it along with any of the event listeners you have that are updating the value.

willasm:
I resubmitted this plugin as v1.1.0 with the completed changes, but it is saying I cannot use Obsidian in the ID?
How do I fix this?

liamcain:
- [getWeatherString](https://github.com/willasm/obsidian-open-weather/blob/ed925e9d68ec444db0a706c7202303aa285135bd/main.ts#L298) you removed the try catch around getWeather. This means that it will throw an error silently. A common pattern might be to show the user a [`Notice`](https://github.com/obsidianmd/obsidian-api/blob/606526e212cb0018690a729d7f0d1c8797071d6a/obsidian.d.ts#L2542) when an error occurs so that the plugin doesn't just fail silently.

> [view.setViewData(editor,false);](https://github.com/willasm/obsidian-open-weather/blob/e66bda521490853dd31f23716850bd65fb12b56f/main.ts#L724) Just to be clear: if the user is in Preview mode, this will only affect the output display and the change won't be saved to the file. I don't think that's what you want to do. I think if the user is in Preview mode, you instead want to [modify](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L3760) the file. Optionally, just add a check here to make sure the user is in edit mode.

This issue was not addressed.

liamcain:
Once you've addressed the issues above, feel free to ping me to re-review. Thanks!

willasm:
@liamcain 
This has been updated in release 1.2.0.
I have inserted the try catch which shows a notice on error. It still returns an empty string so the calling code knows that an error has occurred.
I have also added a check for edit mode with the other issue as well as changed it to use modify.

liamcain:
Thank you for your patience. Sorry for the long wait, I'm happy to finally get this into the list of Community Plugins 😄

PR #1462, opened 2022-12-22
Add obsidian gpt

jmilldotdev:

## Repo URL

Link to my plugin: https://github.com/jmilldotdev/obsidian-gpt

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


community-plugins.json
```diff
@@ -5563,5 +5563,12 @@
         "description": "This is a markdown export plugin for Obsidian.",
         "author": "bingryan",
         "repo": "bingryan/obsidian-markdown-export-plugin"
+    },
+        {
+        "id": "obsidian-gpt",
+        "name": "Obsidian GPT",
```
liamcain, line 5569: Hi, we're asking plugins to avoid using the word "Obsidian" in their plugin name to avoid over-saturating the brand name and reserve the naming for first-party plugins (like Obsidian Publish, Obsidian Sync, etc).

community-plugins.json
```diff
@@ -3608,7 +3608,8 @@
     {
         "id": "obsidian-textgenerator-plugin",
         "name": "Text Generator",
-        "description": "Text generator is a handy plugin that helps you generate text content using GPT-3 (OpenAI).",
+        "description": "Text generator is a handy plugin that helps you generate text content using
```
liamcain, line 3611:
```suggestion
        "description": "Text generator is a handy plugin that helps you generate text content using GPT-3 (OpenAI).",
```

community-plugins.json
```diff
@@ -3609,6 +3609,7 @@
         "id": "obsidian-textgenerator-plugin",
         "name": "Text Generator",
         "description": "Text generator is a handy plugin that helps you generate text content using GPT-3 (OpenAI).",
+	-3 (OpenAI).",
```
liamcain, line 3612:
```suggestion

```

liamcain:
- [private view: SettingsItemView](https://github.com/jmilldotdev/obsidian-gpt/blob/12e406ed282560ac59768982c88093f92f39a8ad/main.ts#L21) You don't want to store a reference to your view in the plugin. If the user closes the item view, this prevents the view from being garbage collected. It also doesn't look like it's being used at all, so it looks safe to remove.
- [onunload](https://github.com/jmilldotdev/obsidian-gpt/blob/12e406ed282560ac59768982c88093f92f39a8ad/main.ts#L173) I also recommend you don't detach leaves on unload. This works for when the plugin gets disabled/uninstalled, but when the plugin gets updated, it means the view gets moved around in the user's workspace which is often unwanted. I recommend just removing this and letting the user manually close the views if they don't want them anymore.
- [Show Model Settings](https://github.com/jmilldotdev/obsidian-gpt/blob/12e406ed282560ac59768982c88093f92f39a8ad/main.ts#L158) I suggest not using a checkCallback for "Show" commands. "Show" has a predictable meaning for Obsidian views. I made an example repo to show what "Show" commands should do: <https://github.com/liamcain/obsidian-example-custom-view/blob/194d4557abb52a41c0380a2f1ec75334d9ce2775/src/main.ts#L33>

jmilldotdev:
@liamcain addressed all feedback in latest release

liamcain:
Looks good. Do you mind changing the name in the manifest.json file too? https://github.com/jmilldotdev/obsidian-gpt/blob/12a520ad4915eaaeacf499862b44575c19b40093/manifest.json#L3 this is the name that will appear inside the sidebar and plugin list.

jmilldotdev:
done

community-plugins.json
```diff
@@ -5565,6 +5565,12 @@
         "author": "bingryan",
         "repo": "bingryan/obsidian-markdown-export-plugin"
     },
+    {
+        "id": "obsidian-gpt",
+        "name": "GPT",
+        "description": "GPT & Large Language Model completions in Obsidian editor via API",
+        "author": "Jonathan Miller",
+        "repo": "jmilldotdev/obsidian-gpt"
     {
```
liamcain, line 5574:
```suggestion
        "repo": "jmilldotdev/obsidian-gpt"
    },
    {
```

PR #1470, opened 2022-12-26
added clevcode/obsidian-terminal-plugin

clevcode:

## Repo URL

Link to my plugin: https://github.com/clevcode/obsidian-terminal-plugin

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5563,5 +5563,12 @@
         "description": "This is a markdown export plugin for Obsidian.",
         "author": "bingryan",
         "repo": "bingryan/obsidian-markdown-export-plugin"
+    },
+    {
+        "id": "obsidian-terminal-plugin",
+        "name": "Obsidian Terminal Plugin",
+        "description": "This plugin adds two commands, one for opening the current file in a terminal-based editor (Neovim by default) of your choice in a split below the file in question, and one for opening a terminal within the directory of the current note (or the vault root if no note is selected). It should work on Linux and macOS for now. Planning to allow adding custom commands that runs any terminal-based tool in a split, tab or popout window. Could be used to easily add support for powerful command line based tools such as fzf, ripgrep and of course things such as git-integration etc.",
+        "author": "Joel Eriksson",
+        "repo": "clevcode/obsidian-terminal-plugin"
     }
 ]


community-plugins.json
```diff
@@ -5563,5 +5563,12 @@
         "description": "This is a markdown export plugin for Obsidian.",
         "author": "bingryan",
         "repo": "bingryan/obsidian-markdown-export-plugin"
+    },
+    {
+        "id": "obsidian-terminal-plugin",
+        "name": "Obsidian Terminal Plugin",
```
liamcain, line 5569: Hi, we're asking plugins to avoid using the word "Obsidian" in their plugin name to avoid over-saturating the brand name and reserve the naming for first-party plugins (see: Obsidian Publish, Obsidian Sync). Also including the word "Plugin" in your plugin name is a bit redundant.

liamcain:
- [this.app.workspace.setActiveLeaf](https://github.com/clevcode/obsidian-terminal-plugin/blob/8dc60e7f9e29778e5d2a221f87c05c07da43884c/main.ts#L76) this line is redundant. You already passed active: true in the setViewState
- [setActiveLeaf](https://github.com/clevcode/obsidian-terminal-plugin/blob/8dc60e7f9e29778e5d2a221f87c05c07da43884c/main.ts#L69) same here
- [revealLeaf](https://github.com/clevcode/obsidian-terminal-plugin/blob/8dc60e7f9e29778e5d2a221f87c05c07da43884c/main.ts#L68) no need to reveal a leaf that was newly created
- All of this reading of external files and font loading should not be done in the onload function. You should move it all into [`onLayoutReady`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3626). `onload` is called synchronously, so it's best for the function to exit quitly to avoid slowing down Obsidian load times. Wrapping longer processes in onLayoutReady ensures that the workspace is loaded and avoids the app hanging on startup.
- [hideStatusBar](https://github.com/clevcode/obsidian-terminal-plugin/blob/8dc60e7f9e29778e5d2a221f87c05c07da43884c/main.ts#L121) same here, you don't want to be querying the dom until obsidian is done loading.
- [setAttribute](https://github.com/clevcode/obsidian-terminal-plugin/blob/8dc60e7f9e29778e5d2a221f87c05c07da43884c/main.ts#L218) looks like this line is incomplete
- [setAttribute](https://github.com/clevcode/obsidian-terminal-plugin/blob/8dc60e7f9e29778e5d2a221f87c05c07da43884c/main.ts#L240) You should avoid assigning styles via javascript. Instead, just add a class to the element and move all these styles into CSS so that they are more easily adaptable by themes.
- [filePath](https://github.com/clevcode/obsidian-terminal-plugin/blob/8dc60e7f9e29778e5d2a221f87c05c07da43884c/main.ts#L343) why not just get the active file from inside onOpen? then you don't need to track this.

ericaxu:

PR #1471, opened 2022-12-26
Update community-plugins.json - Add Smart Connections

brianpetro:

## Repo URL

Link to my plugin: https://github.com/brianpetro/obsidian-smart-connections

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [view](https://github.com/brianpetro/obsidian-smart-connections/blob/f8f2c363687e1f68bfcd97dc483f669954b50462/main.js#L172) I suggest removing this this function. It looks like you're mostly using it to register events and update the view from the event listeners. Instead, I suggest you move the event registration into your view. That will remove this extra layer of indirection.
- [this.app.workspace.layoutReady](https://github.com/brianpetro/obsidian-smart-connections/blob/f8f2c363687e1f68bfcd97dc483f669954b50462/main.js#L149) this pattern is a bit outdated. Instead you should just wrap everything in [`onLayoutReady`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3626). The callback will get called immediately if the layout is already ready. This will avoid the extra code duplication here.
- [getMostRecentLeaf](https://github.com/brianpetro/obsidian-smart-connections/blob/f8f2c363687e1f68bfcd97dc483f669954b50462/main.js#L1040) don't use getMostRecentLeaf here. Instead you should use the following:
```ts
let mod = Keymap.isModEvent(event);
let leaf = this.app.workspace.getLeaf(mod);
```
This will properly handle if the meta/ctrl key is pressed and create a leaf in the correct spot.

brianpetro:
Hi @liamcain 

I was able to address all three of your suggestions.

1. I removed the `view()` function from the Plugin class and moved all the view-dependent functions into the View class, including the `file-open` event registration.
2. The `layoutReady` logic was replaced with `onLayoutReady` as described.
3. And `getMostRecentLeaf` was also removed in favor of the suggested code.

Thanks for the feedback. The plugin seems snappier already :)

PR #1473, opened 2022-12-27
[PLUGIN] Adding Canvas CSS Class

Lisandra-dev:

## Repo URL

Link to my plugin: https://github.com/Lisandra-dev/obsidian-canvas-css-class/

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
Just a heads up, you don't need to worry about keeping your branch up to date with master. I'll rebase it once I review it.

Lisandra-dev:
Oh okay ! Thanks for the information :)

liamcain:
- [document](https://github.com/Lisandra-dev/obsidian-canvas-css-class/blob/750308db5e432991c35ca8eff3cb656ec35db4dc/plugin/main.ts#L188) this isn't the correct way to get the currently active leaf. Instead, you should just check for `activeLeaf?.view?.type === 'canvas'`. querying the document won't work with popout windows, it's also not an efficient or accurate way to get the view data.
- [workspace-leaf.mod-active .view-content](https://github.com/Lisandra-dev/obsidian-canvas-css-class/blob/750308db5e432991c35ca8eff3cb656ec35db4dc/plugin/utils.ts#L74) this is not a good way to access the current active leaf. You're better off accessing workspace.activeLeaf.contentEl
- [querySelector](https://github.com/Lisandra-dev/obsidian-canvas-css-class/blob/750308db5e432991c35ca8eff3cb656ec35db4dc/plugin/utils.ts#L86) a canvas can be open in more than one leaf so this approach won't work. You could change this to use [`iterateAllLeaves`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4157), then check if leaf.view instance FileView and leaf.view.type === 'canvas'.
- [addCanvasPathAndCanvasFile](https://github.com/Lisandra-dev/obsidian-canvas-css-class/blob/750308db5e432991c35ca8eff3cb656ec35db4dc/plugin/utils.ts#L129) if this function is just for querying to find canvas files, I don't think you need it. You're better off using [`iterateAllLeaves`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4157) instead.

I think we'll make some changes to the canvas API so that you can just add the cssClass into the canvas file directly. That way you don't need to save it into your plugin settings. For now, your current approach is the best I can offer, but hopefully this can be simplified in the near future.

Lisandra-dev:
I just want to clarify for the second point and last point.

1. [This function](https://github.com/Lisandra-dev/obsidian-canvas-css-class/blob/750308db5e432991c35ca8eff3cb656ec35db4dc/plugin/utils.ts#L70) is to get where I need to put the class, in the active `.view-content` or in `body`
2. The [addCanvasPathAndCanvasFile](https://github.com/Lisandra-dev/obsidian-canvas-css-class/blob/750308db5e432991c35ca8eff3cb656ec35db4dc/plugin/utils.ts#L129) check the append method from settings (`body` or `view-content`) and add the class `.canvas-file` and the `[data-canvas-path]` informations. 


Also, for the [iterateAllLeaves](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4157) function : what is returned ? One leaf ? All canvas opened ? 
Thanks you!

Lisandra-dev:
For the part `leaf instanceof FileView` it always returns false. It's normal ? Or I miss something ?

Lisandra-dev:
I choose to use 
```js
const viewState = leaf.getViewState();
if (viewState.type === "canvas" && viewState.state.file === filePath)
```
instead.

Change requested was made

Lisandra-dev:
Requested changes was made!

liamcain:
- [workspace: Workspace](https://github.com/Lisandra-dev/obsidian-canvas-css-class/blob/c445e1cb41012783ffa45809cff7bef080f43c35/plugin/main.ts#L60) small nitpick but you don't really need to pass `workspace` in here. You can access it from `this.app.workspace`
- [allSpecificLeafs](https://github.com/Lisandra-dev/obsidian-canvas-css-class/blob/c445e1cb41012783ffa45809cff7bef080f43c35/plugin/main.ts#L63) this check is unnecessary, you won't iterate over the same leaf multiple times.
- [getSpecificLeaf](https://github.com/Lisandra-dev/obsidian-canvas-css-class/blob/c445e1cb41012783ffa45809cff7bef080f43c35/plugin/main.ts#L60) just a style nitpick but you call this "getSpecificLeaf" but multiple leaves can match so this is a bit confusing.
- [leavesTypes](https://github.com/Lisandra-dev/obsidian-canvas-css-class/blob/c445e1cb41012783ffa45809cff7bef080f43c35/plugin/main.ts#L206) this is plural but doesn't need to be
- [async](https://github.com/Lisandra-dev/obsidian-canvas-css-class/blob/c445e1cb41012783ffa45809cff7bef080f43c35/plugin/modals/addClass.ts#L37) this doesn't need to be marked as async
- [then](https://github.com/Lisandra-dev/obsidian-canvas-css-class/blob/c445e1cb41012783ffa45809cff7bef080f43c35/plugin/main.ts#L33) this .then() doesn't doing anything
- [then](https://github.com/Lisandra-dev/obsidian-canvas-css-class/blob/c445e1cb41012783ffa45809cff7bef080f43c35/plugin/main.ts#L50) same here

Lisandra-dev:
Thank you!

PR #1474, opened 2022-12-27
Order List Plugin

lizard-heart:

## Repo URL

Link to my plugin: https://github.com/lizard-heart/obsidian-order-list-plugin

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [var](https://github.com/lizard-heart/obsidian-order-list-plugin/blob/e47abb432852f591509fcb59d54711b969b4c249/main.ts#L125) I recommend you change all instances of `var` to either `const` or `let`. `var` has function-level scope, so it can easily lead to bugs if you're not careful. Here's a pretty good guide with examples of why not to use `var`: [why `var` is obsolete](https://javascript.plainenglish.io/4-reasons-why-var-is-considered-obsolete-in-modern-javascript-a30296b5f08f)
- [cmEditor](https://github.com/lizard-heart/obsidian-order-list-plugin/blob/e47abb432852f591509fcb59d54711b969b4c249/main.ts#L87) cmEditor is deprecated. You should use [`editor`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L2185) instead.
- [SelectionRange](https://github.com/lizard-heart/obsidian-order-list-plugin/blob/e47abb432852f591509fcb59d54711b969b4c249/main.ts#L16) you can just use [EditorRange](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L986). No need to define your own interface for this.
- [callback](https://github.com/lizard-heart/obsidian-order-list-plugin/blob/e47abb432852f591509fcb59d54711b969b4c249/main.ts#L30) you should use [editorCallback](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L614) here instead. That will ensure that the command is only available when the user is in edit mode. And you'll be passed the editor object automatically so you won't need to define a getEditor function.
- [console.log(linesList](https://github.com/lizard-heart/obsidian-order-list-plugin/blob/e47abb432852f591509fcb59d54711b969b4c249/main.ts#L74) you should remove all the console.logs from your plugin to avoid spamming the console. This will be beneficial when other plugin devs need to do their own debugging or when users are asked to investigate a plugin issue from the console.

lizard-heart:
Ok, I have made all the changes.

liamcain:
- [editorCallback: () => this.findAndReplace()](https://github.com/lizard-heart/obsidian-order-list-plugin/blob/94ace87581759d98ce6e178a69b3b082a7269ce7/main.ts#L30) you might want to change this to:

```ts
editorCallback: (editor) => this.findAndReplace(editor)
```

Then you don't need to check if there is an editor inside your findAndReplace function, it will be passed in as an argument.

PR #1482, opened 2022-12-28
Add email-block-plugin plugin to community-plugins

joleaf:

## Repo URL

Link to my plugin:
https://github.com/joleaf/obsidian-email-block-plugin

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [getActiveFile](https://github.com/joleaf/obsidian-email-block-plugin/blob/08ed247d01cdc20cb867d1f35412959865bd51f2/main.ts#L92) I think you're better off using [`getAbstractFileByPath`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L3612) and finding the file via the ctx.sourcePath instead of assuming that the file being rendered is the activeFile.
- [""](https://github.com/joleaf/obsidian-email-block-plugin/blob/08ed247d01cdc20cb867d1f35412959865bd51f2/main.ts#L123) you are passing in empty string for the source path. This should be ctx.sourcePath instead. A way you can test to make sure this is correct is to try it on an "ambiguous" link. So if you have two files with the same name in different folders, it's important to know the current file that you're linking from so that the link lookup can be find the correct file.

joleaf:
Thank you @liamcain ! I appreciate your comments and suggestions, thank you!
I fixed both issues and hope that the plugin is okay now.

liamcain:
Sorry, one more small thing: https://github.com/joleaf/obsidian-email-block-plugin/commit/dba36786a973e95879bf42962ec92c96c57d5352#diff-564c860307257bfac9176f554a35aa858c350a59d75c0dcd0a5d63a480805bd0R93

I know it's a bit weird since the second parameter calls for the sourcePath, but in this case you actually want to pass "" as the second parameter. The sourcePath in this case will be a full path from the vault root, meaning the sourcePath would be the vault root or "" for short.

joleaf:
@liamcain Okay lol :D Thank you again for the review! This is fixed as well.

liamcain:
Maybe you forgot to push? I don't see any changes since 0.3.1

joleaf:
-.-  yes, now it is pushed..

PR #1488, opened 2022-12-29
Update community-plugins.json

wenlzhang:

## Repo URL

Link to my plugin: https://github.com/wenlzhang/obsidian-task-marker

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [this.addCommand](https://github.com/wenlzhang/obsidian-task-marker/blob/36e1cf01eb4b4e301920d58f7d38290a84780c80/src/taskmarker-Plugin.ts#L344) I recommend just defining the command inside the addCommand argument. So `addCommand({ id: '', name: '', editorCallback: () => {} })`. Doing it in 2 steps where the addCommand is so far away from the command itself risks you forgetting to add the command after creating it
- [this.taskMarker.settings.supportCanceledTasks](https://github.com/wenlzhang/obsidian-task-marker/blob/36e1cf01eb4b4e301920d58f7d38290a84780c80/src/taskmarker-Plugin.ts#L346) Instead of conditionally adding the command, just use [`editorCheckCallback`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L518) and only show the command if "supportCanceledTasks" is enabled. That way the command will properly appear after the setting is toggled.

PR #1490, opened 2022-12-31
Add plugin: obsidian-ivre-plugin

p-l-:

## Repo URL

Link to my plugin: https://github.com/ivre/obsidian-ivre-plugin

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  ~Android _(if applicable)_~
  - [ ]  ~iOS _(if applicable)_~
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6102,5 +6102,13 @@
         "description": "Manage your Plugins with groups: Enable & disable multiple plugins at once or delay their startup to speed up your Obsidian start up time.",
         "author": "Mocca101",
         "repo": "Mocca101/obsidian-plugin-groups"
+    },
+    {
+        "id": "obsidian-ivre-plugin",
+        "name": "IVRE",
+        "description": "IVRE integration for Obsidian: grab data from IVRE and bring it into Obsidian notes.",
+        "author": "The IVRE contributors",
+        "repo": "ivre/obsidian-ivre-plugin",
+        "donation": "https://github.com/sponsors/p-l-"
     }
 ]


liamcain:
- [this.app](https://github.com/ivre/obsidian-ivre-plugin/blob/d2b199e3ca750e536a21abf1a1bce7e0d0503640/main.ts#L158) this is not defined. It might be working but it's falling back to the global scope, which I don't think you should be relying on. It's currently accessible from the global context for debugging purposes, and might be removed in the future. Instead, you should use the reference to App that you have from your plugin instance and just pass that into your functions.
- [this.app](https://github.com/ivre/obsidian-ivre-plugin/blob/d2b199e3ca750e536a21abf1a1bce7e0d0503640/main.ts#L209) same thing
- [this.app](https://github.com/ivre/obsidian-ivre-plugin/blob/d2b199e3ca750e536a21abf1a1bce7e0d0503640/main.ts#L219) here too
- [this.app.vault](https://github.com/ivre/obsidian-ivre-plugin/blob/d2b199e3ca750e536a21abf1a1bce7e0d0503640/main.ts#L225) and here
- [this.app.](https://github.com/ivre/obsidian-ivre-plugin/blob/d2b199e3ca750e536a21abf1a1bce7e0d0503640/main.ts#L240) hmm, okay there's a lot of them. I won't point them all out
- [getActiveFile](https://github.com/ivre/obsidian-ivre-plugin/blob/d2b199e3ca750e536a21abf1a1bce7e0d0503640/main.ts#L792) this can return null. Also, since you're passing in a MarkdownView to this function, you should be comparing it to the view.file and not the active file
- [ivre](https://github.com/ivre/obsidian-ivre-plugin/blob/d2b199e3ca750e536a21abf1a1bce7e0d0503640/main.ts#L890) you don't need to prefix the ID with your plugin name, it will get automatically prefixed with it by the addCommand

p-l-:
Thanks a lot for your comment @liamcain! I'm a total noob to both TypeScript & the Obsidian API. What could I use rather than `this.app`? Is there a clean way to reference the app from the global scope?

p-l-:
@liamcain thanks a lot for your review! Sorry for all the mistakes, I'm a total noob in both TypeScript and the Obsidian API!

Hope it's better now, let me know if there are other things to fix. I'm preparing the next release.

liamcain:
- [this.app.workspace.getActiveFile](https://github.com/ivre/obsidian-ivre-plugin/blob/5175d5d8b476b922fbb72ee1cdce7d914319916e/main.ts#L882) sorry, I might not have been clear here. I was suggesting you don't call getActiveFile. While under normal circumstances, it would be the same as the file in the MarkdownView, you don't really want to make that assumption. So I would change this line to be `view.file` (you have access to file from the FileView (<https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L1372>)
- [this.app.workspace.getActiveFile](https://github.com/ivre/obsidian-ivre-plugin/blob/5175d5d8b476b922fbb72ee1cdce7d914319916e/main.ts#L935) same here.

p-l-:
Thanks again @liamcain! That's done. Do you see something else that should be done?

liamcain:
LGTM, merged!

p-l-:
Thanks!

PR #1491, opened 2022-12-31
Added Plugin Groups to Community-plugins.json

Mocca101:

## Repo URL

Link to my plugin:
[Plugin Groups Github](https://github.com/Mocca101/obsidian-plugin-groups
)
## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


liamcain:
- [app](https://github.com/Mocca101/obsidian-plugin-groups/blob/949c1dbb2b47a3b0c01b44cfea5ea664890baa04/src/Managers/PluginManager.ts#L124) It's highly discouraged that you use the global app instance in your plugin. It's currently accessible from the global context for debugging purposes, and might be removed in the future. Instead, you should use the reference to App that you have from your plugin instance.
- [...installedPlugins](https://github.com/Mocca101/obsidian-plugin-groups/blob/949c1dbb2b47a3b0c01b44cfea5ea664890baa04/src/Managers/PluginManager.ts#L49) filter already creates a copy of the list, no need to clone the list first
- [window.performance.now() / 1000](https://github.com/Mocca101/obsidian-plugin-groups/blob/949c1dbb2b47a3b0c01b44cfea5ea664890baa04/main.ts#L33) performance.now is more expensive than Date.now, so unless you need the precision, I suggest you switch to Date.now()

Mocca101:
Thanks for the notes! I've adressed them and they will be in the next release.
Unfortunately I need [window.performance.now() / 1000](https://github.com/Mocca101/obsidian-plugin-groups/blob/949c1dbb2b47a3b0c01b44cfea5ea664890baa04/main.ts#L33) for now. As I haven't found another way to find out how long obsidian has been running ( it is used to check how long obsidian has been running as to not enable "startup" plugins when plugin-groups is being enabled while Obsidian has already been in use for some time e.g. when dis-/re-enabling the plugin). If there is a better way please let me know!

liamcain:
You're saying that `Date.now()` doesn't work for your use case?

Mocca101:
Unfortunately no, Date.now() gives me the current time (since the epoch), but for this usecase I need the time that has passed since Obsidian has started. 
If there is a way to get that or the timestamp of when Obsidian has started I could use that.

PR #1493, opened 2023-01-01
adds obsidian-link-exploder to community-plugins.json

benhughes:

## Repo URL

Link to my plugin: https://github.com/benhughes/obsidian-link-exploder

---

**NOTE:** Plugin is not tested on Windows or Linux, am only using the following APIs:
- `this.app.metadataCache.getCache`
- `this.app.vault.create`
- `this.app.workspace.getLeaf().openFile`

So have assumed this is fine, people have been using this plugin using BRAT. As per manifest this plugin is desktop only until canvas is supported on mobile.

---

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [doesFileExist](https://github.com/benhughes/obsidian-link-exploder/blob/526c7983fbe55d6d02f000530a237fcb9407abd2/main.ts#L19) You should use [`getAbstractFileByPath`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L3612) instead of getCache here to more reliably check if a file exists.
- [log.warn](https://github.com/benhughes/obsidian-link-exploder/blob/526c7983fbe55d6d02f000530a237fcb9407abd2/lib/canvas/canvas.ts#L297) You should probably return null or throw an error here since you're not handling the case where there's a dupe filename in the caller.
- [-canvas](https://github.com/benhughes/obsidian-link-exploder/blob/526c7983fbe55d6d02f000530a237fcb9407abd2/lib/canvas/canvas.ts#L251) why add -canvas when it's already a .canvas extension

benhughes:
@liamcain thanks for your suggestions, they have now been implemented.

I've updated the plugin ids to not include obsidian, think the workflow needs to be run again though

PR #1510, opened 2023-01-05
Add Plugin: Khoj 🦅

debanjum:
### Overview
[Khoj](https://github.com/debanjum/khoj) enables incremental, natural language search on your Obsidian notes.

### Features
- **Natural**: Advanced natural language understanding using Transformer based ML Models
- **Local**: Your personal data stays local. All search, indexing is done on your machine[*](https://github.com/debanjum/khoj#miscellaneous)
- **Incremental**: Incremental search for a fast, search-as-you-type experience

### Code Layout
  The Obsidian plugin code for khoj is [available here](https://github.com/debanjum/khoj/tree/master/src/interface/obsidian). The plugin interacts with your (local) khoj server over an HTTP API
  - [main.ts](https://github.com/debanjum/khoj/blob/master/src/interface/obsidian/src/main.ts): *plugin entrypoint*
     - registers command, ribbon and settings tab
  - [modal.ts](https://github.com/debanjum/khoj/blob/master/src/interface/obsidian/src/modal.ts): *khoj search modal*
    - search, view and jump to notes from modal
  - [settings.ts](https://github.com/debanjum/khoj/blob/master/src/interface/obsidian/src/settings.ts): *khoj plugin settings pane*
     - set khoj server URL, results count and update index
  - [utils.ts](https://github.com/debanjum/khoj/blob/master/src/interface/obsidian/src/utils.ts): *common utility methods*
     - methods to configure khoj server from plugin

For more details see [demo](https://github.com/debanjum/khoj/tree/master/src/interface/obsidian#demo), [modal screenshot](https://github.com/debanjum/khoj/tree/master/src/interface/obsidian#interface) and the khoj obsidian [codebase visualization](https://github.com/debanjum/khoj/tree/master/src/interface/obsidian#Visualize-Codebase)

## Repo URL

Link to my plugin: https://github.com/debanjum/khoj

## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


liamcain:
- [Find Similar Notes](https://github.com/debanjum/khoj/blob/e18124ef6f8522fd48b2494c677a62635c5c090b/src/interface/obsidian/src/main.ts#L27) nitpick: commands are usually Sentence case.
- [this.inputEl.value = await this.app.vault.read](https://github.com/debanjum/khoj/blob/e18124ef6f8522fd48b2494c677a62635c5c090b/src/interface/obsidian/src/modal.ts#L65) loading the entire file contents into the input seems a bit clunky. You'll also hit a hard cap encoding the entire file contents into the query url.
- [mdFiles = this.app.vault.getMarkdownFiles](https://github.com/debanjum/khoj/blob/e18124ef6f8522fd48b2494c677a62635c5c090b/src/interface/obsidian/src/modal.ts#L99) this is anot a good way to find a file match. Instead you should use use [`getAbstractFileByPath`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L3612).
- [return](https://github.com/debanjum/khoj/blob/e18124ef6f8522fd48b2494c677a62635c5c090b/src/interface/obsidian/src/modal.ts#L114) just a heads up, but this "return" isn't returning from the parent function, it's just exiting the forEach callback so this is still very inefficient.
- [finally](https://github.com/debanjum/khoj/blob/e18124ef6f8522fd48b2494c677a62635c5c090b/src/interface/obsidian/src/settings.ts#L40) there's no point to using both async/await _and_ finally. If you're already awaiting the function call, then this can just appear in the next line without the finally.
- [this](https://github.com/debanjum/khoj/blob/e18124ef6f8522fd48b2494c677a62635c5c090b/src/interface/obsidian/src/utils.ts#L5) `this` is not defined. It might not be breaking in your testing since it's resolving to the global scope. But It's highly discouraged that you use the global app instance in your plugin. It's currently accessible from the global context for debugging purposes, and might be removed in the future. Instead, you should use the reference to App that you have from your plugin instance and just pass it to the functions that need it.

debanjum:
@liamcain: Thanks a ton for the detailed feedback. It very much appreciated.
I've fixed up most of your suggestions in [this Khoj PR](https://github.com/debanjum/khoj/pull/127/). Specifically:

- https://github.com/debanjum/khoj/commit/c203c6a3fd9e74dcea4614c5e2b2e67ae93727fe Use Sentence case for Find similar note Obsidian command name
  > [Find Similar Notes](https://github.com/debanjum/khoj/blob/e18124ef6f8522fd48b2494c677a62635c5c090b/src/interface/obsidian/src/main.ts#L27) nitpick: commands are usually Sentence case.

- https://github.com/debanjum/khoj/commit/37063f6a38e0959f7c7465cf05e010ba1d8511a5 Truncate query to 8k chars for find similar notes from Obsidian plugin
  > [this.inputEl.value = await this.app.vault.read](https://github.com/debanjum/khoj/blob/e18124ef6f8522fd48b2494c677a62635c5c090b/src/interface/obsidian/src/modal.ts#L65) loading the entire file contents into the input seems a bit clunky. You'll also hit a hard cap encoding the entire file contents into the query url.

- https://github.com/debanjum/khoj/commit/b415f87093364279606d388c104de808b8927693 Split find and jump to note code in `onChooseSuggestion` method in the modal
  > [return](https://github.com/debanjum/khoj/blob/e18124ef6f8522fd48b2494c677a62635c5c090b/src/interface/obsidian/src/modal.ts#L114) just a heads up, but this "return" isn't returning from the parent function, it's just exiting the forEach callback so this is still very inefficient.

- https://github.com/debanjum/khoj/commit/4456cf5c8fe6e9a3ddcf8aa10d80df8ab931132b No need to use `then` or `finally` in `async` functions after an `await`
  > [finally](https://github.com/debanjum/khoj/blob/e18124ef6f8522fd48b2494c677a62635c5c090b/src/interface/obsidian/src/settings.ts#L40) there's no point to using both async/await _and_ finally. If you're already awaiting the function call, then this can just appear in the next line without the finally.

- https://github.com/debanjum/khoj/commit/4070be637c61038607d088322db5106315f30800 Pass app object from plugin instance to child objects and functions
  > [this](https://github.com/debanjum/khoj/blob/e18124ef6f8522fd48b2494c677a62635c5c090b/src/interface/obsidian/src/utils.ts#L5) `this` is not defined. It might not be breaking in your testing since it's resolving to the global scope. But It's highly discouraged that you use the global app instance in your plugin. It's currently accessible from the global context for debugging purposes, and might be removed in the future. Instead, you should use the reference to App that you have from your plugin instance and just pass it to the functions that need it.

- The code is trying to match files in the frontend vault to files indexed on the khoj backend server. The file paths on the two need not be same if the server is on a different machine. For this the `getAbstractFileByPath` wouldn't work. Of course, I'm open to suggestions if there are better ways to go about this.

  > [mdFiles = this.app.vault.getMarkdownFiles](https://github.com/debanjum/khoj/blob/e18124ef6f8522fd48b2494c677a62635c5c090b/src/interface/obsidian/src/modal.ts#L99) this is anot a good way to find a file match. Instead you should use use [`getAbstractFileByPath`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L3612).

PR #1501, opened 2023-01-03
Add "Obsidian Antidote Plugin"

Samuel-Martineau:
# I am submitting a new Community Plugin

## Repo URL

Link to my plugin: [Samuel-Martineau/obsidian-antidote-plugin](https://github.com/Samuel-Martineau/obsidian-antidote-plugin) 

## Release Checklist
- [x] I have tested the plugin on
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
- [x] My GitHub release contains all required files
  - [x] `main.js`
  - [x] `manifest.json`
- [x] GitHub release name matches the exact version number specified in my manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
- [x] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [x] My README.md describes the plugin's purpose and provides clear usage instructions.
- [x] I have read the tips in https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md and have self-reviewed my plugin to avoid these common pitfalls.
- [x] I have added a license in the LICENSE file.
- [x] My project respects and is compatible with the original license of any code from other plugins that I'm using.
      I have given proper attribution to these other projects in my `README.md`.

PR #1051, opened 2022-07-12
Update community-plugins.json

phibr0:
# I am submitting a new Community Plugin: Commander

In this PR I also removed my old Customizable Sidebar Plugin. This one is a much improved rewrite with a lot more features!

## Repo URL

Link to my plugin: https://github.com/phibr0/obsidian-commander

image

## Release Checklist
- [ ] I have tested the plugin on
  - [ ]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  Android _(if applicable)_
  - [x]  iOS _(if applicable)_
- [x] My GitHub release contains all required files
  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css` _(optional)_
- [x] GitHub release name matches the exact version number specified in my manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
- [x] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [x] My README.md describes the plugin's purpose and provides clear usage instructions.
- [x] I have read the tips in https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md and have self-reviewed my plugin to avoid these common pitfalls.
- [x] I have added a license in the LICENSE file.
- [x] My project respects and is compatible with the original license of any code from other plugins that I'm using.
      I have given proper attribution to these other projects in my `README.md`.

PR #1511, opened 2023-01-05
add "Antidote Grammar Checker Integration" plugin

Heziode:

## Repo URL

Link to my plugin: https://github.com/Heziode/obsidian-antidote

## Release Checklist
  - [X]  Windows: tested with Antidote 11 and Connectix 11
  - [X]  macOS: tested with Antidote 11 and Connectix 11
  - [X]  Linux: tested with Antidote 10 (latest supported version of Antidote on this OS), but require Connectix version 11 or higher
  - [X] `main.js`
  - [X] `manifest.json`


joethei:
Are you aware of the other Antidote plugin that is currently waiting for review?
See #1501, maybe you two can collaborate.

Including `obsidian`  in the plugin ID is unnecessary.
I would also suggest a different name, as just the word “Antidote” might get you into legal trouble.

Heziode:
Hi,

Yes I have seen #1051, and I contactes @Samuel-Martineau, it will merge feature of his plugin into this one.

> Including obsidian in the plugin ID is unnecessary.

Did you mean the ID int the manifest.json?

> I would also suggest a different name, as just the word “Antidote” might get you into legal trouble.

Something like:

- antidote unofficial
- antidote grammar checker

?

Heziode:
@joethei I have updated my repo.

I have renamed the plugin for legal purpose. I have change the plugin ID both, on my `manifest.json` and into `community-plugins.json`.

I have integrate features of #1501 (commands).

I have updated the code to support Linux.
Now, everything is good on my side.

Heziode:
@joethei my PR is ready to merged.

My code has been reviewed by a member of @DruideInformatiqueInc (the company behind Antidote), and updated to use the public API.

The name  the plugin as been renamed for legal purpose, and the manifest too.

liamcain:
- [style.color](https://github.com/Heziode/obsidian-antidote/blob/0a9d1207ac87cf0ed185ecac74007586148be68c/src/main.ts#L105) You should avoid assigning styles via javascript. Instead, just add a class to the element and move all these styles into CSS so that they are more easily adaptable by themes.
- [active-leaf-change](https://github.com/Heziode/obsidian-antidote/blob/0a9d1207ac87cf0ed185ecac74007586148be68c/src/main.ts#L153) I suggest listening for [`file-open`](https://github.com/obsidianmd/obsidian-api/blob/ec589e9762a1d7e2faad01f894cb34c41b10ecaf/obsidian.d.ts#L4189) instead. It will still be called every time the active leaf changes but it will ensure you have an active file. It will also handle some edge cases like when the user focuses the sidebar but they still want the UI to reflect the last focused file.
- [layout-change](https://github.com/Heziode/obsidian-antidote/blob/0a9d1207ac87cf0ed185ecac74007586148be68c/src/main.ts#L166) you can remove this if you just listen for file-open
- [getLeaf](https://github.com/Heziode/obsidian-antidote/blob/0a9d1207ac87cf0ed185ecac74007586148be68c/src/main.ts#L373) contrary to the name, this doesn't get you the currently active leaf. It also creates leaves in some conditions. I suggest instead using `activeEditor?.file`
- [Map<WorkspaceLeaf, AgentConnectix>](https://github.com/Heziode/obsidian-antidote/blob/0a9d1207ac87cf0ed185ecac74007586148be68c/src/main.ts#L20) oh no! don't do this, this is a memory leak waiting to happen. You should be using a WeakMap to avoid keeping references to workspace leafs. That said, I'm not sure you need this map. Looking more closely at this file, I think this approach isn't safe and will lead to memory leaks. I think a better, much safer approach is to create a codemirror widget with access to the Antidote API. Obsidian will manage the references for you so you don't need to worry about memory management. The downside is it's a bit of a challenge learning the CodeMirror api. There should be a number of plugins out there to draw inspiration from though. For example, the Creases plugin creates a custom editor widget: <https://github.com/liamcain/obsidian-creases/blob/a657e88951c64d57211953f44dd57c538f37b9d4/src/main.ts#L55>. Once you register your editor extension, it should be a lot simpler to manage. Feel free to also reach out on discord for help in the #plugin-dev channel.

Heziode:
> * [style.color](https://github.com/Heziode/obsidian-antidote/blob/0a9d1207ac87cf0ed185ecac74007586148be68c/src/main.ts#L105) You should avoid assigning styles via javascript. Instead, just add a class to the element and move all these styles into CSS so that they are more easily adaptable by themes.

Done ✅ 

> * [active-leaf-change](https://github.com/Heziode/obsidian-antidote/blob/0a9d1207ac87cf0ed185ecac74007586148be68c/src/main.ts#L153) I suggest listening for [`file-open`](https://github.com/obsidianmd/obsidian-api/blob/ec589e9762a1d7e2faad01f894cb34c41b10ecaf/obsidian.d.ts#L4189) instead. It will still be called every time the active leaf changes but it will ensure you have an active file. It will also handle some edge cases like when the user focuses the sidebar but they still want the UI to reflect the last focused file.
> * [layout-change](https://github.com/Heziode/obsidian-antidote/blob/0a9d1207ac87cf0ed185ecac74007586148be68c/src/main.ts#L166) you can remove this if you just listen for file-open

I don't think that the use of `file-open` is a good idea in my case. It is right, that I need to ensure that te file is open, but, I also need to know if I can edit this file (edit mode) and it does not seem possible to obtain this information with TFile.

> * [getLeaf](https://github.com/Heziode/obsidian-antidote/blob/0a9d1207ac87cf0ed185ecac74007586148be68c/src/main.ts#L373) contrary to the name, this doesn't get you the currently active leaf. It also creates leaves in some conditions. I suggest instead using `activeEditor?.file`

It also does not seems relevant here (for the same reason as above).

> * [Map<WorkspaceLeaf, AgentConnectix>](https://github.com/Heziode/obsidian-antidote/blob/0a9d1207ac87cf0ed185ecac74007586148be68c/src/main.ts#L20) oh no! don't do this

I currently changed it to WeakMap, thanks to your remark. I will take a look to refactoring it to use the CodeMirror API, but it can take a while. I don't know if this is a blocking change for this PR

liamcain:
> [getLeaf](https://github.com/Heziode/obsidian-antidote/blob/0a9d1207ac87cf0ed185ecac74007586148be68c/src/main.ts#L373) contrary to the name, this doesn't get you the currently active leaf. It also creates leaves in some conditions. I suggest instead using activeEditor?.file

If you don't want to use activeEditor, you should at least use workspace.activeLeaf. Calling getLeaf will create issues if the user pins their active view, since getLeaf will create a new leaf. So calling this in layout-change will create a new view every time the user focuses a pinned leaf.

I think that's the only thing blocking this from going in.

---

It's also worth noting that this plugin won't work from within the Canvas plugin. active-leaf-change isn't fired when a canvas card is focused. Only file-open is.

Heziode:
> If you don't want to use activeEditor, you should at least use workspace.activeLeaf. Calling getLeaf will create issues if the user pins their active view, since getLeaf will create a new leaf. So calling this in layout-change will create a new view every time the user focuses a pinned leaf.
> 
> I think that's the only thing blocking this from going in.

Ok, I will make the change probably this week-end.



> It's also worth noting that this plugin won't work from within the Canvas plugin. active-leaf-change isn't fired when a canvas card is focused. Only file-open is.

It is not a problem. Supporting canvas is not the main goal.

---

Nevertheless, I have a problem for which I can't find a solution, and no help on the Discord: it is about: https://github.com/Heziode/obsidian-antidote/issues/7

My plugin does not seem to load properly when Obsidian starts. The settings are not displayed. I have to deactivate and reactivate the plugin to get it to load properly.

Did you have any idea on how to fix this issue?

liamcain:
Did you check the dev tools for errors? It throws an error because you're trying to call getLeaf before the workspace is done loading. If you make the change to workspace.activeLeaf, it should fix this issue too.

Heziode:
From [obsidian-api](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L3976-L3989):

```
    /**
     * Indicates the currently focused leaf, if one exists.
     *
     * Please avoid using `activeLeaf` directly, especially without checking whether
     * `activeLeaf` is null.
     *
     * The recommended alternatives are:
     * - If you need information about the current view, use {@link getActiveViewOfType}.
     * - If you need to open a new file or navigate a view, use {@link getLeaf}.
     *
     * @public
     * @deprecated - The use of this field is discouraged.
     */
    activeLeaf: WorkspaceLeaf | null;
```

Should I still use `activeLeaf`, even if it is deprecated?

liamcain:
Oh, since you don't actually need the leaf itself, just the view, then yea. Don't use activeLeaf, just use getActiveViewOfType(MarkdownView). Good point!

Heziode:
Ok, I just released an update with your remark, plus other fixes and improvements.

community-plugins.json
```diff
@@ -4562,7 +4562,9 @@
         "name": "Todoist completed tasks",
         "author": "Andrey Kulishov",
         "description": "Add completed Todoist tasks to your Obsidian notes",
-        "repo": "Ledaryy/obsidian-todoist-completed-tasks"
+        "repo": "Ledaryy/obsidian-todoist-completed-tasks",
+        "donation": "https://www.buymeacoffee.com/ledaryy",
+        "branch": "master"
```
liamcain, line 4567:
```suggestion
        "repo": "Ledaryy/obsidian-todoist-completed-tasks"
```

community-plugins.json
```diff
@@ -5846,5 +5846,13 @@
...
         "description": "Embeddable Pikchr diagrams renderer plugin for Obsidian",
         "repo": "notlibrary/obsidian-adamantine-pick"
+    },
+    {
+         "id": "antidote-grammar-checker-integration",
+         "name": "Antidote Grammar Checker Integration",
+         "author": "Heziode",
+         "description": "Unofficial integration of Antidote, a powerful English and French grammar checker",
+         "repo": "heziode/obsidian-antidote",
+         "donation": "https://www.buymeacoffee.com/Heziode"
```
liamcain, line 5856:
```suggestion
         "repo": "heziode/obsidian-antidote"
```

Heziode:
`donation` property removed

liamcain:
@Heziode I just realized that there's no documentation about it (sorry, will fix!) but the donation link belongs in the manifest.json file within your plugin repo.

PR #1513, opened 2023-01-06
Add Habit Calendar Plugin

hedonihilist:

## Repo URL

Link to my plugin: https://github.com/hedonihilist/obsidian-habit-calendar

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


hedonihilist:
Thanks for spending time reviewing my plugin. This plugin uses `innerHTML` in 

https://github.com/hedonihilist/obsidian-habit-calendar/blob/0887d8b4f537c66b1aa1f41cacb48cf9da83192c/main.ts#L263

but it's guarded by an `enableHTML` setting, which is `false` by default. This will give the freedom of editing DOM to the users knowing what they are doing. 

If the plugin still violates the principles, I will dop this feature in the next PR

hedonihilist:
> 

Sorry I forgot to delete the `obsidian.d.ts` file. It's there for reference when writing code.

liamcain:
- [const pm = source.match](https://github.com/hedonihilist/obsidian-habit-calendar/blob/ebda8f16fca903b114c5788c77e8ed93bcb2cecf/main.ts#L182) looks like this section can be removed, I don't think it's doing anything, right?
- [innerHTML](https://github.com/hedonihilist/obsidian-habit-calendar/blob/ebda8f16fca903b114c5788c77e8ed93bcb2cecf/main.ts#L263) Even though it's guarded behind a flag, I still recommend wrapping `input` in [sanitizeHTMLToDom](https://github.com/obsidianmd/obsidian-api/blob/6b2138aa28fb6d6f58f78e6fabec02186fdc0f82/obsidian.d.ts#L2977) for added security

hedonihilist:
> * [const pm = source.match](https://github.com/hedonihilist/obsidian-habit-calendar/blob/ebda8f16fca903b114c5788c77e8ed93bcb2cecf/main.ts#L182) looks like this section can be removed, I don't think it's doing anything, right?
> * [innerHTML](https://github.com/hedonihilist/obsidian-habit-calendar/blob/ebda8f16fca903b114c5788c77e8ed93bcb2cecf/main.ts#L263) Even though it's guarded behind a flag, I still recommend wrapping `input` in [sanitizeHTMLToDom](https://github.com/obsidianmd/obsidian-api/blob/6b2138aa28fb6d6f58f78e6fabec02186fdc0f82/obsidian.d.ts#L2977) for added security

Thanks for spending time reviewing my plugin. I will clean the unused code and check out the `sanitizeHTMLToDom` api, which I didn't know before.

PR #1514, opened 2023-01-06
Add Meld Build Plugin

meld-cp:

## Repo URL

Link to my plugin: https://github.com/meld-cp/obsidian-build

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[interface MeldBuildPluginSettings {](https://github.com/meld-cp/obsidian-build/blob/24a3de716a65839870747ba5226d5c5a4945f881/src/main.ts#L7-L13)
Since you are not using settings you can remove this and any references to it.

[while(!this.completed){](https://github.com/meld-cp/obsidian-build/blob/24a3de716a65839870747ba5226d5c5a4945f881/src/modal-ask.ts#L33-L37)
I would recommend using a callback approach or similar here, instead of constantly waiting.

[app](https://github.com/meld-cp/obsidian-build/blob/24a3de716a65839870747ba5226d5c5a4945f881/src/rci-io.ts#L101)

[public static delay(ms: number) : Promise<()=>void> {](https://github.com/meld-cp/obsidian-build/blob/24a3de716a65839870747ba5226d5c5a4945f881/src/utils.ts#L23-L25)
The Obsidian already provide a `sleep` method for this, which does basically the same.

meld-cp:
Thanks for the review, those changes have been made.

liamcain:
- [innerHTML](https://github.com/meld-cp/obsidian-build/blob/71c45bdf0790959a7a15833742dd60c86ba37a20/src/modal-message.ts#L35) [innerHTML is a potential security risk](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml). It's hard to trace through the code to see what's creating these messages. If it's something you control, you should just create the html yourself. If it's generated by the user, then I suggest limiting it to strings and using setText()
- [read](https://github.com/meld-cp/obsidian-build/blob/71c45bdf0790959a7a15833742dd60c86ba37a20/src/rci-io.ts#L97) if you aren't writing the file contents back to the disk, you should use [cachedRead](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3373) instead. It's faster since you're just reading the file contents from memory.

meld-cp:
Thank you!

- [innerHTML](https://github.com/meld-cp/obsidian-build/commit/214df0d6fb92c09efa810c9b82e2630d86974155) has been replaced
- [read replaced with cachedRead](https://github.com/meld-cp/obsidian-build/commit/26181f02dc48685bbff797cb82cc7c33e6bf2946)

community-plugins.json
```diff
@@ -6386,5 +6386,13 @@
...
         "author": "froehlichA",
         "repo": "froehlichA/obsidian-avatar"
+    },
+    {
+        "id": "meld-build",
+        "name": "Meld Build",
+        "description": "Write and execute (sandboxed) JavaScript to render templates, query DataView and create dynamic notes.",
+        "author": "meld-cp",
+        "repo": "meld-cp/obsidian-build",
+        "branch": "master"
```
liamcain, line 6396:
```suggestion
        "repo": "meld-cp/obsidian-build"
```

meld-cp:
Thank you guys 👍

PR #1516, opened 2023-01-06
Add obsidian-image-embedder

sascharucks:

## Repo URL

Link to my plugin: https://github.com/sascharucks/obsidian-image-embedder/tree/0.1.0

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[name: 'Embed external image based on Content-Type and file extension',](https://github.com/sascharucks/obsidian-image-embedder/blob/5c523c2dc7b12878956aae2ec85744463a84f995/main.ts#L11)
This name doesn't really fit what the code is actually doing.
You can also change the command id to something simpler like `embed`, Obsidian will automatically prefix it with the plugin id.

[this.app.workspace.on("editor-paste", this.pasteHandler);](https://github.com/sascharucks/obsidian-image-embedder/blob/5c523c2dc7b12878956aae2ec85744463a84f995/main.ts#L20)
You can also wrap this in a `this.registerEvent(yourCode)`, then Obsidian will handle the rest and you don't need the code in `onunload`

[await fetch](https://github.com/sascharucks/obsidian-image-embedder/blob/5c523c2dc7b12878956aae2ec85744463a84f995/main.ts#L92)
I would generally recommend using the `requestUrl` function from the Obsidian API, as that handles CORS for you, one caveat is that it does not support `HEAD` requests

ericaxu:

PR #1518, opened 2023-01-07
Adding tolino-notes-import plugin

juergenbr:
Adding my plugin to allow importing notes from Tolino E-Reader into Obsidian

## Repo URL

Link to my plugin: https://github.com/juergenbr/obsidian-tolino-notes-import

## Release Checklist
  - [X]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


joethei:
[// Remember to rename these classes and interfaces!](https://github.com/juergenbr/obsidian-tolino-notes-import/blob/a4b850672d8b501eb1c12630aeb3fb85d77b324d/main.ts#L11)
[// When registering intervals, this function will automatically clear the interval when the plugin is disabled.](https://github.com/juergenbr/obsidian-tolino-notes-import/blob/a4b850672d8b501eb1c12630aeb3fb85d77b324d/main.ts#L41-L44)
[console.log(this.settings.tolinoDriveSetting);](https://github.com/juergenbr/obsidian-tolino-notes-import/blob/a4b850672d8b501eb1c12630aeb3fb85d77b324d/src/ImportModal.ts#L23)
Remove this.

[const statusBarItemEl = this.addStatusBarItem();](https://github.com/juergenbr/obsidian-tolino-notes-import/blob/a4b850672d8b501eb1c12630aeb3fb85d77b324d/main.ts#L26-L27)
I would remove this, it just clutters the status bar.

[export function normalizeFilename(fileName: string): string {](https://github.com/juergenbr/obsidian-tolino-notes-import/blob/a4b850672d8b501eb1c12630aeb3fb85d77b324d/src/FileUtils.ts#L4-L15)
The Obsidian API provides the `normalizePath` function for this.
This will also fix the problem with the slashes.

[static parseGermanNote(note: string): TolinoNoteModel {](https://github.com/juergenbr/obsidian-tolino-notes-import/blob/a4b850672d8b501eb1c12630aeb3fb85d77b324d/src/NoteParser.ts#L17)
This function name is confusing me a bit, is there some distinction between german or english notes for example?
AFAIK tolino devices are only sold in germany (Hallo übrigens)

juergenbr:
@joethei thanks for taking a look at it!
I updates my version to 1.0.1. 

Regarding the function name: Tolino stores notes in a .txt file on the device. Depending on the language setting, there are certain hard-coded strings inserted around each note. If device language is changed, existing notes are not updated, so multiple languages are mixed together in the same file.
E.g if you create a Highlight in German, it starts with ``Markierung auf Seite``.
If you change the language to English and add another note, the string starts with ``Highlight on page ``.

Here is an extract of my notes.txt file showing two notes, first one created when device language was German, and second one created after I switched the language to English:
```
Chip War (Miller, Chris)
Markierung auf Seite 680-681: "So great is China’s government support for YMTC that even during the COVID lockdown it was allowed to keep working, according to Nikkei Asia, a Japanese newspaper with some of the best coverage of China’s chip industry."
Hinzugefügt am 31.01.2023 | 20:51

-----------------------------------

Code (Petzold, Charles)
Highlight on page 462: "The light goes off because power is no longer being supplied to the
    second relay. Similarly, if the bottom switch is closed, the light is also
    off:"
Added on 02/03/2023 | 17:03

-----------------------------------
```

joethei:
Ah, well that is frustrating from a parsing perspective.

liamcain:
- [this.registerInterval](https://github.com/juergenbr/obsidian-tolino-notes-import/blob/69a6c3f1eb640acfe010ad0d4c19da72cec95f81/main.ts#L38) You still didn't remove this. It's not doing anything.
- [NoteCreationService](https://github.com/juergenbr/obsidian-tolino-notes-import/blob/69a6c3f1eb640acfe010ad0d4c19da72cec95f81/src/NoteCreationService.ts#L9) nitpick but I don't see any reason this needs to be a class. I recommend just making this a function. It also seems like fileSystem isn't even initialized, and this.app is unused too...

juergenbr:
@liamcain thanks for the feedback, I added your suggested changes and also removed the word "Plugin" from the name as suggested by Github Actions. However, the action still seems to pick up the old name. My latest plugin version is 1.0.2 now, and it contains the manifest with the correct name. Any idea what's wrong here?

joethei:
Sorry for taking so long

juergenbr:
> Sorry for taking so long to fix this issue

@joethei No worries! Anything else that needs fixing before this can be merged?

community-plugins.json
```diff
@@ -6606,5 +6606,12 @@
...
     "repo": "jia6y/flomo-to-obsidian"
-  }
+  },
+  {
+    "id": "tolino-notes-import",
+    "name": "Tolino notes Importer",
+    "author": "juergenbr",
+    "description": "This is a plugin for Obsidian to import notes from a Tolino E-Reader",
+    "repo": "juergenbr/obsidian-tolino-notes-import"
+  },
```
liamcain, line 6616:
```suggestion
  }
```

liamcain:
Really sorry for such a long wait with re-reviewing this plugin!! Not sure how it fell through the cracks for so long 😞 that's totally my fault. We're currently working on ways we can improve the plugin review process and avoid these long wait times.

Thanks for your patience 🙏

PR #1520, opened 2023-01-07
Add YTranscript plugin

lstrzepek:

## Repo URL

Link to my plugin: https://github.com/lstrzepek/obsidian-yt-transcript

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[console.log('Timestamp:', value);](https://github.com/lstrzepek/obsidian-yt-transcript/blob/9ebf23fe7a6bf9e22e1c81736d7a3b2908a87fd9/main.ts#L90)
[console.log('open!');](https://github.com/lstrzepek/obsidian-yt-transcript/blob/9ebf23fe7a6bf9e22e1c81736d7a3b2908a87fd9/transcript-view.ts#L37)
I would suggest removing these.

lstrzepek:
> [console.log('Timestamp:', value);](https://github.com/lstrzepek/obsidian-yt-transcript/blob/9ebf23fe7a6bf9e22e1c81736d7a3b2908a87fd9/main.ts#L90)
> [console.log('open!');](https://github.com/lstrzepek/obsidian-yt-transcript/blob/9ebf23fe7a6bf9e22e1c81736d7a3b2908a87fd9/transcript-view.ts#L37)
> I would suggest removing these.

Done!

liamcain:
- [occure](https://github.com/lstrzepek/obsidian-yt-transcript/blob/e86da84f5c81b2f9a424fb7b9deaf050d48a3711/main.ts#L85) typo: occur
- [Timestamp mod](https://github.com/lstrzepek/obsidian-yt-transcript/blob/e86da84f5c81b2f9a424fb7b9deaf050d48a3711/main.ts#L84) maybe it's just me, but it's unclear what "mod" means in this setting name. Does it mean "timestamp modifier"? Even then, I think it might make more sense to call it something like "interval"

PR #1523, opened 2023-01-08
Update community-plugins.json

kiichi:

## Repo URL

Link to my plugin: https://github.com/kiichi/3-D-viewer

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5928,6 +5928,13 @@
         "description": "Quickly move a file to a new or existing folder, then open the next file.",
         "author": "Ken Lim",
         "repo": "kenlim/file-chucker-plugin"
+    },
+    {
+        "id": "three-d-file-viewer",
+        "name": "3-D File Viewer",
+        "description": "Drag and drop 3-D model files, such as .glb, .gltf, and .obj file, then you can reder in reading view.",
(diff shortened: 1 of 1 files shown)


joethei:
[import * as path from 'path';](https://github.com/kiichi/3-D-viewer/blob/860dc17cd6240e3d418b079dcf0feaf06e481e8e/main.ts#L3)
This will crash on mobile, as the mobile app does not include Node API's.
Here is a version that is using the Obsidian API(haven't tested it, but should work):
```
const file = this.app.vault.getAbstractFileByPath(context.sourcePath);
if(file instanceof TFile) {
    const url = this.app.vault.getResourcePath(file);
    // create vierwer
}
```

[if (src?.endsWith(".glb") || src?.endsWith(".gltf") || src?.endsWith(".obj")){](https://github.com/kiichi/3-D-viewer/blob/860dc17cd6240e3d418b079dcf0feaf06e481e8e/main.ts#L21)
I would move this check a few lines up, as you are currently creating a div, that might not even be used.

ericaxu:

PR #1524, opened 2023-01-08
Update community-plugins.json

alvaro-cas:

## Repo URL

Link to my plugin:
https://github.com/alvaro-cas/kindle-csv-converter-obsidian

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[id: 'kindle-csv-converter',](https://github.com/alvaro-cas/kindle-csv-converter-obsidian/blob/0a46287142b33090b6538806e1eb675f04783cc4/main.ts#L15)
This id does not need to be unique, I would recommend an id of `choose-file`, Obsidian will automatically prepend this with the plugin id.


[const files = (this.app.vault.adapter as any).files as AdapterFile[];](https://github.com/alvaro-cas/kindle-csv-converter-obsidian/blob/0a46287142b33090b6538806e1eb675f04783cc4/settings.ts#L40-L44)
Don't do this, do this instead.
```ts
for (let file of this.app.vault.getAllLoadedFiles()) {
            if (file instanceof TFolder) {
                files.push(file);
            }
        }
```

[import { Modal, Notice, setIcon } from 'obsidian';](https://github.com/alvaro-cas/kindle-csv-converter-obsidian/blob/0a46287142b33090b6538806e1eb675f04783cc4/modal_read.ts#L1-L4)
I would recommend sticking to one import style.

[document.createElement('input');](https://github.com/alvaro-cas/kindle-csv-converter-obsidian/blob/0a46287142b33090b6538806e1eb675f04783cc4/modal_read.ts#L31)
Use `activeDocument` instead, as `document` is only for the main window.

[fs.createReadStream(input.files[0]['path']);](https://github.com/alvaro-cas/kindle-csv-converter-obsidian/blob/0a46287142b33090b6538806e1eb675f04783cc4/modal_read.ts#L35)
Where would this file be located?
If its inside the vault I recommend using `app.vault.readCached(tfile);`



## Style Nitpicks:

I would generally recommend moving all code files into a separate `src` folder, that makes it easier to review and maintain.

[kindleCsvPlugin](https://github.com/alvaro-cas/kindle-csv-converter-obsidian/blob/0a46287142b33090b6538806e1eb675f04783cc4/main.ts#L6)
[kindleSettings](https://github.com/alvaro-cas/kindle-csv-converter-obsidian/blob/0a46287142b33090b6538806e1eb675f04783cc4/main.ts#L29)
[reader](https://github.com/alvaro-cas/kindle-csv-converter-obsidian/blob/0a46287142b33090b6538806e1eb675f04783cc4/main.ts#L21)
Classes are generally named with the first letter in caps.

alvaro-cas:
- Where would this file be located?
This is the .csv input file, so it can be anywhere outside the vault depending on where the user download it.

The other requested changes are already updated in the [plugin repo](https://github.com/alvaro-cas/kindle-csv-converter-obsidian/). Thank you for reviewing my code! 😃

liamcain:
[view](https://github.com/alvaro-cas/kindle-csv-converter-obsidian/blob/645bd2b2932a49e9c0f7c1822cf705e71b8d2756/main.ts#L20) why is the command only available if the active view is a markdown file? Seems like it doesn't matter what the active file is

alvaro-cas:
Yeah, actually it doesn't matter. I thought the command could not trigger at all, but actually I checked and I think it also works on other views not sure why. Do I remove that?

liamcain:
I think it would make more sense if you remove the check.

PR #1527, opened 2023-01-09
Weather and Moon Phases plugin

mubarizahmed:

## Repo URL

Link to my plugin: https://github.com/mubarizahmed/obsidian-weather-moon

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5965,6 +5965,6 @@
         "description": "This plugin allows you to insert weather and moon phase information into your notes.",
         "author": "Mubariz Ahmed",
         "repo": "mubarizahmed/obsidian-weather-moon"
-    },
+    }
 ]
 
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5958,6 +5958,13 @@
         "description": "Add random notes from your vault to the Obsidian canvas",
         "author": "jmilldotdev",
         "repo": "jmilldotdev/obsidian-canvas-randomnote"
-    }
+    },
+    {
+        "id": "obsidian-weather-moon",
+        "name": "Weather & Moon Phases",
+        "description": "This plugin allows you to insert weather and moon phase information into your notes.",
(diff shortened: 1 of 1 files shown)


joethei:
I would generally recommend moving all code files into a separate `src` folder, that makes it easier to review and maintain.

["dependencies": {](https://github.com/mubarizahmed/obsidian-weather-moon/blob/917010554e8508472e8c006e232356a0f7a44418/package.json#L29-L144)
Why do you include so many packages that you don't actually use?

[// Remember to rename these classes and interfaces!](https://github.com/mubarizahmed/obsidian-weather-moon/blob/917010554e8508472e8c006e232356a0f7a44418/main.ts#L3)
You can remove this.

[id: 'current-weather-test',](https://github.com/mubarizahmed/obsidian-weather-moon/blob/917010554e8508472e8c006e232356a0f7a44418/main.ts#L27-L28)
I would recommend removing the `-test`

[await fetch('https://wttr.in/'+this.settings.location+'?'+this.settings.unit+'&format='+this.settings.inlineFormatWeather+'&lang=en');](https://github.com/mubarizahmed/obsidian-weather-moon/blob/917010554e8508472e8c006e232356a0f7a44418/main.ts#L109)
I recommend using the `requestUrl` function from the Obsidian API

[console.log(data);](https://github.com/mubarizahmed/obsidian-weather-moon/blob/917010554e8508472e8c006e232356a0f7a44418/main.ts#L111)
Remove this.

[const data = await response.text();](https://github.com/mubarizahmed/obsidian-weather-moon/blob/917010554e8508472e8c006e232356a0f7a44418/main.ts#L136-L137)
You can just use `response.json()´, same applies for the `requestUrl` function.

PR #1528, opened 2023-01-09
Add 🪝 Grappling Hook Plugin to Community Store

chrisgrieser:
The plugin is basically an implementation of vim's alternate buffer (`:buffer #`) as well as a simplified implementation of the [Harpoon plugin for neovim](https://github.com/ThePrimeagen/harpoon).

## Repo URL

Link to my plugin: https://github.com/chrisgrieser/grappling-hook

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[const recentFiles = this.app.workspace.lastOpenFiles;](https://github.com/chrisgrieser/grappling-hook/blob/fef9d47304f608cf7855fee6edca0348e79fd8dd/main.ts#L102)
Use `app.workspace.getLastOpenFiles()` instead.

chrisgrieser:
> [const recentFiles = this.app.workspace.lastOpenFiles;](https://github.com/chrisgrieser/grappling-hook/blob/fef9d47304f608cf7855fee6edca0348e79fd8dd/main.ts#L102)
> Use `app.workspace.getLastOpenFiles()` instead.

done 👍

liamcain:
- [grapplingHookPlugin](https://github.com/chrisgrieser/grappling-hook/blob/92d79b26269b45cdec828f5c264422dd731eca05/main.ts#L28) nit: capitalize your class name :)
- [getLeaf](https://github.com/chrisgrieser/grappling-hook/blob/92d79b26269b45cdec828f5c264422dd731eca05/main.ts#L64) I wouldn't call this helper function "activeLeaf". Get Leaf doesn't always return the active leaf. For example, if the current leaf is pinned, it will create a new leaf. The usage seems fine, I'm just worried about the terms getting conflated.
- [let nextIndex = currentIndex + 1;](https://github.com/chrisgrieser/grappling-hook/blob/92d79b26269b45cdec828f5c264422dd731eca05/main.ts#L91) a trick you could do here is let nextIndex = (currentIndex + 1) % filePathArray.length;
- [pathToTFile](https://github.com/chrisgrieser/grappling-hook/blob/92d79b26269b45cdec828f5c264422dd731eca05/main.ts#L66) I recommend always returning a value from a function instead of relying on the function exiting and returning undefined. It makes it easier to infer the return type of the function. So add `return null` at the end here.
- [nextFile](https://github.com/chrisgrieser/grappling-hook/blob/92d79b26269b45cdec828f5c264422dd731eca05/main.ts#L164) nextFile can be undefined.

chrisgrieser:
Thanks for the thorough review! Being a hobby coder, it's super valuable to get feedback on code I wrote 😊 

All implemented in latest commit (except `nextFile`, since the earlier guard clauses on starred files should already prevent `nextFile` ever returning undefined)

PR #1534, opened 2023-01-12
Add obsidian-babashka

filipesilva:

## Repo URL

Link to my plugin: https://github.com/filipesilva/obsidian-babashka

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[id: 'obsidian-babashka-eval-codeblock',](https://github.com/filipesilva/obsidian-babashka/blob/991fd90ec8d04328d76845c29b27689d2e66e8d2/main.ts#L386)
You don't need to include your plugin name in the command id,
Obsidian prefix it automatically with the plugin id.

filipesilva:
@joethei thanks for taking a look! I'd removed the prefix in https://github.com/filipesilva/obsidian-babashka/commit/ee368a585edd53cdf74cbd01c9ed32f0c36398e7, and created a 1.0.5 release.

liamcain:
- [if (validateAndSetup(codeblock, view, settings)) {](https://github.com/filipesilva/obsidian-babashka/blob/2ec7e15f8a8e2a87c053a3422efc3c13f46aaab4/main.ts#L276) just a code style suggestion. I recommend changing this to:

> if (codeblock && validateAndSetup(codeblock, view, settings)) {

Then you don't need the type assertion below. If you still want to keep the Notice, then you can move the notice into the caller fn. I think in general you just want to avoid the `as`

filipesilva:
Thanks for the suggestion, I tried it in https://github.com/filipesilva/obsidian-babashka/commit/1cdbe5d04644c76080c39726c70b37d9e29baf5e and it did make the types nicer to work with!

PR #1535, opened 2023-01-12
Shiba Inu Theme 

faroukx:
Change name/id Shiba Uni to Shiba Inu.


ericaxu:

faroukx:
Will do that.

PR #1544, opened 2023-01-15
Add ActivityWatch watcher [time tracker] to community-plugins.json

LordGrimmauld:
Add a plugin to track time spent in Obsidian tracking info like vault and file names.

## Repo URL

Link to my plugin:
https://github.com/LordGrimmauld/aw-watcher-obsidian

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ] ~~Android _(if applicable)_~~
  - [ ] ~~iOS _(if applicable)_~~
  - [x] `main.js`
  - [x] `manifest.json`
  - [ ] ~~`styles.css` _(optional)_~~


LordGrimmauld:
Update: Testing on Windows shows this working. I am not in a position to test macOS, but considering all APIs this depends on should be available there too I'll assume it working until I receive a report that says otherwise.

liamcain:
- [app](https://github.com/LordGrimmauld/aw-watcher-obsidian/blob/3fe4d23a37fc77b9a2e3a48eca691b7a25e389a6/main.ts#L133) this is referencing the global app instance. You should reference app via `this.app`
- [create](https://github.com/LordGrimmauld/aw-watcher-obsidian/blob/3fe4d23a37fc77b9a2e3a48eca691b7a25e389a6/main.ts#L141) So, it's a bit weird, but the `create` event is actually fired when the vault is initialized. So this will lead to a huge surge of calls with each file in your vault. You should add a [`onLayoutReady`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3626) callback and subscribe to `create` _after_ the vault is done initializing.
- [sleeptime](https://github.com/LordGrimmauld/aw-watcher-obsidian/blob/3fe4d23a37fc77b9a2e3a48eca691b7a25e389a6/main.ts#L149) maybe add a check to make sure that sleeptime isn't 0 to avoid sending heartbeats constantly (I see it's not configurable, so maybe this is just a future improvement)
- [oldPath](https://github.com/LordGrimmauld/aw-watcher-obsidian/blob/3fe4d23a37fc77b9a2e3a48eca691b7a25e389a6/main.ts#L122) oldPath is only the argument name for rename
- [sendHeartbeatData](https://github.com/LordGrimmauld/aw-watcher-obsidian/blob/3fe4d23a37fc77b9a2e3a48eca691b7a25e389a6/main.ts#L91) why are these passed as heartbeats? Shouldn't only the 5 second heartbeat be sent as a heartbeat and the file events are something else?
- [create](https://github.com/LordGrimmauld/aw-watcher-obsidian/blob/3fe4d23a37fc77b9a2e3a48eca691b7a25e389a6/main.ts#L141) just a heads up, `create` is also called for folders. You might want to check to make sure its a file and not a folder.

LordGrimmauld:
- All references to the global app instance have been replaced
- Thank you for bringing that weirdness with create to my attention. That _definitely_ was a bug that needed addressing! I could both confirm and fix this.
- sleeptime is currently not configurable mainly because i did not yet find a nice way to have the user enter numbers into config fields. If there is something i just missed, do let me know, as it is on my list of things to do. A sanity check is a good idea either way, so that has been added, if for now only to future-proof.
- oldPath was an artifact left over by copy/pasting the other very similar file events and has been cleaned up.
- sendHeartbeatData: According to ActivityWatch contributors, sending zero-duration "heartbeats" is a valid way to send events that interrupt the timeline at that point in time, which is what i want. After a rename, the file should no longer appear as being active under the old name. The naming however was hard to understand, the function has therefore been renamed to just sendData.
- create being triggered on folders as well as files is not wrong. In fact i quite like it that way, there is no reason why folders should _not_ be tracked.

Thank you for the thorough review, and after the changes i made i am looking forward to see this approved.

PR #1545, opened 2023-01-15
add plugin: graph presets

ycnmhd:

## Repo URL

Link to my plugin: https://github.com/ycnmhd/obsidian-graph-presets

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[menu.items.unshift(menu.items.pop());](https://github.com/ycnmhd/obsidian-graph-presets/blob/51e53730244a1e22de8130775a2a4393f1609b48/src/context-menu-items/apply-preset-menu-item.ts#L14)
What is this intended to solve?

[const file = await app.vault.create(fileName, content);](https://github.com/ycnmhd/obsidian-graph-presets/blob/51e53730244a1e22de8130775a2a4393f1609b48/src/helpers/obsidian/fs/create-file.ts#L14)``

[const folder = vault.getAbstractFileByPathInsensitive(folderPath);](https://github.com/ycnmhd/obsidian-graph-presets/blob/51e53730244a1e22de8130775a2a4393f1609b48/src/helpers/obsidian/fs/create-folder.ts#L10)
Why are you using the case insensitive version here?

[style={{](https://github.com/ycnmhd/obsidian-graph-presets/blob/51e53730244a1e22de8130775a2a4393f1609b48/src/modals/graph-selecton-modal/graph-leaf-list-item.tsx#L34-L41)
Don't hardcode the style like this, moving it into a separate class allows users to target it with CSS easier.

joethei:
1. use sections for that, instead of the `action` section use your own.

2. No, don't do that, using globals is a bad practice, instead you can pass the plugin/app instance arround.

4. Please do.

liamcain:
@ycnmhd 
The code looks good except for relying so heavily on the global `app` instance as joethei pointed out. Once you address the validation errors, I think we can merge this. However, I would appreciate it if you switched to passing around the `app` instance instead of using the global one. Thanks!

PR #1552, opened 2023-01-18
Adding obsidian-testing-vault plugin to help plugin developers make u…

pedersen:
…seful and safe vaults for testing purposes.

## pedersen/obsidian-testing-vault

Link to my plugin:

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [X]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


pedersen:
> I would generally recommend moving all code files into a separate `src` folder, that makes it easier to review and maintain.
> 
> [id: 'testing-vault-single-note',](https://github.com/pedersen/obsidian-testing-vault/blob/0295ff43461f936f1b79ec0a903983fe9de57dbf/main.ts#L22) Obsidian will automatically prefix the command id with the plugin id.

Thank you. I've updated this package and made a new release that addresses these comments.

liamcain:
- [await this.app.vault.delete](https://github.com/pedersen/obsidian-testing-vault/blob/fc52e45cc26ffb34e452721b7e6edf165df6914e/src/main.ts#L15) this doesn't do what you're expecting it to. This just maps each file to a promise, but it doesn't actually wait for each promise to resolve before executing the next one. If you're trying to run this in serial, you should use a regular for loop instead of map.
- [getFiles](https://github.com/pedersen/obsidian-testing-vault/blob/fc52e45cc26ffb34e452721b7e6edf165df6914e/src/main.ts#L14) I don't think you want to use getFiles here. If you ever extend this to also create folders, they won't be cleaned up too. You probably want to use getRoot() and iterate over all the top folders children and delete those.

pedersen:
Thank you. I've updated the testing vault to remove async/await, as well as making the destroyVault method recursive.

pedersen:
Thank you. I have removed `obsidian-` from the plugin id.

pedersen:
I've just published a new version using the fixed id. Will close and reopen to see if validation will now pass.

PR #1560, opened 2023-01-19
Add dmn-eval-plugin

joleaf:

## Repo URL

Link to my plugin:
https://github.com/joleaf/obsidian-dmn-eval-plugin/

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### Notes for the reviewer
- Java is required for testing this plugin
- The jar is BASE64 encoded. However, the created BASE64 String of the jar is created in the build process and built from the java source files (Maven support).


joethei:
[this.app.vault.adapter.basePath](https://github.com/joleaf/obsidian-dmn-eval-plugin/blob/b473ff70c70d373bbef5d75058961b89f7f980d5/main.ts#L39)
Use `this.app.vault.adapter.getBasePath()` instead.

[app.vault.configDir + path.sep + "plugins" + path.sep + "dmn-eval-plugin";](https://github.com/joleaf/obsidian-dmn-eval-plugin/blob/b473ff70c70d373bbef5d75058961b89f7f980d5/main.ts#L93)
You can use `this.manifest.dir` instead.

joleaf:
@joethei Thank you for the review.

I updated the code as you suggested.
However, with only `this.manifest.dir`, the functions `existsSync`, `unlinkSync`, and `appendFileSync` does not work (I don't know why but I guess that this has to do with the working directory).
Thus, the current code is a combination of `this.manifest.dir` and `this.app.vault.adapter.getBasePath()` to get the absolute path.

joethei:
That's what I meant, see the link text.
I can only link to entire lines in GH.

joleaf:
Yes thank you! Are the changes sufficient?

liamcain:
- [createDMNEvaluatorJar](https://github.com/joleaf/obsidian-dmn-eval-plugin/blob/b473ff70c70d373bbef5d75058961b89f7f980d5/main.ts#L19) I suggest wrapping this inside a [`onLayoutReady`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3626) callback since filesystem operations can be expensive and onload is called synchronously when Obsidian starts (so it's important for the function to finish executing quickly)
- [basePath](https://github.com/joleaf/obsidian-dmn-eval-plugin/blob/b473ff70c70d373bbef5d75058961b89f7f980d5/main.ts#L93) you should use [`getBasePath`](https://github.com/obsidianmd/obsidian-api/blob/6b2138aa28fb6d6f58f78e6fabec02186fdc0f82/obsidian.d.ts#L1217) instead (it's public in the api)

joleaf:
@liamcain Thank you!

> * [basePath](https://github.com/joleaf/obsidian-dmn-eval-plugin/blob/b473ff70c70d373bbef5d75058961b89f7f980d5/main.ts#L93) you should use [`getBasePath`](https://github.com/obsidianmd/obsidian-api/blob/6b2138aa28fb6d6f58f78e6fabec02186fdc0f82/obsidian.d.ts#L1217) instead (it's public in the api)

This is fixed.

PR #1566, opened 2023-01-22
Add plugin : Note aliases

pulsovi:

## Repo URL

Link to my plugin: https://github.com/pulsovi/obsidian-note-aliases

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
  - [ ] ~~`styles.css` _(optional)_~~
    - [X] I have added a license in the package.json file


liamcain:
- [const targetFile = await this.app.vault.create(targetPath, '')](https://github.com/pulsovi/obsidian-note-aliases/blob/713eb83afd782a13f5da2507308fdb933db175cb/main.ts#L35) I think you're better off using the private api `await app.fileManager.createNewFile(parent, path);` here. There are a few edge cases to consider, like creating parent folders, ensuring the path is formed correctly, and that the string is normalized, all of which createNewFile will take care of.

pulsovi:
> * [const targetFile = await this.app.vault.create(targetPath, '')](https://github.com/pulsovi/obsidian-note-aliases/blob/713eb83afd782a13f5da2507308fdb933db175cb/main.ts#L35) I think you're better off using the private api `await app.fileManager.createNewFile(parent, path);` here. There are a few edge cases to consider, like creating parent folders, ensuring the path is formed correctly, and that the string is normalized, all of which createNewFile will take care of.

Hello, 
Thank you for your suggestion.
What is the "private API" and where can I found the relevant documentation?

PR #1579, opened 2023-01-25
Add Codeblock Customizer to community-plugins.json

mugiwara85:

## Repo URL

Link to my plugin: https://github.com/mugiwara85/CodeblockCustomizer

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[this.app.workspace.on('css-change', this.handleCssChange.bind(this, settingsTab), this);](https://github.com/mugiwara85/CodeblockCustomizer/blob/1aeddae5e6661f8f6f42d45280896d49b0e95d32/src/main.ts#L57)
Please register this event so it is unloaded properly.
<https://marcus.se.net/obsidian-plugin-docs/events>

[MyPluginSettings](https://github.com/mugiwara85/CodeblockCustomizer/blob/1aeddae5e6661f8f6f42d45280896d49b0e95d32/src/Settings.ts#L1)
[MyPluginColors](https://github.com/mugiwara85/CodeblockCustomizer/blob/1aeddae5e6661f8f6f42d45280896d49b0e95d32/src/Settings.ts#L33)
Would recommend renaming these to something more meaningful.

[app.workspace.getActiveViewOfType(MarkdownView)](https://github.com/mugiwara85/CodeblockCustomizer/blob/1aeddae5e6661f8f6f42d45280896d49b0e95d32/src/ReadingView.ts#L23)

[if (!isCodeBlockHeaderEnabled && pluginSettings.bEnableLineNumbers){](https://github.com/mugiwara85/CodeblockCustomizer/blob/1aeddae5e6661f8f6f42d45280896d49b0e95d32/src/ReadingView.ts#L62-L69)
Don't hardcode styling like this, instead assign a class defined in your `styles.css`, that makes it easier for users to change the styling with a snippet.


You are sometimes using semicolons to end statements, sometimes you don't.
I would recommend sticking to one.

mugiwara85:
Thank you for the review @joethei! I will correct the things you described, and report back, once I'm done!

mugiwara85:
@joethei I corrected everything you described, and uploaded a new release (1.0.2). Is there anything else I should do, so the plugin gets released?

liamcain:
- [document.createElement](https://github.com/mugiwara85/CodeblockCustomizer/blob/febc6958dff80aa3b786ff26f561725a7785d014/src/ReadingView.ts#L94) you should just add CSS to target the collapsed state instead of creating new style tags. Your CSS selector is not specific enough and is likely going to collide with other classes or other plugins and cause unexpected bugs.
- [innerHTML](https://github.com/mugiwara85/CodeblockCustomizer/blob/febc6958dff80aa3b786ff26f561725a7785d014/src/ReadingView.ts#L151) you're better off using innerText or .setText() instead. [innerHTML can be dangerous](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml)
- [padding-right](https://github.com/mugiwara85/CodeblockCustomizer/blob/febc6958dff80aa3b786ff26f561725a7785d014/styles.css#L3) this looks like it will break the padding used to make space for fold markers
- [while(!codeElm.classList.contains("is-loaded"))](https://github.com/mugiwara85/CodeblockCustomizer/blob/febc6958dff80aa3b786ff26f561725a7785d014/src/ReadingView.ts#L15) what is this for?

mugiwara85:
> * [document.createElement](https://github.com/mugiwara85/CodeblockCustomizer/blob/febc6958dff80aa3b786ff26f561725a7785d014/src/ReadingView.ts#L94) you should just add CSS to target the collapsed state instead of creating new style tags. Your CSS selector is not specific enough and is likely going to collide with other classes or other plugins and cause unexpected bugs.

Sorry, my CSS skills are not the best. Could you explain it please a little bit more detailed, what the problem this is, and how I could fix it?

> * [innerHTML](https://github.com/mugiwara85/CodeblockCustomizer/blob/febc6958dff80aa3b786ff26f561725a7785d014/src/ReadingView.ts#L151) you're better off using innerText or .setText() instead. [innerHTML can be dangerous](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml)

In this case InnerHTML is set to hold a lineNumber. I am aware of the danger, but since it does not hold user inputted data, I thought it would be ok. But I can change that.

> * [padding-right](https://github.com/mugiwara85/CodeblockCustomizer/blob/febc6958dff80aa3b786ff26f561725a7785d014/styles.css#L3) this looks like it will break the padding used to make space for fold markers

No, it does not. This just "moves" the gutter next to the codeblock. The gutter is only displayed next to codeblock. Here is a picture which shows that the fold marker is displayed correctly. 
image


> * [while(!codeElm.classList.contains("is-loaded"))](https://github.com/mugiwara85/CodeblockCustomizer/blob/febc6958dff80aa3b786ff26f561725a7785d014/src/ReadingView.ts#L15) what is this for?

I didn't find a better solution but maybe you know one. The problem is that if you switch to reading view, then the syntax highlighting is applied. This can take some time (I don't know what it depends on, but it is different. Maybe the document size?). If I apply my styling before the syntax highlighting is applied, then it gets reverted, because of the syntax highlighting. So I have to wait a little, to load the document fully, and only apply my styling then when the syntax highlighting is already applied. If the code block has the syntax highlighting applied then it gets a class "is-loaded". That is what I am checking if the codeblock has that class applied or not, and wait a little.

liamcain:
> [document.createElement](https://github.com/mugiwara85/CodeblockCustomizer/blob/febc6958dff80aa3b786ff26f561725a7785d014/src/ReadingView.ts#L94) you should just add CSS to target the collapsed state instead of creating new style tags. Your CSS selector is not specific enough and is likely going to collide with other classes or other plugins and cause unexpected bugs.

All CSS in the app is applied globally. So if another plugin creates a div with a classname called "collapsed" then your plugin will hide that element. This type of bug will be very difficult to track down for another plugin dev since it's out of their control. For that reason, it's best to write CSS that's more "specific" meaning that the selectors in the CSS are more verbose and apply to fewer elements. You can just add to the styles.css file something like:

```css
.codeblock-customizer-pre.collapse {
  display: none;
}
```

> - [padding-right](https://github.com/mugiwara85/CodeblockCustomizer/blob/febc6958dff80aa3b786ff26f561725a7785d014/styles.css#L3) this looks like it will break the padding used to make space for fold markers

The gutter is used for more than just the fold indicators. You can also enable line numbers within the app. Also some other plugins add elements to the gutter (for example, the Lapel plugin). For this reason, it's best to avoid overriding this style. Also, shouldn't you be targeting `codeblock-customizer-gutterElements` instead?

> I didn't find a better solution but maybe you know one. The problem is that if you switch to reading view, then the syntax highlighting is applied. This can take some time (I don't know what it depends on, but it is different. Maybe the document size?). If I apply my styling before the syntax highlighting is applied, then it gets reverted, because of the syntax highlighting. So I have to wait a little, to load the document fully, and only apply my styling then when the syntax highlighting is already applied. If the code block has the syntax highlighting applied then it gets a class "is-loaded". That is what I am checking if the codeblock has that class applied or not, and wait a little.

Oh, okay, you're waiting for PrismJS to finish applying the styles. Okay, I don't have a better solution for you in mind.

liamcain:
I noticed there's something wrong with the ActiveLineHighlights in Live Preview. It's applying styling outside of the codeblock. It's always adding styles to the active line. I'm not even sure you need your own decorator to handle this. You should be able to do this with CSS only. There's already a cm-active class on the active line. So you just need to target the active line within the codeblock.

mugiwara85:
> > [document.createElement](https://github.com/mugiwara85/CodeblockCustomizer/blob/febc6958dff80aa3b786ff26f561725a7785d014/src/ReadingView.ts#L94) you should just add CSS to target the collapsed state instead of creating new style tags. Your CSS selector is not specific enough and is likely going to collide with other classes or other plugins and cause unexpected bugs.
> 
> All CSS in the app is applied globally. So if another plugin creates a div with a classname called "collapsed" then your plugin will hide that element. This type of bug will be very difficult to track down for another plugin dev since it's out of their control. For that reason, it's best to write CSS that's more "specific" meaning that the selectors in the CSS are more verbose and apply to fewer elements. You can just add to the styles.css file something like:
> 
> ```css
> .codeblock-customizer-pre.collapse {
>   display: none;
> }
> ```

Ok, I can correct that. 

> 
> > * [padding-right](https://github.com/mugiwara85/CodeblockCustomizer/blob/febc6958dff80aa3b786ff26f561725a7785d014/styles.css#L3) this looks like it will break the padding used to make space for fold markers
> 
> The gutter is used for more than just the fold indicators. You can also enable line numbers within the app. Also some other plugins add elements to the gutter (for example, the Lapel plugin). For this reason, it's best to avoid overriding this style. Also, shouldn't you be targeting `codeblock-customizer-gutterElements` instead?

I know I can enable linenumbers, but that's not what I want. Enabling line numbers sets linenumbers before every line. My guter only ads linenumbers before codeblock lines. It wouldn't make much sense enabling line numbers for the whole document and use those numbers to highlight lines, because if I insert one line anywhere in the document the wrong lines would be highlighted :). 
No, I use the  right class. If I uncomment that css code, then this is how it looks:
image
I don't do anything with the gutter. I add my own gutter. The problem is that there is one gutter "cm-gutters" which holds every other gutter:
image
And "cm-gutters" has this padding, which I don't need. If you have any other ideas, I am gladly hear it, because this is driving me crazy. My preferred way for adding linenumbers to codeblock would be inside the codeblock, and not before them. I can also do that, and will offer this as an option as well. However, this isn't working 100% correct. One problem is, that if I add a linenumber to a line, and I am deleting from the line with backspace, then the linenumber will jump up line, and there I can delete it. The other problem is that if a line "breaks" because it is too long, then the line number is not displayed correctly. Nevertheless, I will offer this option as well, and everybody can decide which one he prefers. I also asked this on the codemirror forum, and received this answer.
> 
> > I didn't find a better solution but maybe you know one. The problem is that if you switch to reading view, then the syntax highlighting is applied. This can take some time (I don't know what it depends on, but it is different. Maybe the document size?). If I apply my styling before the syntax highlighting is applied, then it gets reverted, because of the syntax highlighting. So I have to wait a little, to load the document fully, and only apply my styling then when the syntax highlighting is already applied. If the code block has the syntax highlighting applied then it gets a class "is-loaded". That is what I am checking if the codeblock has that class applied or not, and wait a little.
> 
> Oh, okay, you're waiting for PrismJS to finish applying the styles. Okay, I don't have a better solution for you in mind.

Exactly :)

mugiwara85:
> I noticed there's something wrong with the ActiveLineHighlights in Live Preview. It's applying styling outside of the codeblock. It's always adding styles to the active line. I'm not even sure you need your own decorator to handle this. You should be able to do this with CSS only. There's already a cm-active class on the active line. So you just need to target the active line within the codeblock.

Well, yes and no. I know there is a cm-activeLine (if I remember correctly) class, which I could use, but the problem is the following. Using the options in the settings tab, you can achieve following results:
- the active line is highlighted, only outside of codeblocks
- the active line is highlighted only inside of codeblocks
- the active line is highlighted with different colors. so it is highlighted with red only outside of codeblocks, but if you click inside a codeblock it is highlighted with a blue
- both inside and outside of codeblocks the active line is highlighted with the same color
The cm-activeLine is applied everywhere, so I couldn't achieve this with that class.

liamcain:
> I know I can enable linenumbers, but that's not what I want. Enabling line numbers sets linenumbers before every line

I'm not suggesting that you enable line numbers. The issue is that your breaking the padding for users that _do_ have line numbers enabled in the app. This is a side effect caused by your plugin that will break the experience for those users.

> The cm-activeLine is applied everywhere, so I couldn't achieve this with that class.

You can apply the styles to be combinatory. So the highlight should only be applied to elements that have both `codeblock-customizer-line-background` and `cm-activeLine`

mugiwara85:
@liamcain @joethei could one you please give an update here?

liamcain:
Thanks, just took another look and the issues with the gutter appear to be addressed. It still causes the active line to get highlighted even outside of code blocks, but that's less problematic. Thanks for your patience and sorry for all the back-and-forths. Glad to finally have this in the community plugin list 🙏

PR #1582, opened 2023-01-26
Add 'Double Colon Conceal' plugin

msrch:

Display double colon (i.e. Dataview inline fields) as a single colon in reading view (preview mode) for more natural experience.

## Repo URL

Link to my plugin: https://github.com/msrch/obsidian-double-colon-conceal

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
  - ~~`styles.css` _(optional)_~~


liamcain:
- [node instanceof HTMLElement](https://github.com/msrch/obsidian-double-colon-conceal/blob/6fc19d1c728f08f2b8cd311d44920293de435aa4/main.ts#L40) Here you'll actually want to use our custom [`instanceOf`](https://github.com/obsidianmd/obsidian-api/blob/606526e212cb0018690a729d7f0d1c8797071d6a/obsidian.d.ts#L53) instead of `instanceof HTMLElement`. So `if (someNode.instanceOf(HTMLElement) {}` The typical `instanceof HTMLElement` will fail in pop-out windows because the `HTMLElement` is defined on the Window instance. So the class definition in one window is actually a different class definition for each popout window. It's a very weird and annoying nuance, I know, but that's why we provide a custom function on the Node prototype.
- [node instanceof HTMLDivElement](https://github.com/msrch/obsidian-double-colon-conceal/blob/6fc19d1c728f08f2b8cd311d44920293de435aa4/main.ts#L32) same here. Use [`instanceOf`](https://github.com/obsidianmd/obsidian-api/blob/606526e212cb0018690a729d7f0d1c8797071d6a/obsidian.d.ts#L53).
- [node instanceof Text](https://github.com/msrch/obsidian-double-colon-conceal/blob/6fc19d1c728f08f2b8cd311d44920293de435aa4/main.ts#L51) here too...
- [node instanceof HTMLDivElement](https://github.com/msrch/obsidian-double-colon-conceal/blob/6fc19d1c728f08f2b8cd311d44920293de435aa4/main.ts#L32) and here

msrch:
@liamcain Thank you for your review!

Changes done as requested, please see:

- commit: https://github.com/msrch/obsidian-double-colon-conceal/commit/58f0d96f108d4c265583b54b1585459692fe7ed5
- new version released: https://github.com/msrch/obsidian-double-colon-conceal/releases/tag/1.1.3

PR #1587, opened 2023-01-27
Add Flomo Importer Plugin

jia6y:

## Repo URL

Link to my plugin: https://github.com/jia6y/flomo-to-obsidian

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [this.fsa.write](https://github.com/jia6y/flomo-to-obsidian/blob/6428c09bf4e5082cd0158d0a6db623a14f3db3e1/lib/impoter_ui.ts#L38) you should avoid doing multiple concurrent writes to the file system. It's less efficient than just writing all the data at once.
- [var](https://github.com/jia6y/flomo-to-obsidian/blob/6428c09bf4e5082cd0158d0a6db623a14f3db3e1/lib/impoter_ui.ts#L51) I recommend you change all instances of `var` to either `const` or `let`. `var` has function-level scope, so it can easily lead to bugs if you're not careful. Here's a pretty good guide with examples of why not to use `var`: [why `var` is obsolete](https://javascript.plainenglish.io/4-reasons-why-var-is-considered-obsolete-in-modern-javascript-a30296b5f08f). It also looks like the var is unused.
- [isDesktopOnly](https://github.com/jia6y/flomo-to-obsidian/blob/6428c09bf4e5082cd0158d0a6db623a14f3db3e1/manifest.json#L9) you will need to mark your plugin as Desktop only since you're using NodeJS specific libraries like `path`, `fs`, and `os`
- [close](https://github.com/jia6y/flomo-to-obsidian/blob/6428c09bf4e5082cd0158d0a6db623a14f3db3e1/lib/impoter_ui.ts#L202) I don't think you want to close the modal when a user presses "import" and there's no file attached.
- [console.log](https://github.com/jia6y/flomo-to-obsidian/blob/6428c09bf4e5082cd0158d0a6db623a14f3db3e1/lib/impoter_ui.ts#L200) you should display this message to the users somehow.

jia6y:
@liamcain Thanks for your comments, I have updated the code as you suggested.
- leveraged a buffer to store the content, writing all the data at once.
- removed the "var"
- set isDesktioOnly to "true"
- a Notice will pop up when no file is selected.

jia6y:
> @liamcain Thanks for your comments, I have updated the code as you suggested.
> 
> * leveraged a buffer to store the content, writing all the data at once.
> * removed the "var"
> * set isDesktioOnly to "true"
> * a Notice will pop up when no file is se

> @liamcain Thanks for your comments, I have updated the code as you suggested.
> 
> * leveraged a buffer to store the content, writing all the data at once.
> * removed the "var"
> * set isDesktioOnly to "true"
> * a Notice will pop up when no file is selected.

Hi Liam, thanks for pointed out the problems earlier, not sure if you could get a chance to look at it now.

jia6y:
Good day @joethei, seems the "plugin" label got removed after a new push, would you mind helping to add it back once you get time.

liamcain:
Really sorry for such a long wait with rereviewing this plugin!! Not sure how it fell through the cracks for so long 😞 that's totally my fault.

Excited to finally have in the community plugin list. Thanks for your patience 🙏

PR #1588, opened 2023-01-27
Add Plugin Better Mathjax

greasycat:

## Repo URL

Link to my plugin:

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[this.app.vault.off("modify", this.userDefinedFileChanged);](https://github.com/greasycat/BetterMathjax/blob/b1cee15a5cc5ae08b6399ea12c378ca848e38d70/main.ts#L47)
Instead of doing this you can wrap [this code](https://github.com/greasycat/BetterMathjax/blob/b1cee15a5cc5ae08b6399ea12c378ca848e38d70/main.ts#L43) inside of a `this.registerEvent()`, Obsidian will take care of the rest.

[id: 'better-mathjax-select-next-suggestion',](https://github.com/greasycat/BetterMathjax/blob/b1cee15a5cc5ae08b6399ea12c378ca848e38d70/src/commands.ts#L9)
Obsidian will automatically prefix the command id with the plugin id, you don't need to do it yourself.

[hotkeys: [](https://github.com/greasycat/BetterMathjax/blob/b1cee15a5cc5ae08b6399ea12c378ca848e38d70/src/commands.ts#L11-L16)
Avoid setting default hot keys for plugins that you intend for others to use.
Hot keys are highly likely to conflict with those defined by other plugins or by the user themselves.

[if (this.consoleLogEnabled) {](https://github.com/greasycat/BetterMathjax/blob/b1cee15a5cc5ae08b6399ea12c378ca848e38d70/src/logger.ts#L29)
I would not guard error messages behind a debug toggle.

[math.style.paddingLeft = "20px";](https://github.com/greasycat/BetterMathjax/blob/b1cee15a5cc5ae08b6399ea12c378ca848e38d70/src/mathjax-helper.ts#L50)
You should avoid assigning styles via javascript and instead move all these styles into CSS so that they are more easily adaptable by themes.


[const content = await this.app.vault.adapter.read(file.path);](https://github.com/greasycat/BetterMathjax/blob/b1cee15a5cc5ae08b6399ea12c378ca848e38d70/src/mathjax-helper.ts#L151)
[Prefer the Vault API over the Adapter API](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#prefer-the-vault-api-over-the-adapter-api)

Would also recommend moving the `main.ts` file inside the `src` folder.

PR #1589, opened 2023-01-27
Adds Backslash LaTeX Macro Plugin

Bjarno:

## Repo URL

Link to my plugin: https://github.com/Bjarno/obsidian-backslash-latex

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux -- will test later
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
  - [ ] `styles.css` _(optional)_ -- no `styles.css` present
      I have given proper attribution to these other projects in my `README.md`. -- This plugin does not depend on any other plugin: only on Obsidian itself.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5796,5 +5796,12 @@
         "author": "Michael Hansen",
         "description": "Render NPC and adversary statblocks for The One Ring 2E roleplaying game.",
         "repo": "modality/obsidian-the-one-ring-2e-statblocks"
+    },
+    {
+        "id": "backslash-latex-macro",
+        "name": "Backslash LaTeX Macro",
+        "author": "Bjarno Oeyen",
+        "description": "Replace LaTeX macros with (Unicode) characters. Inspired by DrRacket.",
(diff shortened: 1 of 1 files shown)


joethei:
[const platform = process.platform;](https://github.com/Bjarno/obsidian-backslash-latex/blob/d9b763bb15e6199219821509959a1a96914a3951/main.ts#L195)
Use [Platform](https://github.com/obsidianmd/obsidian-api/blob/bceb489fc25ceba5973119d6e57759d64850f90d/obsidian.d.ts#L2632-L2679) from the Obsidian API instead, that will also make this plugin be mobile compatible.

[hotkeys: [](https://github.com/Bjarno/obsidian-backslash-latex/blob/d9b763bb15e6199219821509959a1a96914a3951/main.ts#L202-L204)
Avoid setting default hot keys for plugins that you intend for others to use.
Hot keys are highly likely to conflict with those defined by other plugins or by the user themselves.

ericaxu:

PR #1590, opened 2023-01-27
Add Pinterest Widgets Plugin

AllJavi:

## Repo URL

Link to my plugin: https://github.com/AllJavi/pinterest-widget-obsidian

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[const yaml = Yaml.parse(source);](https://github.com/AllJavi/pinterest-widget-obsidian/blob/f3ce5c86d552e4a1079203291ae068ce0dc2ccad/main.ts#L50)
You don't need to import a seperate package for YAML parsing, the Obsidian API provides the [parseYaml](https://github.com/obsidianmd/obsidian-api/blob/bceb489fc25ceba5973119d6e57759d64850f90d/obsidian.d.ts#L2630) function.

[container.innerHTML = `<a](https://github.com/AllJavi/pinterest-widget-obsidian/blob/f3ce5c86d552e4a1079203291ae068ce0dc2ccad/main.ts#L62-L66)
[Avoid using `innerHTML`](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml)

[await fetch](https://github.com/AllJavi/pinterest-widget-obsidian/blob/f3ce5c86d552e4a1079203291ae068ce0dc2ccad/main.ts#L70)

[(0, eval)(clearData);](https://github.com/AllJavi/pinterest-widget-obsidian/blob/f3ce5c86d552e4a1079203291ae068ce0dc2ccad/main.ts#L76)
This is pretty dangerous, if someone takes over that pinterest server they can cause Obsidian to execute all kinds of nefarious code.
Please at least include a warning in the README.

AllJavi:
Thanks for the suggestions. I already made all the changes, and as I didn't know about the guidelines document, I made some recommended changes too. 
I really don't like using the eval call, but unfortunately it's the only way they let you render the embeds. So I just added a warning on the README

liamcain:
- Instead of requesting the remote copy from pinterest, I suggest bundling your own copy of pinit with the plugin. It appears to be available in an unminified version here: <https://github.com/pinterest/widgets/blob/master/pinit_main.js>
- [EventTarget.prototype.addEventListener](https://github.com/AllJavi/pinterest-widget-obsidian/blob/c7e1a5b174f0b9e98e2baed00b82f197e540bf21/main.ts#L10) what is this for? Ideally you don't set anything on the EventTarget prototype to avoid an unexpected side effects. There's also an issue that this won't work with [pop-out windows](https://help.obsidian.md/User+interface/Pop-out+windows) because they use a different `window` and `global` instance.

ericaxu:

PR #1591, opened 2023-01-27
Update description of plugin `terminal`

polyipseity:


ericaxu:
Would you consider describing it as "Integrate consoles, shells, and terminals inside Obsidian"?

polyipseity:
That's a great idea. Changed!

PR #1595, opened 2023-01-29
Add "air-quotes" plugin

alangrainger:

## Repo URL

Link to my plugin: https://github.com/alangrainger/obsidian-air-quotes

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[id: 'air-quote-insert',](https://github.com/alangrainger/obsidian-air-quotes/blob/cefdbb38c8c539fbc6054f58ff447893f0a3a01f/src/main.ts#L17)
you don't need to do it yourself.

[const child_process = spawn(pandocCommand, {](https://github.com/alangrainger/obsidian-air-quotes/blob/cefdbb38c8c539fbc6054f58ff447893f0a3a01f/src/pandoc.ts#L39-L43)
You don't check if the device is actually running Windows.

alangrainger:
@joethei I've made the required changes for the command ID and limiting the command to Windows only.

Regarding the semicolons - I don't see this convention enabled in the Obsidian sample plugin `.eslintrc`:

https://github.com/obsidianmd/obsidian-sample-plugin/blob/master/.eslintrc

I can add these if needed, but I'm just confused as to whether this is a requirement or if you were merely mentioning it as an aside. Please let me know.

joethei:
The semicolon part is only an aside, no need to fix it.

alangrainger:
Gotcha. It was very interesting to know! I've only been working with JS for 5 years, but I didn't realize there were situations like that where the code wouldn't parse correctly. 

Stylistically I prefer it without semicolons, but that's only because I'm used to it that way (JS Standard style). I might have to try it the other way.

alangrainger:
@joethei is there anything else I need to do to re-submit this for review?

liamcain:
- [bookPath](https://github.com/alangrainger/obsidian-air-quotes/blob/980ebfbde48f185d4b0182f14fd51386e1f5eece/src/main.ts#L29) you should use [`getFileCache`](https://github.com/obsidianmd/obsidian-api/blob/bceb489fc25ceba5973119d6e57759d64850f90d/obsidian.d.ts#L2378) and read the frontmatter that's already cached. That way you don't need to parse the file yourself. Since you're also supporting Dataview, I suggest using the dataview API instead of searching for :: key value pairs yourself.
- [case 'ArrowDown':](https://github.com/alangrainger/obsidian-air-quotes/blob/980ebfbde48f185d4b0182f14fd51386e1f5eece/src/search.ts#L116-L126) I suggest you add a modifier key in addition to listening for arrow keys. Otherwise you won't be able to use the arrow keys to move the cursor which sounds like a bad user experience.
- [window.addEventListener](https://github.com/alangrainger/obsidian-air-quotes/blob/980ebfbde48f185d4b0182f14fd51386e1f5eece/src/search.ts#L86) I suggest adding these to the current scope instead of registering the listener with the window. The scope is accessible [here](https://github.com/obsidianmd/obsidian-api/blob/bceb489fc25ceba5973119d6e57759d64850f90d/obsidian.d.ts#L2446). You can use this.scope.register([], 'Arrowdown', () => {})
- [onClose](https://github.com/alangrainger/obsidian-air-quotes/blob/980ebfbde48f185d4b0182f14fd51386e1f5eece/src/search.ts#L176) you also need to call this.component.unload() in the close to properly cleanup the rendered markdown.
- [basePath](https://github.com/alangrainger/obsidian-air-quotes/blob/980ebfbde48f185d4b0182f14fd51386e1f5eece/src/pandoc.ts#L37) you should use getBasePath instead. Then you won't need ts-ignore

ericaxu:
Hi @alangrainger , did you see Liam's comments above?

ericaxu:

alangrainger:
Hi @liamcain @ericaxu - picking this up again. Fixes and comments inline:

> [bookPath](https://github.com/alangrainger/obsidian-air-quotes/blob/980ebfbde48f185d4b0182f14fd51386e1f5eece/src/main.ts#L29) you should use [getFileCache](https://github.com/obsidianmd/obsidian-api/blob/bceb489fc25ceba5973119d6e57759d64850f90d/obsidian.d.ts#L2378) and read the frontmatter that's already cached. That way you don't need to parse the file yourself. Since you're also supporting Dataview, I suggest using the dataview API instead of searching for :: key value pairs yourself.

Using Dataview API means that the plugin will not be able to find the value in an example case where someone using Air Quotes on their mobile device, but doesn't have Dataview plugin installed on that device. This could be quite a common occurrence - someone who sets things up on their laptop (which uses Dataview), but uses their phone alongside their book for the actual quote inserting.

Given that case, and how simple the regex method is to match either Dataview or YAML, it seemed like the logical way to go.

> [case 'ArrowDown':](https://github.com/alangrainger/obsidian-air-quotes/blob/980ebfbde48f185d4b0182f14fd51386e1f5eece/src/search.ts#L116-L126) I suggest you add a modifier key in addition to listening for arrow keys. Otherwise you won't be able to use the arrow keys to move the cursor which sounds like a bad user experience.

There might be some confusion here is to what occurs during the modal popup. There is no cursor to move at the point where these key events are being listened for, so there's nothing to interrupt. Have a play with it and you'll see what I mean - the modal is functioning as designed.

> [window.addEventListener](https://github.com/alangrainger/obsidian-air-quotes/blob/980ebfbde48f185d4b0182f14fd51386e1f5eece/src/search.ts#L86) I suggest adding these to the current scope instead of registering the listener with the window. The scope is accessible [here](https://github.com/obsidianmd/obsidian-api/blob/bceb489fc25ceba5973119d6e57759d64850f90d/obsidian.d.ts#L2446). You can use this.scope.register([], 'Arrowdown', () => {})

Done - thanks for that improvement.

> [onClose](https://github.com/alangrainger/obsidian-air-quotes/blob/980ebfbde48f185d4b0182f14fd51386e1f5eece/src/search.ts#L176) you also need to call this.component.unload() in the close to properly cleanup the rendered markdown.

Done - thanks for that improvement.

> [basePath](https://github.com/alangrainger/obsidian-air-quotes/blob/980ebfbde48f185d4b0182f14fd51386e1f5eece/src/pandoc.ts#L37) you should use getBasePath instead. Then you won't need ts-ignore

If I use `app.vault.adapter.getBasePath()`, I get the error "Property 'getBasePath' does not exist on type 'DataAdapter'". It functions fine, but then I still have to add in a `ts-ignore`. What am I doing wrong here?

ericaxu:
Re-opened, thanks @alangrainger !

liamcain:
> If I use app.vault.adapter.getBasePath(), I get the error "Property 'getBasePath' does not exist on type 'DataAdapter'". It functions fine, but then I still have to add in a ts-ignore. What am I doing wrong here?

https://liamca.in/Obsidian/API+FAQ/filesystem/get+the+absolute+path+of+a+file here's some more info. You should either use getResourcePath() or check if the DataAdapter is a FileSystemAdapter.

> There might be some confusion here is to what occurs during the modal popup. There is no cursor to move at the point where these key events are being listened for, so there's nothing to interrupt. Have a play with it and you'll see what I mean - the modal is functioning as designed.

Oh oops! My mistake

PR #1601, opened 2023-01-29
Add Obsidian reading comments plugin

BumbrT:

## Repo URL

Link to my plugin: https://github.com/BumbrT/obsidian-reading-comments

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [ x] `main.js`
  - [x ] `manifest.json`


liamcain:
- [currentNote: MarkdownView](https://github.com/BumbrT/obsidian-reading-comments/blob/c2b1bb2e794aa8967aee5776ff169871da149b8b/src/obsidianPlugin.ts#L14) you shouldn't be trying to manage a reference to the active view yourself. This will prevent the view from being properly recycled from memory when its supposed to. If you just want the last open focused, then you can use [`getActiveFile`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4146). This should give you the same thing you're trying to capture. Then you can just get the file contents by reading the TFile instead of using getViewData()

BumbrT:
> * [currentNote: MarkdownView](https://github.com/BumbrT/obsidian-reading-comments/blob/c2b1bb2e794aa8967aee5776ff169871da149b8b/src/obsidianPlugin.ts#L14) you shouldn't be trying to manage a reference to the active view yourself. This will prevent the view from being properly recycled from memory when its supposed to. If you just want the last open focused, then you can use [`getActiveFile`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4146). This should give you the same thing you're trying to capture. Then you can just get the file contents by reading the TFile instead of using getViewData()

Hi, Thanks for comment. Is it possible to get file view by TFile, returned by "getActiveFile". After cleanup reference to current note, cant implement logic to navigate in file by click in my plugin window.

liamcain:
You can check if `app.workspace.activeEditor instanceof MarkdownView`.

BumbrT:
Tried this way, but when i click something in my plugin window, `app.workspace.activeEditor` is always null :(

liamcain:
Ah, sorry. That was a bad suggestion on my part then. For now I suggest you use the private API:

```ts
let view = workspace.getActiveFileView();
```

You'll need to tsignore it but it should get you the view you're looking for.

BumbrT:
Thank you, published new version with this active file/view reference approach.

PR #1604, opened 2023-01-30
Add obsidian-inbox plugin to community plugins list

Zachatoo:

## Repo URL

Link to my plugin: https://github.com/Zachatoo/obsidian-inbox

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [getActiveViewOfType](https://github.com/Zachatoo/obsidian-inbox/blob/e232da17364f734fd39eaf71f236c4ba8adbcef7/src/main.ts#L37) if you use [`activeEditor`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4009) instead, then you could support canvas files. Meaning, if a file card is focused within a canvas, the command could still work. You would just need to pass in the file info object instead of leaf
- [detachLeavesOfType](https://github.com/Zachatoo/obsidian-inbox/blob/e232da17364f734fd39eaf71f236c4ba8adbcef7/src/main.ts#L92) I recommend against detaching leaves unless a user specifically requests it. This will change the user's workspace layout when it doesn't need to. I've seen this anti-pattern a lot, so I created a sample repo with a pattern that I think makes more sense: [ensureLeafExists](https://github.com/liamcain/obsidian-example-custom-view/blob/194d4557abb52a41c0380a2f1ec75334d9ce2775/src/main.ts#L44)
- [inboxNotePath](https://github.com/Zachatoo/obsidian-inbox/blob/e232da17364f734fd39eaf71f236c4ba8adbcef7/src/main.ts#L187) why strip the extension when saving to settings?

PR #1621, opened 2023-02-04
Add File Diff plugin

friebetill:

## https://github.com/friebetill/obsidian-file-diff

Link to my plugin: [File Diff](https://github.com/friebetill/obsidian-file-diff)

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -5832,5 +5832,12 @@
         "author": "Boninall",
         "description": "A plugin to make your canvas work like a mindmap.",
         "repo": "quorafind/obsidian-canvas-mindmap"
+    },
+    {
+        "id": "file-diff",
+        "name": "File Diff",
+        "author": "Till Friebe",
+        "description": "View the differences between two files with merge options.",
+        "repo": "friebetill/obsidian-file-diff"
     }
 ]


joethei:
[.button-container {](https://github.com/friebetill/obsidian-file-diff/blob/e48981ab195223707e9dd22c414c60cd66a0f9c2/styles.css#L12-L18)
Don't overwrite Obsidian CSS in a plugin


[id: 'file-diff',](https://github.com/friebetill/obsidian-file-diff/blob/e48981ab195223707e9dd22c414c60cd66a0f9c2/src/main.ts#L13)
you don't need to do it yourself.

[// Show warning when this option is selected for the first time](https://github.com/friebetill/obsidian-file-diff/blob/e48981ab195223707e9dd22c414c60cd66a0f9c2/src/main.ts#L45-L51)
Keep in mind that localstorage is shared between vaults, if you want this to show up in each vault add `this.app.appId` to the key (this is not declared so your editor will show a warning)

[export function delay(ms: number): Promise<void> {](https://github.com/friebetill/obsidian-file-diff/blob/e48981ab195223707e9dd22c414c60cd66a0f9c2/src/utils/delay.ts#L1)
You can use the `sleep` function from the Obsidian API, which does basically the same thing.



---

friebetill:
Thank you for your feedback! I have incorporated your suggestions.

Regarding the note about `localStorage`, I'm not sure how to implement your feedback. However, since I don't think it's that important whether it's displayed once or multiple times, I didn't spend much time searching for a solution and I'm satisfied if it's only displayed once.

liamcain:
- [getMostRecentLeaf](https://github.com/friebetill/obsidian-file-diff/blob/9df590f6b209b50f4005684e7a8e101d596a76d2/src/main.ts#L65) you should use `getLeaf` here. getMostRecentLeaf isn't meant to be used for this case. The most recent leaf might be pinned, in which case, you shouldn't navigate away.
- [new](https://github.com/friebetill/obsidian-file-diff/blob/9df590f6b209b50f4005684e7a8e101d596a76d2/src/main.ts#L30) you shouldn't be constructing a View and opening it yourself. Instead, you should register the view with your plugin using [registerView](https://github.com/obsidianmd/obsidian-api/blob/bceb489fc25ceba5973119d6e57759d64850f90d/obsidian.d.ts#LL2721). There's more info about creating custom views [here](https://marcus.se.net/obsidian-plugin-docs/user-interface/views). If you don't register the view, then the workspace won't be able to properly serialize it.

friebetill:
Thanks again for the feedback!

- I am now using `getMostRecentLeaf`.
- I am now registering the view with `registerView`. Unfortunately, this means that I can no longer display a dynamically generated title in `getDisplayText`. This is because I have to pass a factory method of the view to the `registerView` method. Therefore, the view only has the information once `setState` is called, but `getDisplayText` is called earlier.

liamcain:
- [setState](https://github.com/friebetill/obsidian-file-diff/blob/6ee9e381d4e506a738cccf9d5f222dda4214c408/src/components/differences_view.ts#L43) you shouldn't be overriding setState here. Instead, just register the event listener inside the constructor. Then just add a check to ensure that file1 and file2 are set.
- [read](https://github.com/friebetill/obsidian-file-diff/blob/6ee9e381d4e506a738cccf9d5f222dda4214c408/src/components/differences_view.ts#L64) use [cachedRead](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3373) here instead
- [read](https://github.com/friebetill/obsidian-file-diff/blob/6ee9e381d4e506a738cccf9d5f222dda4214c408/src/components/differences_view.ts#L65) same here.
- [file1: TFile;](https://github.com/friebetill/obsidian-file-diff/blob/6ee9e381d4e506a738cccf9d5f222dda4214c408/src/components/differences_view.ts#L12) I suggest just moving these onto the DifferencesView instead of wrapping them in an object.
- [revealLeaf](https://github.com/friebetill/obsidian-file-diff/blob/6ee9e381d4e506a738cccf9d5f222dda4214c408/src/main.ts#L157) No need to call getLeavesOfType. Just assign the result of getLeaf(true) to a variable. Then call setViewState(...). Then pass it into revealLeaf.

> Unfortunately, this means that I can no longer display a dynamically generated title in getDisplayText. This is because I have to pass a factory method of the view to the registerView method. Therefore, the view only has the information once setState is called, but getDisplayText is called earlier. Do you know a way to give this information before getDisplayText is called?

I suspect that might be because you're overriding setState without calling super.setState().

friebetill:
Thank you for your feedback!

> [setState](https://github.com/friebetill/obsidian-file-diff/blob/6ee9e381d4e506a738cccf9d5f222dda4214c408/src/components/differences_view.ts#L43) you shouldn't be overriding setState here. Instead, just register the event listener inside the constructor. Then just add a check to ensure that file1 and file2 are set

I moved the EventListener to the constructor. However, I still override `setState` because when the FileDiff is initially opened, the state must be set and the view must be built. 

> [file1: TFile;](https://github.com/friebetill/obsidian-file-diff/blob/6ee9e381d4e506a738cccf9d5f222dda4214c408/src/components/differences_view.ts#L12) I suggest just moving these onto the DifferencesView instead of wrapping them in an object.

I don`t understand exactly what you mean. If I move the interface directly into the class, then I can't use the interface in `main.ts` anymore and lose accuracy with the types.

friebetill:
Yes the PR is ready to be merged. Thanks for the feedback!

PR #1634, opened 2023-02-06
Added x86 Assembly Flow Graphing Plugin

dwolfe884:

## Repo URL

Link to my plugin: https://github.com/dwolfe884/obsidian-x86-flow-graph

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[// Remember to rename these classes and interfaces!](https://github.com/dwolfe884/obsidian-x86-flow-graph/blob/ce396c6fd00234d40162fdd3271adcea119433e9/src/index.ts#L3)
Can get removed.

We have defined the Canvas spec here, you can use these data types by importing from `obsidian/canvas`.


I recommend you change all instances of `var` to either `const` or `let`. `var` has function-level scope, so it can easily lead to bugs if you're not careful. Here's a pretty good guide with examples of why not to use `var`: [why `var` is obsolete](https://javascript.plainenglish.io/4-reasons-why-var-is-considered-obsolete-in-modern-javascript-a30296b5f08f)


<https://flaviocopes.com/javascript-automatic-semicolon-insertion>, you don't need to fix this one

dwolfe884:
- [x] Removed unnecessary comments
- [x] Replaced `var` with `let` and `const`
- [x] Imported `obsidian/canvas` and refactored all relevant code to use those interfaces.

Thanks for pointing me in the right direction for the canvas code. I should have figured there was already that code somewhere

liamcain:
- [nodeid](https://github.com/dwolfe884/obsidian-x86-flow-graph/blob/487efeaf31a76c40ae618596a3438d72592b465e/src/index.ts#L5) why are all these variables defined outside your class? This means they'll never get properly cleaned up from memory after your command is executed. Since you have an array of Nodes, this will prevent the nodes from being cleared out of memory properly when they get removed.
- [nodes](https://github.com/dwolfe884/obsidian-x86-flow-graph/blob/487efeaf31a76c40ae618596a3438d72592b465e/src/index.ts#L29) you are assigning to nodes twice in a row
- [nodeid](https://github.com/dwolfe884/obsidian-x86-flow-graph/blob/487efeaf31a76c40ae618596a3438d72592b465e/src/index.ts#L5) you have nanoid in your package.json, maybe use that for generating the IDs instead.
- https://github.com/dwolfe884/obsidian-x86-flow-graph/tree/main/node_modules you shouldn't be commiting your node_modules folder to github
- [import * as internal from 'stream';](https://github.com/dwolfe884/obsidian-x86-flow-graph/blob/487efeaf31a76c40ae618596a3438d72592b465e/src/index.ts#L3) this import is unused and cause the plugin to break on mobile since this is a node.js import.

dwolfe884:
- [x] Moved class variables inside class and replaced references with `this.` instead of file wide globals.
- [x] removed duplicate assignment of nodes array
- [x] I'm using the incremental nature of my node/edge ID's to access previously created nodes, I could probably refactor this to use randomly generated nodeID's if you think that would be best. I'm leaving as is for now since I don't see an issue with it.
- [x] Removed my `node_modules` directory from github
- [x] Removed `import * as internal from 'stream';` from my imports.

Thanks @liamcain for the help! Let me know if I should work more on those ID changes.

liamcain:
Hi, thanks for the bump! Your earlier message somehow slipped through the cracks so I didn't have it marked as ready for re-review. I'll take a look this afternoon and merge it if everything looks good.

liamcain:
You really should fix the indentation in your file, it's very confusing that your class functions are not indented at all.

PR #1635, opened 2023-02-06
Add Image Upload Toolkit plugin

addozhang:

## Repo URL

Link to my plugin: https://github.com/addozhang/obsidian-image-upload-toolkit

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[import {existsSync} from "fs";](https://github.com/addozhang/obsidian-image-upload-toolkit/blob/829c3d21967f569e9bb20eff1ac6d91c810b1343/src/ui/publishSettingTab.ts#L4-L5)
This will chrash on mobile, since there are no node libraries available there.

[if (!existsSync(join((this.app.vault.adapter as FileSystemAdapter).getBasePath(), value))) {](https://github.com/addozhang/obsidian-image-upload-toolkit/blob/829c3d21967f569e9bb20eff1ac6d91c810b1343/src/ui/publishSettingTab.ts#L34)
Better check if `await this.app.vault.getAbstractFileByPath(value)` returns something, this will also be mobile compatible then.


[if (!await this.adapter.exists(path.normalize(image.path))) {](https://github.com/addozhang/obsidian-image-upload-toolkit/blob/829c3d21967f569e9bb20eff1ac6d91c810b1343/src/uploader/imageTagProcessor.ts#L39)
[Prefer the Vault API over the Adapter API](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#prefer-the-vault-api-over-the-adapter-api), better use the function mentioned previously.

[${await this.adapter.exists(path.normalize(image.path))](https://github.com/addozhang/obsidian-image-upload-toolkit/blob/829c3d21967f569e9bb20eff1ac6d91c810b1343/src/uploader/imageTagProcessor.ts#L41)
This will always be false, also code duplication.

addozhang:
@joethei Thanks for your helpful advice and I have updated code following your advice.

liamcain:
- [getActiveViewOfType](https://github.com/addozhang/obsidian-image-upload-toolkit/blob/a95a5c5491e4f1da39534c834f567928dd7ce3d8/src/uploader/imageTagProcessor.ts#L112) you can use [`activeEditor`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4009) instead which will support files embedded in Canvas too.
- [path, {join, parse} from "path";](https://github.com/addozhang/obsidian-image-upload-toolkit/blob/a95a5c5491e4f1da39534c834f567928dd7ce3d8/src/uploader/imageTagProcessor.ts#L2) sadly "path" is a nodejs library so this won't work on mobile. I think you're missing isDesktopOnly from your manifest.json as well, but if you want to use "path" you'll need to set this to false.

PR #1640, opened 2023-02-08
Add HK Code Block

HeekangPark:

## Repo URL

Link to my plugin:[HK Code Block](https://github.com/HeekangPark/obsidian-hk-code-block)

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[frag.createDiv().innerHTML = text.split("\n").map((line) => line.trim()).join("<br>");](https://github.com/HeekangPark/obsidian-hk-code-block/blob/8ace6262e8ad2a28f4f05ef3227e7f9ac9671b85/src/HKCodeBlockSettingTab.ts#L7)
[Avoid using `innerHTML`](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-innerhtml-outerhtml-and-insertadjacenthtml)

[tc.addOptions({](https://github.com/HeekangPark/obsidian-hk-code-block/blob/8ace6262e8ad2a28f4f05ef3227e7f9ac9671b85/src/HKCodeBlockSettingTab.ts#L31-L34)
Would recommend changing this to something more like:
```ts
tc.addOptions({
    "default_off": "default off, but on when specified",
    "always_off": "always off",
})
```
Same thing here, will also make stuff like this easier to read: [if (settings.useTitleGlobal === "default off, but on when specified" && (showTitle === true)) {](https://github.com/HeekangPark/obsidian-hk-code-block/blob/8ace6262e8ad2a28f4f05ef3227e7f9ac9671b85/src/HKCodeBlockProcessor.ts#L360)

[new Setting(containerEl)](https://github.com/HeekangPark/obsidian-hk-code-block/blob/8ace6262e8ad2a28f4f05ef3227e7f9ac9671b85/src/HKCodeBlockSettingTab.ts#L263-L277)
This would be better as a toggle.


[value as ("default off, but on when specified" | "always off");](https://github.com/HeekangPark/obsidian-hk-code-block/blob/8ace6262e8ad2a28f4f05ef3227e7f9ac9671b85/src/HKCodeBlockSettingTab.ts#L37)
This is innecessary

HeekangPark:
- using `innerHTML` in `settingTab.ts` : So far as I know, this is the only method to use html tags(ex. `<b>`) on the setting tab. Please let me know if there are better way to do so without using `innerHTML`.
- long dropdown item names : I believe long names(ex. `default off, but on when specified`) are more straightforward than short names(ex. `default_off`), while it is not very difficult to write and read thanks to IDE's support.
- change binary dropdown to toggle : done ✔️ 
- remove `as` statement : done ✔️

liamcain:
- [app](https://github.com/HeekangPark/obsidian-hk-code-block/blob/d1be28ba7e5902b4bebae5e316335b549d3bcd9e/src/processor.ts#L263) It's highly discouraged that you use the global app instance in your plugin. It's currently accessible from the global context for debugging purposes, and might be removed in the future. Instead, you should use the reference to App that you have from your plugin instance.
- [innerHTML](https://github.com/HeekangPark/obsidian-hk-code-block/blob/d1be28ba7e5902b4bebae5e316335b549d3bcd9e/src/processor.ts#L365) this should use setText() instead of innerhtml
- [view](https://github.com/HeekangPark/obsidian-hk-code-block/blob/d1be28ba7e5902b4bebae5e316335b549d3bcd9e/src/processor.ts#L263) this check will prevent the plugin from working inside of Canvas.

PR #1644, opened 2023-02-10
Add Perilous Writing plugin

sameersismail:

## Repo URL

Link to my plugin: https://github.com/sameersismail/obsidian-perilous-writing

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[id: "perilous-writing-short-session",](https://github.com/sameersismail/obsidian-perilous-writing/blob/52426ad6439073e0856a2c791a57d1e7b30ab4e7/src/main.ts#L20)

[wrapper.id = PROGRESS_BAR_BACKGROUND_ID;](https://github.com/sameersismail/obsidian-perilous-writing/blob/52426ad6439073e0856a2c791a57d1e7b30ab4e7/src/perilous/dom.ts#L86-L92)
You should avoid assigning styles via JavaScript and instead move all these styles into CSS so that they are more easily adaptable by themes and snippets

liamcain:
- [getEditorTabContainer](https://github.com/sameersismail/obsidian-perilous-writing/blob/ff49f0b2053f11cca90ba9f415522f9beff9c8f2/src/perilous/dom.ts#L12) I don't recommend traversing the dom like this to find the tab container. Instead, you already have the View. There is a reference to the tab container element if you try hard enough. The API for that isn't public but it should be accessible with something like `view.parent.tabsContainerEl`.
- [setProgressPercent](https://github.com/sameersismail/obsidian-perilous-writing/blob/ff49f0b2053f11cca90ba9f415522f9beff9c8f2/src/perilous/dom.ts#L29) is there only supposed to be one progressbar total? What happens if the workspace has a split in it?
- I think you want to return the session to your plugin class and make sure to call scheduler.teardown() when the plugin is unloaded. I don't see the scheduler is unloaded currently.

PR #1650, opened 2023-02-12
added obsidian-reloader to community plugins

o-su:

## Repo URL
Link to my plugin: https://github.com/o-su/obsidian-file-reloader

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
Please fill the pull request template.

o-su:
added PR template

liamcain:
Could you provide more information about what this plugin does and its use case in the README? The code doesn't seem to match the current description so I'm wondering if I'm missing something.

o-su:
@liamcain I have obsidian vault stored on a network drive. When I edit the md file on one pc, the change does not automatically appear in obsidian on the other pc without restarting the application. This plugin is a workaround for this issue.

liamcain:
My best guess for why this workaround is happening is because revealLeaf causes a "focus" event to trigger on the active window. But even then, I don't understand why that would cause the file to be processed. Given that this is a very specific use case that's for your own personal use, I don't think it should be included in the list of community plugins. If you can provide some more information or explanation as to the root cause of the bug, then maybe we can patch this into Core Obsidian instead of having it as a plugin.

ericaxu:

PR #1664, opened 2023-02-14
New plugin: Zen

Maxymillion:
# New plugin: Zen

## Repo URL

Link to my plugin: [Maxymillion/zen](https://github.com/Maxymillion/zen)

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[console.log(`${pluginConfig.name}: Loading`);](https://github.com/Maxymillion/zen/blob/704576c35d406a2a157f6dba7059747aacf2dd6c/src/main.ts#L20-L21)
Same as in your other plugin, better to access `this.plugin.manifest.name`;

[id: 'zen-toggle',](https://github.com/Maxymillion/zen/blob/704576c35d406a2a157f6dba7059747aacf2dd6c/src/main.ts#L33)

[document.body.className = document.body.className.split(" ").filter(c => !c.startsWith("zen-module--")).join(" ").trim();](https://github.com/Maxymillion/zen/blob/704576c35d406a2a157f6dba7059747aacf2dd6c/src/main.ts#L76)
This will only work for the main window, any additional popout windows will not be used.

[// @ts-ignore](https://github.com/Maxymillion/zen/blob/704576c35d406a2a157f6dba7059747aacf2dd6c/src/ui/ZenView.ts#L74-L76)
Instead of this I would recommend something like:
```
let inner = headerInner.createSpan();
setIcon(inner, 'eye-off');
```

Maxymillion:
Thanks @joethei,

I've resolved the given issues and retro actively updated all the `document.body` references to a [pop-out window safe](https://forum.obsidian.md/t/plugin-developers-how-to-update-your-plugin-to-support-pop-out-windows/40228) variant.

liamcain:
- [createHeaderIcon](https://github.com/Maxymillion/zen/blob/833eb7f72edd680c2ed20208fcbfb0a445d36429/src/ui/ZenView.ts#L37) I really don't like how hacky this approach is. It breaks a lot of Obsidian user expectations by creating an immovable tab. I think you're better off creating a ribbon item or a statusbar item to toggle Zen mode. Those use existing interface constructs that won't create extra confusion for users.

PR #1672, opened 2023-02-18
Close-similar-tabs

1C0D:
# Close Similar Tabs

## Repo URL

Link to my plugin:

## Release Checklist
- [x ] I have tested the plugin on
  - [x ]  Windows
  - [x ]  macOS
  - [x ]  Linux
  - [x ]  Android _(if applicable)_
  - [x ]  iOS _(if applicable)_
  - [x ] `main.js`
  - [x ] `manifest.json`
- [x ] I have added a license in the LICENSE file.


1C0D:
Close Similar Tabs

liamcain:
I think your approach to finding duplicates is making a couple few assumptions.

- You're assuming that iterating over the leaves will be ordered based on usage. You're detaching the last leaf in the leaves that match, but you're not being consistent about whether you want to keep the oldest or the newest leaf. So this behavior vary which I don't think users will want.
- You're likely overcomplicating things too much. Since the layout will change every time a new leaf gets added/removed, why not just compare leaves against the current activeLeaf? Then you just need to iterate over all leaves and find leaves that have the same matching file.
- You're assuming that leaves with the same file are the same type. An outline view, for example, a file associated. So your plugin will end up closing the outline view too. You should probably check that leaf.view instance MarkdownView to avoid closing things like backlinks, outline view, local graph view, etc.

1C0D:
I fixed the code as asked

liamcain:
- [this.app.workspace.getLeaf](https://github.com/1C0D/Obsidian-Close-Similar-Tabs/blob/dbea5c69ac427903a6523648eaac6d4ebc47f31d/main.ts#L15) I know the documentation has it marked as deprecated, but you should actually be using [activeLeaf](https://github.com/obsidianmd/obsidian-api/blob/bceb489fc25ceba5973119d6e57759d64850f90d/obsidian.d.ts#L3991) here. getLeaf is only meant to be used in combination with openFile. In some cases it can cause a new leaf to get created.
- What happened to your logic to check that you're only closing tabs in the same container as the active leaf?

1C0D:
I have addressed the feedback provided and made the following changes to the plugin:

- Implemented two modes for closing duplicate tabs, one to keep the oldest tab and one to keep the newest tab. 
- Changed the approach for finding duplicates by iterating over all leaves and comparing them against the active leaf. 
- Added a check to ensure that only tabs of the same type are closed to avoid accidentally closing tabs like backlinks, outline view, and local graph view. 
- Fixed the use of deprecated methods and renamed the class to follow naming conventions.
- Added more information to the README to explain what the plugin does and how to use it.
Please let me know if you have any further feedback or suggestions. Thank you for your help in improving the plugin!

1C0D:
I have addressed the feedback provided and made the following changes to the plugin:
Implemented two modes for closing duplicate tabs, one to close duplicates within each window (by default), and one to close in all windows.
Changed the approach for finding duplicates by iterating over all leaves and comparing them against the active leaf.
Please let me know if you have any further feedback or suggestions. Thank you for your help in improving the plugin!

liamcain:
Looks good!

- https://github.com/1C0D/Obsidian-Close-Similar-Tabs/blob/228b3fe246b66ecd21c96cb4ea77b0b8064cb1fd/src/settings.ts#L33 This is just feedback on the user experience, but if you need to add explanation text for what "on" and "off" each mean for a toggle, then you might be better off using a dropdown. Then you could have the options be more explanatory like "Current window only" vs "All windows"
- https://github.com/1C0D/Obsidian-Close-Similar-Tabs/blob/228b3fe246b66ecd21c96cb4ea77b0b8064cb1fd/src/main.ts#L65 you already have this as a variable. You can use leafPath here

1C0D:
Thank you for these improvements. Done

PR #1675, opened 2023-02-18
Add plugin: Tasks Calendar Wrapper

Leonezz:
Add obsidian-tasks-calendar-wrapper

## Repo URL

Link to my plugin: https://github.com/Leonezz/obsidian-tasks-calendar-wrapper

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["fundingUrl": "",](https://github.com/Leonezz/obsidian-tasks-calendar-wrapper/blob/9ba77f1fb92ded73f357eb2d366329ebfe4c5c28/manifest.json#L9)
If you don't have one remove this.

[this.app.workspace.detachLeavesOfType(TIMELINE_VIEW);](https://github.com/Leonezz/obsidian-tasks-calendar-wrapper/blob/9ba77f1fb92ded73f357eb2d366329ebfe4c5c28/src/main.ts#L42)
I recommend you don't detach leaves on unload.
This works for when the plugin gets disabled/uninstalled, but when the plugin gets updated, it means the view gets moved around in the user's workspace which is often unwanted. I recommend just removing this and letting the user manually close the views if they don't want them anymore.

[return t;](https://github.com/Leonezz/obsidian-tasks-calendar-wrapper/blob/9ba77f1fb92ded73f357eb2d366329ebfe4c5c28/src/settings.ts#L605)
[return cp;](https://github.com/Leonezz/obsidian-tasks-calendar-wrapper/blob/9ba77f1fb92ded73f357eb2d366329ebfe4c5c28/src/settings.ts#L610)
You don't need to return here.


[return](https://github.com/Leonezz/obsidian-tasks-calendar-wrapper/blob/9ba77f1fb92ded73f357eb2d366329ebfe4c5c28/src/settings.ts#L676)
This does not make sense, remove the return.

[this.app.metadataCache.off('resolved', this.onReloadTasks);](https://github.com/Leonezz/obsidian-tasks-calendar-wrapper/blob/9ba77f1fb92ded73f357eb2d366329ebfe4c5c28/src/views.tsx#L54)
Since you are already registering this above you don't need this, Obsidian should handle it automatically.

[import moment from "moment";](https://github.com/Leonezz/obsidian-tasks-calendar-wrapper/blob/9ba77f1fb92ded73f357eb2d366329ebfe4c5c28/utils/options.ts#L1)
Better to import moment from the Obsidian package.

Leonezz:
Thanks for the reviewing @joethei ,

I have made some changes accordingly:

1. The empty fundingUrl is now removed (https://github.com/Leonezz/obsidian-tasks-calendar-wrapper/commit/b2741c419cb7284c7b19e2447f390448cb100602)
2. The unnecessary detaching is now removed (https://github.com/Leonezz/obsidian-tasks-calendar-wrapper/commit/fd1867c898e91b666f6783850de24251a218122f)
3. The unnecessary returning is now removed (https://github.com/Leonezz/obsidian-tasks-calendar-wrapper/commit/30e5fcbfccde015e1f0c3c23b2763f89a35b1224)
4. The `return new Notice` is designed to early return the function, so that code outside the `if` scope  can be skipped. But I believe an `else` will also do, hence the `return` is removed in the above commit.
5. The unnecessary unregister is now removed (https://github.com/Leonezz/obsidian-tasks-calendar-wrapper/commit/9698292bf3a66507a28d7ffe911d4b6d2f2c5b1a)
6. `moment` is now importing from `obsidian` (https://github.com/Leonezz/obsidian-tasks-calendar-wrapper/commit/f4a32fb3e0ab4b03556f7f7696b12c5375c72850)

Thanks again for the reviews, I think it's now ready for another round of reviewing.

liamcain:
- [timelineView](https://github.com/Leonezz/obsidian-tasks-calendar-wrapper/blob/6dd608488d53fa9f05ff43f14bf65c3025c5fb3e/src/main.ts#L19) - [Avoid managing references to custom views](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#avoid-managing-references-to-custom-views) Doing so can cause memory leaks or unintended consequences. Instead, a better approach is to have your plugin emit an event and have your View listen for that event. This is a better event flow for your data and will decouple your plugin from your view.

Leonezz:
@liamcain Thanks for the review, I will change that in the next release.

PR #1677, opened 2023-02-19
Update community-plugins.json

popscallion:

## Repo URL

Link to my plugin: [No Empty Windows](https://github.com/popscallion/obsidian-no-empty-windows)

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:

liamcain:
- [tabs.length == 1](https://github.com/popscallion/obsidian-no-empty-windows/blob/c497065e725557f65323c5d341acb9a4219d4e54/main.ts#L11) This assertion isn't correct. A window can have more than one tab group. So this will actually cause the window to close any time you try to close a split tab group.
- There's another reason why this behavior isn't part of the app. Closing the main window will close the whole vault. So if you have popout windows open, then they will also be closed.

popscallion:
Thanks both, I've updated the plugin to be much smarter with how it handles multiple tab groups and poppets and rereleased as v1.1.0

PR #1678, opened 2023-02-20
add o2 plugin

songkg7:

## Repo URL

Link to my plugin: [o2](https://github.com/songkg7/o2)

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[id: 'o2-converting',](https://github.com/songkg7/o2/blob/860c06dc6e22096493d92bc5912739dccf5d8bbc/src/main.ts#L12)


[// If the plugin hooks up any global DOM events (on parts of the app that doesn't belong to this plugin)](https://github.com/songkg7/o2/blob/860c06dc6e22096493d92bc5912739dccf5d8bbc/src/main.ts#L27-L35)
Remove these



songkg7:
@joethei As you advised, I deleted unnecessary code and modified `.eslintrc` to use semicolons. Thank you for your kindness.

liamcain:
- [converting](https://github.com/songkg7/o2/blob/8cd7e361f3fc333f50303beda2488c3449433a54/src/main.ts#L13) the command name here should be more specific. Like "convert to Jekyll"
- [O2Modal](https://github.com/songkg7/o2/blob/8cd7e361f3fc333f50303beda2488c3449433a54/src/o2Modal.ts#L3) you can remove this
- [copyToPublishedDirectory](https://github.com/songkg7/o2/blob/8cd7e361f3fc333f50303beda2488c3449433a54/src/jekyll/chirpy.ts#L41) I don't understand why this is happening at the beginning of this function. Shouldn't the output files be what ends up in the published directory instead of a copy of the original files? Or are these just backup files?
- [this.app](https://github.com/songkg7/o2/blob/8cd7e361f3fc333f50303beda2488c3449433a54/src/jekyll/chirpy.ts#L106) `this` - you're not inside a class scope so `this` is referring to the global scope. It's highly discouraged that you use the global app instance in your plugin. It's currently accessible from the global context for debugging purposes, and might be removed in the future. Instead, you should use plugin.app
- In general I think your plugin is lacking some more usage instructions and information in the readme and in settings. The usage of "Ready directory" and "Published directory" is unclear to me and likely will be to users.

songkg7:
> * [converting](https://github.com/songkg7/o2/blob/8cd7e361f3fc333f50303beda2488c3449433a54/src/main.ts#L13) the command name here should be more specific. Like "convert to Jekyll"

Exactly. The name was quite improvised. I forgot to change this.

> * [O2Modal](https://github.com/songkg7/o2/blob/8cd7e361f3fc333f50303beda2488c3449433a54/src/o2Modal.ts#L3) you can remove this

removed.

> * [copyToPublishedDirectory](https://github.com/songkg7/o2/blob/8cd7e361f3fc333f50303beda2488c3449433a54/src/jekyll/chirpy.ts#L41) I don't understand why this is happening at the beginning of this function. Shouldn't the output files be what ends up in the published directory instead of a copy of the original files? Or are these just backup files?

Even after publishing the note, I wanted to be able to read it inside the obsidian, so I needed to copy it to a specific folder. However, I agreed with you that this process was confusing because it existed before the conversion began, not after the publication. So I changed it to the form of backup.

> * [this.app](https://github.com/songkg7/o2/blob/8cd7e361f3fc333f50303beda2488c3449433a54/src/jekyll/chirpy.ts#L106) `this` - you're not inside a class scope so `this` is referring to the global scope. It's highly discouraged that you use the global app instance in your plugin. It's currently accessible from the global context for debugging purposes, and might be removed in the future. Instead, you should use plugin.app

Modified to use `plugin.app`.

> * In general I think your plugin is lacking some more usage instructions and information in the readme and in settings. The usage of "Ready directory" and "Published directory" is unclear to me and likely will be to users.

I changed the description of 'README.md' file and 'setting' more specifically.

I have incorporated most of the requirements you proposed. Developing the Obsidian plugin and working with TypeScript are both very new to me. However, your review has been very specific, and it has been a great help for my learning. Many thanks.

PR #1688, opened 2023-02-21
Added open-files-with-commands plugin 

LostPaul:

## Repo URL

Link to my plugin: https://github.com/LostPaul/ob-open-files-with-commands

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [find](https://github.com/LostPaul/ob-open-files-with-commands/blob/5f5ca16685ee7b7bb4e2d6ff211ade735b5240c0/src/main.ts#L25) this assumes there's only one command per file. Likely true, but not always true. I suggest just switching this to a for loop and updating the command for each match.
- [find](https://github.com/LostPaul/ob-open-files-with-commands/blob/5f5ca16685ee7b7bb4e2d6ff211ade735b5240c0/src/main.ts#L33) here too
- [file](https://github.com/LostPaul/ob-open-files-with-commands/blob/5f5ca16685ee7b7bb4e2d6ff211ade735b5240c0/src/main.ts#L13) you should check that file instanceof TFile since this menu is also used when right-clicking on a folder.

LostPaul:
> * [find](https://github.com/LostPaul/ob-open-files-with-commands/blob/5f5ca16685ee7b7bb4e2d6ff211ade735b5240c0/src/main.ts#L25) this assumes there's only one command per file. Likely true, but not always true. I suggest just switching this to a for loop and updating the command for each match.
> 
>     * [find](https://github.com/LostPaul/ob-open-files-with-commands/blob/5f5ca16685ee7b7bb4e2d6ff211ade735b5240c0/src/main.ts#L33) here too
> 
>     * [file](https://github.com/LostPaul/ob-open-files-with-commands/blob/5f5ca16685ee7b7bb4e2d6ff211ade735b5240c0/src/main.ts#L13) you should check that file instanceof TFile since this menu is also used when right-clicking on a folder.

When you try to add more than one command to a file it should respond to you with that file already has a command.

> [file](https://github.com/LostPaul/ob-open-files-with-commands/blob/5f5ca16685ee7b7bb4e2d6ff211ade735b5240c0/src/main.ts#L13) you should check that file instanceof TFile since this menu is also used when right-clicking on a folder.
oh yeah I'll change that the problem is already made some other changes that added the option to open the files in the right leaf but currently it opens it again in the leaf so how would I check if the user already opened the file in that leaf? (Already asked here https://discord.com/channels/686053708261228577/840286264964022302/1079104143425278073)

LostPaul:
Anyways I made the change and that's the issue I'm talking about also do you have better name suggestions?

https://user-images.githubusercontent.com/70213368/221383750-e600f8b6-74a0-4368-87b7-0d5d01440ecf.mp4

liamcain:
For names, what about:
- Open
- Open below
- Open to the right
- Open in right sidebar
- Open in left sidebar

The term "leaf" is only really understood by developers. We don't use the term at all for user facing things, so it's better to use the term tab or sidebar when appropriate.

LostPaul:
> For names, what about:
> 
>     * Open
> 
>     * Open below
> 
>     * Open to the right
> 
>     * Open in right sidebar
> 
>     * Open in left sidebar
> 
> 
> The term "leaf" is only really understood by developers. We don't use the term at all for user facing things, so it's better to use the term tab or sidebar when appropriate.

Changed

PR #1693, opened 2023-02-23
Update Nebula theme: comply with 1.0 and remove legacy label

dlccyes:
[Relevant commit](https://github.com/dlccyes/obsidian-nebula/commit/dbbf8891287e12416f6059f0e61355fd3d48a3e4)


joethei:

PR #1696, opened 2023-02-25
Add JichouP/obsidian-marp-plugin

JichouP:

## Repo URL

Link to my plugin: <https://github.com/JichouP/obsidian-marp-plugin>

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[id: 'marp-open-preview',](https://github.com/JichouP/obsidian-marp-plugin/blob/2ab9fe3e813445e488350da9b28feb2dd701ccd0/src/main.ts#L33)

liamcain:
- [FileState](https://github.com/JichouP/obsidian-marp-plugin/blob/2ab9fe3e813445e488350da9b28feb2dd701ccd0/src/fileState.ts#L3) instead of making this a global state for the plugin, I suggest you make this a viewState. That will have a lot of benefits:
  - You'll be able to have multiple previews at once
  - The file will get serialized to the layout so reopening Obsidian so the view can be restored with the same state
  - you could make it work like our existing sidebar views. The preview could reflect the currently focused leaf and pinning it would freeze the state to a given file.

JichouP:
@joethei 
I didn't know that.
I removed the prefix in this PR. <https://github.com/JichouP/obsidian-marp-plugin/pull/2>

JichouP:
@liamcain 
Thanks for pointing this out.
The `registerEvent` has been fixed in this PR. <https://github.com/JichouP/obsidian-marp-plugin/pull/4>

However, for the `FileState`, I have no idea how to pass the `viewState` to the `PreviewView`.
Is it possible to pass the `state` in the following way?
And if so, how do I reference it from `PreviewView`?
``js
setViewState(
  {
    type: MARP_PREVIEW_VIEW_TYPE,
    active: true,
    state: {
      file: ...
    }
  }
)
```

Or did you intend to create a property `this.fileState` and a method `this.setFileState` on `PreviewView` to access the state?

liamcain:
So, I was writing up a response to this when I realized there really isn't enough documentation about storing state on the view. So I wrote up a long post about it here: <https://liamca.in/Obsidian/API+FAQ/views/persisting+your+view+state>

Anything you pass into `state` from setViewState will get passed into `setState` on the View.

JichouP:
@liamcain and @joethei 
Thank you very much for your kind attention.
And now that `viewState` works, I have created a new release `1.0.1`!
<https://github.com/JichouP/obsidian-marp-plugin/releases/tag/1.0.1>

I hope you will take a look at it when you are free.

liamcain:
Merging since I think it's fine to go in now, but I left a comment on the latest commit. You shouldn't need to be calling setState yourself.

PR #1705, opened 2023-02-27
Add file publisher plugin

yiglas:

## Repo URL

Link to my plugin: https://github.com/yiglas/obsidian-file-publisher

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6071,5 +6071,12 @@
         "description": "This is a syntax converting plugin to make obsidian markdown syntax compatible with other markdown syntax.",
         "author": "haril song",
         "repo": "songkg7/o2"
+    },
+    {
+        "id": "file-publisher",
+        "name": "File Publisher",
+        "description": "Publishes a file to a given api.",
+        "author": "Devin Sackett",
+        "repo": "yiglas/obsidian-file-publisher"
     }
 ]


joethei:
[import * as fs from "fs";](https://github.com/yiglas/obsidian-file-publisher/blob/da8a70801ae98a459bc553e9400ea11819bcf922/main.ts#L6)
[import * as path from "path";](https://github.com/yiglas/obsidian-file-publisher/blob/da8a70801ae98a459bc553e9400ea11819bcf922/main.ts#L15-L16)
This will break on mobile, as Node.js is not available there, best to use the Vault API.

[const adapter = this.app.vault.adapter as FileSystemAdapter;](https://github.com/yiglas/obsidian-file-publisher/blob/da8a70801ae98a459bc553e9400ea11819bcf922/main.ts#L35)
This will also break on mobile, as `FileSystemAdapter` is not used there, best to use the Vault API instead of the Adapter API.

[axios.post(url, data, {](https://github.com/yiglas/obsidian-file-publisher/blob/da8a70801ae98a459bc553e9400ea11819bcf922/main.ts#L96-L101)

yiglas:
@joethei thank you for the review. I've updated the release 

1. Removed fs and path references
2. Removed the usage of the `FileSystemAdapter` 

I wasn't able to replace the usage of axios, the requestUrl function wouldn't post the body of the request to the API server. I am still looking into but would like to release this version if possible.

liamcain:
- [rename](https://github.com/yiglas/obsidian-file-publisher/blob/0d79ab63561615add0d4171eae4d193829f6e271/main.ts#L140) you should use [`renameFile`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L1215) instead of `rename` so that links get renamed too.
- [menu.addItem](https://github.com/yiglas/obsidian-file-publisher/blob/0d79ab63561615add0d4171eae4d193829f6e271/main.ts#L52) maybe only add the menu item if the url is configured.

PR #1709, opened 2023-02-28
Add the theme "Elegance"

Victologo:


ericaxu:

Victologo:
Thanks for the advices. I don't know how GitHub and the synchronization with Obsidian works. Thanks.

PR #1710, opened 2023-03-01
Add new TiddlyWiki import/export plugin

lucasbordeau:

## Repo URL

Link to my plugin: https://github.com/lucasbordeau/obsidian-tiddlywiki

## Release Checklist
  - [X]  Windows
  - [x]  macOS
  - [X]  Linux
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[import * as path from 'path';](https://github.com/lucasbordeau/obsidian-tiddlywiki/blob/c95363ec1dd2fcfa9a94235202b42ae1924e20b6/main.ts#L2)
This will break on mobile, as Node.js is not available there.
Either rewrite this to use the Vault/Adapter API from Obsidian, or set `isDesktopOnly` to true in the manifest.

[alert('Success !')](https://github.com/lucasbordeau/obsidian-tiddlywiki/blob/c95363ec1dd2fcfa9a94235202b42ae1924e20b6/main.ts#L58)
Instead of an `alert` I would recommend a Notice from the Obsidian API.

lucasbordeau:
Done in [1.0.3](https://github.com/lucasbordeau/obsidian-tiddlywiki/releases/tag/1.0.3) : 

- Set plugin in desktopOnly mode
- Added a nice Notice instead of alert

lucasbordeau:
@joethei Ready for review ;)

liamcain:
- [SampleSettingTab](https://github.com/lucasbordeau/obsidian-tiddlywiki/blob/228851d2fe01f05ad41e87107e516594c3f2bb99/main.ts#L14)you should rename your class to be more unique (it's not a sample anymore)
- [basePath](https://github.com/lucasbordeau/obsidian-tiddlywiki/blob/228851d2fe01f05ad41e87107e516594c3f2bb99/main.ts#L81) you should use [`getBasePath`](https://github.com/obsidianmd/obsidian-api/blob/6b2138aa28fb6d6f58f78e6fabec02186fdc0f82/obsidian.d.ts#L1217) here instead. It's in the public api spec so no need for ts-ignore
- [basePath](https://github.com/lucasbordeau/obsidian-tiddlywiki/blob/228851d2fe01f05ad41e87107e516594c3f2bb99/main.ts#L52) same here
- [test.json](https://github.com/lucasbordeau/obsidian-tiddlywiki/blob/228851d2fe01f05ad41e87107e516594c3f2bb99/main.ts#L85) you should probably make this a more meaningful name or make it user configurable even. VaultName-timestamp.json is probably best.

PR #1712, opened 2023-03-01
Add global-search-and-replace

MahmoudFawzyKhalil:

## Repo URL

Link to my plugin: https://github.com/MahmoudFawzyKhalil/obsidian-global-search-and-replace/

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[id: "snr-search-and-replace",](https://github.com/MahmoudFawzyKhalil/obsidian-global-search-and-replace/blob/11fb0133bec35823b2737852e7c70aa7f1e4cf4d/src/main.ts#L22)

[this.app = app;](https://github.com/MahmoudFawzyKhalil/obsidian-global-search-and-replace/blob/11fb0133bec35823b2737852e7c70aa7f1e4cf4d/src/domain/file-operator.ts#L22)

MahmoudFawzyKhalil:
Thanks for reviewing my PR.

I created a new release (0.1.1) addressing these comments.

MahmoudFawzyKhalil:
Thank you for merging and for the suggestions. 

I worry with `cachedRead()` if I get a stale result and get the matching range from that result if that would cause me to modify the file in an unintended way?

liamcain:
That's true. The main issue though is that the file could still change between the search and the replace action. So it might make sense to use cachedRead when doing the previewing, and read (as in re-reading the file) when doing the actual replacement.

PR #1716, opened 2023-03-02
New plugin: Home tab

olrenso:

## Repo URL

Link to my plugin: https://github.com/olrenso/obsidian-home-tab

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[enabled: boolean](https://github.com/olrenso/obsidian-home-tab/blob/93e24493f45d2ddffa99621cc39631fb0ec40599/src/main.ts#L19)
This does not exist at this location, it might exist in some plugins, but not in all.
Just check if `getPluginById´ returns null or not.


liamcain:
- [this.app](https://github.com/olrenso/obsidian-home-tab/blob/312f75cdf15d83d577d9fb0d7547e02af0c2e3af/src/ui/starredFile.svelte#L35) this doesn't reference anything so it's falling back to the global app instance. You should access the app instance via the leaf instead. Or pass in app separately.
- [app.workspace](https://github.com/olrenso/obsidian-home-tab/blob/312f75cdf15d83d577d9fb0d7547e02af0c2e3af/src/ui/starredFile.svelte#L23) same here
- [e.ctrlKey](https://github.com/olrenso/obsidian-home-tab/blob/312f75cdf15d83d577d9fb0d7547e02af0c2e3af/src/ui/starredFile.svelte#L43) you should use [`Keymap.isModEvent`](https://github.com/obsidianmd/obsidian-api/blob/009a10f196df551256a5894ea60fdf9ddaab19a1/obsidian.d.ts#L1639) instead. Obsidian doesn't use ctrl-clicking for macOS. Instead it's the command key. isModEvent will handle all the platform specific logic for you and will also help with opening files in a new split or a new window. Just pass the result directly to [`getLeaf`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L3964). So app.workspace.getLeaf(Keymap.isModEvent(evt))

olrenso:
Thank you both for the code review. I updated the plugin to [1.0.1](https://github.com/olrenso/obsidian-home-tab/releases/tag/1.0.1) with the related fixes.

PR #1721, opened 2023-03-05
Add new plugin "Tistory Poster"

idreamer:
# Repo URL
- Link to my plugin: https://github.com/fjdu/obsidian-latex-unicode

# Release Checklist
    - [ ] Windows
    - [x] macOS
    - [ ] Linux
    - [ ] Android (if applicable)
    - [ ] iOS (if applicable)
     - [x] main.js
     - [x] manifest.json
     - [x] styles.css (optional)


joethei:
Please use the correct link to your repo in your pull request message.

["fundingUrl": "",](https://github.com/idreamer/tistory-poster/blob/f97da668f4725a15dcab53896c7c85afc37eb874/manifest.json#L9)
If you don't have one remove this.

["minAppVersion": "0.0.1",](https://github.com/idreamer/tistory-poster/blob/f97da668f4725a15dcab53896c7c85afc37eb874/manifest.json#L5)
Please use a proper version, this plugin uses API's that where not available back then.
Just set it to a recent Obsidian version.

[// Remember to rename these classes and interfaces!](https://github.com/idreamer/tistory-poster/blob/f97da668f4725a15dcab53896c7c85afc37eb874/main.ts#L12)
Remove this.

[await fetch(url, {](https://github.com/idreamer/tistory-poster/blob/f97da668f4725a15dcab53896c7c85afc37eb874/main.ts#L180-L182)
Would recommend using the inbuild `requestUrl´ function from Obsidian.
Also a bit confused, is putting the content inside the query the only optons?
The length of this is limited, so if the API allows would recommend putting it into the body instead.

idreamer:
@joethei Hi Joe, Thanks for your feedback. I updated the codes based on your feedback. 
> Also a bit confused, is putting the content inside the query the only options? The length of this is limited, so if the API allows would recommend putting it into the body instead.
I think that the API specification does not support putting the content into the body.

PR #1724, opened 2023-03-06
add Obsidian ibook plugin

bingryan:

## Repo URL

Link to my plugin: https://github.com/bingryan/obsidian-ibook-plugin

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[id: 'ibook-export-command',](https://github.com/bingryan/obsidian-ibook-plugin/blob/eb37dfd9e97a6cf3846514389fb98ed4d257fd37/src/plugin.ts#L20)

[html2markdown(html: string) {](https://github.com/bingryan/obsidian-ibook-plugin/blob/eb37dfd9e97a6cf3846514389fb98ed4d257fd37/src/export.ts#L82-L84)
Obsidian exposes the [`htmlToMarkdown`](https://github.com/obsidianmd/obsidian-api/blob/bceb489fc25ceba5973119d6e57759d64850f90d/obsidian.d.ts#L1629-L1633) function for this, its using turndown internally also.

bingryan:
@joethei Thanks for your suggestion, I have fixed it 🤓

liamcain:
- [trigger("file-open")](https://github.com/bingryan/obsidian-ibook-plugin/blob/5b2505dbabe7129a8effd2e0aef0dbd8bec87b45/src/plugin.ts#L25) [`file-open`](https://github.com/obsidianmd/obsidian-api/blob/ec589e9762a1d7e2faad01f894cb34c41b10ecaf/obsidian.d.ts#L4189) file-open should also pass in the currently open file. I don't recommend using this approach for refreshing the view since it will have unintended side-effects (many views in the app will react to the file-open event). What are you trying to accomplish with this line exactly?

bingryan:
> 

i want to refresh the left sidebar

bingryan:
> * [trigger("file-open")](https://github.com/bingryan/obsidian-ibook-plugin/blob/5b2505dbabe7129a8effd2e0aef0dbd8bec87b45/src/plugin.ts#L25) [`file-open`](https://github.com/obsidianmd/obsidian-api/blob/ec589e9762a1d7e2faad01f894cb34c41b10ecaf/obsidian.d.ts#L4189) file-open should also pass in the currently open file. I don't recommend using this approach for refreshing the view since it will have unintended side-effects (many views in the app will react to the file-open event). What are you trying to accomplish with this line exactly?

this.app.workspace.trigger("file-explorer:refresh");

a better expression?

liamcain:
The file explorer view should automatically refresh with the newly created file without needing to trigger a refresh automatically. Is this not the behavior you're seeing?

> this.app.workspace.trigger("file-explorer:refresh");

I don't understand. This event doesn't exist.

bingryan:
> The file explorer view should automatically refresh with the newly created file without needing to trigger a refresh automatically. Is this not the behavior you're seeing?

When I was developing, after `export` command action , The ibook directory(the target directory where I generated the document) already existed in the vault , but the left sidebar was not refresh and show this directory. so I added an event to avoid my target directory doesn't show up after command action.

bingryan:
I closed all plugins in the development environment , find: file explorer view automatically refresh 

I have delete that code

liamcain:
Okay, so it sounds like everything is working fine without that code?

bingryan:
> Okay, so it sounds like everything is working fine without that code. Is that right?

yes🤓

PR #1733, opened 2023-03-09
add obsidian-plugin-journal-review to community-plugins.json

Kageetai:

## Repo URL

Link to my plugin: https://github.com/Kageetai/obsidian-plugin-journal-review

## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


joethei:
[list](https://github.com/Kageetai/obsidian-plugin-journal-review/blob/c0f5e05ef12664fd3691169c41bd19b414daebe8/styles.css#L1)
Use a more unique class name, so you don't overwrite some custom styling the user has already created/a different plugin is using.

Kageetai:
> [list](https://github.com/Kageetai/obsidian-plugin-journal-review/blob/c0f5e05ef12664fd3691169c41bd19b414daebe8/styles.css#L1) Use a more unique class name, so you don't overwrite some custom styling the user has already created/a different plugin is using.

@joethei Thanks for the reminder, I wrapped the main container in an ID and used this to wrap all CSS classes

liamcain:
- [new Component()](https://github.com/Kageetai/obsidian-plugin-journal-review/blob/79838a11542d09a9835780688bf8a0127eccf5a6/src/components/NotePreview.tsx#L36) you should pass a reference to your View instance here instead of creating a new component. The reason you pass a component in here is so that the rendered markdown can get properly unloaded when the passed in component calls "unload." By creating a new component and passing it in, you're not actually doing anything other than making typescript happy.

Kageetai:
> * [new Component()](https://github.com/Kageetai/obsidian-plugin-journal-review/blob/79838a11542d09a9835780688bf8a0127eccf5a6/src/components/NotePreview.tsx#L36) you should pass a reference to your View instance here instead of creating a new component. The reason you pass a component in here is so that the rendered markdown can get properly unloaded when the passed in component calls "unload." By creating a new component and passing it in, you're not actually doing anything other than making typescript happy.

I figured back then already that that is probably not the right way, but I wasn't sure what `Component` refers to as the docs are quite vague on that part.
I thought I also read somewhere that you are not supposed to keep references to your views, but let Obsidian handle that when using `this.registerView`, is that correct? And if so, what's the best way to pass a reference to my view into my Preact component? Do you have any suggestions.

> * [onClick](https://github.com/Kageetai/obsidian-plugin-journal-review/blob/79838a11542d09a9835780688bf8a0127eccf5a6/src/components/NotePreview.tsx#L52) I suggest changing this to `.getLeaf(Keymap.isModEvent(evt)).openFile(note)`. This will allow users to command-click to open in new tab or command-opt-click to open in a new split.

Good tip, will do that.

PR #1736, opened 2023-03-10
feat(prioritize): adding prioritize plugin

EloiMusk:

## Repo URL

Link to my plugin: https://github.com/EloiMusk/obsidian-prio-plugin

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[hotkeys: [](https://github.com/EloiMusk/obsidian-prio-plugin/blob/b309c68473033dd01c199bfe6e48de399ded17d2/main.ts#L46-L52)

liamcain:
Your plugin README doesn't really explain what a "Priority level" is. I was only able to figure it out by looking at the code. I assumed it would be something that lives in the note's frontmatter.

- [levelAlias](https://github.com/EloiMusk/obsidian-prio-plugin/blob/625f6f194b878a559b31f690b38b92211f96aaed/utils/Priority.ts#L51) you should add a check here to ensure that levelAlias exists.
- [levelAlias](https://github.com/EloiMusk/obsidian-prio-plugin/blob/625f6f194b878a559b31f690b38b92211f96aaed/utils/Priority.ts#L12) Same here.

PR #1738, opened 2023-03-10
Add theme validation

joethei:
Adding a validation script for new theme submissions.

Also adding a plugin label to the existing plugin validation.


.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,216 @@
...
+
+                try {
+                    const screenshot = await github.rest.repos.getContent({
+                        owner, repo, path: theme.screenshot
+                    });
+                    const imageMeta = await probe(screenshot.data.download_url);
+                    if (imageMeta.type !== "png" && imageMeta.type !== "jpg") {
+                        addError('Theme screenshot is not of filetype `.png` or `.jpg`');
+                    }
+                    const recommendedSize = "we generally recommend a size around 512 by 288px";
```
liamcain, line 141: Nit: looks like you switch between single and double quotes a lot in this file.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,216 @@
...
+                if (/^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$/.test(manifest.author)) {
+                    addWarning('We generally discourage from including email addresses in the `author` field.');
+                }
+
+                try {
+                    const screenshot = await github.rest.repos.getContent({
+                        owner, repo, path: theme.screenshot
+                    });
+                    const imageMeta = await probe(screenshot.data.download_url);
+                    if (imageMeta.type !== "png" && imageMeta.type !== "jpg") {
```
liamcain, line 138: Is it possible for type to return "jpeg" at all?

.github/workflows/validate-theme-entry.yml
liamcain, line 141:
```suggestion
                    const recommendedSize = "we generally recommend a size around 512 × 288 pixels";
```

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,216 @@
...
+                }
+
+                try {
+                    await github.rest.repos.getContent({
+                        owner, repo, path: 'obsidian.css'
+                    });
+                    addWarning('Your repository includes a `obsidian.css` file, this is only used in legacy versions of Obsidian.');
+                } catch (e) { }
+
+                if (!(/^[0-9.]+$/i.test(manifest.version))) {
```
liamcain, line 125: Manifest might be an empty object here.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,264 @@
...
+            await (async () => {
+                if (context.payload.pull_request.changed_files > 1) {
+                    addError('You modified files other than `community-css-themes.json`.');
+                }
+
+                const author = context.payload.pull_request.user.login;
+                const themes = JSON.parse(fs.readFileSync('community-css-themes.json', 'utf8'));
+                const theme = themes[themes.length - 1];
+
+                let prIssue = false;
```
liamcain, line 52: Can we call this `hasPullRequestError` instead? I saw "issue" at first and assumed it was related to Github Issues.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,264 @@
...
+                        path: 'manifest.json',
+                    });
+
+                    manifest = JSON.parse(Buffer.from(manifestFile.data.content, 'base64').toString('utf-8'));
+                } catch (e) {
+                    addError(`You don't have a valid \`manifest.json\` at the root of your repo.`);
+                    return;
+                }
+
+                let manifestIssue = false;
```
liamcain, line 117: `hasManifestError`?

liamcain (approved):
LGTM, just the naming nitpicks

ericaxu:
Is it ready to merge? Anything we should let everyone know?

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+      - uses: actions/setup-node@v2
+      - run: npm install probe-image-size
+      - uses: actions/github-script@v6
+        with:
+          script: |
+            const probe = require('probe-image-size');
+            const fs = require('fs');
+
+            if (context.payload.pull_request.additions <= context.payload.pull_request.deletions) {
+                // Don't run any validation checks if the user is just modifying existing theme config
```
lishid, line 26: Move this up

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+                try {
+                    const themes = JSON.parse(fs.readFileSync('community-css-themes.json', 'utf8'));
+                    theme = themes[themes.length - 1];
+
+                } catch (e) {
+                    addError('Could not parse `community-css-themes.json`, invalid JSON. ' + e.message);
+                    return;
+                }
+
+                const author = context.payload.pull_request.user.login;
```
lishid, line 59: This is not needed until much further, move it down

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+                
+                let validPrKeys = ['name', 'author', 'repo', 'screenshot', 'modes'];
+                for (let key in Object.keys(manifest)) {
```
lishid, line 62: Reverse the order of these two for loops

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+            await (async () => {
+                if (context.payload.pull_request.changed_files > 1) {
+                    addError('You modified files other than `community-css-themes.json`.');
+                }
+
+                let theme;
+
+                try {
+                    const themes = JSON.parse(fs.readFileSync('community-css-themes.json', 'utf8'));
+                    theme = themes[themes.length - 1];
```
lishid, line 52: Move this down out of the try-catch.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+                    if (!manifest.hasOwnProperty(key)) {
+                        addError(`Your PR does not have the required \`${key}\` property.`);
+                    }
+                }
+
+                if (theme.name.toLowerCase().includes('obsidian')) {
+                    addWarning(`We discourage themes from including the word "Obsidian" in their name since it's redundant and makes the theme selection screen harder to visually parse.`);
+                }
+
+                if (themes.filter(t => t.name === theme.name).length > 1) {
```
lishid, line 77: Themes isn't set to a variable in this scope.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+                    addError('There is already a theme with this name');
+                }
+
+                if (themes.filter(t => t.repo === theme.repo).length > 1) {
+                    addError('There is already a entry pointing to the `' + theme.repo + '` repository');
+                }
+
+                // we can't validate further, as there is no repo set
+                if(!theme.hasOwnProperty('repo')) {
```
lishid, line 86: Shouldn't this check be moved a bit more earlier? We're already using `theme.repo` in the previous section.

lishid: Actually no this is not needed since we already do a check for all the required keys.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+                    return;
+                }
+
+                // Validate theme repo
```
lishid, line 90: Move this earlier just like we do for plugins.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+                        path: 'manifest.json',
+                    });
+
+                    manifest = JSON.parse(Buffer.from(manifestFile.data.content, 'base64').toString('utf-8'));
+                } catch (e) {
+                    addError(`You don't have a valid \`manifest.json\` at the root of your repo.`);
+                    return;
+                }
+
+                let validManifestKeys = ['name', 'minAppVersion', 'author', 'authorUrl'];
```
lishid, line 135: I don't think we should check for invalid manifest keys - it should be ok to add stuff to the manifest.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+                for (let key in manifest) {
+                    if (!validKeys.includes(key) && !key === 'version') {
+                        addError(`Your manifest has the invalid \`${key}\` property.`);
+                    }
+                }
+
+                for (let key in validManifestKeys) {
```
lishid, line 142: Again, don't think we need this.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+                    if (!manifest.hasOwnProperty(key)) {
+                        addError(`Your manifest does not have the required \`${key}\` property.`);
+                    }
+                }
+
+                if (manifest.hasOwnProperty('name') && manifest.name != theme.name) {
+                    addError(`Plugin name mismatch, the name in this PR (\`${theme.name}\`) is not the same as the one in your repo (\`${manifest.name}\`). If you just changed your plugin name, remember to change it in the manifest.json in your repo and your latest GitHub release.`);
+                }
+
+                if (manifest.hasOwnProperty('version') && !(/^[0-9.]+$/i.test(manifest.version))) {
```
lishid, line 152: This is a required field so we should remove the optional `hasOwnProperty` check.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+                    addError('Your latest version number is not valid. Only numbers and dots are allowed.');
+                }
+
+                if (manifest.hasOwnProperty('author') && /^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$/.test(manifest.author)) {
```
lishid, line 156: The author check should be moved earlier and should run on the submitted object rather than the manifest.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+                    if (imageMeta.type !== 'png' && imageMeta.type !== 'jpg') {
+                        addError('Theme screenshot is not of filetype `.png` or `.jpg`');
+                    }
+                    const recommendedSize = 'we generally recommend a size around 512 × 288 pixels';                    
+                    if (imageMeta.width > 1000 || imageMeta.height > 500) {
+                        addError(`Your theme screenshot is too big, ${recommendedSize}`);
+                    }
+                    if (imageMeta.width < 250 || imageMeta.height < 100) {
+                        addError(`Your theme screenshot is too small, ${recommendedSize}`);
+                    }
```
lishid, line 190: I would add a warning, if neither is the case, to tell them what the recommended size is. Also these statements should be `else if` rather than two `if`s since they can collide if they submit a really wide or long screenshot - then you'd say it's too big and too small.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+                        });
+
+                        const assets = release.data.assets || [];
+                        if (!assets.find(p => p.name === 'theme.css')) {
+                            addError('Your latest Release is missing the `theme.css` file.');
+                        }
+                        if (!assets.find(p => p.name === 'manifest.json')) {
+                            addError('Your latest Release is missing the `manifest.json` file.');
+                        }
+                    } catch (e) { }
```
lishid, line 211: Shouldn't this give an error?

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+
+                    try {
+                        await github.rest.licenses.getForRepo({ owner, repo });
```
lishid, line 214: This should be outside.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,263 @@
...
+                message.push(`<sup>This check was done automatically. Do <b>NOT</b> open a new PR for re-validation. Instead, to trigger this check again, make a change to your PR and wait a few minutes, or close and re-open it.</sup>`);
+
+
+                await github.rest.issues.createComment({
+                    issue_number: context.issue.number,
+                    owner: context.repo.owner,
+                    repo: context.repo.repo,
+                    body: message
+                });
+                core.setFailed("Failed to validate theme");
```
lishid, line 245: This should only happen if errors.length > 0

lishid (changes requested):
Ok so a lot of small things to tweak. I think this should be as similar to the plugin validation script as possible, including the order of operations. This helps us make sure we are consistent across the two. I've marked any place that diverge either in order of operation so it's just a matter of moving code around.

.github/workflows/validate-theme-entry.yml
lishid: Ok.

.github/workflows/validate-theme-entry.yml
joethei: This check will stop the Action to fail without a message.
The checks after this might fail accessing `undefined` fields, the only message will be in the action logs.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,269 @@
...
+
+                if (!(/^[0-9.]+$/i.test(manifest.version))) {
+                    addError('Your latest version number is not valid. Only numbers and dots are allowed.');
+                }
+
+                if (/^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$/.test(theme.author)) {
+                    addWarning('We generally discourage from including email addresses in the `author` field.');
+                }
+
+                if (theme.name.toLowerCase().includes('obsidian')) {
```
lishid, line 149: This should go further up to match up where we have it in the plugin validation script.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,269 @@
...
+
+                if (manifest.fundingUrl && manifest.fundingUrl === "https://obsidian.md/pricing") {
+                  addError(`The \`fundingUrl\` field in your manifest should not point to the Obsidian Website, If you don't have a link were users can donate to you, you can just omit this.`);
+                }
+
+                if (!(/^[0-9.]+$/i.test(manifest.version))) {
+                    addError('Your latest version number is not valid. Only numbers and dots are allowed.');
+                }
+
+                if (/^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$/.test(theme.author)) {
```
lishid, line 145: This should go further up to match up where we have it in the plugin validation script.

.github/workflows/validate-theme-entry.yml
lishid: In that case we should rename the var to `requiredManifestKeys`.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,269 @@
...
+                    addWarning('We generally discourage from including email addresses in the `author` field.');
+                }
+
+                if (theme.name.toLowerCase().includes('obsidian')) {
+                    addWarning(`We discourage themes from including the word "Obsidian" in their name since it's redundant and makes the theme selection screen harder to visually parse.`);
+                }
+
+                if (themes.filter(t => t.name === theme.name).length > 1) {
```
lishid, line 153: This should go further up to match up where we have it in the plugin validation script.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,269 @@
...
+                    addError('There is already a theme with this name');
+                }
+
+                if (themes.filter(t => t.repo === theme.repo).length > 1) {
```
lishid, line 157: This should go further up to match up where we have it in the plugin validation script.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,269 @@
...
+                    let manifestFile = await github.rest.repos.getContent({
+                        owner,
+                        repo,
+                        path: 'manifest.json',
+                    });
+
+                    manifest = JSON.parse(Buffer.from(manifestFile.data.content, 'base64').toString('utf-8'));
+                } catch (e) {
+                    addError(`You don't have a valid \`manifest.json\` at the root of your repo.`);
+                    return;
```
lishid, line 113: This should not just return - we should deal with missing manifest with an if clause just like plugin reviews.
Doing it this way allows us to continue checking for more things like having a theme.css, so we can report as many problems in one go as possible.

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,269 @@
...
+                        addError('Theme screenshot is not of filetype `.png` or `.jpg`');
+                    }
+                    const recommendedSize = 'we generally recommend a size around 512 × 288 pixels';                    
+                    if (imageMeta.width > 1000 || imageMeta.height > 500) {
+                        addError(`Your theme screenshot is too big, ${recommendedSize}`);
+                    }
+                    else if (imageMeta.width < 250 || imageMeta.height < 100) {
+                        addError(`Your theme screenshot is too small, ${recommendedSize}`);
+                    } else if (imageMeta.width !==512 || image.height !== 288) {
+                        addWarning(`Theme theme screenshot size is not optimal, ${recommendedSize}`);
```
lishid, line 192: Perhaps also add "yours is detected as WxH"

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,269 @@
...
+                    }
+                } catch (e) {
+                    addError('The theme screenshot cannot be found.');
+                }
+
+                // only validate releases if version is included
+                if (manifest.hasOwnProperty('version')) {
```
lishid, line 199: Isn't `version` a required key?

.github/workflows/validate-theme-entry.yml
```diff
@@ -0,0 +1,269 @@
...
+                message.push(`<sup>This check was done automatically. Do <b>NOT</b> open a new PR for re-validation. Instead, to trigger this check again, make a change to your PR and wait a few minutes, or close and re-open it.</sup>`);
+
+
+                await github.rest.issues.createComment({
+                    issue_number: context.issue.number,
+                    owner: context.repo.owner,
+                    repo: context.repo.repo,
+                    body: message
+                });
+                if (warnings.length > 0) {
```
lishid, line 249: This should be errors.length not warnings.length, and it should be outside to match up with where it is in the plugin script.

lishid (changes requested):

.github/workflows/validate-theme-entry.yml
joethei: According to the docs yes, but not according to the code.

lishid (approved):

PR #1741, opened 2023-03-11
Add Source Code Note Plugin

waiting0324:
Add obsidian-code-note plugin

## Repo URL

Link to my plugin: https://github.com/waiting0324/obsidian-code-note

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
Why is the language hard-coded here?

waiting0324:
Oh, I was using Java as a functional test before, I fix this problem now.

liamcain:
- [mmPreview = new SourceCodeView(this.codeBlockContents, preview);](https://github.com/waiting0324/obsidian-code-note/blob/aff1ac6efc51b230ebc24cabd10ae5039010e433/src/main.ts#L61) Don't create instances of your view directly. Instead you should use setViewState({ type: TYPE_OF_YOUR_VIEW }) and let obsidian create the view instance for you. This will ensure it follows the same code path that occurs on startup (since Obsidian will serialize and deserialize the workspace views for you based on the view type.

PR #1755, opened 2023-03-13
Add AI Commander plugin

yzh503:

## Repo URL

Link to my plugin: https://github.com/yzh503/obsidian-aicommander-plugin

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
I generally recommend moving all code files into a separate `src` folder, that makes it easier to review and maintain.

[const YOUR_GENERATED_SECRET = '9VFMTHCukRuT5WqOkAD1:8cde275ebde49165527e9c97ecc96abef1e34473458fe8f9f3ecad177a163538';](https://github.com/yzh503/obsidian-aicommander-plugin/blob/aa6d48b68cbd192c073f3594eabd2aaa06fc5db6/main.ts#L38-L55)
What REST API is that?
It looks like it's a seperate one, so please mention this in the README.
Also, don't hardcode the API key, but make it configurable.

[await axios.post](https://github.com/yzh503/obsidian-aicommander-plugin/blob/aa6d48b68cbd192c073f3594eabd2aaa06fc5db6/main.ts#L52)
Would recommend using the `requestUrl` function from the Obsidian API instead, as that does a few Obsidian specific things.


[async getImageBase64(url: string) {](https://github.com/yzh503/obsidian-aicommander-plugin/blob/aa6d48b68cbd192c073f3594eabd2aaa06fc5db6/main.ts#L111-L125)
Can be replace by:
```ts
const buffer = await requestUrl(url).arrayBuffer;
return arrayBufferToBase64(buffer);
```

[await this.app.vault.adapter.exists(dirPath).then((exists) => {](https://github.com/yzh503/obsidian-aicommander-plugin/blob/aa6d48b68cbd192c073f3594eabd2aaa06fc5db6/main.ts#L183-L186)
It's better to use the Vault API instead of the Adapter API.
This can be replaced with:
```ts
const exists = this.app.vault.getAbstractFileByPath(dirPath) instanceof TFolder;
if(!exits) {
    await this.app.vault.createFolder(dirPath);
}
```

[await this.app.vault.adapter.writeBinary(filepath, buffer);](https://github.com/yzh503/obsidian-aicommander-plugin/blob/aa6d48b68cbd192c073f3594eabd2aaa06fc5db6/main.ts#L189)
Again, it's better to use the Vault API.

[console.log(this.settings.saveImg);](https://github.com/yzh503/obsidian-aicommander-plugin/blob/aa6d48b68cbd192c073f3594eabd2aaa06fc5db6/main.ts#L191)
Remove this.

[return response.data.webPages.value;](https://github.com/yzh503/obsidian-aicommander-plugin/blob/aa6d48b68cbd192c073f3594eabd2aaa06fc5db6/main.ts#L235)
I don't see any validation for this.

[const attachmentFolder = await this.app.vault.adapter.read(`${this.app.vault.configDir}/app.json`).then((content: string) => {](https://github.com/yzh503/obsidian-aicommander-plugin/blob/aa6d48b68cbd192c073f3594eabd2aaa06fc5db6/main.ts#L239)
Don't read from the config file.
Better use the undocumented `this.app.vault.getAvaiablePathForAttachments(base: string, extension: string, currentFile: TFile | null): Promise<string>`
That will return the path your file should be stored at.

[console.log('currentPathString', currentPathString);](https://github.com/yzh503/obsidian-aicommander-plugin/blob/aa6d48b68cbd192c073f3594eabd2aaa06fc5db6/main.ts#L283-L285)
Remove this.


[// If the plugin hooks up any global DOM events (on parts of the app that doesn't belong to this plugin)](https://github.com/yzh503/obsidian-aicommander-plugin/blob/aa6d48b68cbd192c073f3594eabd2aaa06fc5db6/main.ts#L565-L572)
Remove this.

yzh503:
@joethei thank you for the review. I've updated the release. All your suggestions have been addressed.

liamcain:
- [registerInterval](https://github.com/yzh503/obsidian-aicommander-plugin/blob/174cb2fc4755217ce30957a36b75ce017fdb5edd/src/main.ts#L622) remove this
- [toLowerCase().replace](https://github.com/yzh503/obsidian-aicommander-plugin/blob/174cb2fc4755217ce30957a36b75ce017fdb5edd/src/main.ts#L591) command IDs don't actually need to be kebab case, you don't need to do this transformation

PR #1754, opened 2023-03-13
Update community-plugins.json

shmolf:
# I am submitting a new Community Plugin

## Repo URL

Link to my plugin:

## Release Checklist
- [x] I have tested the plugin on
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  Android _(if applicable)_
  - [ ]  iOS _(if applicable)_
- [x] My GitHub release contains all required files
  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css` _(optional)_
- [x] GitHub release name matches the exact version number specified in my manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
- [x] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [x] My README.md describes the plugin's purpose and provides clear usage instructions.
- [x] I have read the tips in https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md and have self-reviewed my plugin to avoid these common pitfalls.
- [x] I have added a license in the LICENSE file.
- [x] My project respects and is compatible with the original license of any code from other plugins that I'm using.
      I have given proper attribution to these other projects in my `README.md`.

PR #1764, opened 2023-03-15
Add wikipedia-search plugin

StrangeGirlMurph:

## Repo URL

Link to my plugin: https://github.com/StrangeGirlMurph/obsidian-wikipedia-search

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


StrangeGirlMurph:
Just read https://github.com/obsidianmd/obsidian-releases/pull/1754#issuecomment-1510438026 ^^
I guess I'll stop now with always fixing the merge conflict. I was just scared :)

liamcain:
- [getActiveViewOfType](https://github.com/StrangeGirlMurph/obsidian-wikipedia-search/blob/a77191ceb0653dc85443457ecc8548b5030c0314/src/main.ts#L17) this extra check is unnecessary. If you have access to an editor, then you already know that your active file is either a MarkdownView or a CanvasView (either of which should be fine for your use case). I would recommend changing this to be just an editorCallback and then you don't need to fetch the view

StrangeGirlMurph:
@liamcain okay, thank you! Will fix it ASAP

PR #1766, opened 2023-03-16
Request to add Soomda plugin

michaellee:

## Repo URL

Link to my plugin: https://github.com/michaellee/soomda

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


michaellee:
> #### Hello!
> **I found the following issues in your plugin submission**
> 
> **Warnings:**
> 
> ⚠️ Your repository does not include a license. It is generally recommended for open-source projects to have a license. Go to https://choosealicense.com/ to compare different open source licenses.
> 
> This check was done automatically. Do NOT open a new PR for re-validation. Instead, to trigger this check again, make a change to your PR and wait a few minutes, or close and re-open it.

🤦🏻‍♂️ I thought I had added a LICENSE when creating this PR. I've now added one.

joethei:
What sets your plugin appart from the other plugins that allow hiding the sidebar/panes?

Why is this marked as desktop only?

[addIcon('icon', icon);](https://github.com/michaellee/soomda/blob/fbfb065299c19a94e6b22f44a6c73bba2f47bb52/main.ts#L27)
Would recommend naming the icon `soomda`, `icon` is too generic.

[id: 'soomda-show-sidebars',](https://github.com/michaellee/soomda/blob/fbfb065299c19a94e6b22f44a6c73bba2f47bb52/main.ts#L39)


---

michaellee:
Hey @joethei Johannes, thanks for taking your time to review my plugin!

> What sets your plugin apart from the other plugins that allow hiding the sidebar/panes?

I didn't realize there were other plugins that did similar things. I did try searching for plugins prior to building this and didn't see anything that stuck out as doing this.

I basically wanted something that toggled the sidebars both at the same time.

One thing I added was an option to remember which pane was last open. In case people don't want for both sidebars to be toggled, but instead be "aware" of last preferred sidebar.

> Why is this marked as desktop only?

Wasn't aware that there were sidebars in mobile? I'm guessing there might be in tablets? Can definitely add support later, I just don't have a tablet to test on at the moment.

Hence why I opted for desktop only.

> Would recommend naming the icon soomda, icon is too generic.

Thanks for calling this out, I've updated this.

> Obsidian will automatically prefix the command id with the plugin id, you don't need to include it yourself.

Noted, I've updated.

> As an aside not using semicolons in JavaScript/TypeScript code can lead to rare

Great article! Not sure if it's too convincing, but a good reminder to run code through formatter. I've run it through Prettier now so that my `"` and `'` are consistent and semi colons are now there :)

Again, thanks, for taking the time to help. Let me know if there's anything else.

liamcain:
> > Would recommend naming the icon soomda, icon is too generic.
> Thanks for calling this out, I've updated this.

joethei's comment here was that the icon name is too generic. It's a bit confusing since the other things are automatically prefixed, but icons are actually all added to the same collection and aren't namespaced. So it would be better to name this something more specific to avoid it overlapping with other plugins.

PR #1767, opened 2023-03-16
New plugin: ring a secretary

vrtmrz:

## Repo URL

Link to my plugin: https://github.com/vrtmrz/ring-a-secretary/

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


vrtmrz:
Thank you for reading this request, let me explain before you ask me.

This plugin is indeed rather similar to other plugins. However, the following features are trivial but useful.
- We can insert conversations in the middle of a note as blocks (Actually, stored as a code block). 
- All inserted conversations are searchable as same as a note, and with the note.
- This plug-in has an explicit chat-oriented UI.

I am looking forward to hearing your suggestions. Thank you.

liamcain:
- [axios](https://github.com/vrtmrz/ring-a-secretary/blob/4fcf0bea63b14cc72ab1aa00fa151d75be329dbf/src/main.ts#L3) is axios even being used?

PR #1771, opened 2023-03-16
New plugin to reveal file clicking title

1C0D:

## Repo URL

Link to my plugin:

## Release Checklist
  - [x ]  Windows
  - [ x]  macOS
  - [x ]  Linux
  - [ x]  Android _(if applicable)_
  - [x ]  iOS _(if applicable)_
  - [x ] `main.js`
  - [x ] `manifest.json`
- [x ] I have added a license in the LICENSE file.


claremacrae:
Just to copy in my comment on the original #1769 

A quick note re credits in the README...

I think this plugin may have arisen out of the following declined pull requests to a similar plugin:

https://github.com/claremacrae/reveal-active-file-button-plugin/pull/6
https://github.com/claremacrae/reveal-active-file-button-plugin/pull/7

Which in turn credits the original author of that code:
https://github.com/OfficerHalf/obsidian-collapse-all

Looking at the code here, I am totally comfortable with neither of those plugins being credited in the README here, as the code here is totally different.

liamcain:
- [active-leaf-change](https://github.com/1C0D/Obsidian-Reveal-File-in-explorer/blob/2eccc8394f02acebc4f6f5516963de6a880a08f9/src/main.ts#L8) There's no need to register add the active-leaf-change event. You can just register it once to the main window on load. When pop-out windows are created, the window click events are automatically copied over, so this will be handled automatically. So it should be safe to remove lines 7 through 10.

1C0D:
done

PR #1780, opened 2023-03-19
new plugin: Easy toggle sidebars

1C0D:
# I am submitting a new Community Plugin(PC only)

## Repo URL

Link to my plugin:

## Release Checklist
  - [x ]  Windows
  - [x ]  macOS
  - [x ]  Linux
  - [x ]  Android _(if applicable)_
  - [ x]  iOS _(if applicable)_
  - [x ] `main.js`
  - [ x] `manifest.json`
- [ x] I have added a license in the LICENSE file.


liamcain:
- [this.leftSplit](https://github.com/1C0D/obsidian-easy-toggle-sidebars/blob/24376fdc34f1c8457c821b0999e47d2eb1c8bd0c/src/main.ts#L68) don't assign these to local variables. That will keep instances of them stored in memory. If the user switches workspaces using the Workspaces core plugin, then the splits are entirely removed and recreated. Your plugin will keep the stale versions in memory instead of referencing the new splits. If the goal is just to simplify accessing the variables, you can create helper functions on your plugin for quick access. That won't store a reference to the elements but will just fetch them as needed.
- [doubleClickTimer](https://github.com/1C0D/obsidian-easy-toggle-sidebars/blob/24376fdc34f1c8457c821b0999e47d2eb1c8bd0c/src/main.ts#L12) unused
- [isDoubleClickTimerStarted](https://github.com/1C0D/obsidian-easy-toggle-sidebars/blob/24376fdc34f1c8457c821b0999e47d2eb1c8bd0c/src/main.ts#L13) unused.

1C0D:
@liamcain 
I've made the changes you suggested and the plugin is now working as intended. 
I've removed the local variables for leftSplit and rightSplit, and instead created helper functions to access them as needed. I also removed junk code.
Thank you again for your help.

PR #1782, opened 2023-03-20
Register obsidian-smart-rename

mnaoumov:

## Repo URL

Link to my plugin: https://github.com/mnaoumov/obsidian-smart-rename

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


joethei:
[id: 'smart-rename-command',](https://github.com/mnaoumov/obsidian-smart-rename/blob/6f638a5ca89394261a7412556112e7cf8059a19b/SmartRenamePlugin.ts#L15)

[this.app.workspace.getActiveFile() as TFile;](https://github.com/mnaoumov/obsidian-smart-rename/blob/6f638a5ca89394261a7412556112e7cf8059a19b/SmartRenamePlugin.ts#L30)
This could be null.

[this.app.metadataCache.offref(eventRef);](https://github.com/mnaoumov/obsidian-smart-rename/blob/6f638a5ca89394261a7412556112e7cf8059a19b/SmartRenamePlugin.ts#L144)
Instead of this wrap the `this.app.metadataCache.on(...` in `this.registerEvent()`

mnaoumov:
@joethei Thanks for your suggestions.

> [id: 'smart-rename-command',](https://github.com/mnaoumov/obsidian-smart-rename/blob/6f638a5ca89394261a7412556112e7cf8059a19b/SmartRenamePlugin.ts#L15) Obsidian will automatically prefix the command id with the plugin id, you don't need to include it yourself.

Well, that's the name of the command, which happens to match the plugin name, because I don't have a better name for it. I will just remove unnecessary `command` suffix.

> [this.app.workspace.getActiveFile() as TFile;](https://github.com/mnaoumov/obsidian-smart-rename/blob/6f638a5ca89394261a7412556112e7cf8059a19b/SmartRenamePlugin.ts#L30) This could be null.
> 

Technically it is impossible due to the check in [Line 21](https://github.com/mnaoumov/obsidian-smart-rename/blob/6f638a5ca89394261a7412556112e7cf8059a19b/SmartRenamePlugin.ts#L21) . I will write an additional check just for the code purity.

> [this.app.metadataCache.offref(eventRef);](https://github.com/mnaoumov/obsidian-smart-rename/blob/6f638a5ca89394261a7412556112e7cf8059a19b/SmartRenamePlugin.ts#L144) Instead of this wrap the `this.app.metadataCache.on(...` in `this.registerEvent()`

Originally I considered that to be a one-time event, subscribed and unsubscribed instantly used only after the individual rename. But with your comment I realized I could make it a long-running event just ensuring it is a no-op after backlinks fixed

liamcain:
- [this.app.workspace.getActiveFile](https://github.com/mnaoumov/obsidian-smart-rename/blob/28f9601da6fcbe01f69ba615de90a897b06ce58c/SmartRenamePlugin.ts#L22) this check should happen _before_ the actual function call you have for !checking. It doesn't make a huge difference since you make the call to getActiveFile again inside smartRename() but you could opt to just pass the TFile instead, that way you don't need to fetch it twice, and you can shortcircuit early if there is no active file.

PR #1789, opened 2023-03-21
Add Page Properties plugin

necauqua:

## Repo URL

Link to my plugin: https://github.com/necauqua/obsidian-page-properties

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["author": "Anton Bulakh <self@necauqua.dev>",](https://github.com/necauqua/obsidian-page-properties/blob/f1633580c3c292175a318e16ef93397d25ee5dba/manifest.json#L7)
We generally discourage including your email addresses in the `author` field.

necauqua:
The email thing I think I saw in Dataview manifest, I've removed that.

For the semicolons thing - I don't think JS and TS are equal in that case, for the most examples of issues outlined TS will just fail to compile, the whole point of it.
And also [`no-unexpected-multiline`](https://eslint.org/docs/latest/rules/no-unexpected-multiline) lint is part of the `eslint:recommended` anyway 🤷

liamcain:
I suggest adding a bit more information in the README about what this plugin actually does. Also the settings could use some more explanatory text, such as "Page prop names are inner links" which offers no explanation.

PR #1790, opened 2023-03-22
Update community-plugins.json

gaetanmuck:

## Repo URL

Link to my plugin: [gaetanmuck/obsidian-flashcard-learning](https://github.com/gaetanmuck/obsidian-flashcard-learning)

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[Flascard](https://github.com/gaetanmuck/obsidian-flashcard-learning/blob/10d60b2cb769f39d6bbf3fb5709bef0e6aa26720/src/main.ts#L28)
Typo, also we recommend using sentence case for command names.

gaetanmuck:
Typo and case corrected

liamcain:
- [replaceSelection](https://github.com/gaetanmuck/obsidian-flashcard-learning/blob/68fdcaefa314f0c01a8ec14e661bb2aa447e6151/src/main.ts#L31) calling replaceSelection multiple times risk overriding the previous selection. I suspect this worked while testing since your selection was empty, but if you select text and then this code executes, then only the last text will appear. Instead, you should concatenate the text into one string first and call replaceSelection once.

gaetanmuck:
`replaceSelection` is now only called once: [see code here](https://github.com/gaetanmuck/obsidian-flashcard-learning/blob/cae316e7191f966d2e16b6f697747b008ccf6c21/src/main.ts#L31)

PR #1800, opened 2023-03-23
Added folder notes plugin

LostPaul:

## Repo URL

Link to my plugin: https://github.com/LostPaul/obsidian-folder-notes

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[.nav-folder-title-content {](https://github.com/LostPaul/obsidian-folder-notes/blob/ecad3ddc31fac806118217f2b6e1d586515a6fad/styles.css#L1-L5)
[.nav-folder-title {](https://github.com/LostPaul/obsidian-folder-notes/blob/ecad3ddc31fac806118217f2b6e1d586515a6fad/styles.css#L14-L17)
Don't overwrite core styling with a plugin.

LostPaul:
> [.nav-folder-title-content {](https://github.com/LostPaul/obsidian-folder-notes/blob/ecad3ddc31fac806118217f2b6e1d586515a6fad/styles.css#L1-L5) [.nav-folder-title {](https://github.com/LostPaul/obsidian-folder-notes/blob/ecad3ddc31fac806118217f2b6e1d586515a6fad/styles.css#L14-L17) Don't overwrite core styling with a plugin.

That's needed in order to avoid collapsing the folder when clicking on whitespace or do you have a better solution?

LostPaul:
@joethei Any updates on this?

LostPaul:
Any updates on this? @joethei

liamcain:
- [openFolderNote](https://github.com/LostPaul/obsidian-folder-notes/blob/08c1218a40a3df9982c7890ad938bb0eda876900/src/main.ts#L172) you should pass the MouseEvent into the openFolderNote function then instead of `getLeaf(false)`, do: `getLeaf(Keymap.isModEvent(evt))`. That will let you support opening the file in a new tab/new window without any extra work
- [file?.path](https://github.com/LostPaul/obsidian-folder-notes/blob/08c1218a40a3df9982c7890ad938bb0eda876900/src/commands.ts#L42) shouldn't need the ? here. Looks confusing since it means file is optional here.
- [file?.path](https://github.com/LostPaul/obsidian-folder-notes/blob/08c1218a40a3df9982c7890ad938bb0eda876900/src/commands.ts#L42) don't cast to TFile, instead just add an instanceof check for better type safety. (It's not impossible for a folder to be called `somefolder.md/`

LostPaul:
> * don't cast to TFile, instead just add an instanceof check for better type safety. (It's not impossible for a folder to be called `somefolder.md/`

uh when I don't do that typescript throws an error one line after that I check if it is a TFile or what do you want me to do?

liamcain:
Yes, but since `file` will never be a TFile, I don't see the point in including that

liamcain:
image
since you're reassigning to `file`, I realize what I said might be slightly confusing. I'm talking about this part. Here you're saying `file = the folder note or if the folder doesn't exist, assign file to be the folder`

LostPaul:
> image since you're reassigning to `file`, I realize what I said might be slightly confusing. I'm talking about this part. Here you're saying `file = the folder note or if the folder doesn't exist, assign file to be the folder`

Yes because the file could be also null ts throws an error I could also just let it ignore but ?? file has the same effect as I'm using instanceof to check if it actually is a TFile
image
image

PR #1805, opened 2023-03-24
feat: add obsidian-code-files

lukasbach:

## Repo URL

Link to my plugin: https://github.com/lukasbach/obsidian-code-files

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[id: 'code-files-create',](https://github.com/lukasbach/obsidian-code-files/blob/29d3c8e1982261c9c7b1408b34be3528b5eb71f2/src/codeFilesPlugin.ts#L36)

[this.iframe.src = `https://embeddable-monaco.lukasbach.com?${queryParameters.toString()}`;](https://github.com/lukasbach/obsidian-code-files/blob/29d3c8e1982261c9c7b1408b34be3528b5eb71f2/src/codeEditorView.ts#L55)
This plugin should not require an internet connection to function.

lukasbach:
> This plugin should not require an internet connection to function.

Yeah I figured that, but unfortunately I really didn't get Monaco to work directly with Obsidian. Monaco has a very strict setup with web workers, which to my knowledge requires multiple files that are available under customizable URLs and which the website hosting the editor needs to load during runtime, and I really didn't find a way to get that to work with Obsidians restriction that the entire plugin code needs to be one single file. 

If you have alternative suggestions I would be happy to give them a try, but otherwise, I would love if we could add the plugin with this online-dependency as a caveat. I can also add this info to the top of the readme to make this caveat more visible to people interested in adding the plugin.

joethei:
You can run a web worker inside of Obsidian, you will need a esbuild plugin that can bundle it inside of the main.js.
I know that a few plugins do that, but I don't recall their names.

If you can't get that to work, please absolutely mention it, prominently mention that it connects to your server in the README.

lukasbach:
Hi @joethei. Something like that would probably be the solution, but I'm still running into issues because of Monaco's expectations of how the worker files are resolved, and the way Monaco expects the loader framework to be augmented.

I might give it another try in the future, but would prefer to leave it as is for now. I documented the caveat in the Readme, and also fixed your suggestion on the command id prefix above.

lukasbach:
Hi @joethei, I don't want to pressure and understand that you have other important things to do, but am wondering if there is something I can do to support getting this merged?

liamcain:
- [this.containerEl.getElementsByClassName("view-content")[0]](https://github.com/lukasbach/obsidian-code-files/blob/884f4684080a5eb258afaa676f8d136f8a80c16e/src/codeEditorView.ts#L60) this can be this.contentEl then you don't need to query the dom for it.
- [existingFile](https://github.com/lukasbach/obsidian-code-files/blob/884f4684080a5eb258afaa676f8d136f8a80c16e/src/createCodeFileModal.ts#L66) you also need to check that existingFile is instanceof TFile (since it can return a TFolder too)
- [obsidian](https://github.com/lukasbach/obsidian-code-files/blob/884f4684080a5eb258afaa676f8d136f8a80c16e/src/codeFilesSettingsTab.ts#L85) Capitalize Obsidian please 😄. Also, you can just check if Obsidian is in dark mode instead of asking users to toggle the switch. Just check if `app.getTheme() === 'obsidian'`

lukasbach:
Hey, thanks for taking your time to review, I've fixed the things you've noted. One side-note, the method `app.getTheme()` does not seem to be included in the official obsidian typings, I think that might be helpful to many others as well if it is visible in the types 🙂

PR #1807, opened 2023-03-25
Add new plugin "Pocketbook Cloud Highlight Importer"

lenalebt:

## Repo URL

Link to my plugin: https://github.com/lenalebt/obsidian-pocketbook-cloud-highlight-importer

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[// Remember to rename these classes and interfaces!](https://github.com/lenalebt/obsidian-pocketbook-cloud-highlight-importer/blob/2b5f7a67cece8232938cba60d38e71cd8154dc09/src/main.ts#L5)
This comment can get removed


[id: 'pocketbook-cloud-importer-perform-import',](https://github.com/lenalebt/obsidian-pocketbook-cloud-highlight-importer/blob/2b5f7a67cece8232938cba60d38e71cd8154dc09/src/main.ts#L17)

[import { stringify } from 'yaml';](https://github.com/lenalebt/obsidian-pocketbook-cloud-highlight-importer/blob/2b5f7a67cece8232938cba60d38e71cd8154dc09/src/import.ts#L2)
Instead of importing an external package, use the [`stringifyYaml`](https://github.com/obsidianmd/obsidian-api/blob/bceb489fc25ceba5973119d6e57759d64850f90d/obsidian.d.ts#L3352) function from the Obsidian API

lenalebt:
Thanks for having a look :smile: . I updated the code accordingly!

liamcain:
- [getAbstractFileByPath](https://github.com/lenalebt/obsidian-pocketbook-cloud-highlight-importer/blob/e0ba9f81d2aab817fe9e7e7ad6e66d31b6dbf56f/src/import.ts#L126) don't cast this to a TFile, add a check to ensure this is instanceof TFile to be extra sure that you don't just have a folder with that path.
- [create(file_name, '');](https://github.com/lenalebt/obsidian-pocketbook-cloud-highlight-importer/blob/e0ba9f81d2aab817fe9e7e7ad6e66d31b6dbf56f/src/import.ts#L128) instead of passing '' as the file content then modifying it after, I suggest just passing in the `content` to the create. Then move the .modify() function call into an else statement. That will avoid doing multiple writes for file creations

PR #1820, opened 2023-03-29
Adding Colorful Note Borders (obsidian-colorful-note-borders) plugin to community-plugins.json

rusi:

## Repo URL

Link to my plugin: https://github.com/rusi/obsidian-colorful-note-borders

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["main": "main.js"](https://github.com/rusi/obsidian-colorful-note-borders/blob/d736ebf425d98807d615b5fcb2e45ac533b05583/manifest.json#L9)
The Obsidian manifest does not contain this value.

It looks like your plugin will not work with pop-out windows.

rusi:
Thanks for the feedback. Are you referring to `manifest.json` that is included in the release? 
Is this (the link below) what you expect to see there? 
https://github.com/rusi/obsidian-colorful-note-borders/blob/master/manifest.json#L9

I also tested the plugin with pop-out windows and the border properly renders (see screenshot). The code uses `this.app.workspace.getActiveViewOfType(MarkdownView);` which is why I think this works in this case. Is there a different, preferred way to do this?

image

joethei:
Sorry, was a bit imprecise.
That `main` key & value should not be in the `manifest.json`, as that is not part of the spec.
 If you look at [our sample plugin](https://github.com/obsidianmd/obsidian-sample-plugin/blob/master/manifest.json) you'll see that it's not there either.

Interesting that it works in pop out windows, looks like Obsidian works a bit different in that regard than I though.
I originally mentioned that because of the calls to `document`, as that only references the main window.

rusi:
Thank you for clarifying and I am sorry I misread your comment. I've removed the "main: main.js" line.

I also fixed how the views are updated so that it works properly when using splits (and popup windows).

liamcain:
- [activeView.containerEl.querySelector(".view-content");](https://github.com/rusi/obsidian-colorful-note-borders/blob/91f980ce90d037934d7961925c4a831f11108651/src/main.ts#L99) this can be view.contentEl instead to avoid querying the dom (might need to cast to `any` if that var is not public
- [document](https://github.com/rusi/obsidian-colorful-note-borders/blob/91f980ce90d037934d7961925c4a831f11108651/src/main.ts#L85) I don't think this will work if you have a [pop-out window](https://help.obsidian.md/User+interface/Pop-out+windows) open.
- [applyRules](https://github.com/rusi/obsidian-colorful-note-borders/blob/91f980ce90d037934d7961925c4a831f11108651/src/main.ts#L93) you should probably [debounce](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L772) applyRules to avoid it getting called to frequently and causing the app to slow down.

PR #1827, opened 2023-04-01
Add Crossbow

shoedler:

Crossbow: Find backlinks in your notes

## Repo URL

Link to my plugin: https://github.com/shoedler/crossbow

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


shoedler:
Hi @joethei 
Thank you for your feedback.

I've refactored some stuff in the plugin and also addressed your suggestions.
I released a new version of my plugin and updated the files accordingly.

The merge conflict in this repo has also been resolved.

liamcain:
- [_currentFile](https://github.com/shoedler/crossbow/blob/4e50ea152669d469d21dcf60b5da9e1a82fd18ec/src/main.ts#L37) I suggest moving the _currentFile and the event listeners onto your View subclass. That way the plugin doesn't really doing any processing if the view is closed. It also is a better hierarchy since you might want to allow users to pin the view to a specific file and allow multiple views in the future.
- [prototype](https://github.com/shoedler/crossbow/blob/4e50ea152669d469d21dcf60b5da9e1a82fd18ec/src/editorExtension.ts#L22) please avoid mutating the class prototypes. While it's additive in this case, if another plugin wanted to add a function with the same name, you could run into problems. Similarly, when the user updates your plugin, the old getWordLookup function will still live on the Editor instance meaning you could run into some bugs there if you aren't careful.
- [tree-item-button](https://github.com/shoedler/crossbow/blob/4e50ea152669d469d21dcf60b5da9e1a82fd18ec/styles.css#L12) you should prefix these classnames to avoid them leaking out of your plugin view. These are very similarly named to Obsidian's tree items if we added a tree-item-button in the core app, you might unexpectedly be overriding those elements.

shoedler:
@liamcain Thanks for your review! I'll update my code accordingly.

PR #1832, opened 2023-04-01
Submitted obsidian-tesseract-ocr plugin

kaffarell:

## Repo URL

Link to my plugin: https://github.com/kaffarell/obsidian-tesseract-ocr

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["fundingUrl": "https://github.com/kaffarell",](https://github.com/kaffarell/obsidian-tesseract-ocr/blob/dc6f53bee193438734815ca1b9643323c693a28f/manifest.json#L9)
This is meant for links to services like _buy me a coffee_, _GitHub Sponsors_ etc, if you don't have such a link omit this.

["isDesktopOnly": false](https://github.com/kaffarell/obsidian-tesseract-ocr/blob/dc6f53bee193438734815ca1b9643323c693a28f/manifest.json#L10)
Your plugin depends on NodeJS only API's, which is only available on desktop, set this to `true`.

[let content = await this.app.vault.adapter.read(file.path)](https://github.com/kaffarell/obsidian-tesseract-ocr/blob/dc6f53bee193438734815ca1b9643323c693a28f/src/main.ts#L58)
Use the vault API wherever possible: `this.app.vault.cachedRead(file)`

[private getAllFiles(): TAbstractFile[] {](https://github.com/kaffarell/obsidian-tesseract-ocr/blob/dc6f53bee193438734815ca1b9643323c693a28f/src/main.ts#L111-L117)
This function can be replaced with a call to `this.app.vault.getAllLoadedFiles()`

[private isImage(fileName: string): boolean {](https://github.com/kaffarell/obsidian-tesseract-ocr/blob/dc6f53bee193438734815ca1b9643323c693a28f/src/main.ts#L134-L140)
Instead of this, do something like this:
```ts
private isImage(file: TAbstractFile) {
    return file instanceof TFile && ['jpg', 'png', 'jpeg'].includes(file.extension);
}
```

I would also recommend making the location of the tesseract executable configurable, while having it in the path works most of the time, there are cases where this does not work.

kaffarell:
Hey @joethei thanks for the guidance! Made the changes and released a new version.

community-plugins.json
```diff
@@ -6337,5 +6337,12 @@
        "description": "Obsidian Unhelper. 2023 April Fool's plugin brought to you by Obsidian.",
        "author": "Obsidian",
        "repo": "ericaxu/Gemmy"
+    },
+    {
+       "id": "image-ocr",
+       "name": "Image OCR",
+       "description": "Runs ocr on images and pastes result in description.",
```
liamcain, line 6344:
```suggestion
       "description": "Runs OCR on images and pastes result in description.",
```

liamcain:
- [Image Path](https://github.com/kaffarell/obsidian-tesseract-ocr/blob/db8ffb818422d5b555cbb06102f46e46cdf575f9/src/main.ts#L227) setting names should use [Sentence case](https://en.wiktionary.org/wiki/sentence_case). This maximizes readability.
- [if(['md'].includes(fileName.split('.')[1])) {](https://github.com/kaffarell/obsidian-tesseract-ocr/blob/db8ffb818422d5b555cbb06102f46e46cdf575f9/src/main.ts#L144) this function could be a lot simpler: `fileName.endsWith('.md')`
- [getAllLoadedFiles](https://github.com/kaffarell/obsidian-tesseract-ocr/blob/db8ffb818422d5b555cbb06102f46e46cdf575f9/src/main.ts#L115) you should use [getFiles](https://github.com/obsidianmd/obsidian-api/blob/d6f80eccf6a6f0d1ec314605bdb7e9489311503f/obsidian.d.ts#L3830) instead of writing your own. You can also use [getMarkdownFiles](https://github.com/obsidianmd/obsidian-api/blob/d6f80eccf6a6f0d1ec314605bdb7e9489311503f/obsidian.d.ts#L3826) to automatically filter to md files
- [adapter.write](https://github.com/kaffarell/obsidian-tesseract-ocr/blob/db8ffb818422d5b555cbb06102f46e46cdf575f9/src/main.ts#L90) [Prefer the Vault API over the Adapter API](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#prefer-the-vault-api-over-the-adapter-api).

community-plugins.json
```diff
@@ -6781,5 +6781,12 @@
     "author": "Dale de Silva",
     "description": "Allows import of Google Keep backup json files and their attachments. Can also be used to import other files.",
     "repo": "daledesilva/obsidian_google-keep-import"
-  }
+  },
+  {
+    "id": "image-ocr",
+    "name": "Image OCR",
+    "description": "Runs ocr on images and pastes result in description.",
```
liamcain, line 6788:
```suggestion
    "description": "Runs OCR on images and pastes result in description.",
```

PR #1834, opened 2023-04-02
Add Character Insertion Plugin

TakamiChie:

## Repo URL

Link to my plugin: https://github.com/TakamiChie/Obsidian_CharacterInsertionPlugin

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[// Remember to rename these classes and interfaces!](https://github.com/TakamiChie/Obsidian_CharacterInsertionPlugin/blob/adb34c4e80e3272b8b1464677000baf4a50a4159/src/main.ts#L4)
Can get removed.

[id: 'character-insertion-insert',](https://github.com/TakamiChie/Obsidian_CharacterInsertionPlugin/blob/adb34c4e80e3272b8b1464677000baf4a50a4159/src/main.ts#L31)

TakamiChie:
> [id: 'character-insertion-insert',](https://github.com/TakamiChie/Obsidian_CharacterInsertionPlugin/blob/adb34c4e80e3272b8b1464677000baf4a50a4159/src/main.ts#L31) Obsidian will automatically prefix the command id with the plugin id, you don't need to include it yourself.

If you remove the plug-in name from this ID, it will be only "insert", but I am thinking about whether this value is okay because it is a very common name.
At what level should this ID be unique?

joethei:
The ID you assign only needs to be unique inside your plugin.
You don't need to worry about conflicts with other plugins.

Internally Obsidian will keep track of commands in the following format `plugin-id:command-id`

TakamiChie:
Thanks for the reply.

Corrected what was pointed out. I want you to check.

liamcain:
- [getActiveViewOfType](https://github.com/TakamiChie/Obsidian_CharacterInsertionPlugin/blob/8fd93498b4313141f32b10beb37e6ec8bd61967a/src/main.ts#L20) you should use [`activeEditor`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4009) instead so that your plugin works with Canvas.
- [setLine](https://github.com/TakamiChie/Obsidian_CharacterInsertionPlugin/blob/8fd93498b4313141f32b10beb37e6ec8bd61967a/src/main.ts#L59) instead of using setLine, a better approach might be to find the range of the starting char and use replaceRange instead. That might avoid the issue you're seeing with the cursor moving.

PR #1837, opened 2023-04-02
submitting get-stock-information plugin

mikejongbloet:

## Repo URL
Link to my plugin: https://github.com/mikejongbloet/obsidian-get-stock-information

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:

[const yahooStockAPI = require("yahoo-stock-api").default;](https://github.com/mikejongbloet/obsidian-get-stock-information/blob/45c4973ea3de210b44b0e6043a607d26ce9d1b40/main.ts#L4)
Would recommend sticking to one import style.

[stock["bid"] = currentStockInfo.response.bid.value](https://github.com/mikejongbloet/obsidian-get-stock-information/blob/45c4973ea3de210b44b0e6043a607d26ce9d1b40/main.ts#L105-L107)
This could get shortened to `currentStockInfo.response.bid.value ?? null`
<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing>

mikejongbloet:
Hey @joethei - thanks for the feedback, I think I've addressed it in my current master branch.

liamcain:
- [async function callCurrentPrices](https://github.com/mikejongbloet/obsidian-get-stock-information/blob/fd2563fc4e917513f1ecc07e2a17025d94255838/src/main.ts#L40) these helper functions not need to be defined within your editor callback. It would be better to define them outside your StockInfoPlugin class to avoid recreating the functions every time the editor callback is called.
- [delete stock](https://github.com/mikejongbloet/obsidian-get-stock-information/blob/fd2563fc4e917513f1ecc07e2a17025d94255838/src/main.ts#L146) why are you deleting all the elements in the object manually?
- [callCurrentPrices](https://github.com/mikejongbloet/obsidian-get-stock-information/blob/fd2563fc4e917513f1ecc07e2a17025d94255838/src/main.ts#L55) it would be better to call the api while the modal is open, that way the user can preview the contents about to be inserted into their file. It would also mean inserting the text in the document will feel faster since the content would have been loaded with the modal up.

PR #1841, opened 2023-04-04
Add Frontmatter Alias Display plugin

muhammadv-i:

## Repo URL

Link to my plugin: https://github.com/muhammadv-i/obsidian-frontmatter-alias-display

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [app](https://github.com/muhammadv-i/obsidian-frontmatter-alias-display/blob/ded6de5d78b8e045c5f7d01e5f8db3b7663189db/main.ts#L15) It's highly discouraged that you use the global app instance in your plugin. It's currently accessible from the global context for debugging purposes, and might be removed in the future. Instead, you should use the reference to App that you have from your plugin instance.
- [layout-change](https://github.com/muhammadv-i/obsidian-frontmatter-alias-display/blob/ded6de5d78b8e045c5f7d01e5f8db3b7663189db/main.ts#L27) this will be very expensive performance-wise doing this on every layoutChange + every editor-change. I suggest at least adding a [debounce](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L772) to your function so it's not causing any slowdowns.

PR #1844, opened 2023-04-06
Add Loom

cosmicoptima:

## Repo URL

Link to my plugin: https://github.com/cosmicoptima/loom

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["fundingUrl": "https://www.patreon.com/parafactual",](https://github.com/cosmicoptima/loom/blob/2c13b15cc3fadbb1586e6d71539a85365b450c7a/manifest.json#L9)
Indentation is of.

[view(): LoomView {](https://github.com/cosmicoptima/loom/blob/2c13b15cc3fadbb1586e6d71539a85365b450c7a/main.ts#L113-L117)
Can be replaced with 
`this.app.workspace.getActiveViewOfType(LoomView);`

[id: "loom-complete",](https://github.com/cosmicoptima/loom/blob/2c13b15cc3fadbb1586e6d71539a85365b450c7a/main.ts#L170)

[hotkeys: [{ modifiers: ["Ctrl"], key: " " }],](https://github.com/cosmicoptima/loom/blob/2c13b15cc3fadbb1586e6d71539a85365b450c7a/main.ts#L174)

[callback: async () => this.complete(),](https://github.com/cosmicoptima/loom/blob/2c13b15cc3fadbb1586e6d71539a85365b450c7a/main.ts#L173)
You should use a `checkCallback` here, that way the command will only be visible in the command pallete if the command can be ran.

[const editorView = editor.cm;](https://github.com/cosmicoptima/loom/blob/2c13b15cc3fadbb1586e6d71539a85365b450c7a/main.ts#L394-L395)
Instead of this use the `registerEditorExtension` function: <https://github.com/obsidianmd/obsidian-api/blob/68f8653f79b86db88039f5f2a98d7ba2ddc69ef9/obsidian.d.ts#L2750-L2757>
<https://marcus.se.net/obsidian-plugin-docs/editor/extensions/view-plugins>

[// TextSynth API doesn't have CORS enabled, so we have to proxy](https://github.com/cosmicoptima/loom/blob/2c13b15cc3fadbb1586e6d71539a85365b450c7a/main.ts#L1028)
use the `requestUrl` function from the Obsidian API instead of `fetch`, that will take care of any CORS issues.

Please include information about which external services this plugin is calling inside the readme.

cosmicoptima:
> Can be replaced with
`this.app.workspace.getActiveViewOfType(LoomView);`

I don't think it can; `view()` returns the only `LoomView`, regardless of whether it is focused. There should be exactly one at any given time, and that is the view the method returns. I could take a few more steps to ensure multiple `LoomView`s can't exist, since I'm not sure I've prevented this robustly. (In fact, that will be one of the next things I get to, and I'll let you know here once I've done so.)

To confirm, I tried replacing the code anyway, and it didn't work, instead often returning `null` where it wouldn't have before.

> While there's no hard rule against it, we recommend against providing a default hotkey when possible. The reason being that it's easy to pick a hotkey that a user already has configured and they could get confused when that key doesn't do what they expect. Also, it's hard choosing a safe default hotkey that's available for all operating systems.

I think this plugin should provide default hotkeys for a couple of reasons:
1. It provides a lot of commands that most users will want to access often and quickly, and it would be annoying to manually set each one upon installation.
2. Most users of this plugin thus far have installed Obsidian so they can use this plugin. They sometimes talk to each other about use of the plugin and are often unfamiliar with Obsidian. A shared default set of hotkeys saves a lot of confusion in cases like this.

This is a valid concern though. If there's a way to change hotkeys at runtime, maybe I could make the set of default hotkeys opt-in?

> Instead of this use the `registerEditorExtension` function: https://github.com/obsidianmd/obsidian-api/blob/68f8653f79b86db88039f5f2a98d7ba2ddc69ef9/obsidian.d.ts#L2750-L2757
> https://marcus.se.net/obsidian-plugin-docs/editor/extensions/view-plugins

I'm not sure what you mean here. I _am_ using the `registerEditorExtension` function to register the extension. In the area you link to, I'm accessing the plugin that I registered in the way you're recommending. The way I'm accessing it is the way recommended by the docs you link here. Am I misunderstanding something?

---

I've made all of the other changes you requested, starting in 1.2.1.

cosmicoptima:
as of 1.2.3, `this.view()` doesn't exist anymore; it was only used in `this.view().render()`, which I replaced with `this.renderViews()`, which maps over `LoomView`s and renders each one. this also fixes any broken behavior on zero or multiple `LoomView`s. I made analogous changes for `LoomSiblingsView`.

cosmicoptima:
any update on this?

PR #1847, opened 2023-04-07
Add "No Tabs" plugin

TS-Tobias:

## Repo URL

Link to my plugin:
https://github.com/TS-Tobias/obsidian-no-tabs

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
Another option is probably to remove the "create" event and on file-open, iterate all the tabs in activeLeaf.parent.children, then close all the leaves that aren't the current and aren't pinned.

TS-Tobias:
> Another option is probably to remove the "create" event and on file-open, iterate all the tabs in activeLeaf.parent.children, then close all the leaves that aren't the current and aren't pinned.

You are right, thank you for the suggestion!
I might add a second mode of operation in the future.

The current implementation leaves both pinned and (manually created) non-active tabs open. This allows you to still use tabs when needed.

Removing all inactive & unpinned items would strictly enforce a one-tab policy (plus the pinned tabs, of course).

I would have preferred to simply create new notes myself (instead of closing the previous active tab) and place the cursor focus inside the filename field - but as far as I have seen, this is currently not possible with the API.

---

Could you elaborate a little more on the "removing the `create` event" part? Are you referring to [this](https://marcus.se.net/obsidian-plugin-docs/reference/typescript/classes/Vault#create) method?

liamcain:
> The current implementation leaves both pinned and (manually created) non-active tabs open. This allows you to still use tabs when needed.

Oh okay, from the name "No tabs" I assumed the goal was to enforce 1 tab.

> Could you elaborate a little more on the "removing the create event" part? Are you referring to [this](https://marcus.se.net/obsidian-plugin-docs/reference/typescript/classes/Vault#create) method?

I was just referring to your "create" event listener here: https://github.com/TS-Tobias/obsidian-no-tabs/blob/384278f2e2a7dc9e86271208bbb40dec9c3299af/main.ts#L23-L26

Since my suggest is unrelated to file creation and more about tab creation, that code block would be unnecessary.

PR #1851, opened 2023-04-08
Add Obsidian Personal Assistant plugin

edonyzpc:

## Repo URL

Link to my plugin: https://github.com/edonyzpc/personal-assistant

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[// Perform additional things with the ribbon](https://github.com/edonyzpc/personal-assistant/blob/906b91dc3554c68c1ebfc9c27817e3ecd4296b18/src/plugin.ts#L56-L57)
Remove this.

[id: 'assistant-memos',](https://github.com/edonyzpc/personal-assistant/blob/906b91dc3554c68c1ebfc9c27817e3ecd4296b18/src/plugin.ts#L83)

[hotkeys: [{ modifiers: ["Mod", "Shift"], key: "m" }],](https://github.com/edonyzpc/personal-assistant/blob/906b91dc3554c68c1ebfc9c27817e3ecd4296b18/src/plugin.ts#L85)

[const directoryExists = await adapter.exists(directoryPath);](https://github.com/edonyzpc/personal-assistant/blob/906b91dc3554c68c1ebfc9c27817e3ecd4296b18/src/plugin.ts#L151-L182)
This block of code can be replaced with:
```ts
if(vault.getAbstractFileByPath(directoryPath) === undefined) {
    vault.createFolder(directoryPath);
}
```

[const fileExists = await adapter.exists(filePath);](https://github.com/edonyzpc/personal-assistant/blob/906b91dc3554c68c1ebfc9c27817e3ecd4296b18/src/plugin.ts#L198)
It's better to check if `this.app.vault.getAbstractFileByPath(filePath) instanceof TFile`.

edonyzpc:
@joethei Thank you for reviewing, I have fixed them.

liamcain:
Reviewing the README, it seems that this plugin does several very distinct things. I think it would be appropriate for you to split this into multiple plugins so that each plugin has a clearly defined scope.

---

- [window.app](https://github.com/edonyzpc/personal-assistant/blob/be38b7c2d9e7caa5792f0e6dde58e17cf5268d7a/src/modal.ts#L29) It's highly discouraged that you use the global app instance in your plugin. It's currently accessible from the global context for debugging purposes, and might be removed in the future. Instead, you should use the reference to App that you have from your plugin instance.

community-plugins.json
```diff
@@ -6697,5 +6697,12 @@
     "author": "celeste",
     "description": "Loom in Obsidian",
     "repo": "cosmicoptima/loom"
+  },
+  {
+    "id": "personal-assistant",
+    "name": "Personal Assistant",
+    "author": "edony",
+    "description": "Obsidian management tool for plugins, notes and shortcuts which help you to use Obsidian efficiently.",
```
liamcain, line 6705: We have recently published a guide for plugin descriptions to follow here: <https://docs.obsidian.md/Plugins/Releasing/Submission+requirements+for+plugins#Keep+plugin+descriptions+short+and+simple> I suggest you reword the description to more accurately capture what this plugin does. You should also not need to mention "Obsidian" in the description since that information is already a given.

Let me know once you've updated the description to something matching our guidelines. Thanks!

edonyzpc: @liamcain Thank you for your comments. It's great suggestion about summarizing the scope of plugin, I think it's some streamlined workflows of plugins and memos management to improve efficiency. And I deleted the command which is not related and update the description of plugin, maybe I will startup a new plugin. ^-^

I have already finished a new PR, please review it again.

liamcain: Looks great, thanks!

PR #1864, opened 2023-04-11
Adding select current line plugin

gokulk16:
Repo URL
Link to my plugin: https://github.com/gokulk16/select-current-line-plugin/tree/main

Release Checklist
    - [x] Windows
    - [x] macOS
    - [x] Linux

    - [x] main.js
    - [x] manifest.json
    - [x] styles.css (optional)


joethei:

gokulk16:
> While there's no hard rule against it, we recommend against providing a default hotkey when possible. The reason being that it's easy to pick a hotkey that a user already has configured and they could get confused when that key doesn't do what they expect. Also, it's hard choosing a safe default hotkey that's available for all operating systems.

Thanks for letting know the recommendations. I'm using the 'Esc' key which is currently not assigned with any default behaviour in the editor. 

Also, double checked the 'esc' key behaviour in other places like settings and other dialogs in obsidian. Everything seems to be working fine from my testing. 

Additionally, 'Esc' should be available in all operations systems and 'Esc' cannot be chosen as default hotkey by the user. Since I was moving away from Roam and Notion, which had same functionality for the 'esc' key and wanted to have the same which I felt is more comfortable.

PR #1868, opened 2023-04-12
Plugin: Unique ID

omdenis:

## Repo URL

How to use:
https://github.com/omdenis/obsidian-id

Link to my plugin:
https://github.com/omdenis/obsidian-id/releases/tag/0.0.1

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6416,7 +6416,7 @@
         "repo": "1C0D/obsidian-easy-toggle-sidebars"
     },
     {
-        "id": "obsidian-id",
+        "id": "unique-id",
         "name": "Unique ID",
         "author": "Denis Miller",
         "description": "Add the unique ID as meta-data",
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6419,7 +6419,7 @@
         "id": "unique-id",
         "name": "Unique ID",
         "author": "Denis Miller",
-        "description": "Add the unique ID as meta-data",
+        "description": "Add the unique ID as metadata",
         "repo": "omdenis/obsidian-id"
     }
 ]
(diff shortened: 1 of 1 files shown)


joethei:
[id: 'obsidian-id-create-sub-task-command',](https://github.com/omdenis/obsidian-id/blob/d70c1729d80ed1ebc864a5b255826ca742c62ef7/main.ts#L19)
you don't need to include it yourself. You can also remove the `-command` part, as that is unecessary.

[callback: () =>](https://github.com/omdenis/obsidian-id/blob/d70c1729d80ed1ebc864a5b255826ca742c62ef7/main.ts#L22)
Use a `checkCallback` here: <https://marcus.se.net/obsidian-plugin-docs/user-interface/commands#conditional-commands>

[this.app.vault.adapter.write(active_note.path, active_note_content).then(() =>](https://github.com/omdenis/obsidian-id/blob/d70c1729d80ed1ebc864a5b255826ca742c62ef7/main.ts#L49-L52)
[Prefer the Vault API over the Adapter API](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#prefer-the-vault-api-over-the-adapter-api)

[if(!await this.is_valid_note(file))](https://github.com/omdenis/obsidian-id/blob/d70c1729d80ed1ebc864a5b255826ca742c62ef7/main.ts#L71-L80)
Would use the new `processFrontmatter` function for this: <https://github.com/obsidianmd/obsidian-api/blob/master/CHANGELOG.md#new-metadata-api>

[if (!text.startsWith('---'))](https://github.com/omdenis/obsidian-id/blob/d70c1729d80ed1ebc864a5b255826ca742c62ef7/main.ts#L117-L127)
Instead of parsing the frontmatter yourself use something like this:
```ts
const frontmatter = this.app.metadataCache.getFileCache(file)?.frontmatter;
if (!frontmatter) return false;
if (!frontmatter.id) return false;
```

The "Create subtask" command is not described in you readme.

ericaxu:

PR #1871, opened 2023-04-13
Add Copilot plugin

logancyang:

## Repo URL

Link to my plugin: https://github.com/logancyang/obsidian-copilot

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
What sets your plugin apart from the other existing Chat-GPT plugins?

[id: 'copilot-chat-toggle-window',](https://github.com/logancyang/obsidian-copilot/blob/f440e96b32f7c47609506aa1a9106b05554457ba/src/main.ts#L35)

[onst res = await fetch](https://github.com/logancyang/obsidian-copilot/blob/f440e96b32f7c47609506aa1a9106b05554457ba/src/openAiStream.ts#L46)
Use `requestUrl` from the Obsidian API instead.

[export const formatDateTime = (now: Date, timezone: 'local' | 'utc' = 'local') => {](https://github.com/logancyang/obsidian-copilot/blob/f440e96b32f7c47609506aa1a9106b05554457ba/src/utils.ts#L23-L40)
Instead of this use moment, which Obsidian exports:
```ts
import {moment} from "obsidian";
moment().format("YYYY_MM_DD-HH_mm");
```

logancyang:
@joethei Thanks for the review! I have some questions regarding the changes, but first let me answer your question.

> What sets your plugin apart from the other existing Chat-GPT plugins?

This plugin is a fully-featured ChatGPT UI as a side panel inside Obsidian. Users can interact with ChatGPT and pass texts between the chat UI and the note seamlessly without ever switching away from Obsidian, customize the params for calling OpenAI API. And it has support for various prompts (customizable prompts next) to make powerful AI features native in Obsidian. Personally I've been using custom ChatGPT UI apps on the web and on my mac, but I feel it helps me a lot if I don't need to switch back and forth, manually copying stuff and constructing prompts in another UI. I didn't find an existing plugin that fits this need yet, so I built this plugin to help my own workflow. Hopefully it may help others too.

There are more details and demos on the readme of this project: https://github.com/logancyang/obsidian-copilot

---

I removed the plugin id from the command ids. Questions about the other 2 comments:

1. I tried switching from `fetch` to `requestUrl`, but couldn't figure out how to keep the support for streaming. Since the response of ChatGPT can be quite slow and verbose (especially in the case of GPT4), it's a much better user experience to see the response streamed out than waiting for the full message. Is there a way to support that with `requestUrl`? I also don't see how it can support the abort signal.

2. I switched to using `moment` from `obsidian` for the util, thanks for the suggestion! But just curious, when I try testing using jest, the utils.test.ts complains about not having the `obsidian` module. I tried a bit to make a mock but couldn't figure out the right way. Also didn't see an example from other repos. What's the best practice to make tests for a plugin?

I'm quite new to this, would love to learn more. Thanks in advance!

logancyang:
@joethei thanks for the answer in Discord! I addressed the reviews and updated the repo to release 1.1.1. Please let me know if there's anything else.

I added more descriptions and included more gifs as example usage for this plugin. I see some other GPT-related plugins but the goal of this project is to be the most straightforward and seamless integration of ChatGPT inside Obsidian. Any more suggestions are welcome!

liamcain:
Hi @logancyang, I noticed that while the plugin is named "Copilot" in the plugin submission, you refer to the plugin as "Obsidian Copilot" in places throughout the app as well as in some of the marketing content. Hi, we're asking plugins to avoid calling themselves "Obsidian X" to reserve the naming convention for first-party plugins (see: Obsidian Publish, Obsidian Sync).

logancyang:
Thank you @liamcain ! Sorry for the late reply as I was traveling last week. Will change it to "Copilot for Obsidian" instead!

PR #1876, opened 2023-04-14
Add GitHub Issue Augmentation plugin

samprintz:
Add github-issue-augmentation-plugin

## Repo URL

Link to my plugin: https://github.com/samprintz/obsidian-issue-augmentation-plugin

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["fundingUrl": "",](https://github.com/samprintz/obsidian-issue-augmentation-plugin/blob/aadd853dfa4a77d9f50413666d4ae2a9926304a7/manifest.json#L9)

[const data = await fs.promises.readFile(path, 'utf-8');](https://github.com/samprintz/obsidian-issue-augmentation-plugin/blob/aadd853dfa4a77d9f50413666d4ae2a9926304a7/main.ts#L96)
Instead of `fs` you can use the Vault API, <https://github.com/obsidianmd/obsidian-api/blob/68f8653f79b86db88039f5f2a98d7ba2ddc69ef9/obsidian.d.ts#L3683-L3826>
It would be something like this:
```ts
const file = this.app.vault.getAbstractFileByPath(this.settings.issueFilePath);
if(file instanceof TFile) {
	const content = await this.app.vault.cachedRead(file);
	const rows = content.split(`\n`);
       ...
}
```

liamcain:
- [IssueWidget](https://github.com/samprintz/obsidian-issue-augmentation-plugin/blob/3280fde02e86b09370de2db57425cde405ebb579/issue-widget.ts#L3) why isn't this file inside /src/?
- [color: #888](https://github.com/samprintz/obsidian-issue-augmentation-plugin/blob/3280fde02e86b09370de2db57425cde405ebb579/styles.css#L2) consider using one of Obsidian's CSS vars for text colors: <https://docs.obsidian.md/Reference/CSS+variables/Foundations/Colors#Text+colors>

PR #1882, opened 2023-04-16
added entry for April-Gras/obsidian-auto-timelines

April-Gras:

## Repo URL

Link to my plugin: https://github.com/April-Gras/obsidian-auto-timelines

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["fundingUrl": "",](https://github.com/April-Gras/obsidian-auto-timelines/blob/77eba575e072fd00eecbf4dab3fb270a131baff2/manifest.json#L9)

[return `app://local/home/mgras/book/Book/${encodeURI(](https://github.com/April-Gras/obsidian-auto-timelines/blob/77eba575e072fd00eecbf4dab3fb270a131baff2/src/cardData.ts#L107-L109)
This will only work on your system, you should do this instead:
```ts
const file = this.app.metadataCache.getFirstLinkpathDest(linkText, pathOfCurrentFile);
if (file instanceof TFile) {
return this.app.vault.getResourcePath(file);
}
```

liamcain:
- [await Promise.all(vault.getMarkdownFiles());](https://github.com/April-Gras/obsidian-auto-timelines/blob/e64db325e596453001db78ed29437259b1a238e5/src/timelineMarkup.ts#L18) getMarkdownFiles isn't async. No need for the await promise.all

PR #1901, opened 2023-04-22
Plugin Html Server

Pr0dt0s:

## Repo URL

Link to my plugin: [https://github.com/Pr0dt0s/obsidian-html-server](https://github.com/Pr0dt0s/obsidian-html-server)

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[stopButton.style.color = 'var(--text-accent)';](https://github.com/Pr0dt0s/obsidian-html-server/blob/35dc678f40dc6e72508bc9300a5e62b77b93f9ae/src/plugin/uiSetup.ts#L37-L38)
You should avoid assigning styles via JavaScript and instead move all these styles into CSS so that they are more easily adaptable by themes and snippets.

[invalidPortElement.innerHTML =](https://github.com/Pr0dt0s/obsidian-html-server/blob/35dc678f40dc6e72508bc9300a5e62b77b93f9ae/src/plugin/settings/settingsTab.ts#L70-L71)
I know this specific instance is not problemantic, but please [avoid `innerHTML`](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Avoid+%60innerHTML%60%2C+%60outerHTML%60+and+%60insertAdjacentHTML%60)

Pr0dt0s:
Hi @joethei, Thanks for the review, I just released a new version with the requested changes, please let me know if there is anything else needed to be able to publish the plugin.

liamcain:
- [await new Promise<void>((resolve) =>](https://github.com/Pr0dt0s/obsidian-html-server/blob/9c7c9a0510174c7de43bfd9be2bbb2e5fc5f14b5/src/plugin/main.ts#L17-L19) Looks like this is an attempt to delay the onload until the workspace is finished loading. Don't do this. If there's anything in your onload that needs to happen after the workspace is ready, move it into a thunk inside the onLayoutReady call. But don't delay the onload from completing. Plugins are initialized in series, so any delay where will delay loading other plugins and cause the whole startup to be slower.
- [stopButton](https://github.com/Pr0dt0s/obsidian-html-server/blob/9c7c9a0510174c7de43bfd9be2bbb2e5fc5f14b5/src/plugin/uiSetup.ts#L35) I would suggest just adding your own css classname here, as joethei suggested above.

Pr0dt0s:
Hi @liamcain, I just submitted a new version on my git repo that addressed the requested changes.

Thanks again for the feedback!, and should anything be changed please let me know.

PR #1907, opened 2023-04-23
Linked Data Vocabularies plugin

kometenstaub:

I removed this plugin last year from the community plugins because the data it relied on changed, and I didn't have the time to adapt it. The plugin is now rewritten to use Obsidian's frontmatter API and the `activeEditor` property. The UI code is also rewritten.

This plugin needs to be submitted in tandem with #1908 because it needs the data from the helper plugin to work.

It uses the adapter API against recommendation because that was the original implementation and I believe that JSON files are not cached by Obsidian, contrary to Markdown files. 

## Repo URL

Link to my plugin: https://github.com/kometenstaub/linked-data-vocabularies

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- The start of your readme should clearly articulate the purpose of the plugin and the functionality it provides.
- [narrowerMax:](https://github.com/kometenstaub/linked-data-vocabularies/blob/eb278b22715d610bf1637a01d8a506e656cec169/src/main.ts#L21) why are you storing numbers as strings?
- [document](https://github.com/kometenstaub/linked-data-vocabularies/blob/eb278b22715d610bf1637a01d8a506e656cec169/src/ui/LOC/utils.ts#L14) this should be activeDocument

kometenstaub:
> The start of your readme should clearly articulate the purpose of the plugin and the functionality it provides.

image

Is that not enough? It doesn't look good on GitHub, but it should be eye-catching in the plugin browser.


> [narrowerMax:](https://github.com/kometenstaub/linked-data-vocabularies/blob/eb278b22715d610bf1637a01d8a506e656cec169/src/main.ts#L21) why are you storing numbers as strings?

Probably when I first wrote that code, I didn't think about converting the input string, and then just never changed it. But it would make sense to change it.

> [document](https://github.com/kometenstaub/linked-data-vocabularies/blob/eb278b22715d610bf1637a01d8a506e656cec169/src/ui/LOC/utils.ts#L14) this should be activeDocument

Yes.


---

I can fix these issues in about a week.

liamcain:
> Is that not enough? It doesn't look good on GitHub, but it should be eye-catching in the plugin browser

That just says "why use this plugin" but nothing in the README provides an overview of the actual functionality. You just jump right into explaining "how" to use it and "why" to use it without ever providing "what"

kometenstaub:
Isn't "What" the explanation of Linked Data that is provided further down?

I'm not sure what I should put there, except maybe a screenshot? I've probably sat too long in front of it so that I cannot see the forest for the trees.

image

image

kometenstaub:
Regarding

> [narrowerMax:](https://github.com/kometenstaub/linked-data-vocabularies/blob/eb278b22715d610bf1637a01d8a506e656cec169/src/main.ts#L21) why are you storing numbers as strings?

If I were to change that now, how would it affect existing users? I suspect they'd need to reconfigure it, but without an Obsidian plugin release changelog, I couldn't tell them about it and they'd all get errors. 
So, I'd need to write code to convert the data and keep it for a long time because not every user constantly updates their plugins.

liamcain:
> I'd need to write code to convert the data and keep it for a long time because not every user constantly updates their plugins.

Yeah, just do the conversion onload. Should be pretty straightforward

PR #1910, opened 2023-04-25
Add colored-text plugin

erincayaz:

## Repo URL

Link to my plugin: https://github.com/erincayaz/obsidian-colored-text

## Release Checklist
  - [X]  Windows
  - [X]  macOS
  - [X]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


joethei:
["author": "Erinc Ayaz <erincayaz852@gmail.com>",](https://github.com/erincayaz/obsidian-colored-text/blob/86fe13ccbc9cc673692b3f10699d30c92be644a6/manifest.json#L7)
The email address will be public on the plugin page, I would recommend to remove it.

Why is this plugin marked as destkop only?
I don't see any indication why it won't work on mobile.

[hotkeys: [{ modifiers: ["Mod", "Shift"], key: "x" }],](https://github.com/erincayaz/obsidian-colored-text/blob/86fe13ccbc9cc673692b3f10699d30c92be644a6/src/main.ts#L33)

erincayaz:
Hi, I updated manifest.json according to your comments, deleted my mail and made desktop only false. I also updated the code and removed the default hotkeys.

liamcain:
- [var](https://github.com/erincayaz/obsidian-colored-text/blob/86fe13ccbc9cc673692b3f10699d30c92be644a6/src/main.ts#L95) I recommend you change all instances of `var` to either `const` or `let`. `var` has function-level scope, so it can easily lead to bugs if you're not careful. Here's a pretty good guide with examples of why not to use `var`: [why `var` is obsolete](https://javascript.plainenglish.io/4-reasons-why-var-is-considered-obsolete-in-modern-javascript-a30296b5f08f)
- [var](https://github.com/erincayaz/obsidian-colored-text/blob/86fe13ccbc9cc673692b3f10699d30c92be644a6/src/main.ts#L91) same here
- [hotkeys: [{modifiers: ["Mod", "Shift"], key: "b"}],](https://github.com/erincayaz/obsidian-colored-text/blob/86fe13ccbc9cc673692b3f10699d30c92be644a6/src/main.ts#L46) While there's no hard rule against it, we recommend against providing a default hotkey when possible. The reason being that it's easy to pick a hotkey that a user already has configured and they could get confused when that key doesn't do what they expect. Also, it's hard choosing a safe default hotkey that's available for all operating systems.
- [export class StatusBar](https://github.com/erincayaz/obsidian-colored-text/blob/86fe13ccbc9cc673692b3f10699d30c92be644a6/src/statusBar.ts#L1) looks like this class is unused.

erincayaz:
Changed the code according to your review, deleted unnecessary files. Also pushed the code, realized I forgot to do that.

PR #1913, opened 2023-04-27
Update community-plugins.json

calomancer:

## Repo URL

Link to my plugin: [/calomancer/obsidian.nomedia/](https://github.com/calomancer/obsidian.nomedia)

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6678,11 +6678,10 @@
     "repo": "tmfelwu/obsidian-inbox"
   },
   {
-    "id": "obsidian-nomedia",
-    "name": "Obsidian nomedia",
+    "id": "android-nomedia",
+    "name": "Android nomedia",
     "author": "JakeisAwesome",
     "description": "Hides media on Android devices by adding the 'nomedia' file to each folder.",
-    "repo": "calomancer/obsidian.nomedia",
-    "branch": "main"
+    "repo": "calomancer/android.nomedia",
   }
 ]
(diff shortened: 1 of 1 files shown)


joethei:
["fundingUrl": "",](https://github.com/calomancer/android.nomedia/blob/066bfdb6fe655085206edfe2f9199c3ce2a87395/manifest.json#L9)
This is meant for links to services like _Buy me a coffee_, _GitHub sponsors_ and so on, if you don't have such a link omit this.

[import * as path from 'path';](https://github.com/calomancer/android.nomedia/blob/066bfdb6fe655085206edfe2f9199c3ce2a87395/main.ts#L7)
This will crash on mobile, as this library is only available on desktop, defeating the point of this plugin.

Checking the folder structure every few seconds is a bad idea, as this can be quite the expensive operation.
As you go through **all** files in the vault and then check the parent folder you might be checking the same folder a few hundred times every second.

React to the file creation event instead.
```ts
this.registerEvent(this.app.vault.on('create', (abstractFile) => {
    if (abstractFile instanceof TFolder) {
        this.app.vault.create(abstractFile.path + "/.nomedia", "");
   }
});
```

calomancer:
I appreciate the feedback. Thanks.

calomancer:
The updates suggested have been made. The file in the release has been updated.

PR #1924, opened 2023-04-30
Adding VC Pilot Plugin

OmarHedeya95:
A plugin that can summarize notes about startups and VCs (GPT-4) and automatically push them to Affinity (CRM)

Updated the manifest.json

## Release Checklist
Release Checklist
    - [x] macOS
   - [x] main.js
   - [x] manifest.json
   - [x] styles.css (optional)
-[x] GitHub release name matches the exact version number specified in my manifest.json (Note: Use the exact version number, don't include a prefix v)

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6681,7 +6681,7 @@
     "id": "vc-copilot",
     "name": "VC Copilot",
     "author": "Omar Hedeya",
-    "description": "A plugin that can summarize notes about startups and VCs (using GPT-4) and allows you to push them to Affinity (CRM)",
+    "description": "A plugin that can summarize notes about startups and VCs (using GPT-4) and allows you to push them to Affinity (CRM).",
     "repo": "OmarHedeya95/vc-copilot"
   }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6681,7 +6681,7 @@
     "id": "vc-copilot",
     "name": "VC Copilot",
     "author": "Omar Hedeya",
-    "description": "A plugin that can summarize notes about startups and VCs (GPT-4) and automatically push them to Affinity (CRM)",
+    "description": "A plugin that can summarize notes about startups and VCs (GPT-4) and allows you to push them to Affinity (CRM)",
     "repo": "OmarHedeya95/vc-copilot"
   }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6676,5 +6676,12 @@
     "author": "Sparsh Yadav",
     "description": "Quickly navigate and create notes through OS shortcut keys.",
     "repo": "tmfelwu/obsidian-inbox"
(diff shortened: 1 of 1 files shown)


OmarHedeya95:
I just updated the manifest.json file

OmarHedeya95:
Just removed the "v" from the releae version name. Should be correct now

OmarHedeya95:
Thanks a lot for the feedback. I implemented the requested changes now

OmarHedeya95:
Following up shortly here

joethei:
Please remove the main.js file from the repo and add it to your .gitignore - compiled release files shouldn't get mixed with source code in your commits and should go to the releases files instead.

joethei:
[.quote {](https://github.com/OmarHedeya95/vc-copilot/blob/d457f8b151c249b74663a4cac29f1f9bb10cd7bd/styles.css#L1)
[.heading {](https://github.com/OmarHedeya95/vc-copilot/blob/d457f8b151c249b74663a4cac29f1f9bb10cd7bd/styles.css#L10)
These class names are pretty generic, would recommend making them more unique.

[import * as fs from 'fs';](https://github.com/OmarHedeya95/vc-copilot/blob/d457f8b151c249b74663a4cac29f1f9bb10cd7bd/src/main.ts#L3-L4)
Unused imports.
These imports are only available from a nodeJS runtime, which means that this would throw errors on Mobile (that's a problem since you have `isDesktopOnly` marked as `false` in your manifest.json file.


[scriptPath_AI = this.settings.vaultPath + '.obsidian/plugins/vc_copilot'](https://github.com/OmarHedeya95/vc-copilot/blob/d457f8b151c249b74663a4cac29f1f9bb10cd7bd/src/main.ts#L447)
Looks like this setting is also unused.

[console.log(`Title: ${title}`)](https://github.com/OmarHedeya95/vc-copilot/blob/d457f8b151c249b74663a4cac29f1f9bb10cd7bd/src/main.ts#L146-L147)

[containerEl.createEl('h2', {text: 'Settings for your copilot'});](https://github.com/OmarHedeya95/vc-copilot/blob/d457f8b151c249b74663a4cac29f1f9bb10cd7bd/src/main.ts#L550)

[submitButton.style.position = 'absolute';](https://github.com/OmarHedeya95/vc-copilot/blob/d457f8b151c249b74663a4cac29f1f9bb10cd7bd/src/modal.ts#L31)

[let r: any = await request({](https://github.com/OmarHedeya95/vc-copilot/blob/d457f8b151c249b74663a4cac29f1f9bb10cd7bd/src/utils.ts#L61-L66)
Alternatively you can use:
```ts
 const r = await request({
	url: full_url,
	method: "GET",
	headers: headers
});
 
const response = r.json();
```
Also you should not type `r` as `any´ here.

ericaxu:

PR #1925, opened 2023-04-30
Add auto-hide-cursor plugin

moismat:

## Repo URL

https://github.com/moismat/obsidian-auto-hide-cursor

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["fundingUrl": "",](https://github.com/moismat/obsidian-auto-hide-cursor/blob/1fe92e084e409709c45ec8c0ab4fc2d7227a1275/manifest.json#L9)

Why is this plugin marked as desktop only?
I don't see any reason for it, and you say you already tested on mobile.

[document.body.addEventListener(](https://github.com/moismat/obsidian-auto-hide-cursor/blob/1fe92e084e409709c45ec8c0ab4fc2d7227a1275/main.ts#L5-L19)
Instead of this use the following:
```ts
this.registerDomEvent(document, "scroll", () => {
...
});
```
And remove the code in `onunload` that removes the event listeners, as Obsidian will handle this when using the `registerDomEvent` function.

Your plugin won't work in pop-out windows.

moismat:
hey @joethei , updated  [here](https://github.com/moismat/obsidian-auto-hide-cursor/pull/1/files#r1186104600) 😄

moismat:
updated. now also works on popout windows 😄

liamcain:
- [layout-change](https://github.com/moismat/obsidian-auto-hide-cursor/blob/7ab81ea3950ed07830dae358b0660da0601c63f8/main.ts#L8) instead of doing this on every layout change (which can happen frequently and will mean that the listener will get registered more than it needs to), you should change this to [`window-open`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L4102). That way it will only register it for new pop-out windows.

PR #1937, opened 2023-05-05
Remove Emoji in 🔔 Chime theme.

Bluemoondragon07:
Some people have issues syncing the theme with Dropbox because of the emoji in the title, so I was wondering if I could remove that stylization.


joethei:
This is possible, but there will two issues:
- existing users won't have the ability to update directly, they need to download the theme from the theme list again.
- the download count will be reset

PR #1959, opened 2023-05-16
Add plugin `show-dotfiles`

polyipseity:

(Side note, there are many trailing whitespaces and other types of inconsistent formatting in `community-plugins.json` that you may want to fix.)

## Repo URL

Link to my plugin: https://github.com/polyipseity/obsidian-show-dotfiles

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6851,5 +6851,12 @@
     "author": "Numeroflip",
     "description": "Automatically prompt for a template, when creating a note.",
     "repo": "numeroflip/obsidian-auto-template-prompt"
+  },
+  {
+    "id": "show-dotfiles",
+    "name": "Show Dotfiles",
+    "author": "polyipseity",
+    "description": "Show dotfiles in Obsidian.",
+    "repo": "polyipseity/obsidian-show-dotfiles"
   }
 ]


Bluemoondragon07:
@polyipseity  Thank you for making this plug-in! I've been looking for a plug-in just like this for so long. Now I can edit css files IN Obsidian! Which is an oddly specific reason for wanting this plug-in, but anyways, thank you so much!

joethei:
This is quite the hacky way to achieve this.
This will break all core plugins that rely on this behavior, Obsidian Sync specifically will be badly affected by this hack.

mwoz123:
@joethei what if the plugin contains warning  "it's not compatible with/cannot be used with Obsidian Sync (and possible other plugin). Use at your risk." Or something similar? 
Will that make you accept it? 

@joethei do you know a less "hacky" way to achieve it? If yes could you share it? 

Cause I'd also love to see such feature in Obsidian.

joethei:
Sorry, we can't accept this plugin in its current state, even with a warning, it's just too dangerous.

Rather than overwriting core behavior of functions provided by JavaScript, I would recommend only monkey-patching the specific functions in Obsidian that are relevant here.

joethei:
Closing this due to no activity in the last month.

Feel free to create a new PR after you have updated your plugin.

PR #1960, opened 2023-05-16
Add plugin `Supsub`

wjgoarxiv:

## Repo URL

Link to my plugin: https://github.com/wjgoarxiv/obsidian-supsub

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["description": "`Supsub` is a plugin designed for Obsidian.md that adds functionality to wrap selected text with](https://github.com/wjgoarxiv/obsidian-supsub/blob/ba63a5756c871d9be2f2ae638cefc10910d46b7e/manifest.json#L6)
We have recently published a guide for plugin descriptions to follow here: https://docs.obsidian.md/Plugins/Releasing/Submission+requirements+for+plugins#Keep+plugin+descriptions+short+and+simple.
Avoid including sentences like "This is a plugin for Obsidian" in your description.

Where is your source code?
Or have you made all your changes in the `main.js` file directly?


[hotkeys: [](https://github.com/wjgoarxiv/obsidian-supsub/blob/ba63a5756c871d9be2f2ae638cefc10910d46b7e/main.js#L38-L43)

[const activeLeaf = this.app.workspace.activeLeaf;](https://github.com/wjgoarxiv/obsidian-supsub/blob/ba63a5756c871d9be2f2ae638cefc10910d46b7e/main.js#L59-L63)
`activeLeaf` is deprecated and should not be used.

[const editor = activeLeaf.view.sourceMode.cmEditor;](https://github.com/wjgoarxiv/obsidian-supsub/blob/ba63a5756c871d9be2f2ae638cefc10910d46b7e/main.js#L62)
`cmEditor` is only for the legacy editor, and should not be used.

Instead of a simple `callback` for your commands you should use a `editorCallback` instead, this will solve the two issues above.
https://docs.obsidian.md/Plugins/User+interface/Commands#Editor+commands

wjgoarxiv:
**Where is your source code?**
→ Yes, I've made all my changes in the `main.js` file directly. 

**`activeLeaf` is deprecated and should not be used.** && **`cmEditor` is only for the legacy editor, and should not be used.**
→ I revised the script; I removed the hotkeys and switched from using `activeLeaf` and `cmEditor` to using the editor passed into the `editorCallback` function. May work with both the legacy and the new editor in Obsidian.

joethei:
Editing the compiled `main.js` directly is not recommended.
Please either use TypeScript following our template: <https://github.com/obsidianmd/obsidian-sample-plugin>, or update your code to use JavaScript properly.
Currently the first ~30 lines of your code are not needed.

PR #1984, opened 2023-05-27
Add Iconoir Icons plugin

gapmiss:

## Repo URL

Link to my plugin: https://github.com/gapmiss/iconoir-icons

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:

It looks like your plugin won't work in pop out windows.

marcusolsson (approved):

liamcain:
- [let foo='bar';](https://github.com/gapmiss/iconoir-icons/blob/9cdf13823d79ddde9841a7c7ed8823ae3a2cb7d4/src/main.ts#L97) looks like this is testing code that made it in
- [var](https://github.com/gapmiss/iconoir-icons/blob/9cdf13823d79ddde9841a7c7ed8823ae3a2cb7d4/src/main.ts#L183) no need for this to be 'var'. You can change this to 'const' or 'let'

gapmiss:
Changes made, tested and version conflict resolved

PR #1986, opened 2023-05-28
Add note-archiver plugin

thenomadlad:

## Repo URL

Link to my plugin: https://github.com/thenomadlad/obsidian-note-archiver

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


thenomadlad:
Hi @joethei thank you for the review and feedback!

> [import { join, dirname } from "path";](https://github.com/thenomadlad/obsidian-note-archiver/blob/669afa991eb4e8a9c5aea02fd2bb1c4a5e2dd7b1/main.ts#L12) This will break on mobile, as Node.js is not available there.

Hmm I'm primarily using join and dirname to take into account the fact that windows uses `\` and prefixes file paths with a drive name:
* this is actually a problem right? or do normal paths like `Archive/2023/my-note.md` work fine in windows as well? (I should create a windows vm to test my plugin..)
* do you have a suggestion for a method or library to use that helps with that?
* is there a way for the `app` object to tell me which OS a user is using so that I can use the right directory separator?

> [id: "note-archiver-archive-current-note",](https://github.com/thenomadlad/obsidian-note-archiver/blob/669afa991eb4e8a9c5aea02fd2bb1c4a5e2dd7b1/main.ts#L37) Obsidian will automatically prefix the command id with the plugin id, you don't need to include it yourself.

Fixed.
 
> Why do you include the version number in the settings object?

I'm planning to change the structure of the plugin's settings in future versions, and I'm just not sure what the new settings will look like: for example, today you can only specify one archive folder, but in the future I want to make it so some sections of your vault can be archived to different folders (and I'm still working on how to represent this)

However, I don't want users to lose their old settings when I do so, so I will use the version field to translate their settings for them on upgrading

joethei:
As you are using the Vault API you don't need to care about OS differences, as all paths are vault absolute, and using forward slashes is recommended here.

When you are passing in user input, use `normalizePath` it will clean up the path to a format that Obsidian understands: <https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Use+%60normalizePath()%60+to+clean+up+user-defined+paths>.

thenomadlad:
> As you are using the Vault API you don't need to care about OS differences, as all paths are vault absolute, and using forward slashes is recommended here.

Good to know, thanks very much, I think I didn't pick this up when I read the doc for normalizePath 😅. Fixed usage of node-api in this commit (and updated the release as well): https://github.com/thenomadlad/obsidian-note-archiver/commit/2c169bce3cdb99b1a0b4a5687835bab545453d6c

Last thing left - test this plugin on the other platforms

thenomadlad:
@joethei and @marcusolsson I've tested it on all platforms, this plugin is ready for primetime. thanks so much for taking a look!

PR #1994, opened 2023-06-01
Add `obsidian-file-include-plugin` community plugin.

tillahoffmann:

## Repo URL

Link to my plugin: https://github.com/tillahoffmann/obsidian-file-include

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[import * as path from 'path';](https://github.com/tillahoffmann/obsidian-file-include/blob/2eef34c32550b76a3231bbfb50957cedce1715fd/main.ts#L2)
This will break on mobile, as Node.js is not available there.


[// Resolve the file relative to the current document.](https://github.com/tillahoffmann/obsidian-file-include/blob/2eef34c32550b76a3231bbfb50957cedce1715fd/main.ts#L17-L25)
You should be able to use the [`normalizePath`](https://github.com/obsidianmd/obsidian-api/blob/583ba39e3f6c0546de5e5e8742256a60e2d78ebc/obsidian.d.ts#L2541) function here instead.

tillahoffmann:
Thanks for the review. Addressed in 1.0.1 by resolving paths manually.

joethei:
You should still use the `normalizePath` function from our API, as it will do that for you: <https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Use+%60normalizePath()%60+to+clean+up+user-defined+paths>

tillahoffmann:
I've added a `normalizePath` call to 1.0.2. However, I've still resolved paths manually using `resolvePath` because `normalizePath` does not handle relative paths. E.g.,

```js
> normalizePath('foo/../bar')
'foo/../bar'
> resolvePath('', 'foo/../bar')
'bar'
```

The reason I resolve manually is that `getAbstractFileByPath` does not handle `..`. E.g.,

```js
> this.app.vault.getAbstractFileByPath('foo/../hello.py')
null
> this.app.vault.getAbstractFileByPath('hello.py')
t {parent: t, deleted: false, vault: t, path: 'hello.py', name: 'hello.py', …}
```

tillahoffmann:
Following a suggestion by @mayurankv, the plugin now uses `@/` to denote paths relative to the vault root rather than `/`. This also means we can apply `normalizePath` before and after `resolvePath` to make sure the user-supplied path has the expected format.

PR #2030, opened 2023-06-11
Added Vault 2 Book by Mitra98t

Mitra98t:

## Repo URL

Link to my plugin:

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6920,6 +6920,6 @@
     "name": "Vault 2 Book",
     "description": "Converts your notes to a book",
     "author": "Mitra98t",
-    "repo": "Mitra98t/obsidian2book-plugin"
+    "repo": "Mitra98t/vault2book-plugin"
   }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6914,5 +6914,12 @@
     "description": "Format interlinear glosses used in linguistics texts.",
     "author": "Mijyuoon",
     "repo": "Mijyuoon/obsidian-ling-gloss"
+  },
+  {
+    "id": "obsidian-2-book",
+    "name": "Obsidian 2 Book",
+    "description": "Converts your notes to a book",
+    "author": "Mitra98t",
+    "repo": "Mitra98t/obsidian2book-plugin"
   }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6918,7 +6918,7 @@
   {
     "id": "vault-2-book",
     "name": "Vault 2 Book",
-    "description": "Converts your notes to a book",
+    "description": "Converts your notes to a book creating a single file containing all the notes linked",
     "author": "Mitra98t",
     "repo": "Mitra98t/vault2book-plugin"
   }
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6916,8 +6916,8 @@
     "repo": "Mijyuoon/obsidian-ling-gloss"
   },
   {
-    "id": "obsidian-2-book",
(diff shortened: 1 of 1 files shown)


joethei:
[this.registerInterval(window.setInterval(() => 5 * 60 * 1000));](https://github.com/Mitra98t/vault2book-plugin/blob/05ce43b6d52cff6fe65f99dc6e1c3b413a0b25ef/main.ts#L91)
You can remove this, since it's unused.

[if (this.settings.showRibbonButton) {](https://github.com/Mitra98t/vault2book-plugin/blob/05ce43b6d52cff6fe65f99dc6e1c3b413a0b25ef/main.ts#L95-L108)

[function isDocFile(file: TAbstractFile): boolean {](https://github.com/Mitra98t/vault2book-plugin/blob/05ce43b6d52cff6fe65f99dc6e1c3b413a0b25ef/main.ts#L152-L158)
Do a `instanceof` check instead: https://docs.obsidian.md/Plugins/Vault#Is+it+a+file+or+folder%3F

[const fileContent = await app.vault.read(file);](https://github.com/Mitra98t/vault2book-plugin/blob/05ce43b6d52cff6fe65f99dc6e1c3b413a0b25ef/main.ts#L186-L194)
To check if a file has a certain tag check:
```ts
getAllTags(app.metadataCache.getFileCache(file));
```


[const file: TFile = fileStr as TFile;](https://github.com/Mitra98t/vault2book-plugin/blob/05ce43b6d52cff6fe65f99dc6e1c3b413a0b25ef/main.ts#L239)
Instead of casting to `TFile` you should remove the function call and do the `instanceof` check here, as typescript automatically performs a cast then.


[const files: TAbstractFile | null = await vault.getAbstractFileByPath("/");](https://github.com/Mitra98t/vault2book-plugin/blob/05ce43b6d52cff6fe65f99dc6e1c3b413a0b25ef/main.ts#L361)
`getAbstractFileByPath` is not an async function.

[const fileExists = await adapter.exists(fileName);](https://github.com/Mitra98t/vault2book-plugin/blob/05ce43b6d52cff6fe65f99dc6e1c3b413a0b25ef/main.ts#L427)
Better would be using the Vault API, just check if `app.vault.getAbstractFileByPath` returns something.

[containerEl.createEl("h2", { text: "Settings for my awesome plugin." });](https://github.com/Mitra98t/vault2book-plugin/blob/05ce43b6d52cff6fe65f99dc6e1c3b413a0b25ef/main.ts#L553)
Remove this, see https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Only+use+headings+under+settings+if+you+have+more+than+one+section.

Mitra98t:
Triggering checks

Mitra98t:
I should have fixed everything you requested now

liamcain:
- [fileExists](https://github.com/Mitra98t/vault2book-plugin/blob/2f8712713b1903d73fd1311c6858213853b933c0/main.ts#L412) you should do an instanceof TFile check instead of casting to TFile. getAbstractFileByPath can also return a TFolder.
- [content](https://github.com/Mitra98t/vault2book-plugin/blob/2f8712713b1903d73fd1311c6858213853b933c0/main.ts#L363) you might consider storing your content as a string[] and pushing content to the array as you go. Then you can do content.join() at the end. This should be slightly faster than concatenating the strings as you go.

PR #2049, opened 2023-06-15
add guid-front-matter plugin

llimllib:

## Repo URL

Link to my plugin: [llimllib/obsidian-guid-plugin](https://github.com/llimllib/obsidian-guid-plugin)

## Release Checklist

  - [ ] Windows
  - [x] macOS
  - [ ] Linux
  - [ ] Android _(if applicable)_
  - [ ] iOS _(if applicable)_
  - [x] on my desktop, but with local mode enabled
  - [x] `main.js`
  - [x] `manifest.json`


pjeby:
You can simplify your implementation by using `app.vault.process()` or better yet, `app.fileManager.processFrontMatter()` to make the changes.  Your `_addId` function would then look something like:

```ts
async function _addID(app: App, f: TFile): Promise<void> {
    const key = "id";
    if (!app.metadataCache.getFileCache(f)?.frontmatter?.[key]) {
        await app.fileManager.processFrontMatter(data => { data[key] = ulid(); });
    }
}
```

Checking the cache before performing the operation will prevent unnecessary file reads, and the `processFrontMatter()`  API will avoid the need to use grey-matter.  (It will also benefit from any future improvements to Obsidian's front matter handling.)

By the way, `app` is available as a global variable in Obsidian now, so you do not need to have it as a parameter.  I *suspect* that if you check metadata as shown and use `processFrontMatter()` to do the writing, then all your setTimeout and debounce calls will also be unnecessary,  But you should check whether that's actually the case, if you added them due to issues you discovered.  If those were copied from another plugin, do note that older plugins may have been written before the existence of some of the APIs I'm suggesting you use here (and before `app` was a global variable)`, so the plugin you used as an example may have needed the timeouts and debouncing for older versions of Obsidian.

llimllib:
Thanks! published 1.1.0 with your suggestions.

a bit of feedback: it was extremely confusing trying to figure out how to do this from the docs. Not hating, just didn't find what I was looking for there

pjeby:
> It's highly discouraged that you use the global app instance in your plugin. It's currently accessible from the global context for debugging purposes, and might be removed in the future.

It was added to the official API in 0.16.3 (about 9 months ago) and is not marked as deprecated.

pjeby:
Well, I didn't actually hear about it via the API on Github -- I had a [Discord convo w/Licat last March](https://discord.com/channels/686053708261228577/716028884885307432/958116305448337428) where he said he was considering making it global, wasn't sure why it was deprecated, and that he was going to put it in the next d.ts distribution -- which then [happened three days later](https://github.com/obsidianmd/obsidian-api/blame/3770989c2364ead5429eb7e4475e095588bbe914/obsidian.d.ts#L4196) and was officially announced in [the release notes for 0.14.4](https://obsidian.md/changelog/2022-03-31-desktop-v0.14.4/), but I don't see any re-deprecation in any subsequent release notes.

llimllib:
OK, I published 1.2.0 with no console logging and without using the global app object. I truly don't care, so just let me know how you think it ought to be

llimllib:
Anything I could do to push this PR along?

llimllib:
I re-merged master, still looking for any feedback or to get this merged

PR #2051, opened 2023-06-17
Add VOX Transcription

vincentbavitz:
Obsidian VOX allows you to automatically transcribe and categorize your voice notes.

## Repo URL

Link to my plugin: https://github.com/vincentbavitz/obsidian-vox/

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [X]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6284,7 +6284,7 @@
     "author": "1C0D",
     "description": "toggle both sidebars (double middle mouse click or own shortcut) or each sidebar clicking middle mouse button and moving mouse towards the corresponding sidebar.",
     "repo": "1C0D/obsidian-easy-toggle-sidebars"
-  },
+  },  
   {
     "id": "x86-flow-graphing",
     "name": "x86 Assembly Flow Graphing",
@@ -6467,10 +6467,10 @@
     "description": "Search for Wikipedia articles and link them.",
     "repo": "StrangeGirlMurph/obsidian-wikipedia-search"
   },
-  {
+  { 
     "id": "vim-toggle",
     "name": "Vim Toggle",
-    "description": "(It toggles vim on/off) Allows for the the toggling and customization of said toggling within settings.",
+    "description": "(It toggles vim on/off) Allows for the the toggling and customization of said toggling within settings.", 
     "author": "Conner Ohnesorge",
     "repo": "conneroisu/vim-toggle"
   },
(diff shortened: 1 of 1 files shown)


joethei:
You need to make it more prominent in your README that the plugin requires a server component.

[Copyright (c) 2020 Vinzent03, Denis Olehov](https://github.com/vincentbavitz/obsidian-vox/blob/d55f5ce1a99502026669a1e921891badcb11d261/LICENSE#L3)
Please update this with your own info, and update the year number.

Your `package.json` contains a few dependencies that don't make sense in your plugin, looks like you copied those from the Git plugin.

[// Give the app time to load in plugins and run its index check.](https://github.com/vincentbavitz/obsidian-vox/blob/d55f5ce1a99502026669a1e921891badcb11d261/src/main.ts#L27-L29)
Use `Workspace#onLayoutReady` instead:
```ts
    this.watchUnprocessedDirectory();
});
```

[const unprocessedFiles = this.app.vault](https://github.com/vincentbavitz/obsidian-vox/blob/d55f5ce1a99502026669a1e921891badcb11d261/src/main.ts#L45-L48)
You should not filter through all files for this, rather use something like this:
```ts
const paths: string[] = [];
    const folder = this.app.vault.getAbstractFileByPath(this.settings.watchDirectory);
    if(folder instanceof TFolder) {
      Vault.recurseChildren(folder, (file) => {
        if(file instanceof TFile) {
          paths.push(file.path);
        }
      });
    }
```


[// Then watch for any changes...](https://github.com/vincentbavitz/obsidian-vox/blob/d55f5ce1a99502026669a1e921891badcb11d261/src/main.ts#L52-L63)
You should use the Obsidian Event system instead, something like:
```ts
 this.registerEvent(this.app.vault.on('create', (file) => {
      if(file.path.includes(this.settings.watchDirectory)) {
        this.processor.queueFiles([file.path]);
      }
    }));
```
You'll want to listen for `create`, `modify`, `delete` and `rename`.

[// If settings change quickly, we don't want to spam `watchUnprocessedDirectory`;](https://github.com/vincentbavitz/obsidian-vox/blob/d55f5ce1a99502026669a1e921891badcb11d261/src/main.ts#L67-L74)
You can use a `debounce` here:
```ts
debounce(this.watchUnprocessedDirectory, WATCHER_DELAY_MS)();
```

[import manifest from "../../manifest.json";](https://github.com/vincentbavitz/obsidian-vox/blob/d55f5ce1a99502026669a1e921891badcb11d261/src/utils/various.ts#L1)
This information is already available from the plugin instance: <https://docs.obsidian.md/Reference/TypeScript+API/Plugin/Plugin>

[export function timer(ms: number) {](https://github.com/vincentbavitz/obsidian-vox/blob/d55f5ce1a99502026669a1e921891badcb11d261/src/utils/various.ts#L3-L5)
Obsidian exports the `sleep` function, which does the same thing.


[ this.addSettingsCategoryHeading("General Settings");](https://github.com/vincentbavitz/obsidian-vox/blob/d55f5ce1a99502026669a1e921891badcb11d261/src/settings/index.ts#L68)

You should not rely on the `fs` package as much, to me it looks like you can replace all of those with the Vault API from Obsidian.
You should also [prefer the Vault API over the Adapter API](https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md#prefer-the-vault-api-over-the-adapter-api).

vincentbavitz:
@joethei Thanks a lot for your review -- I'll amend the plugin shortly. :)

vincentbavitz:
Dear @joethei, I've amended the plugin and updated WRT your review.

liamcain:
- [debounce](https://github.com/vincentbavitz/obsidian-vox/blob/19d19faed5df0cc22b72abf60b2941c5c249f2da/src/main.ts#L66) this doesn't do what you think it does. The debounce is applied to the wrapper function and not watchUnprocessedDirectory. So it will never actually be debounced. Instead you'll want to do something like:

```
onload() {
  this.watchUnprocessedDirectory = debounce(this.watchUnprocessedDirectory, WATCHER_DELAY);
}

saveSettings() {
  this.watchUnprocessedDirectory();
}
```

- [this.registerEvent(this.app.vault.on('create', queueFromWatcher));](https://github.com/vincentbavitz/obsidian-vox/blob/19d19faed5df0cc22b72abf60b2941c5c249f2da/src/main.ts#L59-L60) why are you registering these event handlers ever time the watchUnprocessedDirectory is called? This is going to cause the functions to get called way more than you expect.

liamcain:
Sounds good, let me know when it's ready for another look!

vincentbavitz:
> Sounds good, let me know when it's ready for another look!

Hey @liamcain 
I've amended the plugin now to reflect the PR changes. :)

PR #2067, opened 2023-06-20
Add Memorization Plugin to Obsidian Plugins

nwindian:

---

Repo URL
Link to my plugin: https://github.com/nwindian/Memorization-Plugin

Release Checklist
     - [x] Windows
     - [x] macOS
     - [ ] Linux
     - [ ] Android (if applicable)
     - [ ] iOS (if applicable)
     - [x] main.js
     - [x] manifest.json
     - [x] styles.css (optional)

 - [x] I have read the tips in https://github.com/obsidianmd/obsidian-releases/blob/master/plugin-review.md and have self-reviewed my plugin to avoid these common pitfalls. (💡 See below)
    Note: I do use adapter in my project. But only twice to check if a file/folder exists. I decided to continue with this change, but will of course open to feedback from the reviewer.


joethei:
Please fill the pull request template

nwindian:
@joethei Filled out the template 👍

joethei:
["description": "This is an index card plugin for memorizing your notes. This plugin utilizes a spaced repetion (SM-2) algorithm.",](https://github.com/nwindian/Memorization-Plugin/blob/6c90e4fe5af8b606ce20ada9fae63c8c3811edf6/manifest.json#L6)
We have recently published a guide for plugin descriptions to follow here: https://docs.obsidian.md/Plugins/Releasing/Submission+requirements+for+plugins#Keep+plugin+descriptions+short+and+simple.
Avoid including sentences like "This is a plugin" in your description.

Your Readme should contain some screenshots.

[const files =  this.app.vault.getMarkdownFiles()](https://github.com/nwindian/Memorization-Plugin/blob/6c90e4fe5af8b606ce20ada9fae63c8c3811edf6/main.ts#L70-L94)
This code is duplicated from above, you should move this into a function.

[containerEl.createEl('h2', {text: 'Settings'});](https://github.com/nwindian/Memorization-Plugin/blob/6c90e4fe5af8b606ce20ada9fae63c8c3811edf6/main.ts#L175)


[this.contentEl.style.height = '500px'](https://github.com/nwindian/Memorization-Plugin/blob/6c90e4fe5af8b606ce20ada9fae63c8c3811edf6/src/Modals/PromptModal.ts#L61)

[const originalFile = this.app.vault.getAbstractFileByPath(this.originalTitle)](https://github.com/nwindian/Memorization-Plugin/blob/6c90e4fe5af8b606ce20ada9fae63c8c3811edf6/src/StudyNote.ts#L40-L41)
You should check first if `originalFile instanceof TFile`, instead of blindly casting.

In general I would recommend creating a custom view for your plugin, instead of modifying a note, which is a hacky way achieve this.
<https://docs.obsidian.md/Plugins/User+interface/Views>

nwindian:
@joethei Thanks for the feedback 👍 . 
Regarding the recommendation to use a view instead of a note - I actually didn't think it was hacky. I experimented with both a view and a note and I ended up preferring the latter because I thought it was helpful to have my "index cards" stored as notes that I could see and also I'd see them in my graph view under the same tag.

nwindian:
@joethei I updated the repo according to your comments (except for changing the notes to a view).

PR #2076, opened 2023-06-23
Add Send to Ghost

Southpaw1496:

## Repo URL

Link to my plugin: https://github.com/Southpaw1496/obsidian-send-to-ghost/

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
Please follow the pull request template.

Southpaw1496:
Do I have to re-open the pull request for that?

joethei:
No, but replace your first message here with the template and fill it.
You can get the template here: https://raw.githubusercontent.com/obsidianmd/obsidian-releases/master/.github/PULL_REQUEST_TEMPLATE/plugin.md

Southpaw1496:
Done

joethei:
[containerEl.createEl("h2", { text: "Send to Ghost" });](https://github.com/Southpaw1496/obsidian-send-to-ghost/blob/a767ae59fd2a07ca75a597b5a4e6cf6a65c83d53/src/settingTab/index.ts#L16)

[const noteFile = app.workspace.getActiveFile();](https://github.com/Southpaw1496/obsidian-send-to-ghost/blob/a767ae59fd2a07ca75a597b5a4e6cf6a65c83d53/src/methods/publishPost.ts#L38)

[const result = await request({](https://github.com/Southpaw1496/obsidian-send-to-ghost/blob/a767ae59fd2a07ca75a597b5a4e6cf6a65c83d53/src/methods/publishPost.ts#L51-L63)
You can use this instead:
```ts
const result = await requestUrl(<same as current code>);
const json = result.json();
```

Southpaw1496:
I've addressed the first and third issue, but the third one is confusing: I'm very new to JS and plugin development and just trying to keep this one alive. Do you know of any relevant documentation I could read or code examples I could use to help me understand it?

joethei:
Generally you would pass in the plugin instance as an additional parameter, and then call `plugin.app.....` in your code.

In the specific case you can take the app from the view though. i.e `view.app.workspace.getActiveFile()`;

Southpaw1496:
I've addressed all of your comments.

liamcain:
- [MarkdownView](https://github.com/Southpaw1496/obsidian-send-to-ghost/blob/05022a1b44eebb87234796a83688dc1b2341776a/src/main.ts#L31) the typing here is wrong. This argument is not always a markdownview. If the active view is a CanvasView, it could be the MarkdownFileInfo for the actively focused card in canvas.
- [Ghostt](https://github.com/Southpaw1496/obsidian-send-to-ghost/blob/05022a1b44eebb87234796a83688dc1b2341776a/src/main.ts#L34) typo
- [requestUrl](https://github.com/Southpaw1496/obsidian-send-to-ghost/blob/05022a1b44eebb87234796a83688dc1b2341776a/src/methods/publishPost.ts#L51) you should wrap this in a try/catch to properly handle errors
- [adminToken](https://github.com/Southpaw1496/obsidian-send-to-ghost/blob/05022a1b44eebb87234796a83688dc1b2341776a/src/methods/publishPost.ts#L26) you should make sure that the token isn't empty

Southpaw1496:
In regards to the third issue, is there any documentation on handling errors from `requestUrl`? The official documentation doesn't have much information about what errors it can throw.

I'd want to mainly catch 404 and 401 errors, as those would be the errors if the user mistyped the API URL or API key, respectively.

liamcain:
requestUrl is basically the same as fetch, except that the requests are funneled through the electron main process, to bypass CORS errors. The error types/response will depend on the API you're hitting, so I'd recommend just testing it locally and seeing what object shape you get back.

Southpaw1496:
I have addressed the second, third, and fourth comments. I haven't yet pushed a new tag, I will do that when all of your comments have been addressed.

In regards to the first comment, would [TextFileView](https://docs.obsidian.md/Reference/TypeScript+API/TextFileView/TextFileView) be the appropriate type to use, or is there something else?

Southpaw1496:
I will wait until the plugin's issues are resolved to add the new tag.

liamcain:
> In regards to the first comment, would [TextFileView](https://docs.obsidian.md/Reference/TypeScript+API/TextFileView/TextFileView) be the appropriate type to use, or is there something else?

Sorry, I missed this comment.

the type is `MarkdownView | MarkdownFileInfo` meaning that you're passed either a markdown view or a markdownfileinfo object. So you can do something like:

```
		this.addCommand({
			id: "publish",
			name: "Send to Ghost",
			editorCallback: (_, view: MarkdownView | MarkdownFileInfo) => {
				if (!(view instanceof MarkdownView)) {
					new Notice(
						"You must open the note that you want to send to Ghostt"
					);
					return;
				}

				publishPost(view, this.settings);
			},
```

Southpaw1496:
I have addressed all your comments

PR #2079, opened 2023-06-24
Rename Notion-Like Tables to Dashboards

trey-wallis:
# I am updating an existing Community Plugin

Here is the my new repo URL:
https://github.com/trey-wallis/obsidian-dashboards

I didn't change the id because I want to retain my download count

image

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -3725,10 +3725,10 @@
   },
   {
     "id": "notion-like-tables",
-    "name": "Notion-Like Tables",
+    "name": "Dashboards",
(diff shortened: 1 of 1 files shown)


joethei:
Please don't reformat the file, only change your own entry.

trey-wallis:
@joethei Updated :)

joethei:
It looks like you referenced the wrong repository.
`trey-wallis/dashboards` does not exist, but `trey-wallis/obsidian-dashboards` does.

I've fixed this issue for you.

PR #2080, opened 2023-06-24
index-checker plugin submitted

pavloDeshko:

## Repo URL

Link to my plugin: [https://github.com/pavloDeshko/obsidian-index-checker](https://github.com/pavloDeshko/obsidian-index-checker)

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


pavloDeshko:
> "description": "Simple and unopinionated plugin which will ensure your index (MOC) files contain all links they should contain, while leaving structure and formatting up to you."
> You should simplify your plugin description.

changed it to "Make sure your index (MOC) files contain all links they should contain." 
https://github.com/pavloDeshko/obsidian-index-checker/blob/5b8a1d0f41a7e85b5ea7fb1a57cc0314774d8e26/manifest.json#L6


> Wrap this in this.registerEvent() so it can be properly unloaded.
> this.app.workspace.on('layout-change', ()=>this.setContainers())
> Register this event handler, like the others below.

fixed both, my bad
https://github.com/pavloDeshko/obsidian-index-checker/blob/5b8a1d0f41a7e85b5ea7fb1a57cc0314774d8e26/scr/Plugin.ts#L54C4-L54C4
https://github.com/pavloDeshko/obsidian-index-checker/blob/5b8a1d0f41a7e85b5ea7fb1a57cc0314774d8e26/scr/Marker.ts#L32

> process.env.NODE_ENV=='development' && addTestCommands(this)//TODO remove?
> This will cause the plugin to break on mobile, as process is a Node.JS library.

Removed test utils imports and calls (commented out for use in development only)
https://github.com/pavloDeshko/obsidian-index-checker/blob/5b8a1d0f41a7e85b5ea7fb1a57cc0314774d8e26/scr/Plugin.ts#L15
However, it should not have breaked on mobile cause process.env.NODE_ENV is not making it to main.js as variable - it's a mere placeholder for ESBuild to be substituted for value at build time, see [esbuild.config.mjs](https://github.com/pavloDeshko/obsidian-index-checker/blob/5b8a1d0f41a7e85b5ea7fb1a57cc0314774d8e26/esbuild.config.mjs#L41C3-L41C3). Anyway, its out

> // Delay check for at least 4 seconds after last input and displays 'in process' notice to be change to result later
> You should consider to use debounce here.

[here](https://github.com/pavloDeshko/obsidian-index-checker/blob/5b8a1d0f41a7e85b5ea7fb1a57cc0314774d8e26/scr/Plugin.ts#L187)'s corrected description of this functionality
// Postpones check until DELAY (4 seconds) has passed from last user's editor input and displays 'in process' notice to be change to result later. ie if no input was recorded in last 4 seconds the check is performed immidiately. Gives Obsidian some time to update cache and resolve links.
'debounce' will not do the job here

>  Why 4 seconds?

seems to be enough for obsidian to update cache after input and so avoid degrading user experience by performing check on older versions of files.  I wish Obsidian API had some way to check if cache is up to date or trigger cache update manually, but it seems it currently doesn't :(

PR #2105, opened 2023-07-03
Add Tokei Plugin

ms3056:

## Repo URL
https://github.com/ms3056/Tokei

Link to my plugin:

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[const ClockViewType = "my-clock-view";](https://github.com/ms3056/Tokei/blob/7edfd5669df5073678a24fabea20eecc15b908a2/main.ts#L13)
Would recommend renaming this to to "tokei" to make it more unique

[private updateInterval: NodeJS.Timeout | null = null;](https://github.com/ms3056/Tokei/blob/7edfd5669df5073678a24fabea20eecc15b908a2/main.ts#L18)
This should just be of the type `number`

[const checkLayoutInterval = setInterval(async () => {](https://github.com/ms3056/Tokei/blob/7edfd5669df5073678a24fabea20eecc15b908a2/main.ts#L286-L292)
The code here is duplicated, you should remove this one,
Then you can also remove the interval and the `isViewInitialized` variable

[donateImage.style.height = "60px";](https://github.com/ms3056/Tokei/blob/7edfd5669df5073678a24fabea20eecc15b908a2/main.ts#L385-L386)
You should avoid assigning styles via JavaScript or in HTML and instead move all these styles into CSS.

ms3056:
Changes implemented in the 0.5.0 release. 

- [x] const ClockViewType = "my-clock-view";
  - Would recommend renaming this to to "tokei" to make it more unique

- [x] private updateInterval: NodeJS.Timeout | null = null;
  - This should just be of the type number

- [x] const checkLayoutInterval = setInterval(async () => {
  - The code here is duplicated, you should remove this one,
Then you can also remove the interval and the isViewInitialized variable

- [x] donateImage.style.height = "60px";
  - You should avoid assigning styles via JavaScript or in HTML and instead move all these styles into CSS.

PR #2111, opened 2023-07-04
Add Tokei plugin

ms3056:

## Repo URL
ms3056/Tokei

Link to my plugin: https://github.com/ms3056/Tokei

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[.theme-dark {](https://github.com/ms3056/Tokei/blob/6f0c78e7923a23e7a71b7aae507788072c2464da/styles.css#L1-L15)
Same feedback as with your other plugin, these variables are too generic, and you should use the build in variables from Obsidian if possible.

Your CSS class names are also a bit too generic.

[dateFormatSetting.settingEl.style.borderTop = "none";](https://github.com/ms3056/Tokei/blob/6f0c78e7923a23e7a71b7aae507788072c2464da/main.ts#L535-L537)

[new Notice("Maximum timezone entries added.");](https://github.com/ms3056/Tokei/blob/6f0c78e7923a23e7a71b7aae507788072c2464da/main.ts#L876)
Why is there a limit here?

ms3056:
I have a couple of questions prior to making the changes:

[.theme-dark {](https://github.com/ms3056/Tokei/blob/6f0c78e7923a23e7a71b7aae507788072c2464da/styles.css#L1-L15)
Same feedback as with your other plugin, these variables are too generic, and you should use the build in variables from Obsidian if possible.

Your CSS class names are also a bit too generic.

Question: I have changed the variable names to be specific to the plugin. I have also included code to change the colors via Style Settings. I do not agree, however, with using the standard color variables. Many other plugins drive the colors this way - one is OZ Calendar. This is a design choice, but allowing flexibility in Style Settings is a reasonable compromise, I think. 

```css
/* @settings

name: Tokei
id: tokei
settings:
    -
        id: tokei-background
        title: Background Color
        description: Set the color of the timezone block background
        type: variable-themed-color
        opacity: true
        format: hex
        default-light: '#d7dff0'
        default-dark: '#1c2732'

    -
        id: tokei-main-text
        title: Main Text Color
        description: Set the color of the main text 
        type: variable-themed-color
        opacity: true
        format: hex
        default-dark: '#aabed8'
        default-light: '#497fa0'
        ```
Is this 

[new Notice("Maximum timezone entries added.");](https://github.com/ms3056/Tokei/blob/6f0c78e7923a23e7a71b7aae507788072c2464da/main.ts#L876)
Why is there a limit here?

**Question:** This is a design choice. More than five entries take up too much space. Should users decide they want more than 5, I will consider it. I am not inclined to change this without user feedback. My test users have said 2 or 3 is all they use. 

Please let me know if these compromises are acceptable and I update both Tokei and Tenki and resubmit.

ms3056:
changes made - please review.

joethei:
[updateInterval: NodeJS.Timeout | null = null;](https://github.com/ms3056/Tokei/blob/512065798dc47f0cfa6a1d5820b972e81df1161a/main.ts#L265)
Use a regular number here and use `window.setTimeout` everywhere you use this (same as in the other plugin)

[containerEl.createEl("h1", { text: "Tokei" });](https://github.com/ms3056/Tokei/blob/512065798dc47f0cfa6a1d5820b972e81df1161a/main.ts#L381)
[Don't include a header with your plugin name in the settings ](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Only+use+headings+under+settings+if+you+have+more+than+one+section.)

I would also recommend to move the donation information at the bottom.


The reason I recommend using the Obsidian CSS variables as a default is so that the plugin looks good in as many themes as possible out of the box, without having to install Style settings or write a snippet.

ms3056:
#2111 

[updateInterval: NodeJS.Timeout | null = null;](https://github.com/ms3056/Tokei/blob/512065798dc47f0cfa6a1d5820b972e81df1161a/main.ts#L265)
Use a regular number here and use window.setTimeout everywhere you use this (same as in the other plugin)

Response: code change implemented in 0.6.5

[containerEl.createEl("h1", { text: "Tokei" });](https://github.com/ms3056/Tokei/blob/512065798dc47f0cfa6a1d5820b972e81df1161a/main.ts#L381)
[Don't include a header with your plugin name in the settings](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Only+use+headings+under+settings+if+you+have+more+than+one+section.)

I would also recommend to move the donation information at the bottom.

Response: code change implemented in 0.6.5

liamcain:
- [updateInterval](https://github.com/ms3056/Tokei/blob/08eed63ef614e238bf2202a3238e35507899585d/main.ts#L322) I think it makes more sense for the updateInterval to live on the View class. That way when the view is closed, you don't need to worry about the interval.

PR #2156, opened 2023-07-12
Add Simple Theme

diegoeis:
New theme

The Simple is a theme for Obsidian that prioritizes reading and writing. 

It has minimal customizations and a simple design to avoid distractions and noise while taking notes and writing long-form texts.


diegoeis (commented):
moving to the last entry

PR #2162, opened 2023-07-15
Update community-plugins.json

Cyfine:

## Repo URL

Link to my plugin: https://github.com/Cyfine/EmoTagsTitler

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
Please remove the `main.js` from the repo and add it to your .gitignore - compiled release files shouldn't get mixed with source code in your commits and should go to the releases files instead.

[import {join} from 'path';](https://github.com/Cyfine/EmoTagsTitler/blob/e3491df852d5af29d6cb73d2dab5a527ef85bf91/main.ts#L2)
These imports are only available from a nodeJS runtime, which means that this will throw errors on Mobile (that's a problem since you have `isDesktopOnly` marked as `false` in your manifest.json file.

[this.app.fileManager.renameFile(file, join(dir, newNoteTitle + '.md'));](https://github.com/Cyfine/EmoTagsTitler/blob/e3491df852d5af29d6cb73d2dab5a527ef85bf91/main.ts#L38)
you can just join them with a slash `/´


[const tags = this.app.metadataCache.getFileCache(file).tags;](https://github.com/Cyfine/EmoTagsTitler/blob/e3491df852d5af29d6cb73d2dab5a527ef85bf91/main.ts#L10)


One thing you might want to mention in your readme:
Dropbox does not like emoji's, any files with emoji in the filename won't be synced.

Cyfine:
*  main.js removed from repo 
*  main.ts refactored 
    *  avoid using path.join
    * the functions moved from global scope to plugin class to avoid referencing app from global context 
* readme.md updated to remind Dropbox users 
* release
    * release tag fast forwarded to the latest commit
    * binaries updated

PR #2174, opened 2023-07-18
Add plugin Lovely-Mindmap

xincan1949:

## Repo URL

Link to my plugin: https://github.com/xincan1949/lovely-mindmap

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -7159,5 +7159,12 @@
     "author": "Onur Ayçiçek",
     "description": "Search movies info and trailers.",
     "repo": "onuraycicek/obsidian-movie"
+  },
+  {
+    "id": "lovely-mindmap",
+    "name": "Lovely-Mindmap",
+    "author": "shaun",
+    "description": "Build your own knowledge graph with smiles :-)",
+    "repo": "xincan1949/lovely-mindmap"
   }
 ]


xincan1949:
Hi @joethei, the source code of #2128 is made public. 
Hope to be adopted soon~

joethei:
[createCanvasInstance() {](https://github.com/xincan1949/lovely-mindmap/blob/ce88cd6b0765dcf3bf9224015bfc0fc56ca1e699/src/main.ts#L25-L37)
Don't do this, just use a regular method to get the current canvas instead.
Your current code waits for a canvas file to be opened, which is horribly inefficient.
It also only works for the very first canvas a user opens in a given session.

[function debounce(delay: number = 100): MethodDecorator {](https://github.com/xincan1949/lovely-mindmap/blob/ce88cd6b0765dcf3bf9224015bfc0fc56ca1e699/src/tool/index.ts#L1-L26)
Obsidian already exports a function of the same name,

[console.log('this:\n', this)](https://github.com/xincan1949/lovely-mindmap/blob/ce88cd6b0765dcf3bf9224015bfc0fc56ca1e699/src/module/keymap.ts#L26-L29)

xincan1949:
Hey @joethei , thanks for your carefully review, here are my question and answer

1. Is there any API can get canvas instance?

I do it because sometimes when users create canvas, the canvas may not create in first time, so polling is required

2. `@debounce` is not a function, it's decorator in JavaScript Class. 
 
Different with normal debounce(execute the event after N seconds), the `@debounce` decorator execute a event immediately and rest for N seconds. Maybe I should change the decorator name from `@debounce` to `@rest` 🤔

liamcain:
Added here https://github.com/obsidianmd/obsidian-releases/commit/f8a30a360be5d17dae97fe784a10297680764192!

PR #2197, opened 2023-07-24
Add plugin `modules`

polyipseity:

(Sorry if submitting two plugins at about the same time is too much workload, I won't be making any more for the time being. 😅 (though most code is the same as the last plugin because I use a template to standardize the plugins))

## Repo URL

Link to my plugin: https://github.com/polyipseity/obsidian-modules

## Release Checklist
  - [x] Windows
  - [x] macOS
  - [x] Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


polyipseity:
> What sets your plugin apart from the _CustomJS_ plugin?

The most significant ones for basic users are:
- Use Markdown files with JS code blocks as JS modules, which enables editing inside Obsidian with syntax highlighting and Markdown comments surrounding the code blocks.
- Not requiring configuration (adding the JS files/folders to settings).

For advanced users, that would also mean the module syntax is very familiar (essentially the same as `require` and dynamic `import`).

> There is quite a bit of boilerplate, and some code that looks like it's not even used. Consider removing everything that's not used. The `statusBarHider` for example looks like it's not needed.

Removed `statusBarHider`, `icons.ts`, and unused Svelte-related dependencies.

PR #2204, opened 2023-07-26
Add Fantasy name generator plugin

Lukewh:

## Repo URL

Link to my plugin: https://github.com/lukewh/fantasy-name

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


ericaxu:
Hi, have you looked at the Fantasy Content Generator plugin? Is your plugin significantly different?

Lukewh:
> Hi, have you looked at the Fantasy Content Generator plugin? Is your plugin significantly different?

Hi! I have looked at the Fantasy Content Generator plugin. On the surface this plugin does indeed seem very similar, however I find the Fantasy Content Generator plugin too heavy handed when I need a quick name, so this plugin provides a no-configuration option :).

joethei:
Please remove the `main.js` from the repo and add it to your .gitignore - compiled release files shouldn't get mixed with source code in your commits and should go to the releases files instead.

[const resp = await fetch("https://fantasyname.lukewh.com?family=t");](https://github.com/Lukewh/fantasy-name/blob/e1b20df9cf96678d7494f48d5eedd6559cfc7944/index.ts#L16)

I personally don't see why the name generation code needs to be hosted on a server.

Lukewh:
> Please remove the `main.js` from the repo and add it to your .gitignore - compiled release files shouldn't get mixed with source code in your commits and should go to the releases files instead.
> 
> [const resp = await fetch("https://fantasyname.lukewh.com?family=t");](https://github.com/Lukewh/fantasy-name/blob/e1b20df9cf96678d7494f48d5eedd6559cfc7944/index.ts#L16) I would recommend using the `requestUrl` function from the Obsidian API instead, it will handle some things like CORS automatically.
> 
> I personally don't see why the name generation code needs to be hosted on a server.

Thanks for the review and feedback, I've made the changes you suggested :)!

I'm using an API that I created as it's used by people that don't use Obsidian and honestly, I don't want to duplicate my effort :D!

Lukewh:
Resolved conflict :rocket:

PR #2208, opened 2023-07-26
add ics plugin

muness:

## Repo URL

Link to my plugin: https://github.com/muness/obsidian-ics

## Release Checklist
  - [X]  Windows
  - [X]  macOS
  - [x]  Linux
  - [X]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


joethei:
["description": "This is a ICS plugin for Obsidian.",](https://github.com/muness/obsidian-ics/blob/5b0847186cf9a040c3e49421d69fb007198151ae/manifest.json#L6)
We have  published a guide for plugin descriptions to follow here: https://docs.obsidian.md/Plugins/Releasing/Submission+requirements+for+plugins#Keep+plugin+descriptions+short+and+simple.
Avoid including sentences like "This is a plugin for Obsidian" in your description.

["node-ical": "^0.16.1",](https://github.com/muness/obsidian-ics/blob/5b0847186cf9a040c3e49421d69fb007198151ae/package.json#L16)
This library is only available from a nodeJS runtime, which means that this will throw errors on Mobile (that's a problem since you have `isDesktopOnly` marked as `false` in your manifest.json file, or am I missing something?

[callback: async () => {](https://github.com/muness/obsidian-ics/blob/5b0847186cf9a040c3e49421d69fb007198151ae/src/main.ts#L85-L95)
You should use a editor callback here, that way the command will only be available if there is a editor available: <https://docs.obsidian.md/Plugins/User+interface/Commands#Editor+commands>

[const rangeStart = window.moment(dayToMatch);](https://github.com/muness/obsidian-ics/blob/5b0847186cf9a040c3e49421d69fb007198151ae/src/icalUtils.ts#L21)
Would be nicer to use `import {moment} from "obsidian";` here.

[containerEl.createEl("h2", {](https://github.com/muness/obsidian-ics/blob/5b0847186cf9a040c3e49421d69fb007198151ae/src/settings/ICSSettingsTab.ts#L46-L49)


[.setName("Add New")](https://github.com/muness/obsidian-ics/blob/5b0847186cf9a040c3e49421d69fb007198151ae/src/settings/ICSSettingsTab.ts#L54)

muness:
Thanks for the feedback, @joethei ! I'll work on it, hopefully this weekend.

One thing I am also confused by is this:

> ["node-ical": "^0.16.1",](https://github.com/muness/obsidian-ics/blob/5b0847186cf9a040c3e49421d69fb007198151ae/package.json#L16)
> This library is only available from a nodeJS runtime, which means that this will throw errors on Mobile (that's a problem since you have isDesktopOnly marked as false in your manifest.json file, or am I missing something?

I expected the same issue too, but I've tested on my iPhone, iPad and @TopherMan tested it on [Android](https://github.com/muness/obsidian-ics/pull/46#issue-1818908445) and it works fine.

What do you suggest? Flip it back to desktop only? Or just go with it?

muness:
Applied the feedback suggested in https://github.com/muness/obsidian-ics/pull/52 . The changes are in https://github.com/muness/obsidian-ics/releases/tag/1.0.1 .

Note that I did not disable mobile since the plugin works there. I wonder if this is because the build is [configured](https://github.com/muness/obsidian-ics/blob/5b0847186cf9a040c3e49421d69fb007198151ae/esbuild.config.mjs#L19) to [bundle](https://esbuild.github.io/api/#bundle) dependencies and so they're inlined in main.js?

muness:
Hi @joethei or anyone else, any chance my updates suffice? Thank you!

PR #2228, opened 2023-08-01
Add Zotero Sync plugin

frthjf:

## Repo URL

Link to my plugin: https://github.com/frthjf/obsidian-zotero-sync-client

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[export default class MyPlugin](https://github.com/frthjf/obsidian-zotero-sync-client/blob/125cd2d30236c2f1376a84f1fa162644189f244e/main.ts#L109)
Rename this class.

[if (this.settings.show_ribbon_icon) {](https://github.com/frthjf/obsidian-zotero-sync-client/blob/125cd2d30236c2f1376a84f1fa162644189f244e/main.ts#L191)

[await this.app.vault.rename(fn, newPath)](https://github.com/frthjf/obsidian-zotero-sync-client/blob/125cd2d30236c2f1376a84f1fa162644189f244e/main.ts#L402)
I would recommend calling `await this.app.filemanager.renameFile(fn, newPath);` instead.
That will make sure all links to this file are updated.

frthjf:
Thanks for the great feedback!

I've addressed these issues now and released a new version (1.0.7).

frthjf:
I've released a fixed version.

liamcain:
- [borderColor](https://github.com/frthjf/obsidian-zotero-sync-client/blob/8f3ae4339d4501d7b147f92b602476f9aa610029/main.ts#L795) You should avoid assigning styles via javascript. Instead, just add a class to the element and move all these styles into CSS so that they are more easily adaptable by themes.

PR #2248, opened 2023-08-05
plugin: add PKMer plugin to obsidian

windily-cloud:

## Repo URL

Link to my plugin: [Pkmer-Obsidian](https://github.com/PKM-er/Pkmer-Obsidian)

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -7542,7 +7542,7 @@
     "id": "pkmer",
     "name": "PKMer",
     "author": "PKMer(windily-cloud&cumany)",
-    "description": "A marketplace of third-party plugins and themes developed by Obsidian enthusiasts in China to enhance the experience of using Obsidian in China.",
+    "description": "A marketplace of third-party plugins and themes developed by Obsidian enthusiasts in China. Aim to enhance the experience of using Obsidian in China.",
     "repo": "PKM-er/Pkmer-Obsidian"
   }
 ]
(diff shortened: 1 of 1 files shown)


windily-cloud:
Looks like I don't have to worry about the error.........😣

image

joethei:
The validation needs to pass at least once.

windily-cloud:
> The validation needs to pass at least once.

Okay, I guess I did something right, ha ha😊！

joethei:
after careful consideration we have decided not to allow this plugin into the official community plugins list. The reason why is that it creates a way for developers to bypass the [developer policies](https://docs.obsidian.md/Developer+policies).

Users are free to install any plugin manually. The purpose of the community directory is to enforce certain precautions around security, disclosures, licensing, etc. The PKMer marketplace creates a loophole that we are not comfortable with. Currently PKMer crawls the official list of approved plugins/themes, but there's no way to ensure that unapproved plugins/themes cannot be added.

In the future we hope to address some of the limitations with the official plugin/theme directory that your plugin aims to solve. In the meantime, users can always install PKMer manually if they want to.

Sorry to be the bearer of bad news. I hope our reasoning makes sense to you.

PR #2263, opened 2023-08-07
Add Handwritten Notes

FBarrca:

## Repo URL

Link to my plugin:
https://github.com/FBarrca/obsidian-handwritten-notes/

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [x]  Android 
  - [ ]  iOS 
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["js": "main.js"](https://github.com/FBarrca/obsidian-handwritten-notes/blob/c0102fad52a1dae12679322569e7c1706df1142b/manifest.json#L10)
This is not a part of the manifest spec.

[console.log(`Using relative paths: ${this.settings.useRelativePaths}`);](https://github.com/FBarrca/obsidian-handwritten-notes/blob/c0102fad52a1dae12679322569e7c1706df1142b/src/main.ts#L79)

[MyPluginSettingTab](https://github.com/FBarrca/obsidian-handwritten-notes/blob/c0102fad52a1dae12679322569e7c1706df1142b/src/settings.ts#L65)
Rename this class.

[private readonly plugin: any;](https://github.com/FBarrca/obsidian-handwritten-notes/blob/c0102fad52a1dae12679322569e7c1706df1142b/src/settings.ts#L66)
`any` should be avoided as much as possible use `NotePDF` as the type here.

[button.buttonEl.style.padding = "4px 6px";](https://github.com/FBarrca/obsidian-handwritten-notes/blob/c0102fad52a1dae12679322569e7c1706df1142b/src/utils/utils.ts#L91-L93)

[export async function fileExists(app: App, path: string): Promise<boolean> {](https://github.com/FBarrca/obsidian-handwritten-notes/blob/c0102fad52a1dae12679322569e7c1706df1142b/src/utils/utils.ts#L167-L179)
Check if `app.vault.getAbstractFileByPath` returns something instead, this has the advantage that it's cached.

[contentEl.createEl("h1", { text: "Create New note" });](https://github.com/FBarrca/obsidian-handwritten-notes/blob/c0102fad52a1dae12679322569e7c1706df1142b/src/dialogs/CreatorModal.ts#L25)

FBarrca:
Thanks joethei! 
I fixed all the issues you pointed out except the:
-  [export async function fileExists(app: App, path: string): Promise {](https://github.com/FBarrca/obsidian-handwritten-notes/blob/c0102fad52a1dae12679322569e7c1706df1142b/src/utils/utils.ts#L167-L179) 
Check if app.vault.getAbstractFileByPath returns something instead, this has the advantage that it's cached.

The files I want to check are inside the .obsidian folder (a folder inside the plugin's folder). I initially tried getAbstractFileByPath as you are suggesting, but got an error where it said it couldn't cache. I supposed it had to do with the .obsidian folder being a hidden folder, so I bodged this solution.

If you can think of a better method to call, please let me know, I'm pretty new to the Obsidian API. 

My current code works and the function isn't called excessively, so I think it could be left as it is if we dont find a better solution.

Thanks again <3

liamcain:
- [platformContainer.style.display](https://github.com/FBarrca/obsidian-handwritten-notes/blob/96e74fa4619bd533c620e6eb007571066c138eb8/src/dialogs/WelcomeModal.ts#L54) Maybe move all these styles into CSS
- [this.plugin](https://github.com/FBarrca/obsidian-handwritten-notes/blob/96e74fa4619bd533c620e6eb007571066c138eb8/src/dialogs/WelcomeModal.ts#L123) this is not a good Component to use for managing the lifecycle. You should create a new Component and construct and load it within your Modal class. That way the markdown renderer properly unloads when the modal is closed.

PR #2283, opened 2023-08-12
Update community-plugins.json

Ebonsignori:

## Repo URL

Link to my plugin: https://github.com/Ebonsignori/obsidian-auto-journal/tree/main

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -7523,5 +7523,12 @@
     "author": "ddalexb",
     "description": "Quickly fuzzysearch notes, cards and their content and shift focus to them within the currently opened canvas.",
     "repo": "ddalexb/obsidian-simple-canvasearch"
+  },
+    {
+    "id": "auto-journal",
+    "name": "Auto Journal",
+    "author": "Evan Bonsignori",
+    "description": "Opinionated journaling automation like daily notes but with backfills for the days when Obsidian wasn't opened",
+    "repo": "Ebonsignori/obsidian-auto-journal"
   }
 ]


joethei:
[templateContents = await this.app.vault.read(templateFile as TFile);](https://github.com/Ebonsignori/obsidian-auto-journal/blob/31ae981f64da0e438d4e01533f73c804ad7a261f/src/core.ts#L50)
Don't cast blindly, check if `templateFile instanceof TFile` first.

[path.join(](https://github.com/Ebonsignori/obsidian-auto-journal/blob/31ae981f64da0e438d4e01533f73c804ad7a261f/src/core.ts#L84-L87)
Just join them with a simple slash, `/`, no need to import a library for this.

[path.sep](https://github.com/Ebonsignori/obsidian-auto-journal/blob/31ae981f64da0e438d4e01533f73c804ad7a261f/src/core.ts#L273)
Always a slash when using the vault API.

[descEl.append(](https://github.com/Ebonsignori/obsidian-auto-journal/blob/31ae981f64da0e438d4e01533f73c804ad7a261f/src/settings/settings.ts#L143-L152)
Including the plugin name & description in the settings is not recommended, see https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Only+use+headings+under+settings+if+you+have+more+than+one+section.

[// Code derived from https://github.com/farux/obsidian-auto-note-mover](https://github.com/Ebonsignori/obsidian-auto-journal/blob/31ae981f64da0e438d4e01533f73c804ad7a261f/src/settings/file-suggest.ts#L1)
That's not the original source for that code.

Ebonsignori:
Thanks as always @joethei 

Fixed in version [1.2.2](https://github.com/Ebonsignori/obsidian-auto-journal/releases/tag/1.2.2)

https://github.com/Ebonsignori/obsidian-auto-journal/compare/1.2.1...1.2.2

liamcain:
- [Manually Trigger](https://github.com/Ebonsignori/obsidian-auto-journal/blob/caee174cd5970d027cab3893d5dea84901cbb528/src/main.ts#L21) We recommend that all commands use [Sentence case](https://en.wiktionary.org/wiki/sentence_case) for their names. This maximizes readability in the command palette.
- [automaticallyRun](https://github.com/Ebonsignori/obsidian-auto-journal/blob/caee174cd5970d027cab3893d5dea84901cbb528/src/main.ts#L28) you can have the if statement outside the `onLayoutReady` call to avoid creating the callback if you don't need it

PR #2295, opened 2023-08-14
Add Math Booster plugin

RyotaUshio:

## Repo URL

Link to my plugin: https://github.com/RyotaUshio/obsidian-math-booster

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -7523,5 +7523,12 @@
     "author": "Overraddit",
     "description": "Effortlessly generate context-aware answers from ChatGPT within Obsidian, while maintaining a visually clear and organized conversation history.",
     "repo": "OverRaddit/ZettelGPT"
-  }
+  },
+  {
+    "id": "math-booster",
+    "name": "Math Booster",
+    "author": "Ryota Ushio",
+    "description": "Introduce LaTeX-like mathematic note-taking features into Obsidian",
+    "repo": "RyotaUshio/obsidian-math-booster"
+  }  
 ]


RyotaUshio:
Note: `styles.css` is generated from `styles/*.css` using `make_styles.py`, so probably I should include `styles.css` in `.gitignore`. But I will keep it in the repo for the time being to make the review easier.

RyotaUshio:
Update: I modified the description in `manifest.json`, so I updated `community-plugins.json` accordingly.

RyotaUshio:
Re-updated the description.

liamcain:
- [assertDataview](https://github.com/RyotaUshio/obsidian-math-booster/blob/129d9370d191e3b1d6806fcbd013300c77c04c90/src/main.ts#L44) assertDataview is returning a boolean value but you're not using it. You probably want to avoid running additional code is dataview is not enabled, right?
- [cachedRead](https://github.com/RyotaUshio/obsidian-math-booster/blob/129d9370d191e3b1d6806fcbd013300c77c04c90/src/file_io.ts#L103) if this is for IO you might want to use read instead of cachedRead to get the latest version of the file

RyotaUshio:
@liamcain Thank you a lot for the feedback & merging!
I will update the code according to your comments.

PR #2301, opened 2023-08-16
Changed name of theme Aura to Wolf

ashwinjadhav818:
# Changed the name of the theme Aura to Wolf
- [x] Changed the theme name
- [x] Changed the theme repo location

### DIFF
--- a/community-css-themes.json
+++ b/community-css-themes.json
@@ -997,9 +997,9 @@
     "modes": ["dark"]
   },
   {
-    "name": "Aura",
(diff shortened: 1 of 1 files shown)


joethei:
Are you sure that you want to change the name?
All current users will lose the ability to update and the download count will be reset

ashwinjadhav818:
Yes, I am sure about that

ashwinjadhav818:
Anyways, changed my mind, I'll rather keep it the same.

PR #2305, opened 2023-08-17
add the at-people plugin

saibotsivad:

## Repo URL

Link to my plugin: https://github.com/saibotsivad/obsidian-at-people

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
  - ~~[ ] `styles.css` _(optional)_~~


Ebonsignori:
@ericaxu @joethei [Here y'all asked if this was significantly different from another plugin](https://github.com/Ebonsignori/obsidian-at-symbol-linking) and this plugin is a near copy paste of [my at symbol linking](https://github.com/Ebonsignori/obsidian-at-symbol-linking) plugin. It looks as if the author created it after mine around the time I shared it to the forum. It seems like they are sourcing content from my repo too :/

It's all open source so it's a bit confusing why they would do this. If there were any features different than mine I'd understand, but it almost seems like they raced to compete with mine which is silly

Ebonsignori:
Relevant thread https://forum.obsidian.md/t/mention-support-for-people-identities/6163/71?u=ebonsignori

joethei:
It looks like both have the same goal, but the implementation is different.
Your plugin is more complex than this one.
I don't see any copied code.

PR #2306, opened 2023-08-17
Add Pickly PageBlend plugin

dmitrichev:

## Repo URL

Link to my plugin: https://github.com/dmitrichev/pickly-page-blend

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


dmitrichev:
Fixed

joethei:
[async install() {](https://github.com/dmitrichev/pickly-page-blend/blob/8bdff8654b249f9336aff489cb68c413a3d3e9d7/main.ts#L100-L109)
Wouldn't it make more sense to assign the UUID only when a note published?

[const resp = await fetch(`${this.apiBaseUrl}/publish`, {](https://github.com/dmitrichev/pickly-page-blend/blob/8bdff8654b249f9336aff489cb68c413a3d3e9d7/src/publisher.ts#L66-L69)

dmitrichev:
Hi, @joethei 

Thank you for your comment. I've fixed the assignment of uuids for files at the time of publishing. These changes are available in the `master` branch, the release `1.0.1` is ready.

As for sending a request via fetch, I would like to leave it as is. I've looked into the `requestUrl` function. The problem is that it expects the `body` parameter as an `ArrayBuffer` while I'm using `FormData`. As far as I know, there is no general method for converting a `FormData` to an `ArrayBuffer` or other byte array. Switching to using the `requestUrl` function will force me to do some tricks to properly assemble the `multipart/form-data` with all the `boundary` and stuff. I know about the problem of CORS requests, but the service through which the notes are published is also mine. And it is already configured in such a way as to send the correct CORS headers. So, if you don't mind, I would like to continue using `fetch`. 🙂

liamcain:
- [modifiers: ['Mod', 'Shift'],](https://github.com/dmitrichev/pickly-page-blend/blob/ad8535aeceea05195782bf27fb60991d8eb1d74a/main.ts#L31) While there's no hard rule against it, we recommend against providing a default hotkey when possible. The reason being that it's easy to pick a hotkey that a user already has configured and they could get confused when that key doesn't do what they expect. Also, it's hard choosing a safe default hotkey that's available for all operating systems.
- [file](https://github.com/dmitrichev/pickly-page-blend/blob/ad8535aeceea05195782bf27fb60991d8eb1d74a/main.ts#L45) 'delete' is passed a TAbstractFile so you probably want to check if its an instanceof TFile
- [pickly-page-blend](https://github.com/dmitrichev/pickly-page-blend/blob/ad8535aeceea05195782bf27fb60991d8eb1d74a/main.ts#L27) Hotkey IDs are already prefixed with the plugin ID, no need to prefix it yourself here

PR #2316, opened 2023-08-19
Add Todo.txt Codeblocks Plugin

benjamonnguyen:

## Repo URL

Link to my plugin: https://github.com/benjamonnguyen/obsidian-todotxt-codeblocks

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[import { randomUUID } from "crypto";](https://github.com/benjamonnguyen/obsidian-todotxt-codeblocks/blob/f4456a61a10c703d9e5a42aa750d433654c1efbc/src/model/TodoList.ts#L3)

benjamonnguyen:
@joethei I've replaced references to randomUUID from "crypto" as well as direct mutations of innerHTML and outerHTML

benjamonnguyen:
Sorry if this is not proper but bumping in case this has fallen through the cracks. Thank you!

PR #2323, opened 2023-08-20
Add birthday-tracker plugin

Raboro:

## Repo URL

Link to my plugin: https://github.com/Raboro/Obsidian-Birthday-Tracker-Plugin

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -7747,5 +7747,12 @@
     "author": "Lucy Gschwantner",
     "description": "Read any ArchWiki page directly in Obsidian",
     "repo": "Jackboxx/archwiki-obsidian"
+  },
+  {
+    "id": "easy-bake",
+    "name": "Easy Bake",
+    "author": "mgmeyers",
+    "description": "Easily compile many Obsidian notes down to a single file.",
+    "repo": "mgmeyers/obsidian-easy-bake"
   }
 ]


joethei:
["description": "This plugin allows you to keep track of all birthdays of your family and friends.",](https://github.com/Raboro/Obsidian-Birthday-Tracker-Plugin/blob/528b70d17f1eb36b34a2f5b83c59edec08095bb5/manifest.json#L6)
Avoid including sentences like "This is a plugin that does xy" in your description.

[id: 'birthday-tracker-track-birthdays',](https://github.com/Raboro/Obsidian-Birthday-Tracker-Plugin/blob/528b70d17f1eb36b34a2f5b83c59edec08095bb5/src/main.ts#L20)

[this.app.workspace.detachLeavesOfType(BIRTHDAY_TRACKER_VIEW_TYPE);](https://github.com/Raboro/Obsidian-Birthday-Tracker-Plugin/blob/528b70d17f1eb36b34a2f5b83c59edec08095bb5/src/main.ts#L42)
Instead of detaching I would recommend to check first If a view already exist and reveal that instead, this is related to: https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Don't+detach+leaves+in+%60onunload%60> 

[for (const file of this.app.vault.getFiles()) {](https://github.com/Raboro/Obsidian-Birthday-Tracker-Plugin/blob/528b70d17f1eb36b34a2f5b83c59edec08095bb5/src/main.ts#L49-L54)
Make the file that contains birthday data configurable.
The way it's implemented currently is very bad for performance in bigger vaults.

Raboro:
Thank you very much for the quick feedback. I fixed the description as suggested in the guidelines, removed the plugin ID from the command and checked if the view already exists instead of separating it, already fixed.

For my poor performance of the loop iterating over all files, I would add a setting to add the path of the file with the birthday data and use ``app.vault.getAbstractFileByPath()`` instead as suggested in the guidelines.

And for the birthday file configuration. I have plans to make multiple data collections possible. First as now, second by a table. Maybe metadata and from multiple files later.

Does it fit like this?

PR #2330, opened 2023-08-22
[NEW] Adds czottmann/obsidian-mononote

czottmann:

## Repo URL

Link to my plugin: https://github.com/czottmann/obsidian-mononote

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[${PLUGIN_INFO.pluginVersion}](https://github.com/czottmann/obsidian-mononote/blob/07196e805e90b127a972ddc61297d81b02e16595/src/main.ts#L10)
You can access this from `this.manifest.version`

czottmann:
> [${PLUGIN_INFO.pluginVersion}](https://github.com/czottmann/obsidian-mononote/blob/07196e805e90b127a972ddc61297d81b02e16595/src/main.ts#L10) You can access this from `this.manifest.version`

Thanks @joethei, good to know! In the context of this review, is this a showstopper or information?

joethei:
In this case, it's information.

czottmann:
@liamcain @joethei Is there anything else I can help you with regarding this review? It's been a month since our last contact…

liamcain:
- [activeLeaf](https://github.com/czottmann/obsidian-mononote/blob/07196e805e90b127a972ddc61297d81b02e16595/src/main.ts#L35) this seems to be making an implicit assumption that the file-open event was fired from the activeLeaf. That's not necessarily true. You can open files in the background
- I'm not sure what the expected behavior is if you follow a link [[someFileThatIsAlreadyOpen]]. Should this cause the current tab to be closed and the old tab to be focused?
- [leaf != activeLeaf](https://github.com/czottmann/obsidian-mononote/blob/07196e805e90b127a972ddc61297d81b02e16595/src/main.ts#L50) `!==`

liamcain:
Since this works okay for the base case, I'm going to merge this in, but I think you should probably reconsider the approach for handling the more complex cases.

czottmann:
@liamcain Thanks for the merge!

> * [activeLeaf](https://github.com/czottmann/obsidian-mononote/blob/07196e805e90b127a972ddc61297d81b02e16595/src/main.ts#L35) this seems to be making an implicit assumption that the file-open event was fired from the activeLeaf. That's not necessarily true. You can open files in the background

How would I get notified about those?

> * I'm not sure what the expected behavior is if you follow a link [[someFileThatIsAlreadyOpen]]. Should this cause the current tab to be closed and the old tab to be focused?

Exactly. That's the whole reason for this plugin. If the file's already open, don't open it in another tab, just focus the existing one.

> * [leaf != activeLeaf](https://github.com/czottmann/obsidian-mononote/blob/07196e805e90b127a972ddc61297d81b02e16595/src/main.ts#L50) `!==`

Typo, will be fixed, thanks! No idea why VS Code didn't complain about that weak check, it's set to do it.

liamcain:
> How would I get notified about those

image

Try with this setting disabled, the file-open event should still fire, but the file won't be the activeLeaf. You'll need to handle that case separately.

> Exactly. That's the whole reason for this plugin. If the file's already open, don't open it in another tab, just focus the existing one.

But in that case, I wouldn't want my current tab to be closed. I would expect it to just refocus the existing tab and my old tab would still be in the same position.

czottmann:
> Try with this setting disabled, the file-open event should still fire, but the file won't be the activeLeaf. You'll need to handle that case separately.

Will try that, thanks!

> But in that case, I wouldn't want my current tab to be closed. I would expect it to just refocus the existing tab and my old tab would still be in the same position.

As an example, let's assume you're coming in via an XCU call, the call opens a note, but that note is already open in another tab. Mononote gets the "file opened" notice for the new tab, so it'll check the list of open tabs, it will find the existing tab from earlier, focus that and close the new opened tab.

liamcain:
What's XCU?

czottmann:
Ah sorry, too many acronyms 😅 XCU = x-callback-url, i.e. `obsidian://`.

PR #2337, opened 2023-08-23
Adding Serika Theme

Warrobot10:
# I am submitting a new Community Theme

## Repo URL

Link to my theme: https://github.com/Warrobot10/Serika-for-obsidian

## Theme checklist

- [x] My repo contains all required files (please *do not* add them to this `obsidian-releases` repo).
  - [x] `manifest.json`
  - [x] `theme.css`
  - [x] The screenshot file (16:9 aspect ratio, recommended size is 512px by 288px for fast loading).
- [x] I have indicated which modes (dark, light, or both) are compatible with my theme.
- [x] I have read the tips in https://docs.obsidian.md/Themes/App+themes/Theme+guidelines and have self-reviewed my theme to avoid these common pitfalls.
- [x] My project respects and is compatible with the original license of any code from other themes that I'm using. I have given proper attribution to these other themes in my `README.md`.


joethei:
[--version: "You are using serika v1.0.0 \a\ If you find any issues, make a discussion at the serika github.\a\ github.com/serika"](https://github.com/Warrobot10/Serika-for-obsidian/blob/01ed4ca0c6a2d3c155deb465c665b2ac2f7597da/theme.css#L24C1-L24C130)
The link goes to some other users github profile.
Also, what is that variable used for?

You should consider making a better screenshot, showing the different modes next to each other instead with proper rotation, something like this for example: 


screenshot-hybrid

Warrobot10:
Uhm, so for the version thing, I was looking at the theme sanctum, to see different variables I could use and how they would be used, and I saw that. But sanctum relies on some community plugins to extend its functionality, so that is probably a plugin variable. I will remove that. I love the way that the example screenshot looks, but I have no clue on how to do image editing.
Any tips?
Thanks for your help!

joethei:
The version variable is used inside of the sanctum theme in a `content`: https://github.com/jdanielmourao/obsidian-sanctum/blob/ac69e5992a66d2aeabb30d8c2d90c636d155fc25/obsidian.css#L2457C27-L2457C27
You can find documentation about all the different CSS variables here: <https://docs.obsidian.md>

I am not a designer, but the way I would create such a image is this:
- take a screenshot of your theme in both light and dark mode
- use a specialized online tool to create a side-by-side image
You could also use a proper image editor, but that might require a bit of experience.

Warrobot10:
Done, added a better screenshot.

PR #2339, opened 2023-08-23
Add Unitade plugin

Falcion:

## Repo URL

Link to my plugin: https://github.com/Falcion/UnitadeOBSIDIAN

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[containerEl.createEl('h2', { text: 'Unitade\'s Plugin Settings' });](https://github.com/Falcion/UnitadeOBSIDIAN/blob/6f0f210e42c76ce7dfb7a8914cb8568e9a49401d/main.ts#L57)

Falcion:
> [containerEl.createEl('h2', { text: 'Unitade's Plugin Settings' });](https://github.com/Falcion/UnitadeOBSIDIAN/blob/6f0f210e42c76ce7dfb7a8914cb8568e9a49401d/main.ts#L57) [Only use headings under settings if you have more than one section. ](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Only+use+headings+under+settings+if+you+have+more+than+one+section.)

Made! Now in the settings there are just options without headers.

joethei:
One thing you should look into is what happens when you add the extension of a binary filetype, like png for example, and see if that causes problems.

Falcion:
> One thing you should look into is what happens when you add the extension of a binary filetype, like png for example, and see if that causes problems.

Tested on different binaries, the results are as follows: 
- images: just show inside the interface, as I remember obsidian provides for normal image viewing;
- files like .EXE, .MSI, .JAR: runs them directly (maybe in the potential future I'll try to fix this somehow);
- ZIP files open in Explorer;
- all sorts of video files, like MP4, are simply played in the application;

But all sorts of formats, let's say we create a file and rename it to [NAME].dictionary, will open it as a plain text file.

BUT! if we rename the binary file to a custom extension (same .dictionary), then a file will be able to edit the binary

The plugin allows opening files as notes, only those that don't have any default applications to run (or don't run anything themselves).

And thanks for testing reminder!

image
**^this is .EXE file which was renamed into .DICTIONARY file**
image
**^this is image-binaries file, works on JPG, JPEG, PNG**

PR #2340, opened 2023-08-24
Add "Quick Settings Switchs for Plugins"

Hugo-COLLIN:
# I am submitting a new Community Plugin : <br>Quick Settings Switchs for Plugins

## Repo URL

Link to my plugin: https://github.com/Hugo-COLLIN/obsidian-quick-settings-switchs

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -7775,5 +7775,12 @@
     "author": "williamechols",
     "description": "lightweight Python code snippet executor with runtime input support",
     "repo": "williamechols/obsidian-execute-python"
+  },
+  {
+    "id": "quick-settings-switchs",
+    "name": "Quick Settings Switchs for Plugins",
+    "author": "Hugo COLLIN",
+    "description": "Enable and disable your plugins quickly from the settings side menu.",
+    "repo": "Hugo-COLLIN/obsidian-quick-settings-switchs"
   }
 ]


joethei:
[import * as fs from "fs";](https://github.com/Hugo-COLLIN/obsidian-quick-settings-switchs/blob/fb422de6bae00414e6ef1a64aca27c0147f3b3c5/src/main.ts#L2-L3

[plugin.style.display = 'flex';](https://github.com/Hugo-COLLIN/obsidian-quick-settings-switchs/blob/fb422de6bae00414e6ef1a64aca27c0147f3b3c5/src/main.ts#L89-L91)

[async savePluginState(app: App, pluginId: string, enabled: boolean) {](https://github.com/Hugo-COLLIN/obsidian-quick-settings-switchs/blob/fb422de6bae00414e6ef1a64aca27c0147f3b3c5/src/main.ts#L133-L156)
There are additional `enablePluginAndSave`/`disablePluginAndSave` functions that you can use to replace this.

[containerEl.createEl("h2", {text: "Plugin Toggle Settings"});](https://github.com/Hugo-COLLIN/obsidian-quick-settings-switchs/blob/fb422de6bae00414e6ef1a64aca27c0147f3b3c5/src/settingsTab.ts#L21)

What is your settings tab for?
It looks like it's the same content as the Community plugins tab.

Hugo-COLLIN:
@joethei Thank you for your review.

As you suggested, I used `enablePluginAndSave` and `disablePluginAndSave`, so I don't need to import fs anymore and I can keep `isDesktopOnly` marked as `false`.

I also moved my styles in the styles.css file.

The settings tab was just for testing, I didn't import this part of the code in main.ts. Now I deleted it for more convenience.

Hugo-COLLIN:
Don't understand why checks are failing.

joethei:
That's a git error we see showing up every now and again, not sure what is causing it.

Hugo-COLLIN:
Hello @joethei @liamcain,
A simple message to ask you if it's normal the revision lasts so long : it's about 1 month now!
I understand you have a lot of plugins to check. My only question is to kwow when my plugin will be reviewed.
I know validating plugins is complex and time-consuming, so thank you for your work!

liamcain:
- [registerDomEvent](https://github.com/Hugo-COLLIN/obsidian-quick-settings-switchs/blob/39d1af035363aff9fe88dbba2791094021a35adb/src/main.ts#L13) why are you reregistering this event every time the layout changes? This will keep adding more and more event listeners to the sidedock. You are also not checking to make sure that the querySelector resolves. If this returns null, then the registerDomEvent will error
- [innerHTML](https://github.com/Hugo-COLLIN/obsidian-quick-settings-switchs/blob/39d1af035363aff9fe88dbba2791094021a35adb/src/main.ts#L22) You do not need to inject your css this way. Just add the CSS to styles.css

PR #2342, opened 2023-08-24
Add URL Display plugin

lin-stephanie:

## Repo URL

Link to my plugin: https://github.com/lin-stephanie/obsidian-url-display

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lin-stephanie:
update release

joethei:
[Copyright (c) 2021 CattailNu](https://github.com/lin-stephanie/obsidian-url-display/blob/54f28f96f22e342637d21959ee6fa6bb2eb9084f/LICENSE#L3-L4)
Update this.

[id: 'url-display-open',](https://github.com/lin-stephanie/obsidian-url-display/blob/54f28f96f22e342637d21959ee6fa6bb2eb9084f/src/main.ts#L35)

[private readonly isOpen = () => {](https://github.com/lin-stephanie/obsidian-url-display/blob/54f28f96f22e342637d21959ee6fa6bb2eb9084f/src/main.ts#L89-L95)
Generally when you have a function named `isXyz` it's expected that the function returns a boolean and does not change anything about the current state.

[private static debounce = (fn: () => void, ms = 1000) => {](https://github.com/lin-stephanie/obsidian-url-display/blob/54f28f96f22e342637d21959ee6fa6bb2eb9084f/src/main.ts#L237-L243)
The Obsidian API already includes a function of the same name.

[export class MicroLinkParser extends Parser {](https://github.com/lin-stephanie/obsidian-url-display/blob/54f28f96f22e342637d21959ee6fa6bb2eb9084f/src/parser.ts#L54-L80)
In your readme, please mention that you are using these REST API's, and why.

[urldispaly](https://github.com/lin-stephanie/obsidian-url-display/blob/54f28f96f22e342637d21959ee6fa6bb2eb9084f/src/cache.ts#L9)
Typo.

liamcain:
- [activateView](https://github.com/lin-stephanie/obsidian-url-display/blob/7cffcfd74d30766da09b941abd8c51af697b92cc/src/main.ts#L97) Consider the following code instead of using getLeavesOfType:
```
			let leaf = this.app.workspace.getRightLeaf(false);
			await leaf.setViewState({ type: VIEW_TYPE, active: true });
			this.app.workspace.revealLeaf(leaf);
```

- [getLeavesOfType(VIEW_TYPE)[0]](https://github.com/lin-stephanie/obsidian-url-display/blob/7cffcfd74d30766da09b941abd8c51af697b92cc/src/main.ts#L86) checking the first index of the array is unnecessary. Just check the length
- [active-leaf-change](https://github.com/lin-stephanie/obsidian-url-display/blob/7cffcfd74d30766da09b941abd8c51af697b92cc/src/main.ts#L64) why not just listen for file-open instead?
- [this.app.workspace.getActiveFile()](https://github.com/lin-stephanie/obsidian-url-display/blob/7cffcfd74d30766da09b941abd8c51af697b92cc/src/main.ts#L119) this approach is overly complicated. Instead, you should add a file-open event listener to your UrlDisplayView class. In the file-open handler, just load the passed in TFile into the view. Don't worry about if the active view is a MarkdownView and stop handling the logic in the plugin then passing the data to the view. Just do all the work in the View class.

lin-stephanie:
Thank you a lot for detailed suggestions and merging my plugin. I will make these changes in later versions.

PR #2344, opened 2023-08-24
Add plugin Tags Overview

christianwannerstedt:

## Repo URL

Link to my plugin:
https://github.com/christianwannerstedt/obsidian-tags-overview

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[this.app.workspace.detachLeavesOfType(VIEW_TYPE);](https://github.com/christianwannerstedt/obsidian-tags-overview/blob/cd21330ac673b03b967ec91abc07418c8dee2e16/src/main.ts#L27)
You should not detach the leaves here, but reveal the existing one if there is one.
If a user moves it somewhere else, the position will reset every time they press that ribbon icon.

[import moment from "moment";](https://github.com/christianwannerstedt/obsidian-tags-overview/blob/cd21330ac673b03b967ec91abc07418c8dee2e16/src/utils.ts#L2)
Import the moment instance from the Obsidian API instead.

[import * as fs from "fs";](https://github.com/christianwannerstedt/obsidian-tags-overview/blob/cd21330ac673b03b967ec91abc07418c8dee2e16/src/utils.ts#L1)
Remove any unused imports

christianwannerstedt:
Good feedback, I've now addressed the three points you mentioned and updated the code. 
Feel free to review this PR again.

liamcain:
- [activateView](https://github.com/christianwannerstedt/obsidian-tags-overview/blob/1fd29393e42cba96032ec480b4206107bfde771f/src/main.ts#L26) I suggest rewriting this as:
image
- [this.app.workspace](https://github.com/christianwannerstedt/obsidian-tags-overview/blob/1fd29393e42cba96032ec480b4206107bfde771f/src/main.ts#L37-L39) don't detach leaves on unload. This will cause the leaf positions to get reset when the plugin is updated which isn't always desirable.
- [modify](https://github.com/christianwannerstedt/obsidian-tags-overview/blob/1fd29393e42cba96032ec480b4206107bfde771f/src/views/root-view.tsx#L34) you shouldn't listen for "modify" since you're using the tags from the metadataCache. The modify event will fire before the metadataCache is updated so your tags will be stale. Instead, you should listen for metadataCache's "changed" event <https://github.com/obsidianmd/obsidian-api/blob/249eb773347588f4e9fec7c600e1904cd9240fcb/obsidian.d.ts#L2574>
- [TaggedFile](https://github.com/christianwannerstedt/obsidian-tags-overview/blob/1fd29393e42cba96032ec480b4206107bfde771f/src/views/root-view.tsx#L16) you might want to store this in a map or a weakmap, keyed by the TFile or the file path
- [statSync](https://github.com/christianwannerstedt/obsidian-tags-overview/blob/1fd29393e42cba96032ec480b4206107bfde771f/src/views/tags-view.tsx#L89) TFile already stores this data in TFile.stat. You probably don't even need the TaggedFile abstraction at all.

christianwannerstedt:
Thanks, good ones! I had missed that TFile already had the FileStats available, very convenient. I've updated my plugin (0.1.2) and addressed all the points you did mention in your comment. Feel free to review it again.

liamcain:
Looks like 0.1.2 isn't released on github yet. Mind fixing that and pinging me when its released? Then I can merge. Thanks!

christianwannerstedt:
Absolutely, the [release](https://github.com/christianwannerstedt/obsidian-tags-overview/releases/tag/0.1.2) is now available on github. @liamcain

PR #2353, opened 2023-08-25
add rev-obsidian-sync-plugin

acheong08:

## Repo URL

Link to my plugin: https://github.com/acheong08/rev-obsidian-sync-plugin
- https://github.com/acheong08/rev-obsidian-sync

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.ts`
  - [x] `manifest.json`


acheong08:
To add a little bit of context:
https://news.ycombinator.com/item?id=37247767

I reverse engineered the native sync's server. At first you had to edit the electron JS files to swap the endpoint but that was very inconvenient, especially for IOS. This allows you to use a self hosted server on all (I think) clients without breaking code signing stuff. Credits to @tuxmachine for the idea.

kepano:
Thanks for your submission — as I mentioned on HN it's an impressive technical feat! Unfortunately we cannot accept your plugin into the directory. While we have no qualms with reverse engineering as a playground for experimentation, Obsidian Sync is a service we intend to keep first-party only for the foreseeable future.

acheong08:
Is there a way to load a local plugin into IOS?

PR #2349, opened 2023-08-25
add image-memo plugin

Ender-Wiggin2019:
# I am submitting a new Community Plugin

## Repo URL

Link to my plugin:

## Release Checklist
- [ ] I have tested the plugin on
  - [X]  Windows (using parallel on Mac OS)
  - [X]  macOS
  - [ ]  Linux
  - [ ]  Android _(if applicable)_
  - [ ]  iOS _(if applicable)_
- [ ] My GitHub release contains all required files
  - [X] `main.js`
  - [X] `manifest.json`
  - [X] `styles.css` _(optional)_
- [X] GitHub release name matches the exact version number specified in my manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
- [X] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [X] My README.md describes the plugin's purpose and provides clear usage instructions.
- [X] I have read the tips in https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines and have self-reviewed my plugin to avoid these common pitfalls.
- [X] I have added a license in the LICENSE file.
- [X] My project respects and is compatible with the original license of any code from other plugins that I'm using.
      I have given proper attribution to these other projects in my `README.md`.

PR #2354, opened 2023-08-25
Add Image Memo Plugin

Ender-Wiggin2019:

## Repo URL

Link to my plugin:
https://github.com/Ender-Wiggin2019/obsidian-image-memo

## Release Checklist
  - [X]  Windows (using parallel on Mac OS)
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

---

I have created a PR https://github.com/obsidianmd/obsidian-releases/pull/2349 yesterday, but I cannot pass the testing because of `fatal: couldn't find remote ref refs/pull/2349/merge`. So I close the previous one and create a new one, while update my code according to the feedback.


liamcain:
- [this.view](https://github.com/Ender-Wiggin2019/obsidian-image-memo/blob/361d603f9d4f9600668d6eab3187f522df4a96d4/src/main.ts#L19) this is a potential memory leak, don't keep a reference to the view in your plugin class. Instead, to reload the view when data is saved, I recommend triggering an event in the Plugin class and listening for the event in the View instance. This also scales better if you have multiple instances of the View in your workspace.
- [saveSettings](https://github.com/Ender-Wiggin2019/obsidian-image-memo/blob/361d603f9d4f9600668d6eab3187f522df4a96d4/src/main.ts#L33) I suggest not calling "save" in onload.

ericaxu:
To keep things tidy, we're going to close issues that we don't hear back after a month of inactivity.

Feel free to tell us to re-open it when you're ready to reply and we'll reopen the issue right away. Thanks for your understanding!

PR #2355, opened 2023-08-25
Add plugin: Desk

davidlandry93:

## Repo URL

Link to my plugin: https://github.com/davidlandry93/obsidian-desk

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["author": "David Landry <davidlandry93@gmail.com>",](https://github.com/davidlandry93/obsidian-desk/blob/7d09af3df7bcb16a935f08a80eabdb96d6e1aa5e/manifest.json#L7)
This field is displayed in the sidebar and in the description for the plugin details page. Are you sure you want to have your email listed so prominently?


[// This adds a status bar item to the bottom of the app. Does not work on mobile apps.](https://github.com/davidlandry93/obsidian-desk/blob/7d09af3df7bcb16a935f08a80eabdb96d6e1aa5e/main.ts#L18-L20)
Remove this sample code.

[this.app.workspace.detachLeavesOfType(VIEW_TYPE_DESK)](https://github.com/davidlandry93/obsidian-desk/blob/7d09af3df7bcb16a935f08a80eabdb96d6e1aa5e/main.ts#L40)

davidlandry93:
Hello Johannes, 
Thanks for the review. I implemented your suggestions.

liamcain:
- [evt.stopPropagation](https://github.com/davidlandry93/obsidian-desk/blob/216c4d3b730d6346ee6c2a7df003d7584cd819e8/main.ts#L20) Please don't do this. What is this for?

davidlandry93:
Hello Liam, thanks for having a look.

Yes indeed that is a little intrusive and I removed it. It didn't even do what I wanted. 

Actually maybe you can help me with this. I don't fully understand the behavior of the escape key. I was trying to prevent the escape key from moving away from my own tab. In my tests this morning, if I accidentally press escape when using the extension, focus moves away from the tab towards the "New Tab" page. During testing I found that I would sometimes want to press escape while I was typing in my filters, so I figure users would do it too.

Cheers

liamcain:
> Actually maybe you can help me with this. I don't fully understand the behavior of the escape key. I was trying to prevent the escape key from moving away from my own tab. In my tests this morning, if I accidentally press escape when using the extension, focus moves away from the tab towards the "New Tab" page. During testing I found that I would sometimes want to press escape while I was typing in my filters, so I figure users would do it too

"escape" will attempt to give focus back to the last active non-sidebar tab. You should be able to override this behavior by registering an event handler to "escape" via Scope.

In your view, do:

```
		this.scope.register([], 'Escape', () => {
                     // This can be left blank, or do something custom. Either way, this will override the focus behavior.
		});
```

liamcain:
> Yes indeed that is a little intrusive and I removed it

I don't see that change yet on your branch. Did you forget to push?

davidlandry93:
Thanks for the advice. Still no luck though, you can see what I tried here:

https://github.com/davidlandry93/obsidian-desk/blob/master/deskview.tsx#L36

With this my handler doesn't fire when I press escape (if I try to print a console log inside).

Did I miss something?
I don't mind that behavior too much though, if it's difficult to go around it I would be happy to publish as is.

liamcain:
You should be registering to the view's scope.

```
this.escapeHandler = this.app.scope.register([], 'Escape', () => { });
```

Change to:

```
this.escapeHandler = this.scope.register([], 'Escape', () => { });
```

Since it's on the view scope, you wouldn't need to unregister it either.

davidlandry93:
Hello Liam, sorry for the delay

I implemented your suggestion but it gives an error message:

```
app.js:1 Failed to open view TypeError: Cannot read properties of undefined (reading 'register')
```

I tried to update my `obsidian` dependency to the latest version. I inspected my view object but could not find a scope property?

You can see what I tried here: 
https://github.com/davidlandry93/obsidian-desk/blob/d9d4c286ae1ea0f31f30d91f0c0e5b5200bf7572/deskview.tsx#L38

Sorry again for the delay, and thanks for any help
David

PR #2389, opened 2023-09-02
Add plugin: Todo.txt Codeblocks

benjamonnguyen:

## Repo URL

Link to my plugin: https://github.com/benjamonnguyen/obsidian-todotxt-codeblocks

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[Copyright (c) [year] [fullname]](https://github.com/benjamonnguyen/obsidian-todotxt-codeblocks/blob/aeac664aa2fbf2ad243e42a154454030785b6817/LICENSE#L3)
Please fill this.

[static NAME = 'obsidian-todotxt-codeblocks';](https://github.com/benjamonnguyen/obsidian-todotxt-codeblocks/blob/aeac664aa2fbf2ad243e42a154454030785b6817/src/main.ts#L19)
Please use the actual plugin name here, and not the repo name.

[id: 'create-new-todotxt-task-cmd',](https://github.com/benjamonnguyen/obsidian-todotxt-codeblocks/blob/aeac664aa2fbf2ad243e42a154454030785b6817/src/command/createNewTaskCmd.ts#L7)
You don't need to include the plugin identifier here, the id can just be `create-new-task`

benjamonnguyen:
@joethei Thanks for the review, I've resolved the change requests.

liamcain:
- [await this.saveData(this.settings).then(](https://github.com/benjamonnguyen/obsidian-todotxt-codeblocks/blob/3da757dee037f5c5783249b54608e3e680dea1d8/src/main.ts#L84) why are you mixing async/await with .then()? If you're using await, you can omit the promise chaining. Also, there's really no reason your readonly version of settings needs to be assigned _after_ the save resolves
- [this.registerDomEvent(document, 'keyup', (event: KeyboardEvent) => {](https://github.com/benjamonnguyen/obsidian-todotxt-codeblocks/blob/3da757dee037f5c5783249b54608e3e680dea1d8/src/main.ts#L47) you should remove this. This isn't related to your plugin and will interact with all buttons in the app and will likely cause hard to track down bugs

benjamonnguyen:
@liamcain The UX feels really clunky without the "Enter" to submit behavior, especially when creating many todos.

I've added a check in case the "mod-cta" class shows up anywhere else:
```js
this.registerDomEvent(document, 'keyup', (event: KeyboardEvent) => {
	const modalSubmitBtn = (
		document.getElementsByClassName('todotxt-modal').item(0) as HTMLElement | null
	)
		?.getElementsByClassName('mod-cta')
		.item(0) as HTMLButtonElement | null;
	if (modalSubmitBtn && event.key === 'Enter') {
		modalSubmitBtn.click();
	}
});
```

Let me know if this is acceptable.

liamcain:
Oh, I didn't understand what it was for. If this is for a modal, you should use Modal.scope <https://github.com/obsidianmd/obsidian-api/blob/249eb773347588f4e9fec7c600e1904cd9240fcb/obsidian.d.ts#L2606>

Something like this:

```
		scope.register([], 'Enter', (evt: KeyboardEvent) => {
			if (evt.isComposing) {
				return;
			}
				const actionBtn = document
					.getElementsByClassName('mod-cta')
					.item(0) as HTMLButtonElement | null;
				actionBtn?.click();
		});
```

You can add that to your base modal class

benjamonnguyen:
@liamcain Thanks! I've pushed the changes and updated my release

PR #2400, opened 2023-09-04
Add Text Transform plugin

ipshing:

## Repo URL

Link to my plugin: https://github.com/ipshing/obsidian-text-transform

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -8055,5 +8055,12 @@
 	"author": "julix14",
 	"description": "Let your AI assistant ChatGPT define words and concepts for you.",
 	"repo": "julix14/chatGPT-Obsidian"
-  }
+  },
+  {
+		"id": "text-transform",
+		"name": "Text Transform",
+		"author": "ipshing",
+		"description": "This plugin adds options to transform text to different casings using keyboard shortcuts.",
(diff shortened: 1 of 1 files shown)


joethei:
["description": "This plugin adds options to transform text to different casings using keyboard shortcuts.",](https://github.com/ipshing/obsidian-text-transform/blob/1ab6ceb0dba5122edf77a2f212fe3d8d07d50470/manifest.json#L6)
Avoid including sentences like "This plugin does xyz" in your description.

[id: "text-transform-uppercase",](https://github.com/ipshing/obsidian-text-transform/blob/1ab6ceb0dba5122edf77a2f212fe3d8d07d50470/src/main.ts#L26)

[hotkeys: [](https://github.com/ipshing/obsidian-text-transform/blob/1ab6ceb0dba5122edf77a2f212fe3d8d07d50470/src/main.ts#L28-L33)

ipshing:
I have made the requested changes, are there other steps I need to follow to continue the process?

PR #2403, opened 2023-09-04
Add plugin Copy Metadata

wenlzhang:
Adding the plugin Copy Metadata.

## Repo URL

Link to my plugin: https://github.com/wenlzhang/obsidian-copy-metadata

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
The plugin name could be confusing for some, as properties are often referred to as metadata.

[if (!this.settings.useKeyupEvents) {](https://github.com/wenlzhang/obsidian-copy-metadata/blob/929649e3c185cad935915d137663a02203591cec/main.ts#L28-L66)
This code does nothing, remove it.

[this.containerEl.createEl("h1", { text: "Copy Metadata" });](https://github.com/wenlzhang/obsidian-copy-metadata/blob/929649e3c185cad935915d137663a02203591cec/main.ts#L152)

wenlzhang:
Thank you for the comment!

> The plugin name could be confusing for some, as properties are often referred to as metadata.

In the future, the features can be extended to operate on info such as properties. On the other hand, the following plugin also includes the word metadata in its name.
https://github.com/mdelobelle/metadatamenu

Or, do you have further suggestions on how to rename the plugin?

liamcain:
- [this.app.vault.adapter.stat(activeFile.path);](https://github.com/wenlzhang/obsidian-copy-metadata/blob/37ac12c21c2e2344371c3122b97e4e7e68ae26e1/main.ts#L47) if you have a TFile already, you can access the ctime via tfile.stat.ctime. No need for the extra stat call.

PR #2410, opened 2023-09-05
Add plugin: iDoRecall

zhukoffsky:

## Repo URL

Link to my plugin: https://github.com/iDoRecall/idorecall

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
You don't need to handle the merge conflicts on this PR, we will do that once the plugin is ready to be included.

[// TODO: add licence](https://github.com/iDoRecall/idorecall/blob/b336dabcde4d9aa5670ad939eea217a1c8caf094/LICENSE#L1-L2)
Please do this.

[return file.stat.ctime === Number(e.file);](https://github.com/iDoRecall/idorecall/blob/b336dabcde4d9aa5670ad939eea217a1c8caf094/src/main.ts#L46)
ctime is not stable when syncing between different operating systems, you should use the path to the file instead, this will also make this operation faster, as you can make use of Obsidian's file cache (Vault#getAbstractFileByPath())

[this.app.workspace.detachLeavesOfType('idr-view');](https://github.com/iDoRecall/idorecall/blob/b336dabcde4d9aa5670ad939eea217a1c8caf094/src/main.ts#L158)

[this.app.workspace.detachLeavesOfType('idr-view');](https://github.com/iDoRecall/idorecall/blob/b336dabcde4d9aa5670ad939eea217a1c8caf094/src/main.ts#L128)
Also don't detach the leaves here, focus on the leaf that already exists, if there is one.

[this.app.workspace.on('editor-change', (editor: Editor) => {](https://github.com/iDoRecall/idorecall/blob/b336dabcde4d9aa5670ad939eea217a1c8caf094/src/main.ts#L139-L142)
You forgot to register this event.

[export const debounce = (f: () => void, ms: number) => {](https://github.com/iDoRecall/idorecall/blob/b336dabcde4d9aa5670ad939eea217a1c8caf094/src/utils/debounce.ts#L1-L14)
The Obsidian API already exports a debounce function.

[plug.innerHTML = innerContent;](https://github.com/iDoRecall/idorecall/blob/b336dabcde4d9aa5670ad939eea217a1c8caf094/src/utils/getFragment.ts#L4)
[Avoid using `innerHTML`](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Avoid+%60innerHTML%60%2C+%60outerHTML%60+and+%60insertAdjacentHTML%60)

[containerEl.createEl('h2', { text: 'Settings for IDR plugin' });](https://github.com/iDoRecall/idorecall/blob/b336dabcde4d9aa5670ad939eea217a1c8caf094/src/utils/settings.ts#L19)

[Codemirror tooltip](https://github.com/iDoRecall/idorecall/blob/b336dabcde4d9aa5670ad939eea217a1c8caf094/src/cm-extensions/tooltip.ts#L5)
Tooltips have been migrated to the `@codemirror/view` plugin, use that instead of keeping a version of the old package around.

---

[import axios, { AxiosError, AxiosInstance, AxiosRequestConfig } from 'axios';](https://github.com/iDoRecall/idorecall/blob/b336dabcde4d9aa5670ad939eea217a1c8caf094/src/api/restService.ts#L1)
Instead of axios you should use the  `requestUrl` function from the Obsidian API instead, it will handle some things like CORS automatically.


['iDoRecall Plugin'](https://github.com/iDoRecall/idorecall/blob/b336dabcde4d9aa5670ad939eea217a1c8caf094/src/main.ts#L33)
The word "Plugin" is unnecessary here.

[hotkeys: [{ modifiers: ['Mod'], key: 't' }],](https://github.com/iDoRecall/idorecall/blob/b336dabcde4d9aa5670ad939eea217a1c8caf094/src/constants/command-list.ts#L14)

zhukoffsky:
Hi @joethei! Thank you for the review.
All your comments are fixed.

Except this one
[this.app.workspace.detachLeavesOfType('idr-view');](https://github.com/iDoRecall/idorecall/blob/b336dabcde4d9aa5670ad939eea217a1c8caf094/src/main.ts#L128)
Also don't detach the leaves here, focus on the leaf that already exists, if there is one.

We can't handle this, because we need to close/destroy all forms before creating a new one.

joethei:
The way I read your code, there can only be ever one leaf with your view, right?
In that case, you should not detach any leaves, because every time you do, the new leaf might initialize at a different location as to where the user moved it manually.
Looking at your code that would happen every time a different file is opened.

This should be removed accordingly:
<https://github.com/iDoRecall/idorecall/blob/011c147f48586e657ab7bed008f2e83dee4b0abd/src/main.ts#L96C21-L96C21>

[ this.app.workspace.on('editor-change', (editor: Editor) => {](https://github.com/iDoRecall/idorecall/blob/011c147f48586e657ab7bed008f2e83dee4b0abd/src/main.ts#L160C10-L165)
Also, what is this for?
It looks useless to me.

zhukoffsky:
@joethei Thanks for comments!
Please check. We removed detachLeavesOfType().

[this.app.workspace.on('editor-change', (editor: Editor) => {](https://github.com/iDoRecall/idorecall/blob/011c147f48586e657ab7bed008f2e83dee4b0abd/src/main.ts#L160C10-L165)
Without this our text selection works not stable

joethei:
That sounds more like you have a bug somewhere in your code, and that is just masking it.

Can you clarify what the bug is?

zhukoffsky:
https://github.com/obsidianmd/obsidian-releases/assets/71514058/a6756032-b451-4db2-841f-342fee37272c

Without editor-change event handler our first text selection after opening recall creation form disappears

zhukoffsky:
@joethei so what are our next steps then?

liamcain:
- [is-flashing](https://github.com/iDoRecall/idorecall/blob/2aa27d4453c822fb8fbb4aa7fffc6ffbaa3a00f9/src/styles/global.scss#L45) Why are you overriding this? It doesn't seem related to your plugin even.

zhukoffsky:
Hi @liamcain! Thank you for the review
Your comments resolved, code pushed, please take a look

PR #2415, opened 2023-09-07
Add "Obsidian Webviews" Plugin to Community Tab

Aeases:

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
What sets your plugin appart from the existing OpenGate plugin?
Please include some information in the readme.
Please also credit the OpenGate author in the readme.

Remove the `main.js` file from the repo and add it to your .gitignore., release files should only be included in the releases tab.

["authorUrl": "https://mas.to/aeases, https://twitter.com/duocdev"](https://github.com/Aeases/obsidian-webviews/blob/b9333af636d12a336c826e43fec39eeff7d3f5df/manifest.json#L8)
This only accepts a single link, remove the other link.

[import { BrowserWindow } from 'electron/main'](https://github.com/Aeases/obsidian-webviews/blob/b9333af636d12a336c826e43fec39eeff7d3f5df/src/main.ts#L14-L15)
These imports are only available from a Node.js runtime, which means that this will throw errors on Mobile (that's a problem since you `isDesktopOnly` is missing from your manifest.

[id: `open-gate-create-new`,](https://github.com/Aeases/obsidian-webviews/blob/b9333af636d12a336c826e43fec39eeff7d3f5df/src/main.ts#L74)

[hotkeys: [{ modifiers: ['Mod', 'Shift'], key: 'g' }],](https://github.com/Aeases/obsidian-webviews/blob/b9333af636d12a336c826e43fec39eeff7d3f5df/src/main.ts#L90)

[export class Suggest<T> {](https://github.com/Aeases/obsidian-webviews/blob/b9333af636d12a336c826e43fec39eeff7d3f5df/src/fns/suggest.ts#L5)
Please credit the original author of this code.

[frag.createDiv().innerHTML = html](https://github.com/Aeases/obsidian-webviews/blob/b9333af636d12a336c826e43fec39eeff7d3f5df/src/fns/fragwithhtml.ts#L2)
[Avoid using `innerHTML`](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Avoid+%60innerHTML%60%2C+%60outerHTML%60+and+%60insertAdjacentHTML%60)

[.setName('Single Webview Only')](https://github.com/Aeases/obsidian-webviews/blob/b9333af636d12a336c826e43fec39eeff7d3f5df/src/fns/formEditGate.ts#L59)

[let response = await fetch(url)](https://github.com/Aeases/obsidian-webviews/blob/b9333af636d12a336c826e43fec39eeff7d3f5df/src/fns/fetchTitle.ts#L2)

PR #2417, opened 2023-09-07
Add "Share Note" plugin

alangrainger:

## Repo URL

Link to my plugin: https://github.com/alangrainger/obsidian-share

## 👉 Notes for Obsidian Team

This plugin is different from the two existing plugins QuickShare and OzanShare. QuickShare does not support the theme of your note, nor images. OzanShare is a paid service.

Regarding the code, I will be very interested to get your feedback. What I've done works perfectly, but I'm not sure how "best practice" my methods are, for example how I get the pixel-perfect rendered HTML here: https://github.com/alangrainger/obsidian-share/blob/2195942681f768075e82a91960a61225f9b5246d/src/note.ts#L46

Keen to hear feedback. Maybe you'll like how I did it!

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -8055,5 +8055,12 @@
 	"author": "julix14",
 	"description": "Let your AI assistant ChatGPT define words and concepts for you.",
 	"repo": "julix14/chatGPT-Obsidian"
+  },
+  {
+    "id": "share-note",
+    "name": "Share Note",
+    "author": "Alan Grainger",
+    "description": "Share a note publicly with full styling. Data is stored encrypted on the server and only you have the key.",
+    "repo": "alangrainger/obsidian-share"
   }
 ]


joethei:
[if (!file) {](https://github.com/alangrainger/obsidian-share/blob/103ace321afa08124f5a5a3fce931346ebd2869d/src/main.ts#L141-L148)
The command that calls this function should use a `checkCallback`, then you can remove this code.

[new Setting(containerEl)](https://github.com/alangrainger/obsidian-share/blob/103ace321afa08124f5a5a3fce931346ebd2869d/src/settings.ts#L48-L50)
The first section in settings should not have a heading: <https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Only+use+headings+under+settings+if+you+have+more+than+one+section.>

alangrainger:
Hi @joethei - changes have been made and published in `0.3.15`.

`checkCallback` is added.

The section title is there because I have 3 sections:

image

alangrainger:
Sorry - I understand now, I was mis-reading the docs.

Updated in `0.3.16` with both changes.

liamcain:
Hi Alan, thanks for your plugin submission!

We noticed that when users share notes using your plugin they are hosted at `obsidianshare.com`. 

The word "Obsidian" is a trademark, and therefore you cannot use it this way. The problem with `obsidianshare.com` is that it may confuse people into thinking that pages are hosted on an official Obsidian service.

Please update your domain to not include the word `obsidian` and we will be happy to approve the plugin to the community directory. Thank you!

alangrainger:
Hi @liamcain I have migrated to a new domain:

https://share.note.sx/8HitpABlzq2mOu4W

Please note that existing users will need to be given a chance to update their links before I turn that server off. I hope you can understand that from their perspective.

alangrainger:
@liamcain I see the 👍 - I'm just checking you're not waiting on anything from me here before approving?

liamcain:
Sorry, gave the thumbs up, but was waiting to get a response from the rest of the team (Monday was a holiday for us, and Tuesday was a bit busy with releases).

- [fs](https://github.com/alangrainger/obsidian-share/blob/9dc9f0088d1cab16e57425e0de2136cd0b458ee6/src/note.ts#L4) this import is a Node.JS import so it will fail on mobile.
- [fs.readFileSync](https://github.com/alangrainger/obsidian-share/blob/9dc9f0088d1cab16e57425e0de2136cd0b458ee6/src/note.ts#L260) you will need to use the Filesystem adapter instead, but we don't have a synchronous version of read exposed
- [domCopy](https://github.com/alangrainger/obsidian-share/blob/9dc9f0088d1cab16e57425e0de2136cd0b458ee6/src/note.ts#L70) Why not use [render](https://github.com/obsidianmd/obsidian-api/blob/249eb773347588f4e9fec7c600e1904cd9240fcb/obsidian.d.ts#L2245) instead?

I think the 'fs' import is worth fixing before we merge this. Let me know if you have any questions!

alangrainger:
`fs` replaced with `FileSystemAdapter`: https://github.com/alangrainger/obsidian-share/commit/90c9155855761936f9b2026113f8f1cc829365e2

I don't understand regarding [render](https://github.com/obsidianmd/obsidian-api/blob/249eb773347588f4e9fec7c600e1904cd9240fcb/obsidian.d.ts#L2245) sorry 😬 I'm **reading** the total document into a DOM parser so I can get the Document `<body>` classes and some other attributes.

`render()` appears to be a function for converting Markdown to HTML.

liamcain:
> I'm reading the total document into a DOM parser so I can get the Document <body> classes/styles and some other element classes/styles, plus be able to manipulate that DOM without affecting the user's Obsidian instance.

Yes, I was suggesting you convert the md file to HTML directly instead of ripping the HTML from the DOM.

alangrainger:
@liamcain thanks for the suggestion regarding `render()`. I've found an issue which prevents me from using it, and could use some assistance.

If I render this simple markdown for a callout:

```markdown
> [!NOTE] Title
> Contents
```

I get this HTML which does not include the SVG callout icon - it's just an empty `<svg>` element:

```html
<div data-callout-metadata="" data-callout-fold="" data-callout="note" class="callout node-insert-event">
  <div class="callout-title">
    <div class="callout-icon">
      <svg width="16" height="16"></svg> <!-- 👈 Empty svg here -->
    </div>
    <div class="callout-title-inner">Title</div>
  </div>
  <div class="callout-content">
    <p>Contents</p>
  </div>
</div>
```

What am I doing wrong here? Thank you!

PR #2428, opened 2023-09-09
Add timer plugin

Raboro:

## Repo URL

Link to my plugin: https://github.com/Raboro/obsidian-timer-plugin

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -8146,5 +8146,12 @@
     "author": "Divam Gupta",
     "description": "Create and edit advanced spreadsheets with Obsidian. Rich formatting, formulas, cells, filtering and more!",
     "repo": "divamgupta/obsidian-spreadsheets"
+  },
+  {
+    "id": "timer",
+    "name": "Timer",
+    "author": "Marius Wörfel",
+    "description": "Allows you to measure time.",
+    "repo": "Raboro/obsidian-timer-plugin"
   }
 ]


joethei:
[this.addRibbonIcon('alarm-clock', 'Open Timer', async () => await this.openView());](https://github.com/Raboro/obsidian-timer-plugin/blob/82cd1ea5123d30ac0f26d833d3e00d6c3d8acba9/src/main.ts#L15)

Raboro:
fixed to -> Open timer

liamcain:
- [timer: string](https://github.com/Raboro/obsidian-timer-plugin/blob/8e53cba34fce252e6e7e22d62e44df29b0286e30/src/main.ts#L93) nitpick, it's hard to tell what "timer" is when you have a class "timer" and a string "timer"
- [leaves[0]](https://github.com/Raboro/obsidian-timer-plugin/blob/8e53cba34fce252e6e7e22d62e44df29b0286e30/src/main.ts#L56) don't push to a list this way. You should instead either use leaves.push() or just have a reference to "leaf" and not push to the array at all.
- [timer: Timer](https://github.com/Raboro/obsidian-timer-plugin/blob/8e53cba34fce252e6e7e22d62e44df29b0286e30/src/main.ts#L10) why is there a Timer instance on the Plugin? Shouldn't these only live in the View? How does this behave with multiple Timer views? Is the timer on the view just a reference to the latest timer?
- [AbstractTimerModal](https://github.com/Raboro/obsidian-timer-plugin/blob/8e53cba34fce252e6e7e22d62e44df29b0286e30/src/modals/abstractTimerModal.ts#L3) nit: your Timer modal isn't an abstract class, so it doesn't really make sense to call it AbstractTimerModal

Raboro:
Thank for your suggestions. I fixed those in the latest [release](https://github.com/Raboro/obsidian-timer-plugin/releases/tag/1.0.3).

PR #2429, opened 2023-09-10
Add plugin Cardify

joshuakto:

## Repo URL

Link to my plugin: https://github.com/joshuakto/obsidian-cardify

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
- Obsidian sample plugin


joethei:
["description": "This is a plugin to cardify markdown contents into subsequent markdown files.",](https://github.com/joshuakto/obsidian-cardify/blob/235f89608b8a899313761b44e6ed88e71dd114c0/manifest.json#L6)
Avoid including sentences like "This is a plugin that does xyz" in your description.

["authorUrl": "joshuakto@gmail.com",](https://github.com/joshuakto/obsidian-cardify/blob/235f89608b8a899313761b44e6ed88e71dd114c0/manifest.json#L8)
This only accepts links to websites, if you don't have one use the url to your github profile.

[import * as path from "path";](https://github.com/joshuakto/obsidian-cardify/blob/235f89608b8a899313761b44e6ed88e71dd114c0/src/main.ts#L2)
This import is only available from a Node.js runtime, which means that this will throw errors on Mobile (that's a problem since you have `isDesktopOnly` marked as `false` in your manifest.json file.)

[const ribbonIconEl = this.addRibbonIcon('dice', 'Cardify', async (evt: MouseEvent) => {](https://github.com/joshuakto/obsidian-cardify/blob/235f89608b8a899313761b44e6ed88e71dd114c0/src/main.ts#L38)
Would recommend using a different icon, you can use most from here: <https://lucide.dev/>, or add one manually.

[// Perform additional things with the ribbon](https://github.com/joshuakto/obsidian-cardify/blob/235f89608b8a899313761b44e6ed88e71dd114c0/src/main.ts#L90-L95)
Remove this sample code.

[checkCallback: (checking: boolean) => {](https://github.com/joshuakto/obsidian-cardify/blob/235f89608b8a899313761b44e6ed88e71dd114c0/src/main.ts#L118-L130)
You are missing a `return false`;

[plugin: MyPlugin;](https://github.com/joshuakto/obsidian-cardify/blob/235f89608b8a899313761b44e6ed88e71dd114c0/src/class/CardifySettingTabClass.ts#L5)
This is using the wrong class name



[export default class CardifyTutorialModal extends Modal {](https://github.com/joshuakto/obsidian-cardify/blob/235f89608b8a899313761b44e6ed88e71dd114c0/src/class/CardifyTutorialModalClass.ts#L3)
Remove this, as well as the two commands that reference it.

joshuakto:
Hi, thanks for the review! I have completed all action items following the advice (all changes in release 1.0.2). Please let me know if any follow-ups are needed.

liamcain:
- [fileContent](https://github.com/joshuakto/obsidian-cardify/blob/889a7c8e59c1ad30006528a6c0a940ff80aa336d/src/main.ts#L121) you should use the data passed into the editor-change event. So instead of reading the active file, I'd suggest you use editor.getValue(). That will be faster since you don't need to read the file from disk, it's already in memory.

PR #2432, opened 2023-09-11
Add plugin: Recipe view

lachholden:

## Repo URL

Link to my plugin: https://github.com/lachholden/obsidian-recipe-view

## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [X]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


liamcain:
- [toggleView(this: RecipeViewPlugin, checking: boolean) {](https://github.com/lachholden/obsidian-recipe-view/blob/6f759423664617de9d274e42a9e3f9c5acda332c/src/main.ts#L48) I recommend avoiding passing `this` as an argument since it shadows the `this` reference to the current plugin and makes it really confusing to trace in the code.
- [this.toggleView(c),](https://github.com/lachholden/obsidian-recipe-view/blob/6f759423664617de9d274e42a9e3f9c5acda332c/src/main.ts#L34) the function call signature looks wrong here since the first argument its expecting is the plugin
- [SampleSettingTab](https://github.com/lachholden/obsidian-recipe-view/blob/6f759423664617de9d274e42a9e3f9c5acda332c/src/main.ts#L94) rename this to something else
- [store.plugin.set(this.plugin);](https://github.com/lachholden/obsidian-recipe-view/blob/6f759423664617de9d274e42a9e3f9c5acda332c/src/recipe-view.ts#L31) this should probably happen in the Plugin itself, not in the View, which could get opened multiple times (and it will stay the same plugin)
- [inNewLeaf](https://github.com/lachholden/obsidian-recipe-view/blob/6f759423664617de9d274e42a9e3f9c5acda332c/src/recipe-view.ts#L47) you can use [`Keymap.isModEvent`](https://github.com/obsidianmd/obsidian-api/blob/009a10f196df551256a5894ea60fdf9ddaab19a1/obsidian.d.ts#L1639) instead

lachholden:
@liamcain Great, thanks for that! I've made those changes and included them in an 0.3.1 release.

lachholden:
@liamcain just checking – is there anything else I need to do to get this re-reviewed? Doesn't look like I can change any of the labels etc.

liamcain:
Looks good, sorry for the delay!

lachholden:
All good, thank you! :)

PR #2441, opened 2023-09-12
Add plugin: RescueTime

Tatz884:

## Repo URL

Link to my plugin:
https://github.com/Tatz884/RescueTime-Obsidian/
## Release Checklist
- [ X ] I have tested the plugin on
  - [ X ]  Windows
  - [ X ]  macOS
  - [ X ]  Linux
  - [ ]  iOS _(if applicable)_
- [ X ] My GitHub release contains all required files
  - [ X ] `main.js`
  - [ X ] `manifest.json`
  - [ X ] `styles.css` _(optional)_
- [ X ] GitHub release name matches the exact version number specified in my manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
- [ X ] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [ X ] My README.md describes the plugin's purpose and provides clear usage instructions.
- [ X ] I have read the tips in https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines and have self-reviewed my plugin to avoid these common pitfalls.
- [ X ] I have added a license in the LICENSE file.
- [ X ] My project respects and is compatible with the original license of any code from other plugins that I'm using.
- [ X ] I have given proper attribution to these other projects in my `README.md`.


joethei:
["description": "View your RescueTime data in obsidian.",](https://github.com/Tatz884/RescueTime-Obsidian/blob/b428a4a8b780a02082ee293d530bec5d1ba01c4c/manifest.json#L6)
`Obsidian` needs to be capitalized

["acorn": "^8.10.0",](https://github.com/Tatz884/RescueTime-Obsidian/blob/b428a4a8b780a02082ee293d530bec5d1ba01c4c/package.json#L16-L131)
Why do you have so many dependencies that you don't use, please remove all unnecessary dependencies

[// Remember to rename these classes and interfaces!](https://github.com/Tatz884/RescueTime-Obsidian/blob/b428a4a8b780a02082ee293d530bec5d1ba01c4c/main.ts#L11)
Remove this comment.

[// When registering intervals, this function will automatically clear the interval when the plugin is disabled.](https://github.com/Tatz884/RescueTime-Obsidian/blob/b428a4a8b780a02082ee293d530bec5d1ba01c4c/main.ts#L28-L29)
Remove this sample code

[//Detaches all leaves with the custom view.](https://github.com/Tatz884/RescueTime-Obsidian/blob/b428a4a8b780a02082ee293d530bec5d1ba01c4c/main.ts#L71-L72)
Don't detach leaves with your custom view, before creating a new leaf, this is an antipattern, see: <https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Don't+detach+leaves+in+%60onunload%60>

[export const RIGHT_PANE_VIEW = "right-pane-view";](https://github.com/Tatz884/RescueTime-Obsidian/blob/b428a4a8b780a02082ee293d530bec5d1ba01c4c/src/ui/view/RightPaneView.ts#L14)
Make this view id more unique.

[const barCategoryTitle = wrapper.createEl("div", { text: "Category vs Time Spent", cls: "barCategoryTitle" });](https://github.com/Tatz884/RescueTime-Obsidian/blob/b428a4a8b780a02082ee293d530bec5d1ba01c4c/src/ui/view/RightPaneView.ts#L53)

Tatz884:
Actually my question above was solved. I don't need to detach the leaf anymore. On `activateView()`, I activate the view only if the view is not activated yet `(!this.app.workspace.getLeavesOfType(RESCUE_TIME_RIGHT_PANE_VIEW).length)`.

liamcain:
- [throw new Error("Failed to fetch data today.");](https://github.com/Tatz884/RescueTime-Obsidian/blob/6e5894a02cb78ad822454dd5710f8b7ada45e1a9/main.ts#L36) you're throwing an error inside onload. This will means if the API fails when the plugin tries to load, the plugin will fail to load.
- [response](https://github.com/Tatz884/RescueTime-Obsidian/blob/6e5894a02cb78ad822454dd5710f8b7ada45e1a9/main.ts#L34) onload is called synchronously, which means that this API call will block Obsidian from loading until it gets a response from the rescuetime API. It appears to only be used for adding a statusbar element, so you should consider making this api call asynchronous so that the plugin load can happen more quickly
- [fetchData](https://github.com/Tatz884/RescueTime-Obsidian/blob/6e5894a02cb78ad822454dd5710f8b7ada45e1a9/main.ts#L34) it looks like this API call gets made even if apiToken is not set

Tatz884:
@liamcain Thank you for the review! I fixed the issues as per your comments.

PR #2447, opened 2023-09-14
Add plugin: Markdown table checkboxes

DylanGiesberts:

## Repo URL

Link to my plugin: https://github.com/DylanGiesberts/obsidian-table-checkboxes

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[let view: MarkdownView | null = this.app.workspace.getActiveViewOfType(MarkdownView);](https://github.com/DylanGiesberts/obsidian-table-checkboxes/blob/b1e84cb593f01420afbd485fc3e34912729ea413/main.ts#L6-L7)
Remove this and get a new reference to the view inside the events, your current approach won't work once there is more than one note open, or if the view that was open when the plugin was loaded is closed.

DylanGiesberts:
> [let view: MarkdownView | null = this.app.workspace.getActiveViewOfType(MarkdownView);](https://github.com/DylanGiesberts/obsidian-table-checkboxes/blob/b1e84cb593f01420afbd485fc3e34912729ea413/main.ts#L6-L7) Remove this and get a new reference to the view inside the events, your current approach won't work once there is more than one note open, or if the view that was open when the plugin was loaded is closed.

Changed, thanks for the review

liamcain:
- [evt.key == "]"](https://github.com/DylanGiesberts/obsidian-table-checkboxes/blob/c6226e15b48251c796e78f08403ee68f9b5c2f5d/main.ts#L8) you should do this check before ever calling getActiveViewOfType to avoid doing an extra instanceof check on every keystroke
- [view.editor](https://github.com/DylanGiesberts/obsidian-table-checkboxes/blob/c6226e15b48251c796e78f08403ee68f9b5c2f5d/main.ts#L9) you are better off using [`activeEditor`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4009) which will work inside a markdown file embedded within Canvas
- [instanceof HTMLInputElement](https://github.com/DylanGiesberts/obsidian-table-checkboxes/blob/c6226e15b48251c796e78f08403ee68f9b5c2f5d/main.ts#L25) this won't work inside [pop-out windows](https://help.obsidian.md/User+interface/Pop-out+windows), so you'll need to use `evt.target.instanceOf(HTMLInputElement)` instead
- [evt](https://github.com/DylanGiesberts/obsidian-table-checkboxes/blob/c6226e15b48251c796e78f08403ee68f9b5c2f5d/main.ts#L25) you should also confirm that the event target is within the active editor. For example, you might want to test to see what happens if you have the same file open in multiple tabs side-by-side
- [editor.replaceSelection](https://github.com/DylanGiesberts/obsidian-table-checkboxes/blob/c6226e15b48251c796e78f08403ee68f9b5c2f5d/main.ts#L17) this will only work in edit mode. In preview mode, these changes won't get saved.

DylanGiesberts:
> * [editor.replaceSelection](https://github.com/DylanGiesberts/obsidian-table-checkboxes/blob/c6226e15b48251c796e78f08403ee68f9b5c2f5d/main.ts#L17) this will only work in edit mode. In preview mode, these changes won't get saved.

Isn't this fine? It's meant to replace `- [ ]` with an HTML checkbox when you're typing it, which should always be in edit mode afaik?

DylanGiesberts:
@liamcain I cannot for the life of me get a `change` event to fire from a pop-out window. I've tried `this.registerDomEvent(activeWindow, "change", ...)`, `activeWindow.onchange = ...` and `activeWindow.addEventListener("change", ...)`. I also tried a select event, but to no avail.

Do you know if this is an issue within Obsidian?

Nevertheless thanks for the review, I've changed the other issues, minus the last bullet point.

liamcain:
You can't attach an event listener to a pop-up window until it exists. You will need to listen for the window-open event which is fired when a pop-out window is created. Add your event listener there.

DylanGiesberts:
Ahhh gotcha, I thought it happened automatically since the keyup event continued to work and with `activeWindow` existing.

Changed it to listen for `window-open`, thanks!

liamcain:
> Isn't this fine? It's meant to replace - [ ] with an HTML checkbox when you're typing it, which should always be in edit mode afaik?

Oh I see, yes that's fine.

PR #2458, opened 2023-09-19
Update taskbone plugin description

schlundd:
Update the plugin description to include additional features.


joethei:
You should consider splitting up your plugin into two.
I really don't like how this plugin does two totally unrelated things, OCR and Todoist Sync.

schlundd:
I don't like it either. On the other hand, it actually *is* the same plugin as both of them use the same backend service and different versions would only differ in their initial configuration.

How about this:

- I create seperate repositories
- I use these repos for the metadata (manifest, readme) and releases
- The code stays in one central repository that link to from the readme

This way I don't need to duplicate the code, the plugin repo stays clean and the plugins are actually discoverable.

liamcain:
> This way I don't need to duplicate the code

Since the concern is code duplication, I would suggest you handle that issue locally. Create a shared git submodule, yarn link a shared repo, make a private npm module, or even just symlink the shared code between the two plugins.

There are plenty of to handle this issue that don't involve reusing the same plugin.

schlundd:
I updated the description again and will open a seperate request for Todoist.

community-plugins.json
```diff
@@ -1471,8 +1471,8 @@
   },
   {
     "id": "taskbone-ocr-plugin",
-    "name": "Taskbone OCR",
-    "description": "Extract text from images and make it available for search.",
+    "name": "Taskbone",
+    "description": "Extract text and equations from images and pdfs and make it available for search.",
```
liamcain:
```suggestion
    "description": "Extract text and equations from images and PDFs and make it available for search.",
```

PR #2470, opened 2023-09-21
Add plugin: File Tree Generator

Unarray:

## Repo URL

Link to my plugin: https://github.com/Unarray/FileTreeGenerator

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["name": "📁 File Tree Generator",](https://github.com/Unarray/FileTreeGenerator/blob/cd5c52740ba2f5e9635ceec7c4019cfe09871ebd/manifest.json#L3)
Please remove the emoji

["isDesktopOnly": false](https://github.com/Unarray/FileTreeGenerator/blob/cd5c52740ba2f5e9635ceec7c4019cfe09871ebd/manifest.json#L9)
As your plugin is using the FS API, set this to true.

[export const sleep = (ms: number): Promise<void> => {](https://github.com/Unarray/FileTreeGenerator/blob/cd5c52740ba2f5e9635ceec7c4019cfe09871ebd/src/utils/sleep/sleep.ts#L1-L3)
The Obsidian API already includes this function.

Unarray:
Hello @joethei 👋 
- [x] remove emoji from name
- [x] use obsidian built-in `sleep` function

For the ["isDesktopOnly": false](https://github.com/Unarray/FileTreeGenerator/blob/cd5c52740ba2f5e9635ceec7c4019cfe09871ebd/manifest.json#L9) my plugin is mobile compatible. The button using the electron dialog [isn't rendered on mobile](https://github.com/Unarray/FileTreeGenerator/blob/cd5c52740ba2f5e9635ceec7c4019cfe09871ebd/src/modals/generate-tree/GenerateTree.ts#L65).
Should I do it some other way? Or is it fine as it is?

> [!NOTE]
> Release up to date!

joethei:
You are using the `fs` and `path´ packages in your plugin, just importing them will crash the app on mobile.

Also, you mention in your readme that the plugin will only have functionality on desktop, so setting that variable will mean the plugin can only be installed from desktop.

liamcain:
- [addCommands](https://github.com/Unarray/FileTreeGenerator/blob/ca320c7d34ef6f89ad4ec878ed4e1f7166455e3e/src/FileTreeGenerator.ts#L40) this abstraction seems unnecessary
- [addRibbonIconObject](https://github.com/Unarray/FileTreeGenerator/blob/ca320c7d34ef6f89ad4ec878ed4e1f7166455e3e/src/FileTreeGenerator.ts#L46) this one too
- [addRibbonIconObject](https://github.com/Unarray/FileTreeGenerator/blob/ca320c7d34ef6f89ad4ec878ed4e1f7166455e3e/src/FileTreeGenerator.ts#L46) same here
- [ingore](https://github.com/Unarray/FileTreeGenerator/blob/ca320c7d34ef6f89ad4ec878ed4e1f7166455e3e/src/modals/generate-tree/GenerateTree.ts#L41) typo
- [filesField.infoEl.style.flexGrow](https://github.com/Unarray/FileTreeGenerator/blob/ca320c7d34ef6f89ad4ec878ed4e1f7166455e3e/src/modals/generate-tree/GenerateTree.ts#L61) You should avoid assigning styles via javascript. Instead, just add a class to the element and move all these styles into CSS so that they are more easily adaptable by themes.
- [sep](https://github.com/Unarray/FileTreeGenerator/blob/ca320c7d34ef6f89ad4ec878ed4e1f7166455e3e/src/modals/generate-tree/GenerateTree.ts#L8) since "path" is a node module, I think this will fail on mobile where Node.JS isn't available.
- [fs/promises](https://github.com/Unarray/FileTreeGenerator/blob/ca320c7d34ef6f89ad4ec878ed4e1f7166455e3e/src/utils/path/path.ts#L2) this is unavailable on mobile. It probably doesn't make sense for this to be available on mobile anyway, so just change isDesktopOnly to true in your manifest.json

PR #2478, opened 2023-09-23
Add plugin: Sets

Canna71:

## Repo URL

Link to my plugin: https://github.com/Canna71/obsidian-sets

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


Bluemoondragon07:
Wow, this plugin looks amazing! I've always dreamed of an Anytype-based plugin; thank you for making it come true!

joethei:
[this.app.workspace.on(](https://github.com/Canna71/obsidian-sets/blob/3da76661137fdd5996c6c6bb90dcb76e78838e06/src/main.tsx#L73-L82)
Events need to be registered, so they can be properly unloaded when the plugin is disabled, or alternatively use the `.off´ function in `onunload` <https://docs.obsidian.md/Plugins/Events>

[id: "sets-open-sidebar",](https://github.com/Canna71/obsidian-sets/blob/3da76661137fdd5996c6c6bb90dcb76e78838e06/src/main.tsx#L125) [id: "sets-new-type",](https://github.com/Canna71/obsidian-sets/blob/3da76661137fdd5996c6c6bb90dcb76e78838e06/src/main.tsx#L133), and more

[name: "Open Sets Sidebar",](https://github.com/Canna71/obsidian-sets/blob/3da76661137fdd5996c6c6bb90dcb76e78838e06/src/main.tsx#L126), [name: "Create New Type",](https://github.com/Canna71/obsidian-sets/blob/3da76661137fdd5996c6c6bb90dcb76e78838e06/src/main.tsx#L134) and many more

[this.app.workspace.detachLeavesOfType(SETS_VIEW);](https://github.com/Canna71/obsidian-sets/blob/3da76661137fdd5996c6c6bb90dcb76e78838e06/src/main.tsx#L247)

[this.app.workspace.detachLeavesOfType(SETS_VIEW);](https://github.com/Canna71/obsidian-sets/blob/3da76661137fdd5996c6c6bb90dcb76e78838e06/src/main.tsx#L276)
Also don't detach here, instead, check if there is already a leaf, and either do nothing, or focus on it, depending on where this is called from.

[obsidian-ex.d.ts](https://github.com/Canna71/obsidian-sets/blob/3da76661137fdd5996c6c6bb90dcb76e78838e06/src/obsidian-ex.d.ts)
Please credit the source of this file.

[addRibbonIcon: boolean;](https://github.com/Canna71/obsidian-sets/blob/3da76661137fdd5996c6c6bb90dcb76e78838e06/src/Settings.ts#L17)

[app.workspace.getActiveViewOfType(MarkdownView);](https://github.com/Canna71/obsidian-sets/blob/3da76661137fdd5996c6c6bb90dcb76e78838e06/src/Utils/saveDataIntoBlock.ts#L10)

Canna71:
Thanks a lot for your validation and hard work, most of all because I could have avoided most of that in the first place.
I guess I fixed that now...

liamcain:
- [this.app.workspace.off](https://github.com/Canna71/obsidian-sets/blob/cd9baed210d44061beeb8adec71dc500ec706c92/src/main.tsx#L242) you already registered the event handlers with the plugin, which means they will be removed automatically when the plugin is disabled. You can remove these .off() calls.
- [patchView](https://github.com/Canna71/obsidian-sets/blob/cd9baed210d44061beeb8adec71dc500ec706c92/src/main.tsx#L89) this looks unused. Also I'm not sure what it does, is it copied from another plugin?
- [app.vault.adapter.files](https://github.com/Canna71/obsidian-sets/blob/cd9baed210d44061beeb8adec71dc500ec706c92/src/Utils/getFolders.ts#L5) [`getAllLoadedFiles`](https://github.com/obsidianmd/obsidian-api/blob/902badd38ba907689f0917d7b193f7e33d1284fe/obsidian.d.ts#L3684) and loop through with instanceof TFolder

PR #613, opened 2021-10-31
Add Typing plugin

konodyuk:
# I am submitting a new Community Plugin

## Repo URL

Link to my plugin: https://github.com/konodyuk/obsidian-typing

## Release Checklist

- [x] I have tested this on Windows, macOS, and Linux _(if applicable)_
- [x] My GitHub release contains all required files
  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css` _(optional)_
- [x] GitHub release name matches the exact version number specified in my manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
- [x] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [x] README clearly describes the plugins purpose and provides clear usage instructions.
- [x] I have added a license in the LICENSE file.

PR #2481, opened 2023-09-23
Add plugin: Typing

konodyuk:

## Repo URL

Link to my plugin: https://github.com/konodyuk/obsidian-typing
Link to my docs: https://konodyuk.github.io/obsidian-typing/

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
  - The documentation is temporarily serving the purpose of README. The README will be updated once the docs are established
    - Note: The plugin does not use the code of any other project, although it might share functionality with some of them. Later, I will add a list of plugins that implement similar functionality to my docs.


konodyuk:
Here is my previous PR for reference: https://github.com/obsidianmd/obsidian-releases/pull/613

konodyuk:
After a long pause and a complete reimplementation, I am ready to resubmit the plugin and continue (or more likely restart) the review process.

I have a couple of comments to assist in reviewing the codebase, because the plugin is quite large:
1. The code communicating with Obsidian API is mostly under `src/middleware`, while `src/language`, `src/typing`, and `src/ui` mostly implement pure logic (60% codebase).
2. I have a couple of questions regarding my implementation of marginals, i.e. headers and footers automatically inserted to the top and bottom of each note configured so.

The current implementation ([`src/middleware/marginal_rendering.tsx`](https://github.com/konodyuk/obsidian-typing/blob/master/src/middleware/marginal_rendering.tsx)) inserts them differently into notes in markdown view and hover previews/embeds.
1. Markdown view: I inject codeblocks like
````
```typing-header
```
````
to the top and end of content and then process them. This is needed because if I simply insert them in a markdown postprocessor after the element whose underlying markdown content ends on the last line of the document, then when the user appends some text to the note and that markdown block is no longer the last one, it still won't be rerendered. Thus I insert these codeblocks to markdown content.
This method slightly breaks the highlighting when navigating to a linked block or line in the file, because it adds two lines and the highlighting is shifted 2 lines.

2. Hover preview/embed: since I cannot inject anything into content here, I just insert my elements before the element corresponding to the first markdown block in the document, and after the last. This works, as in preview/transclusion the document is mostly not modified. However, there are some cases when this method fails (like when a note ends with a list or nested list).

What do you think is the best way to implement the headers/footers and what are the possible ways to fix the highlighted problems? In any case, my current implementation works fine and the issues are minor.

Attaching the current documentation (not complete): https://konodyuk.github.io/obsidian-typing/
Please refer to documentation for better understanding of the plugin purpose and features and treat it as a primary source of truth. The README will be updated later and will be based on documentation.

joethei:
Please include at least some basic documentation as to what your plugin does into the readme.

[Copyright (c) 2021 Nikita Konodyuk](https://github.com/konodyuk/obsidian-typing/blob/9815c53ef5abf60c0fbf4f4eb333b94ca0059ffe/LICENSE#L3)
Update the year

[id: "typing-actions",](https://github.com/konodyuk/obsidian-typing/blob/9815c53ef5abf60c0fbf4f4eb333b94ca0059ffe/src/commands.tsx#L41)

[return await this.vault.read(tfile as TFile);](https://github.com/konodyuk/obsidian-typing/blob/9815c53ef5abf60c0fbf4f4eb333b94ca0059ffe/src/utilities/module_manager_sync.ts#L190)
No need to cast here, because you have the check above Typescript already interprets it as a `TFile`

[this.vault.on("modify", async ({ path }) => {](https://github.com/konodyuk/obsidian-typing/blob/9815c53ef5abf60c0fbf4f4eb333b94ca0059ffe/src/utilities/module_manager_sync.ts#L204-L226)
Register the events so they are properly unloaded when the plugin is disabled: <https://docs.obsidian.md/Plugins/Events>

[return `.obsidian/plugins/${gctx.plugin.manifest.id}/font-cache/`;](https://github.com/konodyuk/obsidian-typing/blob/9815c53ef5abf60c0fbf4f4eb333b94ca0059ffe/src/utilities/font_manager.ts#L58)
Users can configure where config files are stored, `.obsidian` is not hard-coded.
You can read the full path for the plugin folder from `plugin.manifest.dir`.

[let folder = gctx.app.vault.getAbstractFileByPath(this.folder) as TFolder;](https://github.com/konodyuk/obsidian-typing/blob/9815c53ef5abf60c0fbf4f4eb333b94ca0059ffe/src/typing/type.ts#L93)
Don't cast blindly, always check if `instanceof TFolder` before using.

[await gctx.app.workspace.getLeaf().openFile(gctx.app.vault.getAbstractFileByPath(this.path) as TFile);](https://github.com/konodyuk/obsidian-typing/blob/9815c53ef5abf60c0fbf4f4eb333b94ca0059ffe/src/typing/note.tsx#L249)
Same here, always check if what `getAbstractFileByPath` returns is what you expect.


[containerEl.createEl("h2").innerText = "Safety";](https://github.com/konodyuk/obsidian-typing/blob/9815c53ef5abf60c0fbf4f4eb333b94ca0059ffe/src/settings/index.tsx#L57)
It's better to use either: `containerEl.createEl("h2", {text: "Your text"})` or `containerEl.createEl("h2").setText("Safety");`

[new Setting(containerEl).setName("OTL Imports Path")](https://github.com/konodyuk/obsidian-typing/blob/9815c53ef5abf60c0fbf4f4eb333b94ca0059ffe/src/settings/index.tsx#L82)

[plugin.register(() => {](https://github.com/konodyuk/obsidian-typing/blob/9815c53ef5abf60c0fbf4f4eb333b94ca0059ffe/src/middleware/title_bar.tsx#L110-L112)
This is not required, as you are already registering the event above.



---

[export function debounce(callback: Callback, wait: number): Callback & { cancel: () => void } {](https://github.com/konodyuk/obsidian-typing/blob/9815c53ef5abf60c0fbf4f4eb333b94ca0059ffe/src/utilities/rate_control.ts#L29-L53)
There is already a function with the same name & function in the Obsidian API.

[addIcon(](https://github.com/konodyuk/obsidian-typing/blob/9815c53ef5abf60c0fbf4f4eb333b94ca0059ffe/src/main.tsx#L46-L50)
This icon looks pretty similar to the already included `grid-2x2`, maybe using that makes sense here.

konodyuk:
Thank you for the review! I updated the plugin in line with your comments, so I hope we can proceed

liamcain:
- [instanceof HTMLAnchorElement](https://github.com/konodyuk/obsidian-typing/blob/1fdce41e2238e45bfe3769b2da4576173a3ef624/src/middleware/link_rendering.tsx#L64) you'll need to use linkEl.instanceOf(HTMLAnchorElement) instead for this to work in [pop-out windows](https://help.obsidian.md/User+interface/Pop-out+windows). Or just check el.tagName === 'A'
- [getActiveViewOfType](https://github.com/konodyuk/obsidian-typing/blob/1fdce41e2238e45bfe3769b2da4576173a3ef624/src/context.ts#L35) for getting the current file, you might want to use [`getActiveFile`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4146) instead. This will help if the active leaf is actually something like the backlinks view (which is still associated with a file)

liamcain:
> What do you think is the best way to implement the headers/footers and what are the possible ways to fix the highlighted problems? In any case, my current implementation works fine and the issues are minor.

I'm not sure I understand the issue, sorry!

PR #2484, opened 2023-09-24
Add plugin: Auto Filename

rcsaquino:
Auto Filename is an Obsidian.md plugin that automatically renames files in Obsidian based on the first x characters of the file, saving you time and effort.

## Repo URL

Link to my plugin: https://github.com/rcsaquino/obsidian-auto-filename

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [x]  Android
  - [ ]  iOS
  - [x] `main.js`
  - [x] `manifest.json`
  - [x] `styles.css`


joethei:
["description": "Auto Filename is an Obsidian.md plugin that automatically renames files in Obsidian based on the first x characters of the file, saving you time and effort.",](https://github.com/rcsaquino/obsidian-auto-filename/blob/c8282e91d2424fac4e318eb36a549f5f48bf926d/manifest.json#L6-L7)
Avoid including sentences like "xyz is a plugin for Obsidian that does " in your description.


[this.app.vault.on("modify", (abstractFile) => {](https://github.com/rcsaquino/obsidian-auto-filename/blob/c8282e91d2424fac4e318eb36a549f5f48bf926d/src/main.ts#L63-L78)
Register your events: <https://docs.obsidian.md/Plugins/Events>, to make sure they get properly unloaded when your plugin is disabled.
Then you can remove your entire `onunload` function.

rcsaquino:
> ["description": "Auto Filename is an Obsidian.md plugin that automatically renames files in Obsidian based on the first x characters of the file, saving you time and effort.",](https://github.com/rcsaquino/obsidian-auto-filename/blob/c8282e91d2424fac4e318eb36a549f5f48bf926d/manifest.json#L6-L7) We have published a guide for plugin descriptions to follow here: https://docs.obsidian.md/Plugins/Releasing/Submission+requirements+for+plugins#Keep+plugin+descriptions+short+and+simple. Avoid including sentences like "xyz is a plugin for Obsidian that does " in your description.
> 
> [this.app.vault.on("modify", (abstractFile) => {](https://github.com/rcsaquino/obsidian-auto-filename/blob/c8282e91d2424fac4e318eb36a549f5f48bf926d/src/main.ts#L63-L78) Register your events: https://docs.obsidian.md/Plugins/Events, to make sure they get properly unloaded when your plugin is disabled. Then you can remove your entire `onunload` function.

Thank you for your feedback. I have updated my plugin description and code to follow your suggestions. I have also added comments to make the code easier to read.

Please let me know if you have any other questions or concerns.

rcsaquino:
Hi! I see that my proposed plugin has conflicts but it said only those with write access to this repository can merge pull requests. Should I do something about this or is this a message for collaborators?

joethei:
We will handle the conflicts once the plugin is ready to be merged.

rcsaquino:
Got it. Thanks!

rcsaquino:
Hi! Just checking in for updates.

liamcain:
- [off](https://github.com/rcsaquino/obsidian-auto-filename/blob/973f736f90275fc11fd5b5af6e687a0db08c12b3/src/main.ts#L104) No need to call off() here. You already registered the event handler, so it will get removed automatically. Also this call is wrong, since you're not passing the same eventref into off() that is passed into on().
- [hideInlineTitle](https://github.com/rcsaquino/obsidian-auto-filename/blob/973f736f90275fc11fd5b5af6e687a0db08c12b3/src/main.ts#L16) this is way more complicated than it needs to be. You shouldn't need to create your own <style> element to manage. You can put the styles into style.css which will be loaded with your plugin.

I suggest creating a class called "hide-inline-title" and applying that class to the document.body, instead of adding and removing the style tag.

- [content.split("");](https://github.com/rcsaquino/obsidian-auto-filename/blob/973f736f90275fc11fd5b5af6e687a0db08c12b3/src/main.ts#L30) don't do this. For larger files, this will be incredibly expensive. Strings are already lists of characters, so you can access characters using content[i] without needing chars.
- [this.renameFile(abstractFile);](https://github.com/rcsaquino/obsidian-auto-filename/blob/973f736f90275fc11fd5b5af6e687a0db08c12b3/src/main.ts#L78) I don't understand, why are you renaming the file on "modify"? 
- [await this.app.fileManager.renameFile(file, newPath);](https://github.com/rcsaquino/obsidian-auto-filename/blob/973f736f90275fc11fd5b5af6e687a0db08c12b3/src/main.ts#L54) it doesn't look like you're even checking to see if the file should/needs to be renamed. You should also debounce renameFile since modify could be called frequently and renaming files can be expensive, especially if there are a lot of files that link to it.
-

rcsaquino:
I appreciate your feedback, @liamcain. I have implemented some modifications based on your suggestions.

> [off](https://github.com/rcsaquino/obsidian-auto-filename/blob/973f736f90275fc11fd5b5af6e687a0db08c12b3/src/main.ts#L104) No need to call off() here.

✅ Removed unecessary `off()` calls.

> [hideInlineTitle](https://github.com/rcsaquino/obsidian-auto-filename/blob/973f736f90275fc11fd5b5af6e687a0db08c12b3/src/main.ts#L16) this is way more complicated than it needs to be.

✅ Modified how "Hide inline title" setting is implemented. Now uses [styles.css](https://github.com/rcsaquino/obsidian-auto-filename/blob/main/src/styles.css) `.hide-inline-title` which is applied to the `.inline-title` div programatically.

> [content.split("");](https://github.com/rcsaquino/obsidian-auto-filename/blob/973f736f90275fc11fd5b5af6e687a0db08c12b3/src/main.ts#L30) don't do this.

✅ Modified `renameFile()` to remove unecessary use of `content.split("")`.

> [this.renameFile(abstractFile);](https://github.com/rcsaquino/obsidian-auto-filename/blob/973f736f90275fc11fd5b5af6e687a0db08c12b3/src/main.ts#L78) I don't understand, why are you renaming the file on "modify"?

- Renaming the file on “modify” is used to automatically update the filename as the user edits a markdown file. This gives better visual feedback and thus provides a better user experience than other methods (such as renaming on close, renaming on open, etc).

> [await this.app.fileManager.renameFile(file, newPath);](https://github.com/rcsaquino/obsidian-auto-filename/blob/973f736f90275fc11fd5b5af6e687a0db08c12b3/src/main.ts#L54) it doesn't look like you're even checking to see if the file should/needs to be renamed. 

- Checks are done at the start of the function at [lines 27-28](https://github.com/rcsaquino/obsidian-auto-filename/blob/main/src/main.ts#L27).
- [Another check](https://github.com/rcsaquino/obsidian-auto-filename/blob/main/src/main.ts#L71) is added to prevent unecessary renaming.

> You should also debounce renameFile

✅ [Debounced](https://github.com/rcsaquino/obsidian-auto-filename/blob/main/src/main.ts#L31) the `renameFile()` function. A [setting](https://github.com/rcsaquino/obsidian-auto-filename/blob/main/src/main.ts#L218) has been added to further tweak this.

rcsaquino:
Thanks again for the feedback, @liamcain.

> [this.app.vault.read](https://github.com/rcsaquino/obsidian-auto-filename/blob/8fdd8e54f6e3859993053b02f2f706c6c8a7fdd8/src/main.ts#L50) Reading the file can be expensive. The "modify" event passes data: string into the event handler. You should use that instead of reading the file.

- `this.app.vault.read` allows getting the first n characters of the markdown file. The first n characters of the md file would then be used as the new filename. `this.app.vault.modify` dosen't return anything. It only modifies the content of the md file which is not what's needed on this case.

> [setTimeout](https://github.com/rcsaquino/obsidian-auto-filename/blob/8fdd8e54f6e3859993053b02f2f706c6c8a7fdd8/src/main.ts#L39) you can use [debounce](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L772) from the Obsidian API instead of implementing it yourself.

- Obsidian's default debounce function does not allow for the nuances required by this plugin. For example, if a user edits a file and then switches to another file during the debounce timeout, the plugin will only perform filename changes on the current file, even if they are also needed on the previous file. The custom implementation [prevents that](https://github.com/rcsaquino/obsidian-auto-filename/blob/main/src/main.ts#L33) from happening.

rcsaquino:
Thanks again for the feedback, @liamcain.

> [this.app.vault.read](https://github.com/rcsaquino/obsidian-auto-filename/blob/8fdd8e54f6e3859993053b02f2f706c6c8a7fdd8/src/main.ts#L50) Reading the file can be expensive.

- [`cachedRead()`](https://github.com/rcsaquino/obsidian-auto-filename/blob/main/src/main.ts#L51) is now used instead of `read()` for better performance.

> The "modify" event passes data: string into the event handler. You should use that instead of reading the file.

- `this.app.vault.cachedRead` returns the content of the file and therefore allows getting its first n characters. This would then be used as the new filename. `this.app.vault.modify` dosen't return anything. It only modifies the content of the md file which is not what's needed on this case.

> [setTimeout](https://github.com/rcsaquino/obsidian-auto-filename/blob/8fdd8e54f6e3859993053b02f2f706c6c8a7fdd8/src/main.ts#L39) you can use [debounce](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L772) from the Obsidian API instead of implementing it yourself.

- Obsidian's default debounce function does not allow for the nuances required by this plugin. For example, if a user edits a file and then switches to another file during the debounce timeout, the plugin will only perform filename changes on the current file, even if they are also needed on the previous file. The custom implementation [prevents that](https://github.com/rcsaquino/obsidian-auto-filename/blob/main/src/main.ts#L33) from happening.

PR #2514, opened 2023-10-02
Add plugin: Gladdis

AurelienStebe:

## Repo URL

Link to my plugin: https://github.com/AurelienStebe/Gladdis

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[import path from 'path'](https://github.com/AurelienStebe/Gladdis/blob/01ecea4fe478d1d92f4f127b827b893364c3a3c1/src/commands.ts#L1-L2)

AurelienStebe:
Apologies, I should have noted in my PR that the `commands.ts` & `service.ts` source files are not part of the Obsidian plugin release. They are for the NPM release, making Obsidian a kind of IDE to write and test the AI prompts, which users can then integrate as-is in any software via the NPM package and the NodeJS API.

You can see the implementation of the `DiskInterface` with Obsidian's Vault API in the [`obsidian.ts` source file, line 176](https://github.com/AurelienStebe/Gladdis/blob/01ecea4fe478d1d92f4f127b827b893364c3a3c1/src/obsidian.ts#L176).

However, I just noticed that I'm still importing `yaml`, which Obsidian already includes. I'm working on a fix right now to exclude it from the Obsidian release. I'll need a similar solution later for `PDF.js` & `Turndown`, in any case. I will also remove the `dotenv` `parse()` import, which unnecessarily requires `path` & `fs`.

AurelienStebe:
I just pushed [a new release](https://github.com/AurelienStebe/Gladdis/releases/tag/0.4.3) fixing the dependencies issues I mentioned. I replaced `dotenv` with a custom `parse()` function and removed `yaml` from the bundle by aliasing it to the Obsidian exported utility functions. There should be only 3 dependencies left in `main.js`: the `deepmerge` utility, the OpenAI NodeJS API and a big JSON file required for the token counter. I plan on making this file an optional download with a button in the settings panel, but it won't be for a few versions.

PR #2519, opened 2023-10-07
Add plugin: Fountain Editor

chuangcaleb:

## Repo URL

Link to my plugin: https://github.com/chuangcaleb/obsidian-fountain-editor

## Notes

There is a [Darakah/obsidian-fountain: Obsidian plugin to edit, write and render Fountain Writing Syntax for screenplays and scripts (implements fountain-js project to parse fountain script).](https://github.com/Darakah/obsidian-fountain) plugin that is very similar.

  - Formats Fountain syntax, just like this plugin
  - But for preview mode only, and must be wrapped in a code block
  - 2.5 years+ old project, hasn't been updated since

True that codemirror 6 wasn't as matured back then as it is now. But this plugin also extends extra functionality, and doesn't actually crossover into preview mode at all.

ALSO I plan to make perf improvements over time, but I'd like to get a public community release out first to allow more people to access and test out this plugin. Comments appreciated!

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(works, need styling enhancements)_
  - [x] `main.js`
  - [x] `manifest.json`


community-plugins.json
```diff
@@ -8564,7 +8564,8 @@
     "author": "Marius Wörfel",
     "description": "Allows you to measure time.",
     "repo": "Raboro/obsidian-timer-plugin"
-  },
+
+  },
```
liamcain, line 8568:
```suggestion
  }, 
```

chuangcaleb:
Thank you for reviewing and merging :DDDDDD AODSISAJDJAS

PR #2523, opened 2023-10-08
Add plugin: Remotely Secure

sboesen:

## Repo URL

Link to my plugin: https://github.com/sboesen/remotely-secure

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


sboesen:
Updated to enable github issues on the repo.

joethei:
What sets your plugin apart from the original remotely-save plugin?
Please make that clear in the readme.

[log.info(plan.mixedStates); // for debugging](https://github.com/sboesen/remotely-secure/blob/1d200370f6a3b493abb134945b0593e7619bb062/src/main.ts#L302)

[`${this.app.vault.configDir}/plugins/${this.manifest.dir}`;](https://github.com/sboesen/remotely-secure/blob/1d200370f6a3b493abb134945b0593e7619bb062/src/main.ts#L1020)
This path is wrong, `manifest.dir` already includes the rest.

[import * as path from "path";](https://github.com/sboesen/remotely-secure/blob/1d200370f6a3b493abb134945b0593e7619bb062/src/misc.ts#L2)

[return window.moment(x).format() as string;](https://github.com/sboesen/remotely-secure/blob/1d200370f6a3b493abb134945b0593e7619bb062/src/misc.ts#L345)
Import `moment` from the Obsidian API instead of taking it from `window`

[await fetch](https://github.com/sboesen/remotely-secure/blob/1d200370f6a3b493abb134945b0593e7619bb062/src/remoteForDropbox.ts#L215)

[const rsp2 = JSON.parse(rsp1);](https://github.com/sboesen/remotely-secure/blob/1d200370f6a3b493abb134945b0593e7619bb062/src/remoteForOnedrive.ts#L141)
Use `requestUrl` above, and then you can use `rsp1.json()` instead of this code.

[containerEl.createEl("h1", { text: "Remotely Secure" });](https://github.com/sboesen/remotely-secure/blob/1d200370f6a3b493abb134945b0593e7619bb062/src/settings.ts#L675)

[const { eye, eyeOff } = getEyesElements();](https://github.com/sboesen/remotely-secure/blob/1d200370f6a3b493abb134945b0593e7619bb062/src/settings.ts#L642-L651)
[Avoid using `innerHTML`](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Avoid+%60innerHTML%60%2C+%60outerHTML%60+and+%60insertAdjacentHTML%60), instead use the `setIcon` function from the Obsidian API, no need to import lucide yourself.

[await mkdirpInVault(r.key, vault); /* should be unnecessary */](https://github.com/sboesen/remotely-secure/blob/1d200370f6a3b493abb134945b0593e7619bb062/src/sync.ts#L1129)
Why is this done if it's unnecessary?

sboesen:
Thank you for the review and feedback! I've updated the [readme](https://github.com/sboesen/remotely-secure#remotely-secure) to call out the differences more clearly and am taking the time to update the plugin to address these issues. Will post another comment once I have an updated PR.

sboesen:
Updated to address your feedback, please let me know if there is any additional feedback. I broke up the commits for each piece of feedback called out. Please see code at [679c1429](https://github.com/sboesen/remotely-secure/commit/679c1429abee8b15460d6bf55ebc99c15874b2a6) for latest version.

PR #2525, opened 2023-10-09
Add plugin: Scholar

lolipopshock:

## Repo URL

Link to my plugin: https://github.com/lolipopshock/obsidian-scholar

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


lolipopshock:
Fix in https://github.com/lolipopshock/obsidian-scholar/releases/tag/v1.2.1

joethei:
[import * as path from "path";](https://github.com/lolipopshock/obsidian-scholar/blob/80ca9ca68fb8cd25165e9cef575debb2fd585297/src/main.ts#L48), [import { exec } from "child_process";](https://github.com/lolipopshock/obsidian-scholar/blob/80ca9ca68fb8cd25165e9cef575debb2fd585297/src/obsidianScholar.ts#L13)

[this.keyListener = async (event: KeyboardEvent) => {](https://github.com/lolipopshock/obsidian-scholar/blob/80ca9ca68fb8cd25165e9cef575debb2fd585297/src/main.ts#L347-L378)
Instead of adding a listener here, I would recommend using scopes.
```ts
this.scope.register(['Shift'], 'Tab', () => {
	// your code here.
})
```

[text: "SemanticScholar Search Results",](https://github.com/lolipopshock/obsidian-scholar/blob/80ca9ca68fb8cd25165e9cef575debb2fd585297/src/main.ts#L405)

[file = this.app.vault.getAbstractFileByPath(file) as TFile;](https://github.com/lolipopshock/obsidian-scholar/blob/80ca9ca68fb8cd25165e9cef575debb2fd585297/src/obsidianScholar.ts#L126)
Don't do this blindly, always do a `instanceof` check before converting types.

[containerEl.createEl("h2", { text: SETTING_HEADER });](https://github.com/lolipopshock/obsidian-scholar/blob/80ca9ca68fb8cd25165e9cef575debb2fd585297/src/settingsTab.ts#L59)

lolipopshock:
Thanks, @joethei ! 

https://github.com/lolipopshock/obsidian-scholar/pull/7 fixes the issues you've mentioned above. 

As for this issue: 
> [import * as path from "path";](https://github.com/lolipopshock/obsidian-scholar/blob/80ca9ca68fb8cd25165e9cef575debb2fd585297/src/main.ts#L48), [import { exec } from "child_process";](https://github.com/lolipopshock/obsidian-scholar/blob/80ca9ca68fb8cd25165e9cef575debb2fd585297/src/obsidianScholar.ts#L13)
These imports are only available from a Node.js runtime, which means that this will throw errors on Mobile (that's a problem since you have isDesktopOnly marked as false in your manifest.json file.

I removed `path` since it's not actually used. There's only one desktop specific command that uses `child_process` (open PDF in system app). And I guard that command with a platform checking: https://github.com/lolipopshock/obsidian-scholar/blob/18d46a3e34dc315609b885b3cfb970caf0ccb305/src/obsidianScholar.ts#L268 . Hope that helps!

lolipopshock:
@joethei  in case you missed it

liamcain:
- [modalInstructionsHTML](https://github.com/lolipopshock/obsidian-scholar/blob/82e6424d8380837eb006cd2e33c7d82c3ae03ef1/src/main.ts#L681) no need to generate the HTML yourself, you can just use [setInstructions](https://github.com/obsidianmd/obsidian-api/blob/791214a68d0dc322b88e5abce617bdf603cc2a2d/obsidian.d.ts#L1174)
- [this.keyListener](https://github.com/lolipopshock/obsidian-scholar/blob/82e6424d8380837eb006cd2e33c7d82c3ae03ef1/src/main.ts#L537) you can use Modal.scope instead and just add a scope handler to Tab and Shift Tab

community-plugins.json
```diff
@@ -8952,5 +8952,11 @@
...
     "repo": "solargate/obsidian-inline-encrypter"
-  }
+  },
+  {
+    "id": "scholar",
+    "name": "Scholar",
+    "author": "Shannon Shen",
+    "description": "Manage research library and streamline research workflow.",
+    "repo": "lolipopshock/obsidian-scholar"
 ]
```
liamcain, line 8962:
```suggestion
    "repo": "lolipopshock/obsidian-scholar"
}
]
```

PR #2534, opened 2023-10-13
Add plugin: Blockier

blorbb:

## Repo URL

Link to my plugin: https://github.com/blorbb/obsidian-blockier

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [this.registerDomEvent(document, "keydown", (ev) => {](https://github.com/blorbb/obsidian-blockier/blob/b04c013c0c1740513bca7d7cdd1fea73f4d87e6f/src/main.ts#L49) this won't work in [pop-out windows](https://help.obsidian.md/User+interface/Pop-out+windows). The better way to handle this would be adding a Codemirror extension that listens for "space."

```ts
	let ext = keymap.of([
		{
			key: 'Space',
			run: (view) => tryReplaceBlock(view),
		},
	]);
```

- [hotkeys: [{ modifiers: ["Mod"], key: "a" }],](https://github.com/blorbb/obsidian-blockier/blob/b04c013c0c1740513bca7d7cdd1fea73f4d87e6f/src/main.ts#L43) This is going to prevent users using mod-A anywhere in the app. This will break things like selecting all properties. Or selecting all text inside a textfield in Settings. Instead of adding a default hotkey, you should add an editor extension that looks for mod-A. I believe the Outliner plugin does something like this if you're looking for examples.

blorbb:
@liamcain thanks for the feedback! I'm not too familiar with the APIs you suggested, so lmk if there's anything else I need to change (should I have _removed_ the ctrl-a command in favour of `registerEditorExtension`?).

liamcain:
> (should I have removed the ctrl-a command in favour of registerEditorExtension?).

You might still want to include a Command that doesn't have a hotkey associated with it. That way users can provide their own hotkey too if they want.

PR #2551, opened 2023-10-16
Add plugin: Global Markdown Encryption

shlemiel:

## Repo URL

Link to my plugin: https://github.com/shlemiel/globaloe

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
Please fix the warnings pointed out by the bot above.

[const buffer = require('buffer');
const crypto = require('crypto');](https://github.com/shlemiel/globaloe/blob/7d1855baa5ee17a687d2a2c35f9c9492f41854b9/main.ts#L1C1-L2C34)

[MyPlugin](https://github.com/shlemiel/globaloe/blob/7d1855baa5ee17a687d2a2c35f9c9492f41854b9/main.ts#L125C22-L125C31)
Please rename this class.

shlemiel:
Thank you for the review. I changed them all.

Below are list of the applied changes:

1. Plugin name is changed to `global-markdown-encrypt`
2. `isDesktopOnly` is now `true` (I forgot this)
3. `MyPlugin` -> `GlobalMarkdownEncrypt`
4. `authorUrl` is changed to my github page, not repository

shlemiel:
I resolved the conflict though, is it correct to be manually doing this?

PR #2552, opened 2023-10-17
Add plugin: GChat Reminder

anil-e:

## Repo URL

Link to my plugin: https://github.com/anil-e/obsidian_gchat_plugin

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[MyPlugin](https://github.com/anil-e/obsidian_gchat_plugin/blob/23d547c6ca9404f871f15e0236f7c066bbb9d64d/main.ts#L11)
Rename this class.

[async checkAllFilesForTasks() {](https://github.com/anil-e/obsidian_gchat_plugin/blob/23d547c6ca9404f871f15e0236f7c066bbb9d64d/main.ts#L52-L58)
This will cause performance issues, especially in larger vaults.
Instead, use the MetadataCache to retrieve all tasks, cache the results, and listen to the events provided on MetadataCache to receive updates to the list of tags.
Then, only check through that list of tasks every few minutes.

[const response = await fetch(](https://github.com/anil-e/obsidian_gchat_plugin/blob/23d547c6ca9404f871f15e0236f7c066bbb9d64d/main.ts#L125-L134)

anil-e:
Thanks for the review and sorry for the late response @joethei, unfortunately i didn't make it earlier.

I have now implemented the requested changes, hope it meets your expectations :)

liamcain:
- [read](https://github.com/anil-e/obsidian_gchat_plugin/blob/4cf1697db960597a8acd7fb5fbd6bd83f2169851/src/main.ts#L56) you should use [cachedRead](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3373) here instead

PR #2558, opened 2023-10-18
Add plugin: ManicTime

manictime:

## Repo URL

Link to my plugin: https://github.com/manictime/manictime-obsidian

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["description": "This extension sends path of active file to locally installed ManicTime client",](https://github.com/manictime/manictime-obsidian/blob/2cdbaf6628170cb26cffcbef18260142311dceff/manifest.json#L6)
Avoid including sentences like "This is a plugin for Obsidian that does ..." in your description.

[this.app.workspace.on('active-leaf-change', () => this.sendData());](https://github.com/manictime/manictime-obsidian/blob/2cdbaf6628170cb26cffcbef18260142311dceff/main.ts#L14-L17)

manictime:
The issues have been fixed.

liamcain:
- [serverManager](https://github.com/manictime/manictime-obsidian/blob/8279924e36d71c4275c4b5daf226b72715aa2522/main.ts#L55) Server manager shouldn't be wrapped in a IIFE. This means that it's lifecycle management is completely independent from the plugin. If the plugin is disabled, it won't be properly cleaned up. Instead, you should either move this into the Plugin class, or make a separate class with a cleanup function to ensure that the interval is removed when the plugin is disabled.

tilen-fiser:
Issue was fixed. The interval is removed when plugin is disabled.

PR #2559, opened 2023-10-18
Add plugin: JSON table

dario-baumberger:

## Repo URL

Link to my plugin:
https://github.com/dario-baumberger/obsidian-json-table

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


dario-baumberger:
I have tested the plugin on iOS. It is a bit cumbersome because you lose the selection in the editor when you open the command palette. But you can add the commands of the plugin to the Mobile Task Bar, that's how it works. I will improve the handling in future versions. I'll carry out the same test on Android soon.

dario-baumberger:
Test now on Android and iOS, works well on both. Tested with a local version with icons in commands. 

Should I merge these changes into 1.0.0, or should I create 1.0.1?

joethei:
Always release new versions, don't overwrite already existing releases.

[const response = await fetch(selection);](https://github.com/dario-baumberger/obsidian-json-table/blob/cbf2895c59f64d5cf1726180611a2a3441b13cc8/src/main.ts#L40)


[containerEl.createEl("h4", {](https://github.com/dario-baumberger/obsidian-json-table/blob/cbf2895c59f64d5cf1726180611a2a3441b13cc8/src/settings.ts#L17-L35)
Remove all the selected code, the commands should be explained in the readme, not in the settings.

[containerEl.createEl("h4", {](https://github.com/dario-baumberger/obsidian-json-table/blob/cbf2895c59f64d5cf1726180611a2a3441b13cc8/src/settings.ts#L37-L39)
Also get rid of this header, the first section should not have one.
(this would be the first header after the removal of the code above)

dario-baumberger:
Thank you for the feedback and the quick review. 

I made the following changes according to your change request: https://github.com/dario-baumberger/obsidian-json-table/pull/1/files?diff=split&w=0

The changes are available in version 1.0.1. 

I asked about the version because I wasn't sure how you would like it. Since this is the first plugin release from the perspective of community plugins, if the pull request is merged.

PR #2587, opened 2023-10-25
Add plugin: iCal

andrewbrereton:

## Repo URL

Link to my plugin: https://github.com/andrewbrereton/obsidian-to-ical-plugin

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[this.registerInterval(](https://github.com/andrewbrereton/obsidian-to-ical-plugin/blob/79be6b376b395cbd635b2601da1d5f1d55287f5d/src/ObsidianIcalPlugin.ts#L87-L93)
Remove this sample code.

[containerEl.createEl('h1', { text: 'Obsidian to iCal'});](https://github.com/andrewbrereton/obsidian-to-ical-plugin/blob/79be6b376b395cbd635b2601da1d5f1d55287f5d/src/ObsidianIcalPlugin.ts#L139-L142)

[return [...markdown.matchAll(this.taskRegExp)]](https://github.com/andrewbrereton/obsidian-to-ical-plugin/blob/79be6b376b395cbd635b2601da1d5f1d55287f5d/src/TaskFinder.ts#L14-L21)
Use the metadataCache instead to look for tasks, your current approach will have performance issues in larger vaults.

---

[import { DateTime } from "luxon";](https://github.com/andrewbrereton/obsidian-to-ical-plugin/blob/79be6b376b395cbd635b2601da1d5f1d55287f5d/src/Model/Task.ts#L3)
Instead of Luxon I would recommend using the `moment` library which is exported by the Obsidian API, this will create a smaller compiled source file.

andrewbrereton:
Thanks for your feedback @joethei.

Your suggestions have been implemented.

https://github.com/andrewbrereton/obsidian-to-ical-plugin/compare/1.0.6...1.1.0

andrewbrereton:
Hi @joethei, sorry to bump this one up. Is this plugin OK to merge now?

andrewbrereton:
Hi @liamcain, is there any further feedback for this? Thanks.

liamcain:
- [import * as path from "path";](https://github.com/andrewbrereton/obsidian-to-ical-plugin/blob/64b1f3589b6a00fe9955b8e6b57207464ea71e0d/src/FileClient.ts#L2) this is a nodejs import so it won't work on mobile. Obsidian will automatically normalize path separators, so just use "/" instead of path.sep here.

PR #2600, opened 2023-10-28
Add plugin: Metal Archives (Unofficial)

vincenzocaputo:

## Repo URL

Link to my plugin:

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["name": "Metal Archives Plugin (Unofficial)",](https://github.com/vincenzocaputo/obsidian-metal-archives-plugin/blob/320c42335c7ccb938312da590b6ada56a9b481aa/manifest.json#L3)
This should not contain the word "Plugin"

[id: "metal-archives-band",](https://github.com/vincenzocaputo/obsidian-metal-archives-plugin/blob/320c42335c7ccb938312da590b6ada56a9b481aa/src/main.ts#L55)

[const fs = require("fs");](https://github.com/vincenzocaputo/obsidian-metal-archives-plugin/blob/320c42335c7ccb938312da590b6ada56a9b481aa/src/main.ts#L91)
This import only available from a Node.js runtime, which means that this will throw errors on Mobile (that's a problem since you have `isDesktopOnly` marked as `false` in your manifest.json file.)
Use the Vault API instead.

[await vault.save(newNote);](https://github.com/vincenzocaputo/obsidian-metal-archives-plugin/blob/320c42335c7ccb938312da590b6ada56a9b481aa/src/note_handler/NoteRenderer.ts#L15)
That function does not exist.

vincenzocaputo:
> ["name": "Metal Archives Plugin (Unofficial)",](https://github.com/vincenzocaputo/obsidian-metal-archives-plugin/blob/320c42335c7ccb938312da590b6ada56a9b481aa/manifest.json#L3) This should not contain the word "Plugin"
> 
> [id: "metal-archives-band",](https://github.com/vincenzocaputo/obsidian-metal-archives-plugin/blob/320c42335c7ccb938312da590b6ada56a9b481aa/src/main.ts#L55) Obsidian will automatically prefix the command id with the plugin id, you don't need to include it yourself.
> 
> [const fs = require("fs");](https://github.com/vincenzocaputo/obsidian-metal-archives-plugin/blob/320c42335c7ccb938312da590b6ada56a9b481aa/src/main.ts#L91) This import only available from a Node.js runtime, which means that this will throw errors on Mobile (that's a problem since you have `isDesktopOnly` marked as `false` in your manifest.json file.) Use the Vault API instead.
> 
> [await vault.save(newNote);](https://github.com/vincenzocaputo/obsidian-metal-archives-plugin/blob/320c42335c7ccb938312da590b6ada56a9b481aa/src/note_handler/NoteRenderer.ts#L15) That function does not exist.

I should have fixed all the points. Many thanks!

PR #2611, opened 2023-10-30
Add plugin: Highlights Panel

frankhaotic:

## Repo URL

Link to my plugin: https://github.com/frankhaotic/obsidian-highlights-panel

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["fundingUrl": "",](https://github.com/frankhaotic/obsidian-highlights-panel/blob/bc08e05f028300265a7f0e2972c0c5342120e56b/manifest.json#L9)
If you don't have one, omit this.

[const activeFile = this.workspace.getActiveFile() as TFile;](https://github.com/frankhaotic/obsidian-highlights-panel/blob/bc08e05f028300265a7f0e2972c0c5342120e56b/src/main.ts#L27)
This could also be null.

[let editor = this.app.workspace.getActiveViewOfType(MarkdownView)?.editor;](https://github.com/frankhaotic/obsidian-highlights-panel/blob/bc08e05f028300265a7f0e2972c0c5342120e56b/src/main.ts#L34)
The active file might not be a markdown file, which would result in this being null.
Use `this.app.workspace.activeEditor` instead.


[this.workspace.detachLeavesOfType(HIGHLIGHTS_VIEW_TYPE);](https://github.com/frankhaotic/obsidian-highlights-panel/blob/bc08e05f028300265a7f0e2972c0c5342120e56b/src/main.ts#L68)
Do not detach leaves with your custom view, this is an antipattern, see: <https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Don't+detach+leaves+in+%60onunload%60>
Instead, only open a new leaf if there is not one already from your plugin.

[this.contentEl.style.position = 'relative';](https://github.com/frankhaotic/obsidian-highlights-panel/blob/bc08e05f028300265a7f0e2972c0c5342120e56b/src/views/highlights-view.ts#L46-L47)

PR #2639, opened 2023-11-06
Add plugin: cursor-goaway

liuxingyu521:

## Repo URL

Link to my plugin: https://github.com/liuxingyu521/obsidian-plugin-cursor-goaway

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[this.app.workspace.on('file-open', (file) => {](https://github.com/liuxingyu521/obsidian-plugin-cursor-goaway/blob/ed8dc64161c240fd0cbb563f37b27e57d9a05937/main.ts#L8)

liuxingyu521:
> [this.app.workspace.on('file-open', (file) => {](https://github.com/liuxingyu521/obsidian-plugin-cursor-goaway/blob/ed8dc64161c240fd0cbb563f37b27e57d9a05937/main.ts#L8) Events need to be registered, so they can be properly unloaded when the plugin is disabled. https://docs.obsidian.md/Plugins/Events

@joethei  I corrected and released a new version. https://github.com/liuxingyu521/obsidian-plugin-cursor-goaway/releases/tag/1.1.0

liamcain:
The code for this plugin doesn't do what the plugin description claims to do. It appears you are setting the cursor position to the last line of the file. That will cause the view to be scrolled to the bottom of the file, which I don't think is the intention.

liuxingyu521:
> The code for this plugin doesn't do what the plugin description claims to do. It appears you are setting the cursor position to the last line of the file. That will cause the view to be scrolled to the bottom of the file, which I don't think is the intention.

@liamcain  Sorry, I was busy and forgot to reply some time ago.
It's exactly what you said. If change it this way, it will automatically scroll to the bottom. I've changed it again and can help you take a look please.

code: https://github.com/liuxingyu521/obsidian-plugin-cursor-goaway/tree/1.2.0

PR #2644, opened 2023-11-08
Add plugin: DIDA(滴答清单)

viduycheung:

## Repo URL

Link to my plugin: https://github.com/viduycheung/dida-obsidian

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


viduycheung:
fixed

joethei:
Why did you create a new plugin for this?
Wouldn't it be a better idea to just offer a configuration option in the settings?

viduycheung:
> Why did you create a new plugin for this? Wouldn't it be a better idea to just offer a configuration option in the settings?

DIDA and TickTick are different services designed for different regions. The majority of users aren’t aware of the other service. Adding an option to switch between services could cause confusion and make the plugin more challenging to search.

joethei:
I don't see the issue in that, update the name & description to include both names and add a configuration option at the top to select the service used.

viduycheung:
Our team has deliberated on whether to create a new plugin or introduce an option to switch accounts. We believe that developing a new plugin would be the most effective method to provide this feature to DIDA users. However, is it not permissible within Obsidian’s guidelines to have two plugins with similar functionality but different account servers?

joethei:
We want to avoid having multiple plugins that share the same functionality.
The only difference I found between the two plugins is the configuration, which means that the plugins are very similar to each other.

how2051:
If this patch isn't allowed to be merged, can I just download some files of yr repository and install it manually? If so, what should I do, what files to download and move to where? @viduycheung

viduycheung:
> If this patch isn't allowed to be merged, can I just download some files of yr repository and install it manually? If so, what should I do, what files to download and move to where? @viduycheung

You can fork the https://github.com/viduycheung/dida-obsidian and run it locally.

https://docs.obsidian.md/Plugins/Getting+started/Build+a+plugin#Step+2+Build+the+plugin

PR #2661, opened 2023-11-13
Add plugin: Autocorrect Formatter

b-yp:

## Repo URL

Link to my plugin: https://github.com/b-yp/obsidian-autocorrect

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["description": "A plugin running on Obsidian that utilizes autocorrect to format Markdown content.",](https://github.com/b-yp/obsidian-autocorrect/blob/342c0ae8c045bf646314dc97172d730245c8d5a6/manifest.json#L6)
Avoid including sentences like "This is a plugin for Obsidian that does ..." in your description.

[MyPlugin](https://github.com/b-yp/obsidian-autocorrect/blob/342c0ae8c045bf646314dc97172d730245c8d5a6/main.ts#L5)
Rename this class.

[name: 'Formatting Content',](https://github.com/b-yp/obsidian-autocorrect/blob/342c0ae8c045bf646314dc97172d730245c8d5a6/main.ts#L16)
Also, it should be something like `Format content`

[await init('https://img.ypll.xyz/dev/autocorrect_wasm/pkg/autocorrect_wasm_bg.wasm')](https://github.com/b-yp/obsidian-autocorrect/blob/342c0ae8c045bf646314dc97172d730245c8d5a6/main.ts#L12)
Bundle this file, don't load it over the network.

b-yp:
Thank you very much for your review. I've addressed all the other issues, but regarding the final one, due to the limitations of the package used, relative addresses are not correctly loaded. Unfortunately, I had to use network addresses. I'm not sure if it's absolutely necessary to change this.

joethei:
Plugins should only load data over the network when absolutely required.
In your case, it should be possible somehow to bundle that file, which is why it's a required change.

b-yp:
Okay, I understand. I'll study it again.

b-yp:
Hello, I've switched to local loading. Please review the PR when you have time.

joethei:
> ["description": "A plugin running on Obsidian that utilizes autocorrect to format Markdown content.",](https://github.com/b-yp/obsidian-autocorrect/blob/342c0ae8c045bf646314dc97172d730245c8d5a6/manifest.json#L6)
Avoid including sentences like "This is a plugin for Obsidian that does ..." in your description.

This has not been addressed.

b-yp:
Thank you for your careful review; it has been corrected.

b-yp:
> * [getActiveViewOfType](https://github.com/b-yp/obsidian-autocorrect/blob/ad24c31299e49bd1e45f2a6d3866ca52002af58f/main.ts#L10) this won't work. You will only resolve the markdownView once, when the plugin loads, but it will become wrong as soon as the user opens another file. This check needs to happen inside the command callback. You can also use [editorCallback](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L614) instead of `callback` in your `addCommand` definition. This will ensure that the command is only available when the user is editing a file.

Thank you for your guidance. I have made the modifications according to your suggestions. If there are any further issues, please feel free to point them out. Thank you.
https://github.com/b-yp/obsidian-autocorrect/commit/da8c7dc9f74037efb8847d311ca1c10483e4e503

liamcain:
```
				const markdownView = this.app.workspace.getActiveViewOfType(MarkdownView);
				if (!markdownView) return

				const editor = markdownView.editor;
```

Just a heads up, editorCallback is passed an editor instance so you don't need any of these lines. You can just use the editor that's passed into the callback function.

b-yp:
> ```ts
> const markdownView = this.app.workspace.getActiveViewOfType(MarkdownView);
> if (!markdownView) return
> 
> const editor = markdownView.editor;
> ```
> 
> [editorCallback](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L614)
> 
> Just a heads up, editorCallback is passed an editor instance so you don't need any of these lines. You can just use the editor that's passed into the callback function.

Thank you for your thorough review. I have addressed and resolved the issue.

PR #2666, opened 2023-11-13
Add plugin: Vim Yank Highlight

aleksey-rowan:

## Repo URL

Link to my plugin: https://github.com/aleksey-rowan/obsidian-vim-yank-highlight

## Release Checklist
  - [X]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[timeoutHandle: NodeJS.Timeout;](https://github.com/aleksey-rowan/obsidian-vim-yank-highlight/blob/b61b7c14a7c5f4849841cf58b7c709b3f2250a55/src/main.ts#L25)
use a regular number here.

[return (</* { editor?: { cm: EditorView } } */ any>this.activeView!)](https://github.com/aleksey-rowan/obsidian-vim-yank-highlight/blob/b61b7c14a7c5f4849841cf58b7c709b3f2250a55/src/main.ts#L32-L33)
use `this.app.workspace.activeEditor` instead, which will make this compatible with Canvas.

[this.timeoutHandle = setTimeout(() => {](https://github.com/aleksey-rowan/obsidian-vim-yank-highlight/blob/b61b7c14a7c5f4849841cf58b7c709b3f2250a55/src/main.ts#L114)
Use `window.setTimeout` here, to avoid conflicts with the NodeJS version.

aleksey-rowan:
> [timeoutHandle: NodeJS.Timeout;](https://github.com/aleksey-rowan/obsidian-vim-yank-highlight/blob/b61b7c14a7c5f4849841cf58b7c709b3f2250a55/src/main.ts#L25) use a regular number here.

Done.
 
> [return (</* { editor?: { cm: EditorView } } */ any>this.activeView!)](https://github.com/aleksey-rowan/obsidian-vim-yank-highlight/blob/b61b7c14a7c5f4849841cf58b7c709b3f2250a55/src/main.ts#L32-L33) use `this.app.workspace.activeEditor` instead, which will make this compatible with Canvas.

This was probably a bad naming of the getter on my part here as I need the `EditorView`, not the `Editor` itself. I need the view first to get the plugin and then run the update on in for this view.

Is there a better way to get the `EditorView`? I was looking at other community plugins and found this way which seems a bit more "clean/logical":

```ts
this.app.workspace.getActiveViewOfType(MarkdownView)?.leaf.view.editor?.cm;
```

Also, I don't think it needs to be compatible with Canvas since it's only going to get used in the markdown editor itself.

I updated the release. Please, let me know if this is okay or still needs to change.

> [this.timeoutHandle = setTimeout(() => {](https://github.com/aleksey-rowan/obsidian-vim-yank-highlight/blob/b61b7c14a7c5f4849841cf58b7c709b3f2250a55/src/main.ts#L114) Use `window.setTimeout` here, to avoid conflicts with the NodeJS version.

Done.

aleksey-rowan:
@liamcain, thank you for taking a look at my submission!

> [bind(this)](https://github.com/aleksey-rowan/obsidian-vim-yank-highlight/blob/933430bb5871961cc00851e2637e31f0fc4535f7/src/main.ts#L60) I don't think this will work.

Good catch! I changed that to class properties defined with arrow function which should get the scope and not cause memory leaks.

> [initialized](https://github.com/aleksey-rowan/obsidian-vim-yank-highlight/blob/933430bb5871961cc00851e2637e31f0fc4535f7/src/main.ts#L15) this is unused

`initialized` is used on this line: https://github.com/aleksey-rowan/obsidian-vim-yank-highlight/blob/933430bb5871961cc00851e2637e31f0fc4535f7/src/main.ts#L49

PR #2678, opened 2023-11-19
Add plugin: Daily ICalendar

mmmykhailo:

## Repo URL

Link to my plugin: https://github.com/mmmykhailo/obsidian-daily-icalendar

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


community-plugins.json
```diff
@@ -9098,5 +9098,12 @@
      "author": "Mgussekloo",
      "description": "Autofocus the editor when opening a new file.",
      "repo": "mgussekloo/obsidian-editor-autofocus"
+  },
+  {
+     "id": "daily-icalendar-plugin",
```
mmmykhailo, line 9103:
```suggestion
     "id": "daily-icalendar",
```

mmmykhailo:
Bruh, github failed uploading them for some reason 🤨

joethei:
[import { VCalendar, VEvent, VTimeZone } from "node-ical";](https://github.com/mmmykhailo/obsidian-daily-icalendar/blob/be02da932ba8f8c9dd82eac469d3fe3bb4369840/utils/html.ts#L1)

mmmykhailo:
I set isDesktopOnly to true for now. Don't have enogh time to find a better library or write my own.
By the way, if you know one that have good enough types - I would be glad if you advised it 🤓

PR #2681, opened 2023-11-20
Add plugin: Custom save

HananoshikaYomaru:

## Repo URL

Link to my plugin: https://github.com/HananoshikaYomaru/obsidian-custom-save

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


HananoshikaYomaru:
The name in LICENSE is `Yeung Man Lung Ken` which is my real name on passport. Don't ask me to change it.

joethei:
["description": "Add custom save actions to obsidian",](https://github.com/HananoshikaYomaru/obsidian-custom-save/blob/fb387ed8cd7dce710fcae8b9af080ff924df0271/manifest.json#L6)
That description (along with the one in your readme) is a bit confusing, I would suggest something like:
> Run any commands when you 

[console.log("loaded: ", loadedData);](https://github.com/HananoshikaYomaru/obsidian-custom-save/blob/fb387ed8cd7dce710fcae8b9af080ff924df0271/src/SettingManager.ts#L60), and more

---

[hotkeys: [](https://github.com/HananoshikaYomaru/obsidian-custom-save/blob/fb387ed8cd7dce710fcae8b9af080ff924df0271/src/main.ts#L32-L37)
Won't this conflict with the default save action, if one does not change the hotkey settings?
(Yes, I know that it's included in the instructions to change it, but users might not read that part)

HananoshikaYomaru:
Ok I will update the setting to make it easier for people to follow. I will also remove the default hotkey. For obsidian, if there is hotkey crash, only the last only will have effect. So you are right, if they don't update there might be a problem. But since two commands are just save command, there won't be problem. But I will change it anyway.

HananoshikaYomaru:
sorry the name of this plugin might need to be changed. I am planning to expand on this plugin.

HananoshikaYomaru:
After thinking a bit, I don't think I will remove the default hot key from the command because it makes sense that if the command is a save command then it should be Mod + S. Instead I added a warning if there is crash of hot key.

CleanShot 2023-11-29 at 20 35 47@2x
 

CleanShot 2023-11-29 at 20 35 22@2x

And I don't plan to change the name of the plugin because I don't have time to expand the scope of this plugin. It is okay to leave it for now. 

I have fixed all issues. so please review again.

PR #2712, opened 2023-11-29
Add Image Convert Contextmenu plugin

byfun:

## Repo URL

Link to my plugin: https://github.com/byfun/obsidian-image-convert-contextmenu

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -9182,5 +9182,12 @@
     "author": "Jose Elias Alvarez",
     "description": "Intelligently paste Markdown links.",
     "repo": "jose-elias-alvarez/obsidian-paste-link"
-  }
+  },
+  {
+    "id": "image-convert-contextmenu",
+    "name": "Image Convert Contextmenu",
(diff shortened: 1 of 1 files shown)


joethei:
[import { basename } from "path";](https://github.com/byfun/obsidian-image-convert-contextmenu/blob/c739b1d7bce8c983583a5539c22fdc3df6708e2c/src/helpers.ts#L2)

[await fetch(image.src, { mode: "no-cors" });](https://github.com/byfun/obsidian-image-convert-contextmenu/blob/c739b1d7bce8c983583a5539c22fdc3df6708e2c/src/helpers.ts#L240-L249)
Use the `requestUrl` function from the Obsidian API instead, it will handle some things like CORS automatically, which means that you can remove the dependence on that external server.

PR #2728, opened 2023-12-01
Add plugin: Multi State CheckBox Switcher

KubaMiszcz:

## Repo URL

Link to my plugin:https://github.com/KubaMiszcz/MultiStateCheckBoxSwitcher

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[fundingUrl": {](https://github.com/KubaMiszcz/MultiStateCheckBoxSwitcher/blob/cccda785bb1c694cd910854b8aeee93ce188cd4c/manifest.json#L9-L12)
All these links point to the homepage of the respective services, either put in links to your profile on these sites, or remove this section entirely.

[this.registerInterval(](https://github.com/KubaMiszcz/MultiStateCheckBoxSwitcher/blob/cccda785bb1c694cd910854b8aeee93ce188cd4c/src/main.ts#L93-L96)
Remove this sample code.

[containerEl.createEl('h1', { text: 'Multi State CheckBox Switcher' });](https://github.com/KubaMiszcz/MultiStateCheckBoxSwitcher/blob/cccda785bb1c694cd910854b8aeee93ce188cd4c/src/SettingsTab.ts#L45)

KubaMiszcz:
fixed according to CR
@joethei

liamcain:
- [currentLine.startsWith](https://github.com/KubaMiszcz/MultiStateCheckBoxSwitcher/blob/8e3aead4c11363a9ca510ed9c9f80f0500610d6d/src/main.ts#L44) this won't work if the list is indented
- [setCursor](https://github.com/KubaMiszcz/MultiStateCheckBoxSwitcher/blob/8e3aead4c11363a9ca510ed9c9f80f0500610d6d/src/main.ts#L60) you are calling setCursor and setSelection. These do the same thing so one of the lines is redundant.
- [mySetting: string](https://github.com/KubaMiszcz/MultiStateCheckBoxSwitcher/blob/8e3aead4c11363a9ca510ed9c9f80f0500610d6d/src/SettingsTab.ts#L10) remove this, it's unused.

KubaMiszcz:
thanks for CR, all fixed

- _[currentLine.startsWith](https://github.com/KubaMiszcz/MultiStateCheckBoxSwitcher/blob/8e3aead4c11363a9ca510ed9c9f80f0500610d6d/src/main.ts#L44) this won't work if the list is indented_
- [x] **fixed**
---

- _[setCursor](https://github.com/KubaMiszcz/MultiStateCheckBoxSwitcher/blob/8e3aead4c11363a9ca510ed9c9f80f0500610d6d/src/main.ts#L60) you are calling setCursor and setSelection. These do the same thing so one of the lines is redundant._

setSelection maintain original selection selected, and setCursor is necessary to keep cursor in original position if you select lines in from down to up direction - eg starts somewhere in line 10 and moves it to somewhere in line 4, without it, cursor moves to line 10 instead stay in line 4
- [x] **checked**
---
 
- _[mySetting: string](https://github.com/KubaMiszcz/MultiStateCheckBoxSwitcher/blob/8e3aead4c11363a9ca510ed9c9f80f0500610d6d/src/SettingsTab.ts#L10) remove this, it's unused._
- [x] **removed**

liamcain:
Sorry for the delay! Looks good to me.

> setSelection maintain original selection selected, and setCursor is necessary to keep cursor in original position if you select lines in from down to up direction - eg starts somewhere in line 10 and moves it to somewhere in line 4, without it, cursor moves to line 10 instead stay in line 4

`setSelection` accepts a "head" and "anchor" so the order of these two arguments changes the order of the selection. If you keep the order correct, the setCursor should be unnecessary.

PR #2783, opened 2023-12-11
Add plugin: File Explorer Reload

mnaoumov:

## Repo URL

Link to my plugin:

https://github.com/mnaoumov/obsidian-file-explorer-reload

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -9399,5 +9399,12 @@
       "author": "mmmykhailo",
       "description": "View calendar events in daily notes.",
       "repo": "mmmykhailo/obsidian-daily-icalendar"
+    },
+    {
+        "id": "file-explorer-reload",
+        "name": "File Explorer Reload",
+        "author": "mnaoumov",
+        "description": "Reloads file explorer pane",
+        "repo": "mnaoumov/obsidian-file-explorer-reload"
     }
 ]


joethei:
Can you link to a bug report for the issue you are attempting to fix?
I can't find a fitting one.

[import * as fsPromises from "fs/promises";](https://github.com/mnaoumov/obsidian-file-explorer-reload/blob/284594ddb20f52d8ee6d876d09368f0b0b3b9a94/FileExplorerReloadPlugin.ts#L8)

[const dir = this.app.vault.getAbstractFileByPath(directoryPath) as TFolder;](https://github.com/mnaoumov/obsidian-file-explorer-reload/blob/284594ddb20f52d8ee6d876d09368f0b0b3b9a94/FileExplorerReloadPlugin.ts#L46)
This should never be casted, always do a `instanceof` check to confirm that it's actually a directory.

joethei:
Also, have you restarted your computer recently?

mnaoumov:
@joethei 

1. I created a [bug report](https://forum.obsidian.md/t/sometimes-changes-made-outside-of-obsidian-are-not-reflected-in-the-files-pane/73451) and added the link to it in the plugin's `README`.
2. Yes, good point, switched to `isDesktopOnly = true` in the release 1.2.0
3. Yes, good point, changed in the release 1.2.0

liamcain:
Thanks for surfacing the bug and creating a testing vault. And nice job with the plugin! I don't think we want to introduce this plugin to the list of Community plugins though since it's our hope to solve the problem on our end without a plugin making calls to private APIs within the file system adapter.

We will investigate further to see if we can find the root cause. Hope you understand!

liamcain:
Hopefully when users encounter a bug in Obsidian, their first instinct isn't to look for a plugin that safely patches into the internals of the app to fix the bug. Instead, they should go through the forums or official communication channels so that we can better understand the total breadth of who is affected by the bug and get more useful metrics and debugging information.

We would rather not promote these workarounds.

mnaoumov:
@liamcain I found the root cause for the bug, the plugin intended to workaround, and provided the suggestion for the team how to fix it.

https://forum.obsidian.md/t/sometimes-changes-made-outside-of-obsidian-are-not-reflected-in-the-files-pane/73451/6

DeanLemans:
bump

PR #2819, opened 2023-12-18
Add plugin: Outline++

RyotaUshio:

## Repo URL

Link to my plugin: https://github.com/RyotaUshio/obsidian-outline-plus

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
Hi, sorry for the delay. The linked bug in Obsidian was fixed in v1.5.4 (currently Insiders only, but expected to be released to public soon). Do you mind if I close this pull request since it only seems to be for working around this bug?

RyotaUshio:
@liamcain Thank you for the update. Although the main purpose of Outline++ was to fix the bug, it also renders markdown in the outline view. So I think it still can be valuable for some users. 

Could you consider adding this to the list? I've already updated the README.

PR #2847, opened 2023-12-25
Add plugin: PDF++

RyotaUshio:

If you have any suggestions, please let me know. Thank you!

## Repo URL

Link to my plugin: https://github.com/RyotaUshio/obsidian-pdf-plus

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


RyotaUshio:
By the way, I'm facing a problem that is likely to require some advanced understanding of the internals of Obsidian to fix. If you have any suggestions, please let me know. (But please feel free to ignore this)

---

This plugin has a feature that "trims" PDF embeds such as `![[file.pdf#page=1&selection=0,1,2,3]]`. More specifically, when embedding a selection or an annotation from a PDF file, this option displays only the selection/annotation and its surroundings rather than the entire page. For example:

https://github.com/RyotaUshio/obsidian-pdf-plus?tab=readme-ov-file#pdf-embeds-are-automatically-trimmed

This feature is implemented essentially by doing `containerEl.style.height = <HEIGHT COMPUTED BASED ON THE TARGET SELECTION/ANNOTATION>` (more accurately, [calling the `setHeight` method of the `ObsidianViewer` class defined in `pdf.viewer.min.js`](https://github.com/RyotaUshio/obsidian-pdf-plus/blob/278ac7124047594d0204efe42c22e05482c7eead/src/patchers/pdf.ts#L114)). 

But the problem is, the viewport of the markdown view cannot be automatically updated, which leads to this issue: https://github.com/RyotaUshio/obsidian-pdf-plus/issues/8

PR #2890, opened 2024-01-03
Add plugin: Image Magician

luxmargos:

## Repo URL

Link to my plugin: https://github.com/luxmargos/obsidian-image-magician-plugin

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


luxmargos:
I have now updated to version 0.1.4

/skip

I have fixed most of the issues based on the guidance from @ObsidianReviewBot. However, a few cases have been enhanced or left as they are. Please see the following explanations.

* This class is only available on desktop, which means that this will throw errors on mobile (that's a problem since you have isDesktopOnly marked as false in your manifest.json file.
  * Explanation: I updated the code case by case for whether it's on desktop or mobile, so there will be no error.

* You should avoid assigning styles via JavaScript or in HTML and instead move all these styles into CSS so that they are more easily adaptable by themes and snippets.
  * Explanation: That is an intended code to position an element dynamically.

luxmargos:
I have now updated to version 0.1.5, including /skip within my code lines.

joethei:
[process.env.NODE_ENV === "development"](https://github.com/luxmargos/obsidian-image-magician-plugin/blob/56cfcba4b990a0014043697693bb5583b5599704/src/main.ts#L126)
`process` is only available on desktop, this will crash your plugin on mobile.

[id: "imgk-plugin-export-image",](https://github.com/luxmargos/obsidian-image-magician-plugin/blob/56cfcba4b990a0014043697693bb5583b5599704/src/main.ts#L203), [id: "imgk-plugin-process-auto-export",](https://github.com/luxmargos/obsidian-image-magician-plugin/blob/56cfcba4b990a0014043697693bb5583b5599704/src/main.ts#L235), [id: "imgk-plugin-process-auto-export-forced",](https://github.com/luxmargos/obsidian-image-magician-plugin/blob/56cfcba4b990a0014043697693bb5583b5599704/src/main.ts#L243)
Don't prefix the command id with the plugin id, that will happen internally automatically.

[findValutFile](https://github.com/luxmargos/obsidian-image-magician-plugin/blob/56cfcba4b990a0014043697693bb5583b5599704/src/vault_util.ts#L28)
Typo.

[let result = files.find((f: TFile) => {](https://github.com/luxmargos/obsidian-image-magician-plugin/blob/56cfcba4b990a0014043697693bb5583b5599704/src/vault_util.ts#L35-L38)
[Avoid iterating all files to find a file by its path]https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Avoid+iterating+all+files+to+find+a+file+by+its+path.

[// I think this is the link search behaviour of obsidian](https://github.com/luxmargos/obsidian-image-magician-plugin/blob/56cfcba4b990a0014043697693bb5583b5599704/src/vault_util.ts#L39-L45)
To find a file if you are only given a linkpath, use this function: <https://docs.obsidian.md/Reference/TypeScript+API/MetadataCache/getFirstLinkpathDest>

luxmargos:
I have now updated version 0.1.6, with fixes are applied.

PR #2908, opened 2024-01-08
Add plugin: Image Inline

ZackaryW:

## Repo URL

Link to my plugin:
https://github.com/ZackaryW/obsidian-image-inline

## Release Checklist
  - [X ]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


ZackaryW:
closing initial pull request due to main branch update https://github.com/obsidianmd/obsidian-releases/pull/2898

joethei:
Why is your plugin marked as desktop only?
I don't see any reason for it to be.

[id: 'convert-image-to-base64-hotkey',](https://github.com/ZackaryW/obsidian-image-inline/blob/7d4908b5140f344cb5a0d3e29d1d4896107f8086/src/main.ts#L25), [name: 'Convert Image to Base64 (Hotkey)',](https://github.com/ZackaryW/obsidian-image-inline/blob/7d4908b5140f344cb5a0d3e29d1d4896107f8086/src/main.ts#L26)
Remove the "Hotkey" portion.

[hotkeys: [{ modifiers: ["Ctrl", "Alt"], key: "v" }],](https://github.com/ZackaryW/obsidian-image-inline/blob/7d4908b5140f344cb5a0d3e29d1d4896107f8086/src/main.ts#L27)

[callback: () => this.handleHotkeyCommand(),](https://github.com/ZackaryW/obsidian-image-inline/blob/7d4908b5140f344cb5a0d3e29d1d4896107f8086/src/main.ts#L28)
use a editorCallback to make sure that command will only show up if there is an active editor.
https://docs.obsidian.md/Plugins/User+interface/Commands#Editor+commands

[const editor = activeView.editor;](https://github.com/ZackaryW/obsidian-image-inline/blob/7d4908b5140f344cb5a0d3e29d1d4896107f8086/src/main.ts#L92)
And use that editor instance passed there here.


[async convertBlobToBase64(blob: Blob): Promise<string> {](https://github.com/ZackaryW/obsidian-image-inline/blob/7d4908b5140f344cb5a0d3e29d1d4896107f8086/src/main.ts#L78)
Instead of this function use:
```ts
arrayBufferToBase64(await getBlobArrayBuffer(blob));
```

[containerEl.createEl('h2', { text: 'Settings for Image to Base64 Plugin' });](https://github.com/ZackaryW/obsidian-image-inline/blob/7d4908b5140f344cb5a0d3e29d1d4896107f8086/src/main.ts#L121)

ZackaryW:
@joethei Thanks for the advice. I have updated the project according to the guidelines. Please let me know what else needs improvement.

ZackaryW:
updated, thank you

ZackaryW:
updated, thank you @ObsidianReviewBot

PR #2911, opened 2024-01-09
Add plugin: Simple Image Inserter

jdholtz:

## Repo URL

Link to my plugin: https://github.com/jdholtz/obsidian-image-inserter

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


liamcain:
- [getAttachmentsDestination](https://github.com/jdholtz/obsidian-image-inserter/blob/5416532431da80a2d65d615a3ee0f2ab32e27f1e/src/file_modal.ts#L48) we've added `getAvailablePathForAttachment` to the API recently for this exact use case.

jdholtz:
> [getAttachmentsDestination](https://github.com/jdholtz/obsidian-image-inserter/blob/5416532431da80a2d65d615a3ee0f2ab32e27f1e/src/file_modal.ts#L48) we've added getAvailablePathForAttachment to the API recently for this exact use case.

Thanks @liamcain, I did not know that existed. However, I've been trying it out and it doesn't return the correct path when the attachment configuration path is set to "In the folder specified below" and that folder does not exist (it returns the path in the vault's base folder).

While this may be expected, this plugin always creates that folder before copying the attachments in. Therefore, I would still need this logic to get the destination (although the de-duplication feature of this function is great). Unfortunately though, even though I make this folder below and then call `getAvailablePathForAttachments` (both within the same command execution phase) it still outputs to the vault's base folder. It only outputs the correct path in the next call after the folder was created last call. Do you have any solutions to this?

PR #2916, opened 2024-01-10
Add plugin: Auto Definition Link

nmcarp99:

## Repo URL

Link to my plugin: https://github.com/nmcarp99/obsidian-auto-definition-link

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


joethei:
You should move all code files into a separate `src` folder, that makes it easier to review and maintain.

["fundingUrl": "https://github.com/nmcarp99",](https://github.com/nmcarp99/Obsidian-Auto-Definition-Link/blob/0139ed08a413e864e4764db40c5a572b2973bb3b/manifest.json#L9)
This is meant for links to services like _Buy me a coffee_, _GitHub sponsors_ and so on, if you don't have such a link omit this.

[if (AutoDefinitionLink.settings.autoRefreshLinks !== 'always') return;](https://github.com/nmcarp99/Obsidian-Auto-Definition-Link/blob/0139ed08a413e864e4764db40c5a572b2973bb3b/main.ts#L153)
Why is this condition different from the other events above?

[.setName('Refresh Links')](https://github.com/nmcarp99/Obsidian-Auto-Definition-Link/blob/0139ed08a413e864e4764db40c5a572b2973bb3b/settings.ts#L110)

nmcarp99:
Those issues should be fixed now. As for your question, the setting is used to specify whether to update links on all (modify, rename, delete), main (rename, delete), never (none of them). The **modify** event should only be called if "always" is selected, while **rename** and **delete** should always be called unless "never" is selected.

liamcain:
- Instead of reading each file, you are probably better off relying on Obsidian's metadataCache APIs which already have the blockIds pulled out. You can use iterateCacheRefs to find them all.

PR #2986, opened 2024-01-23
Rename author and update repo for OfficerHalf plugins

nathonius:
I've renamed my github account and updated all my plugin readmes and manifests accordingly. Now doing the same for my previously released plugins here.


nathonius:
Updated wrt to renamed Trello plugin.

PR #3004, opened 2024-01-27
Update plugin description

KosmosisDire:
Updated the description of my plugin, webpage html export


community-plugins.json
```diff
@@ -5274,7 +5274,7 @@
         "id": "webpage-html-export",
         "name": "Webpage HTML Export",
         "author": "Nathan George",
-        "description": "Exports an Obsidian document, folder, or vault as an HTML document / webpage / website with correct styling. Also has an interactive embedded dark / light theme toggle, and document outline.",
+        "description": "Export html from single files, canvas pages, or whole vaults. Similar to Obsidian's Publish or digital garden, but with direct access to the exported HTML. Focuses on flexibility, features, and style parity.",
```
joethei, line 5277: Please don't make comparisons in the description, those belong in the readme.

KosmosisDire: The main reason I added that is that I have a had a number of requests to add a comparison to the description because they were searching for those key words. I'll try and change it so it includes the key words while actually adding something meaningful other than just a comparison.

KosmosisDire:
Okay I have made that change

PR #3024, opened 2024-01-29
rename dataset aid plugin

conneroisu:
https://github.com/conneroisu/Text-Dataset-Aid-Plugin


joethei:
Don't modify the plugin id.
Any existing users of your plugin won't be able to update

PR #3052, opened 2024-02-05
Add plugin: Set View Mode per Note

AlexDavies8:

## Repo URL

Link to my plugin: https://github.com/AlexDavies8/obsidian-frontmatter-viewmode

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [X]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


joethei:
What sets your plugin apart from the _Force note view mode_ plugin?

[import * as path from 'path';](https://github.com/AlexDavies8/obsidian-frontmatter-viewmode/blob/01ca740e2ffa164f0b64a215f0e486de944cd782/main.ts#L2)

[path.join(this.app.vault.configDir, 'app.json')](https://github.com/AlexDavies8/obsidian-frontmatter-viewmode/blob/01ca740e2ffa164f0b64a215f0e486de944cd782/main.ts#L19)
Just join them with a simple `/`

AlexDavies8:
Thanks for the feedback. I'll fix the issue with path (although strangely it seems to work fine on android for me despite this).

In terms of the difference with Force Note View Mode, that plugin doesn't respect default preferred view style properly, and is also very inconsistent. It seems to trigger on a different event internally which happens before the actual loading of the note. This means that it needs an arbitrary delay before it updates the view mode, so if the note takes too long to load it won't work.

This plugin is simpler, faster and importantly it respects the default editing mode, so you don't have to manually tag notes as preview or source mode, you can just use a single tag.

PR #3053, opened 2024-02-05
Add plugin: Image to text OCR

dario-baumberger:

## Repo URL

Link to my plugin: https://github.com/dario-baumberger/obsidian-image-to-text-ocr

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


dario-baumberger:
In the [plugin guidelines](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Avoid+iterating+all+files+to+find+a+file+by+its+path) it's written "Avoid iterating all files to find a file by its path". 

Unfortunately I wasn't able to find a solution with `getAbstractFileByPath` because I don't know the path. It may be different from vault to vault due to user attachment settings: https://github.com/dario-baumberger/obsidian-image-to-text-ocr/blob/master/src/main.ts#L227. If a correction is necessary here, I would be very grateful for further suggestions

dario-baumberger:
/skip

The ObsidianReviewBot detected FileSystemAdapter in the code and says, it's a problem because isDesktopOnly false. On mobile the plugin won't work with Obsidian Image-Tags, but with all other image tags like html img-tag and Markdown image-tag it works. The plugin displays a notice if the adapter wasn't found. I could write a hint, that it doesn't work on mobile devices.  But from my perspective, It would be nice to have this plugin also available on mobile for other image-tags than Obsidian tags.

The logs are all for debugging purposes and are only executed if logging is enabled in the plugin settings.

dario-baumberger:
I just released a new version with a notice if the FileSystemAdapter isn't available (already existed) but added an extra text: Obsidian images are not supported on mobile devices.

It's also written in the readme. 

If this isn't enough to release it with isDesktopOnly: false, please let me know. Then I will change it to true

joethei:
[name: "Image To Text: Replace Selection",](https://github.com/dario-baumberger/obsidian-image-to-text-ocr/blob/fa42ee9f8a8119f2c889bf72867dbce8a9c437e2/src/main.ts#L45), [name: "Image To Text: Replace Selection - custom language",](https://github.com/dario-baumberger/obsidian-image-to-text-ocr/blob/fa42ee9f8a8119f2c889bf72867dbce8a9c437e2/src/main.ts#L80), [name: "Image To Text: Append Selection",](https://github.com/dario-baumberger/obsidian-image-to-text-ocr/blob/fa42ee9f8a8119f2c889bf72867dbce8a9c437e2/src/main.ts#L127), [name: "Image To Text: Append Selection - custom language",](https://github.com/dario-baumberger/obsidian-image-to-text-ocr/blob/fa42ee9f8a8119f2c889bf72867dbce8a9c437e2/src/main.ts#L161)
Remove the "Image To Text" prefix and [use sentence case in UI](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Use+sentence+case+in+UI)

[const imageFile = files.find((file) => file.name === fileName);](https://github.com/dario-baumberger/obsidian-image-to-text-ocr/blob/fa42ee9f8a8119f2c889bf72867dbce8a9c437e2/src/main.ts#L229)
Use `this.app.metadataCache.getFirstLinkPathDest(filename, '') instead.
<https://docs.obsidian.md/Reference/TypeScript+API/MetadataCache/getFirstLinkpathDest>

[const resourcePath = adapter.getResourcePath(imageFile.path);](https://github.com/dario-baumberger/obsidian-image-to-text-ocr/blob/fa42ee9f8a8119f2c889bf72867dbce8a9c437e2/src/main.ts#L238)
There is also a `getResourcePath` function in the Vault API, no need to use the adapter.
Plus, this will be mobile compatible.

dario-baumberger:
@joethei Thank you very much for the review and the feedback! Especially for the tips regarding `getFirstLinkpathDest` and `getResourcePath`. I didn't find them myself... 

I have made the requested changes and published a new version.

PR #3076, opened 2024-02-12
Add plugin: Quadro

chrisgrieser:

## Repo URL

Link to my plugin: https://github.com/chrisgrieser/obsidian-quadro

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[/* Make them less visible */](https://github.com/chrisgrieser/obsidian-quadro/blob/4b071d25a13a42bccdfb11e024f85882c221e82e/styles.css#L9-L57)
Please don't overwrite core styling in a plugin, add another class so it only applies to your plugin if possible.

[cmdObj.hotkeys = [{ modifiers: ["Mod", "Shift"], key: cmd.hotkeyLetter }];](https://github.com/chrisgrieser/obsidian-quadro/blob/4b071d25a13a42bccdfb11e024f85882c221e82e/src/main.ts#L19)

[export function safelyGetActiveEditor(app: App): Editor | null {](https://github.com/chrisgrieser/obsidian-quadro/blob/4b071d25a13a42bccdfb11e024f85882c221e82e/src/utils.ts#L17-L24)
Use `app.workspace.activeEditor` instead (unless you want to ignore Canvas)

liamcain:
- [safelyGetActiveEditor](https://github.com/chrisgrieser/obsidian-quadro/blob/467987cbe13489b6f5c18dc98a2b067142d431de/src/utils.ts#L23) You should use [`activeEditor`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4009) instead. Otherwise the plugin won't work inside Canvas at all.
- [document.querySelector](https://github.com/chrisgrieser/obsidian-quadro/blob/467987cbe13489b6f5c18dc98a2b067142d431de/src/utils.ts#L67) this won't work inside [pop-out windows](https://help.obsidian.md/User+interface/Pop-out+windows) at all. You'll want to use `activeDocument` instead.

chrisgrieser:
Thanks!

> * [safelyGetActiveEditor](https://github.com/chrisgrieser/obsidian-quadro/blob/467987cbe13489b6f5c18dc98a2b067142d431de/src/utils.ts#L23) You should use [`activeEditor`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4009) instead. Otherwise the plugin won't work inside Canvas at all.

btw, you should update the developer docs for editors, since [`getActiveViewOfType(MarkdownView)` is recommended there](https://docs.obsidian.md/Plugins/Editor/Editor).

PR #3085, opened 2024-02-13
Add plugin: Crafty

liolle:
Add Crafty plugin

## Repo URL

Link to my plugin: https://github.com/liolle/Crafty

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[import { FSWatcher, watch } from "fs";](https://github.com/liolle/Crafty/blob/64ec9a09e21a4397dd1d28e7876d487718ab44fd/main.ts#L10)

[#absFileToFile(file: TAbstractFile) {](https://github.com/liolle/Crafty/blob/64ec9a09e21a4397dd1d28e7876d487718ab44fd/main.ts#L261-L264)
Remove this function and do a `instanceof TFile` check in the place where you are accessing the file.

[static async detachPanelView(plugin: Crafty) {](https://github.com/liolle/Crafty/blob/64ec9a09e21a4397dd1d28e7876d487718ab44fd/dom/handler.ts#L33)
This function name does not match what it's actually doing.

liolle:
> [import { FSWatcher, watch } from "fs";](https://github.com/liolle/Crafty/blob/64ec9a09e21a4397dd1d28e7876d487718ab44fd/main.ts#L10) This import is available from a Node.js runtime, which means that this will throw errors on Mobile (that's a problem since you have `isDesktopOnly` marked as `false` in your manifest.json file.
> 
> [#absFileToFile(file: TAbstractFile) {](https://github.com/liolle/Crafty/blob/64ec9a09e21a4397dd1d28e7876d487718ab44fd/main.ts#L261-L264) Remove this function and do a `instanceof TFile` check in the place where you are accessing the file.
> 
> [static async detachPanelView(plugin: Crafty) {](https://github.com/liolle/Crafty/blob/64ec9a09e21a4397dd1d28e7876d487718ab44fd/dom/handler.ts#L33) This function name does not match what it's actually doing.

Hi, i have done a few changes as requested : 
- **FSWatcher** : I will limit the plugin to desktop only for now, the time to find another way to listen for change on a specific .canvas file, I don't know if the obsidian API has a listener for that?

- **#absFileToFile** : After reading the code back, it seems like i  can just use `this.app.workspace.getActiveFile()` to get the current instance of TFile, I don' t know why i was doing that loop around.

-  **detachPanelView** :  A more appropriate name for this function woul be `showPlacehoderview` since it is the view i am showing for every non .canvas files

PR #3086, opened 2024-02-14
Add plugin: BackItUp

hammadxp:

## Repo URL

https://github.com/hammadxp/back-it-up
Link to my plugin:

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
["description": "Quickly make a copy or snapshot of a note in Obsidian.",](https://github.com/hammadxp/back-it-up/blob/c97a984def8c46c60ea871b04234c20f53a064c8/manifest.json#L6)
Remove the "in Obsidian" part, it should be obvious that this is for Obsidian.

[.hi {](https://github.com/hammadxp/back-it-up/blob/c97a984def8c46c60ea871b04234c20f53a064c8/styles.css)
Remove this file from the repo, or at least omit it from the release so Obsidian does not need to download it on every update.

[this.registerDomEvent(document, "click", (evt: MouseEvent) => {](https://github.com/hammadxp/back-it-up/blob/c97a984def8c46c60ea871b04234c20f53a064c8/main.ts#L22-L28)
Please remove anything from the sample plugin that you don't need.

[import * as path from "path";](https://github.com/hammadxp/back-it-up/blob/c97a984def8c46c60ea871b04234c20f53a064c8/actions.ts#L2-L3)
Instead use the vault API, which has a copy function.

joethei:
[this.registerInterval(](https://github.com/hammadxp/back-it-up/blob/b930bd00f5bbe6d75ca407012ff28062a21b61f2/main.ts#L22-L24)
Remove everything from the sample plugin that you don't need.

[(file as any).basename](https://github.com/hammadxp/back-it-up/blob/b930bd00f5bbe6d75ca407012ff28062a21b61f2/actions.ts#L8), [export async function makeCopyOfNote(file: TAbstractFile) {](https://github.com/hammadxp/back-it-up/blob/b930bd00f5bbe6d75ca407012ff28062a21b61f2/actions.ts#L5)
This will break when you click on a folder, as they don't have a basename.
It would be better if you checked that `file instanceof TFile` before adding the menu item, and updating the type of `file` in this function to be `TFile`.
Then you also won't need the `as any`.

[(file as any).basename](https://github.com/hammadxp/back-it-up/blob/b930bd00f5bbe6d75ca407012ff28062a21b61f2/actions.ts#L25), [export async function makeSnapshotOfNote(file: TAbstractFile) {](https://github.com/hammadxp/back-it-up/blob/b930bd00f5bbe6d75ca407012ff28062a21b61f2/actions.ts#L22)
Same here.

PR #3183, opened 2024-03-04
Add plugin: Metafolders

makary-s:

## Repo URL

Link to my plugin: https://github.com/makary-s/obsidian-metafolders

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


makary-s:
/skip


> [[1](https://github.com/makary-s/obsidian-metafolders/blob/6db83d72e80df90da46a49554ebcb17ef111e9f8/src/utils/hierarchy.ts#L46)]:You should not cast this, instead use a `instanceof` check to make sure that it's actually a file/folder

I was defining the type of an empty array. In any case, I removed the cast.

> You should avoid assigning styles via JavaScript or in HTML

This is not styling, but working with vertical collapse animation.

joethei:
["minAppVersion": "0.15.0",](https://github.com/makary-s/obsidian-metafolders/blob/6b0fc1cbc3ea20ce851f871d5d8ba27226dc5179/manifest.json#L5)

makary-s:
@joethei thank you for the comment. I've checked the changelog and see that I'm indeed using features added in `v1.4.0`, which is a later version than my current one.

I've updated [`scripts/version-bump.mjs`](https://github.com/makary-s/obsidian-metafolders/blob/98318766c12d77afac32599f1e4fc7b07bbce03b/scripts/version-bump.mjs#L11) to take the current version of the obsidian API from `package-lock.json` when publishing.

Now, the version `1.4.11` is specified in `manifest.json` and `versions.json`.

PR #3315, opened 2024-03-30
Add plugin: Typing Sounds

jobedom:

## Repo URL

Link to my plugin: https://github.com/jobedom/obsidian-typing-sounds/

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


joethei:
[import { join } from "path";](https://github.com/jobedom/obsidian-typing-sounds/blob/9e69b52d72f82a9e9849e02794210d1d47f0458e/main.ts#L2)

[join(plugin.app.vault.configDir, "plugins", `obsidian-${plugin.manifest.id}`, filename)](https://github.com/jobedom/obsidian-typing-sounds/blob/9e69b52d72f82a9e9849e02794210d1d47f0458e/main.ts#L22)
The sound files will not be included when somebody downloads your plugin, they should be bundled into the `main.js` file instead.

PR #3318, opened 2024-03-31
Add plugin: Click Clack

Acylation:

## Repo URL

Link to my plugin: https://github.com/Acylation/obsidian-click-clack

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_ (Partly support)
  - [x] `main.js`
  - [x] `manifest.json`
- [x] I have read the tips in https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines and have self-reviewed my plugin to avoid these common pitfalls. (Global `app` reference is still used in this plugin)

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -11107,5 +11107,12 @@
     "author": "Benature",
     "description": "Execute commands on specific events.",
     "repo": "Benature/obsidian-automation"
+  },
+  {
+    "id": "click-clack",
+    "name": "Click Clack",
+    "author": "Acylation",
+    "description": "Simulates typewriter / mechanical keyboard sounds.",
+    "repo": "Acylation/obsidian-click-clack"
   }
 ]


joethei:
[const data = await new Blob([await app.vault.adapter.readBinary(path)], {](https://github.com/Acylation/obsidian-click-clack/blob/a4e00b00fc90d3cd566c5ded6db10a8b95f2085e/src/schemeHelpers.ts#L8), [app.vault.configDir,](https://github.com/Acylation/obsidian-click-clack/blob/a4e00b00fc90d3cd566c5ded6db10a8b95f2085e/src/schemeHelpers.ts#L39), [if (await app.vault.adapter.exists(manifestPath)) {](https://github.com/Acylation/obsidian-click-clack/blob/a4e00b00fc90d3cd566c5ded6db10a8b95f2085e/src/schemeHelpers.ts#L48), [const blob = new Blob([await app.vault.adapter.readBinary(zipFile)], {](https://github.com/Acylation/obsidian-click-clack/blob/a4e00b00fc90d3cd566c5ded6db10a8b95f2085e/src/fetchHelpers.ts#L7) and more

[path.join(](https://github.com/Acylation/obsidian-click-clack/blob/a4e00b00fc90d3cd566c5ded6db10a8b95f2085e/src/schemeHelpers.ts#L38), [key: path.join(schemePath, manifest['key'] as string),](https://github.com/Acylation/obsidian-click-clack/blob/a4e00b00fc90d3cd566c5ded6db10a8b95f2085e/src/schemeHelpers.ts#L54-L58)

[const download = async (name: string, localPath: string) => {](https://github.com/Acylation/obsidian-click-clack/blob/a4e00b00fc90d3cd566c5ded6db10a8b95f2085e/src/fetchHelpers.ts#L64-L93)
The sound should be bundled into the `main.js` file, not downloaded seperately.

Acylation:
Hi, I've released v0.1.2 with modifications upon your suggestions. https://github.com/Acylation/obsidian-click-clack/releases/tag/0.1.2

### Changes

> It's highly discouraged that you use the global app instance in your plugin.

I eliminated the use of global `app`, and pass the `plugin.app` to helper instances.

> This import is only available from a Node.js runtime, which means that this will throw errors on Mobile (that's a problem since you have isDesktopOnly marked as false in your manifest.json file.

I switched to string array `join('/')` and wrapped it up with `normalizePath()`. Now the folder related features work well on both iOS and Windows devices.

### Reserved

> The sound should be bundled into the main.js file, not downloaded seperately.

Actually the plugin have [a built-in one](https://github.com/Acylation/obsidian-click-clack/blob/main/src/defaultSound.ts), encoded in base64. Typing sound feature is literally out-of-the-box.

Fetching extra resources from github release is optional. The separated design of sound resources (stored in the plugin folder, and can acquire from remote) can provide more flexibility as well as keep the main plugin body at a minimal size. As I'm planning to introduce more resources, I think keeping this design is better.

PR #3339, opened 2024-04-05
Add plugin: Random names

palfrey:

## Repo URL

Link to my plugin: https://github.com/palfrey/obsidian-random-names

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -11366,5 +11366,12 @@
         "author": "KraXen72 and Chetachi Ezikeuzor",
         "description": "Paint text different colors.",
         "repo": "KraXen72/obsidian-painter"
+    },
+    {
+        "id": "random-names",
+        "name": "Random names",
+        "author": "Tom Parker-Shemilt",
+        "description": "Generates random names.",
+        "repo": "palfrey/obsidian-random-names"
     }
 ]


joethei:
[id: `random-names-${first}-${second}`,](https://github.com/palfrey/obsidian-random-names/blob/baf3ca590141455086ac332ccc2f37d0e994eb06/main.ts#L43)
Don't prefix the command id with the plugin id, Obsidian will make sure that there are no conflicts with other plugins.

[import * as fs from "fs";](https://github.com/palfrey/obsidian-random-names/blob/baf3ca590141455086ac332ccc2f37d0e994eb06/generate-names.ts#L1-L2)

[const rootName = "node_modules/@xaroth8088/random-names/generators";](https://github.com/palfrey/obsidian-random-names/blob/baf3ca590141455086ac332ccc2f37d0e994eb06/generate-names.ts#L5)
These files won't be shipped with the plugin, you will need to bundle them instead.

palfrey:
> [id: `random-names-${first}-${second}`,](https://github.com/palfrey/obsidian-random-names/blob/baf3ca590141455086ac332ccc2f37d0e994eb06/main.ts#L43) Don't prefix the command id with the plugin id, Obsidian will make sure that there are no conflicts with other plugins.

Fixed in https://github.com/palfrey/obsidian-random-names/commit/8cf7ac71dc676ff16ccde9692d54294b67ec00ec and will do a new release once confirmed no other issues here.

> [import * as fs from "fs";](https://github.com/palfrey/obsidian-random-names/blob/baf3ca590141455086ac332ccc2f37d0e994eb06/generate-names.ts#L1-L2) These imports are only available from a Node.js runtime, which means that this will throw errors on Mobile (that's a problem since you have `isDesktopOnly` marked as `false` in your manifest.json file.
> 
> [const rootName = "node_modules/@xaroth8088/random-names/generators";](https://github.com/palfrey/obsidian-random-names/blob/baf3ca590141455086ac332ccc2f37d0e994eb06/generate-names.ts#L5) These files won't be shipped with the plugin, you will need to bundle them instead.

Both of these aren't actually relevant. So, the `generate-names.ts` is only used during the [postinstall](https://github.com/palfrey/obsidian-random-names/blob/main/package.json#L10) step to generate files with the lists of generators, and none of the `generate-names.ts` code will be run on the actual end-users machines, only on dev or CI systems.

PR #3344, opened 2024-04-06
Upgrade stale action to fix running out of operations

palfrey:
Not a plugin or theme PR. This upgrades the `stale` action to v9, because the action is [running out of operations](https://github.com/obsidianmd/obsidian-releases/actions/runs/8579626949/job/23514812480) and [v9 explicitly fixes this](https://github.com/actions/stale/releases/tag/v9.0.0)


joethei:
Thank you

PR #3380, opened 2024-04-15
Add plugin: neovim-edit

Han8931:

## Repo URL

Link to my plugin: https://github.com/Han8931/neovim-edit

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -11281,6 +11281,6 @@
 		"name": "NeoVim Plugin",
 		"author": "Han",
 		"description": "This plugin allows you to seamlessly use NeoVim to edit your notes within Obsidian",
-		"repo": "Han8931/neovim-edit "
+		"repo": "Han8931/neovim-edit"
 	}
 ]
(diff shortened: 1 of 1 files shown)


joethei:
["description": "This plugin allows you to seamlessly use NeoVim to edit your notes within Obsidian",](https://github.com/Han8931/neovim-edit/blob/9260310e9b3bc80acbd4494f6b2ede0e5558b6dc/manifest.json#L6)
Avoid including sentences like "This is a plugin for Obsidian that does ..." in your description, it should be self-evident that this is a plugin for Obsidian.

[hotkeys: [](https://github.com/Han8931/neovim-edit/blob/9260310e9b3bc80acbd4494f6b2ede0e5558b6dc/main.js#L11-L16)


[callback: () => {](https://github.com/Han8931/neovim-edit/blob/9260310e9b3bc80acbd4494f6b2ede0e5558b6dc/main.js#L17-L19)
This should be a [conditional command](https://docs.obsidian.md/Plugins/User+interface/Commands#Conditional+commands), so that it only shows up if there is a file open.

PR #3538, opened 2024-05-17
Add plugin: Focus Tracker

jeetsukumaran:

## Repo URL

Link to my plugin: https://github.com/jeetsukumaran/obsidian-focus-tracker

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [X]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -11779,5 +11779,12 @@
         "author": "Mara-Li",
         "description": "Add custom CSS classes to text based on regular expressions.",
         "repo": "Mara-Li/obsidian-regex-mark"
+    },
+    {
+        "id": "focus-tracker",
+        "name": "Focus Tracker",
+        "author": "Jeet Sukumaran",
+        "description": "Track and align your focus.",
+        "repo": "jeetsukumaran/obsidian-focus-tracker"
     }
 ]


joethei:
[async getFrontmatter(path: string): Promise<{[key: string]: any}> {](https://github.com/jeetsukumaran/obsidian-focus-tracker/blob/e56b3eff6060aef4d4f57c3494913c234ad8bbfc/src/FocusTracker.ts#L316-L335)
To get the parsed frontmatter use `this.app.metadataCache.getFileCache(file).frontmatter`

joethei:
Could you please explain why you did that change, and then reverted that again?

jeetsukumaran:
> Could you please explain why you did that change, and then reverted that again?

This was a mistake -- I was debugging something else.

I re-implemented the change in the latest release.

Sorry for the confusion.

jeetsukumaran:
> Could you please explain why you did that change, and then reverted that again?

Update: actually, it turns out that there is a functional difference.

When I use:

```
return await this.app.metadataCache.getFileCache(file)?.frontmatter || {};
```

The cache is not updated quickly enough to reflect changes made by to the frontmatter when clicking on an element (as can be seen in release 1.0.4). This results in the display being out of sync.

In release 1.0.5, I re-re-re-reverted :) back to extracting  the string and parsing:

```
            return await this.app.vault.read(file).then((result) => {
                const frontmatter = result.split("---")[1];
                if (!frontmatter) {
                    return {};
                }
                return parseYaml(frontmatter);
            });
```

and you can see the clicks work as expected, as the yaml updates immediately.

PR #3563, opened 2024-05-21
Add plugin: Vault Name

gapmiss:

## Repo URL

Link to my plugin: https://github.com/gapmiss/obsidian-vault-name

## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -11912,5 +11912,12 @@
         "author": "Huajin",
         "description": "Create tabs in your notes.",
         "repo": "xhuajin/obsidian-tabs"
+    },
+    {
+    	"id": "vault-name",
+    	"name": "Vault Name",
+    	"author": "@gapmiss",
+        "description": "Display and customize the vault name (title) in the side navigation file explorer.",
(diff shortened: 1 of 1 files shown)


gapmiss:
@joethei I got to thinking about this plugin and the fact that it will only work with Obsidian version 1.6.0+. I have considered downgrading my install of Obsidian and see if I can make it work for pre-1.6.0 versions, but would rather just leave as is. How can I limit the plugin for only 1.6.0+? If I add the latest version of the API in the manifest.json would that only allow it to be installed for the targeted version(s)? Please advise and thank you.

joethei:
set the `minAppVersion` in the manifest to 1.6.0, then only versions after that will be able to install the latest version of your plugin.
You should also update the `versions.json` file, as Obsidian will check that file to find out which version of the plugin it should install if it's below 1.6.0

PR #3574, opened 2024-05-23
Update Plugin Description: Solve

LiamRiddell:
Hi, 

I'm the owner of Solve which has been approved for over 6 months. I was wondering if I'd be able to update the description with the submitted as it better represents the plugin. Is that possible?


LiamRiddell:
Thank you @joethei!

PR #3611, opened 2024-05-31
Add plugin: More Data

viethung0823:

## Repo URL

Link to my plugin:

## Release Checklist
  - [ x ]  Windows
  - [ x ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


viethung0823:
``This should not be necessary.`` repplied: 

the reason I had this code because when I toggle from the preview mode to the source mode the view state of my plugin just gone so => ``this.app.workspace.getActiveLeavesOfType(MoreDataView);  => return 0;  ``but I dont want it create any new leaf view, just open the same view but in source mode and keeping all the views setting(icons, actions, canAcceptExtension,..) remains . So that was a workaround for me to keeping the view state, I wish there is a better way to do it. Thanks for other comments I fixed them all.
[code](https://github.com/viethung0823/MoreData/blob/48a56d902d2f42132b0b8c05d9acd7cc9585e258/src/main.ts#L46-L56)

joethei:
> the reason I had this [code](https://github.com/viethung0823/MoreData/blob/48a56d902d2f42132b0b8c05d9acd7cc9585e258/src/main.ts#L46-L56) because when I toggle from the preview mode to the source mode the view state of my plugin just gone so => `this.app.workspace.getActiveLeavesOfType(MoreDataView); => return 0; `but I dont want it create any new leaf view, just open the same view but in source mode and keeping all the views setting(icons, actions, canAcceptExtension,..) remains . So that was a workaround for me to keeping the view state, I wish there is a better way to do it. Thanks for other comments I fixed them all.

You need to implement `getViewType()` and `getIcon()` in your view, then remove that code.

PR #3676, opened 2024-06-10
Update the description of the submitted plugin Remotely Save

fyears:
My plugin Remotely Save has been submitted before. Currently I would like to change the description of it.


fyears:
Close then create another pr to follow the "Add plugin" template

PR #3724, opened 2024-06-18
Add plugin: Daily Note Structure

db-developer:

## Repo URL https://github.com/db-developer/daily-note-structure

Link to my plugin:

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -12199,5 +12199,12 @@
         "author": "rooyca",
         "description": "It helps you remember what you were doing.",
         "repo": "rooyca/doing"
+    },
+    {
+      "id": "daily-note-structure",
+      "name": "Daily Note Structure",
+      "author": "db-developer",
+      "description": "One-Click create a structure for and including your daily notes.",
+      "repo": "db-developer/daily-note-structure"
     }
 ]


db-developer:
Released version 1.0.1
- replaced imports of 'moment' by importing moment from 'obsidian'
- removed import of nodejs 'path' module.

joethei:
[new Setting( containerEl )](https://github.com/db-developer/daily-note-structure/blob/ea334693c0131584f4ab8bbb63200ddc41fa5b21/plugin/plugin.ts#L34-L42), [new Setting( containerEl )](https://github.com/db-developer/daily-note-structure/blob/ea334693c0131584f4ab8bbb63200ddc41fa5b21/plugin/plugin.ts#L43-L51), [/* Hide obsidians button for its own 'daily note' in sidebar */](https://github.com/db-developer/daily-note-structure/blob/ea334693c0131584f4ab8bbb63200ddc41fa5b21/styles.css#L10-L55)

db-developer:
Point taken. I did not know about the possibility to customize the ribbon and I prefer the solution provided by obsidians settings panel.
... but ... 😁 ...
My plugin introduces an alternate way to create daily folders. I therefore assume that users will spontaneously decide to remove one or other of the 'Daily Folders' icons in the ribbon. I also think that it is obvious that they expect the options for the ribbon buttons in the settings of my plugin (At least I expect feature requests popping up in my repository)
In order not to present the users with lengthy and repetitive explanations in different places, about where they can find the setting options in obsidian, it would be great if I could reproduce the button marked "1" in the snapshot below in my settings, so that the dialog of obsidian simply pops up and can then take over the job.
I would describe the marked button like the "light switches" at the ends of a long corridor, that lead from different directions to the same consensual result.

Hence the question: How can I call up the dialog (see snapshot) from the obsidian settings in my plugin?

<hr>

The second point is the incorrectly created icon marked with '2'. If you look at the HTML code for the SVG icon in the DOM, you can see that the icon is missing all the necessary size information, which is why it has shrunk to a gray dot in the upper left corner of the button.
The code for this can be found in the 'plugin.ts' in the DailyNoteStructurePlugin class in the member #addRibbons()
To avoid the problem with the icon display in my plugin, I set the properties of the SVGSVGHTMLElement manually.
But now the icon appears in another place (see screenshot) where I don't have this control. I cannot see from the documentation how I can persuade the 'addIcon' function or the 'addRibbonIcon' member from the obsidian API to create the icon with the appropriate attributes on its own.

Is, what is going wrong here, the result of a bug (when creating the SVG node) or do I simply not know a trick that I should know?

... un' wenn die Übersetzung von Deepl net verständlisch woar, gern nochemo uff hessisch 😉

image

db-developer:
Toggles for ribbon items removed. New version 1.1.0

PR #3736, opened 2024-06-21
Add Vimium plugin

karstenpedersen:

## Repo URL

Link to my plugin:

## Release Checklist
  - [X]  Windows
  - [X]  macOS
  - [X]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


joethei:
[const tlds = [](https://github.com/karstenpedersen/obsidian-vimium/blob/49029b893efe77ddf70c5423102a976dff23da68/src/tlds.ts#L1-L754)
What are these for?

karstenpedersen:
Hi, if it is the file that is named `tlds.ts`, then it contains a list of strings that will be put on the marks, so the user can select them. The marks appear on clickable elements, like buttons. It could for example be "ac", so the user would have to press "ac" to click it.

joethei:
Why a hardcoded list?
Why not let the user decide what they want to have as a marker?
It's also a bit weird since that file also contains some longer TLD's .

karstenpedersen:
There is longer TLDS in case there are more clickable elements than the first ones, but it is a good idea with the user being able to modify them. I will have some by default, if that is okay, so the plugin is usable.

karstenpedersen:
The TLDS can't have the same prefix because when the user types then, then they automatically trigger. That is why there are some longer words.

PR #3778, opened 2024-06-27
Add plugin: Iconic

gfxholo:

## Repo URL

Link to my plugin: https://github.com/gfxholo/iconic

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -12437,5 +12437,12 @@
         "author": "Eric Mika",
         "description": "Sync flashcards from a folder in your vault to Anki. Pure Markdown syntax. No fuss.",
         "repo": "kitschpatrol/yanki-obsidian"
+    },
+    {
+        "id": "iconic",
+        "name": "Iconic",
+        "author": "Holo",
+        "description": "Customize your app icons directly from the UI, including tabs, files, bookmarks, properties, and ribbon commands.",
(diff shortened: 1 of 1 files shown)


gfxholo:
/skip

Thanks bot! I was expecting these two points to come up.

1. This plugin sets the `stroke` attribute on individual SVGs to support custom hex colors, which can't really be achieved using hardcoded CSS. It also sets a `filter` attribute on emojis for the same reason.
2. There are 3 casts to `any` in the codebase:
	- 2 in [`IconPicker.ts`](https://github.com/gfxholo/iconic/blob/main/src/IconPicker.ts#L440) to fetch a localization string using a color ID, which is guaranteed to be valid.
	- 1 in [`IconicSettingTab.ts`](https://github.com/gfxholo/iconic/blob/main/src/IconicSettingTab.ts#L16) to avoid some type-checking on an object full of button components, which are always assigned before their first use.

The casts *can* be rewritten in other ways, so let me know if they seem too egregious.

gfxholo:
Big thanks for reviewing this, @joethei 💜

PR #3819, opened 2024-07-04
Add plugin: Code Link

observerw:

## Repo URL

Link to my plugin: https://github.com/observerw/obsidian-code-link

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -12465,5 +12465,12 @@
 	    "author": "jvsteiner",
 	    "description": "Render small latex snippets in your notes as SVG and PNG files.",
 	    "repo": "jvsteiner/obsidian-latex-render"
+    },
+    {
+    	"id": "code-link",
+    	"name": "Code Link",
+    	"author": "observerw",
+    	"description": "Link to code files in your notes",
+    	"repo": "observerw/obsidian-code-link"
     }
 ]


joethei:
[suggest.ts](https://github.com/observerw/obsidian-code-link/blob/fc934794dbef7e10f4836e01729cdecfc780a87d/src/settings/suggester/suggest.ts#L1-L211)

[const relPath = `${configDir}/plugins/obsidian-${pluginId}`;](https://github.com/observerw/obsidian-code-link/blob/fc934794dbef7e10f4836e01729cdecfc780a87d/src/lang/loader.ts#L27)
Wrong path.

[protected async _downloadPackage(pkg: string): Promise<void> {](https://github.com/observerw/obsidian-code-link/blob/fc934794dbef7e10f4836e01729cdecfc780a87d/src/lang/loader.ts#L57-L77)
This should be bundled into the plugin, not downloaded externally.

observerw:
> [suggest.ts](https://github.com/observerw/obsidian-code-link/blob/fc934794dbef7e10f4836e01729cdecfc780a87d/src/settings/suggester/suggest.ts#L1-L211) This code should no longer be used, implement a `AbstractInputSuggest` instead.
> 
> [const relPath = `${configDir}/plugins/obsidian-${pluginId}`;](https://github.com/observerw/obsidian-code-link/blob/fc934794dbef7e10f4836e01729cdecfc780a87d/src/lang/loader.ts#L27) Wrong path.
> 
> [protected async _downloadPackage(pkg: string): Promise {](https://github.com/observerw/obsidian-code-link/blob/fc934794dbef7e10f4836e01729cdecfc780a87d/src/lang/loader.ts#L57-L77) This should be bundled into the plugin, not downloaded externally.

Thank you for your time. 

As for the last problem, tree-sitter needs to be loaded as a `.node` file, so it can't be included as part of the plugin as JavaScript code. If "bundle" means distributing `.node` with the plugin, I'm not quite sure how to do that, since as far as I know plugins are only allowed to include a single `main.js` file.

As for the previous problems, I will fix them soon.

observerw:
In addition, I went and looked up the documentation for esbuild, and there doesn't seem to be a way to inline the binary into the file, and that would apparently make the plugin file very large; also, the tag search feature is optional in the plugin, so having users who need this feature additionally download the tree-sitter related files seems to be the more appropriate way to go.

For these reasons, I think it makes more sense to keep the current code for now 🤔

PR #3877, opened 2024-07-13
Add Location Database Plugin

natedog2424:

## Repo URL

Link to my plugin:
https://github.com/natedog2424/ObsidianLocationDatabase

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

I don't have the ability to test on the other platforms but the plugin is very simple and if it works on desktop and mobile it should work on everything

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1,5 +1,12 @@
 [
     {
+  	"id": "location-database",
+  	"name": "Location Database",
+  	"author": "ND24 Dev",
+  	"description": "Create a database of locations of interest.",
+	"repo": "natedog2424/ObsidianLocationDatabase"
+    },
+    {
         "id": "nldates-obsidian",
         "name": "Natural Language Dates",
         "author": "Argentina Ortega Sainz",


natedog2424:
/skip
I provide instanceof checks before casting so I don't think this is necessary

joethei:
[Copyright (c) [year] [fullname]](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/LICENSE#L3)
Fill this.

["minAppVersion": "0.15.0",](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/manifest.json#L5)

[fundingUrl": "https://github.com/natedog2424",](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/manifest.json#L9)

[// Remember to rename these classes and interfaces!](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L3)
Remove this comment.

[const title = args.get("q") ?? "New Location";](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L22), ['New Location',](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L243)

 [const address = args.get("address") ?? "NO ADDRESS";](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L23-L24)
Don't use caps here.

[const folder = this.app.vault.getAbstractFileByPath(folderPath);](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L73)
use `getFolderByPath` to check if it's actually a folder.

[navigator.clipboard.readText().then(async (text) => {](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L117)
[prefer async/await over Promise](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Prefer%20async/await%20over%20Promise)

[(folder as TFolder)](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L154), [let file = abs_file as TFile;](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L161)
This is not neccessary, TypesScript automatically knows the correct type based on the conditions above.

[const frontmatterRegex = /^---\n([\s\S]*?)\n---\n/;](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L164-L168), [const imageRegex = /image: https:\/\/cartes.io\/api\/maps\/(.*)\/images\/static/;](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L181-L188)
Use the `processFrontMatter` API to modify the frontmatter, instead of manually parsing the YAML.

liamcain:
- [const frontmatterMatch = content.match(frontmatterRegex)](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L165) instead of parsing the frontmatter yourself, you should consider using the frontmatter object stored in the metadataCache. `this.app.metadataCache.getFileCache(file).frontmatter?.location` for example. This will avoid false-positives. For example, your locationRegex needs a "^" token at the start to avoid finding a location property at the non-top level.
- [getAbstractFileByPath(filePath) as TFile](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L104) this is a bad cast, you should still add an instanceof check to ensure you have a TFile
- [overrite existing file](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L103) why are we overwriting here? On L82, there's a check to ensure the file doesn't already exist, so the behavior doesn't seem consistent.
- [readText().then](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L117) your function is marked as async, why not just await the readText() instead? It would improve readability.
- [const folderPath = this.settings.entryFolder](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L72) consider adding another line of validation that checks that folderPath is not an empty string and showing a notice that the plugin is not fully configured. It would be more user friendly than saying "Folder does not exist"
- [if (!folder || !(folder instanceof TFolder)) {](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L150) same as previous comment
- [setPlaceholder('/locations')](https://github.com/natedog2424/ObsidianLocationDatabase/blob/b9ec8689ee9a40cfab17ab7d862ed28f81d2774a/main.ts#L348) if your placeholder text has an actual value here, then this should be the default value. Otherwise, users will be confused when they see the entry folder is "/locations" but the value does not match.

PR #3906, opened 2024-07-19
Add plugin: CSS Inlay Colors

GRA0007:

## Repo URL

Link to my plugin: https://github.com/GRA0007/obsidian-css-inlay-colors

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -12941,5 +12941,12 @@
         "author": "Paco Kwon, Stein Ingebretsen",
         "description": "Easily insert references to scriptures and conference talks from the Church of Jesus Christ of Latterday Saints.",
         "repo": "ingiestein/obsidian-lds-scriptures-plugin"
+    },
+    {
+        "id": "css-inlay-colors",
+        "name": "CSS Inlay Colors",
+        "author": "Benji Grant",
+        "description": "Show inline color hints for CSS colors",
+        "repo": "GRA0007/obsidian-css-inlay-colors"
     }
 ]


joethei:
Here is a proof of concept Liam made that implements a similar thing in Live Preview, it should be easily adaptable:
https://gist.github.com/liamcain/2f003c127efec7c7b089e14178cf5179

GRA0007:
Oh awesome thanks @joethei. I've been looking into regular codemirror plugins that do similar things as well, but this will be really helpful.

GRA0007:
I've released a new version of my plugin that adds support for colors in live preview mode @joethei

liamcain:
- [node.name.includes('inline-code')](https://github.com/GRA0007/obsidian-css-inlay-colors/blob/f5eee7504393729fd349c46177abf1b263bc569b/src/live.ts#L96) instead of using the name, you should use the `node.type.prop(tokenClassNodeProp)` and split that string by `' '` to get the tokens. Then check if any of them are `inline-code`. That will help avoid against false positives. And using an exact match will avoid you running into issues where you get a partial match `something-else-inline-code`
- https://github.com/obsidianmd/obsidian-api/blob/8be9ff1e1b708b9451f35ce2df33e3983e88565b/obsidian.d.ts#L1175 you can use this StateField to determine if live preview is enabled
- [currentTarget instanceof HTMLInputElement](https://github.com/GRA0007/obsidian-css-inlay-colors/blob/f5eee7504393729fd349c46177abf1b263bc569b/src/live.ts#L63) unfortunately, this won't work properly in [pop-out windows](https://help.obsidian.md/User+interface/Pop-out+windows). You will need to adapt this to use `yourElement.instanceOf(HTMLInputElement)` instead, which is our custom function for checking type instances for default elements. more info here: https://github.com/obsidianmd/obsidian-api/blob/8be9ff1e1b708b9451f35ce2df33e3983e88565b/obsidian.d.ts#L59
- [background](https://github.com/GRA0007/obsidian-css-inlay-colors/blob/f5eee7504393729fd349c46177abf1b263bc569b/src/preview.ts#L17) consider passing this as a css variable instead of just setting the background color directly. That would allow for more customization in CSS via themes and snippets.

GRA0007:
Thanks @liamcain! And no worries about the delay. Appreciate the detailed review, I spent hours researching how to achieve certain things like accessing the live preview state but I think the documentation may need some updating. I'll see about fixing these up as soon as I can! 🚀

PR #3923, opened 2024-07-21
Add plugin: Blockreffer

tyler-dot-earth:

## Repo URL

Link to my plugin: https://github.com/tyler-dot-earth/obsidian-blockreffer

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -12990,6 +12990,13 @@
         "author": "datawitch",
         "description": "Recursively copies all Markdown files in a folder, concatenates them, and copies them into the clipboard.",
         "repo": "StructByLightning/obsidian-recursive-copy"
+    },
+    {
+        "id": "blockreffer",
+        "name": "Blockreffer",
+        "author": "tyler.earth",
+        "description": "Search and embed blocks with ^block-references.",
+        "repo": "tyler-dot-earth/obsidian-blockreffer"
     }
 ]
 


joethei:
[This CSS file will be included with your plugin, and](https://github.com/tyler-dot-earth/obsidian-blockreffer/blob/6c2606e0c63252ab5c9b9fa8aadc1682db542ec1/styles.css#L3-L8)
Remove this comment

[.suggestion-content {](https://github.com/tyler-dot-earth/obsidian-blockreffer/blob/6c2606e0c63252ab5c9b9fa8aadc1682db542ec1/styles.css#L10-L14)

[console.log('embed', { item });](https://github.com/tyler-dot-earth/obsidian-blockreffer/blob/6c2606e0c63252ab5c9b9fa8aadc1682db542ec1/main.ts#L199)

[class BlockrefferSettingsTab extends PluginSettingTab {](https://github.com/tyler-dot-earth/obsidian-blockreffer/blob/6c2606e0c63252ab5c9b9fa8aadc1682db542ec1/main.ts#L216-L238)
There are no settings, so your plugin should not have a settings tab.

tyler-dot-earth:
@joethei Thanks for the review, I've made the requested corrections :+1:

liamcain:
- [lines.slice(block.position.start.line, block.position.end.line + 1).join('\n');](https://github.com/tyler-dot-earth/obsidian-blockreffer/blob/0540c783504fd49574fa3f40d9011c2b21f92af9/main.ts#L65) This is a very inefficient way to get a block from the file, splitting a file by newlines is expensive. In the metadataCache, we store blocks using Loc, which has line, col but it also has "offset." This is the offset from the start of the file, you should be using this value instead. The block content is fileContent.slice(start.offset, end.offset)

Everything else looks good though so I just ask you to fix up that one thing, it should make the whole suggest load a lot faster.

PR #3939, opened 2024-07-25
Add plugin: Emera

OlegWock:

## Repo URL

Link to my plugin: https://github.com/OlegWock/obsidian-emera

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -12997,6 +12997,13 @@
         "author": "tyler.earth",
         "description": "Search and embed blocks with ^block-references.",
         "repo": "tyler-dot-earth/obsidian-blockreffer"
+    },
+    {
+        "id": "emera",
+        "name": "Emera",
+        "author": "OlegWock",
+        "description": "Enables you to use custom React components and inline JavaScript, kinda like MDX.",
+        "repo": "OlegWock/obsidian-emera"
     }
 ]
 


OlegWock:
/skip 

I fixed usage of `innerHTML`, but `console.log` isn't really an issue since they will be deleted by esbuild (https://github.com/OlegWock/obsidian-emera/blob/master/esbuild.config.mjs#L41)

OlegWock:
@joethei sorry for ping, but I think there was some bug with this PR. 

From what I see in other PRs, after author sends /skip, bot will remove "Changes requested" and "Additional review required" and add "Ready for review" labels. But this didn't happen with my PR and I think you could miss this PR because of that. 

Let me know if I need to do anything else or if there is any problem with plugin

liamcain:
- [findCurrentView](https://github.com/OlegWock/obsidian-emera/blob/aacdf541e65e74538be5ca6b9489480d4795ddee/src/processors/code-processor.ts#L460) Instead of scanning the entire workspace for the matching view, you can use [this statefield](https://github.com/obsidianmd/obsidian-api/blob/df9434bec6883bd32727f61b1ed1336f7663d0e4/obsidian.d.ts#L1169) to find the MarkdownView
- [Don't do anything in Source mode, we care only about LivePreview](https://github.com/OlegWock/obsidian-emera/blob/aacdf541e65e74538be5ca6b9489480d4795ddee/src/processors/code-processor.ts#L470) don't call getState since that's a bit expensive to serialize everything. Instead just check the result of this [statefield](https://github.com/obsidianmd/obsidian-api/blob/df9434bec6883bd32727f61b1ed1336f7663d0e4/obsidian.d.ts#L1175)
- [node.type.name.startsWith('inline-code')](https://github.com/OlegWock/obsidian-emera/blob/aacdf541e65e74538be5ca6b9489480d4795ddee/src/processors/code-processor.ts#L497) instead of using the name, you might want to use the node.type.prop(tokenClassNodeProp) and split that string by `' '` to get the tokens. Then check if any of them are inline-code. That will help avoid against false positives. And using an exact match will avoid you running into issues where you get a partial match something-else-inline-code
- [node.type.name.includes('HyperMD-codeblock-begin')](https://github.com/OlegWock/obsidian-emera/blob/aacdf541e65e74538be5ca6b9489480d4795ddee/src/processors/code-processor.ts#L507) same for the codeblocks but you'll need to check type.prop(lineClassNodeProp) for those.

PR #3942, opened 2024-07-25
Add plugin: Recent Tab Switcher

samuelrawrs:

## Repo URL

Link to my plugin:

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -12794,5 +12794,12 @@
         "author": "Iulian Onofrei",
         "description": "Share notes via Notepad Tab (https://notepadtab.com).",
         "repo": "revolter/obsidian-share-via-notepad-tab-plugin"
+    },
+    {
+    	"id": "recent-tab-switcher",
+    	"name": "Recent Tab Switcher",
+    	"description": "Switch to the most recently used tab.",
+    	"author": "Samuel Ang",
+    	"repo" : "samuelrawrs/recent-tab-switcher"
     }
 ]


joethei:

[hotkeys: [{ modifiers: ["Alt"], key: "q" }],](https://github.com/samuelrawrs/recent-tab-switcher/blob/17ef7cc3eac0634a284a371af9433285d85125fe/main.ts#L10)

[const leaves = this.app.workspace.getLeavesOfType("markdown");](https://github.com/samuelrawrs/recent-tab-switcher/blob/17ef7cc3eac0634a284a371af9433285d85125fe/main.ts#L13)
Why only Markdown leaves?

samuelrawrs:
> Remove the styles.css file from your release, since it's empty it's just unnecessary bloat.

Will do!

> [hotkeys: [{ modifiers: ["Alt"], key: "q" }],](https://github.com/samuelrawrs/recent-tab-switcher/blob/17ef7cc3eac0634a284a371af9433285d85125fe/main.ts#L10)


Okay, I'll leave the default hotkey out!

> [const leaves = this.app.workspace.getLeavesOfType("markdown");](https://github.com/samuelrawrs/recent-tab-switcher/blob/17ef7cc3eac0634a284a371af9433285d85125fe/main.ts#L13)
Why only Markdown leaves?

No particular reason. Shall I modify this line to just get any leaf type? 

Thanks!

samuelrawrs:
@joethei I've updated the code but couldn't seem to find a suitable replacement for the deprecated `this.app.workspace.activeLeaf`, `getActiveViewOfType` is too specific for my use case and couldn't find a way to get the same output as the deprecated method. Do you have any recommendations or is it okay if I kept using `.activeLeaf` ?

PR #3966, opened 2024-07-28
Add plugin: Print

gitatmax:

## Repo URL
Link to my plugin: https://github.com/gitatmax/printsidian/

## Release Checklist

* [ ]  I have tested the plugin on
  * [x] Windows
  * [x] macOS
  * [x] Linux
  * [ ] Android _(if applicable)_
  * [ ] iOS _(if applicable)_
* [x]  My GitHub release contains all required files (as individual files, not just in the source.zip / source.tar.gz)
 * [x] `main.js`
 * [x] `manifest.json`
 * [ ] `styles.css` _(optional)_
* [x] GitHub release name matches the exact version number specified in my manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
* [x]  I have read the developer policies at https://docs.obsidian.md/Developer+policies, and have assessed my plugins's adherence to these policies.
* [x]  I have read the tips in https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines and have self-reviewed my plugin to avoid these common pitfalls.


gitatmax:
/skip Wondering if this is necessary? I've tried but have struggled getting it rolling, some. I'll keep cracking at it for the next release but, in the meantime, is there any chance this is kosher? Thank you.

joethei:
["id": "printsidian",](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/manifest.json#L2-L3)
Please rename your plugin, the name should not include abreviations of "Obsidian"

["description": "Print from Obsidian",](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/manifest.json#L6)
The description should also not include "Obsidian"

[name: 'Print Note',](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L21), [.setName('Enable Print Option')](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L166), [<title>Print Note: ${title}</title>](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L65)

[hotkeys: [{ modifiers: ["Mod", "Shift"], key: "P" }]](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L23)

[callback: () => this.printCurrentNote(),](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L22)
This should be using [conditional commands](https://docs.obsidian.md/Plugins/User+interface/Commands#Conditional+commands)

[console.log('printCurrentNote called');](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L42), [console.log('Active view found');](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L45), [console.log('Title:', title); // Log the title](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L48), [console.log('Window loaded, title:', document.getElementById('note-title').textContent);](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L129), [console.log('HTML content written to iframe');](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L141), [console.log('Title in iframe:', '${title}');](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L101), [console.log('Title element content:', document.getElementById('note-title').textContent);](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L103)

[<script src="https://cdn.jsdelivr.net/npm/marked/marked.min.js"></script>](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L92-L95)
These resources should be local, and not loaded externally.

[const renderer = new marked.Renderer();](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L113)
Obsidian has https://docs.obsidian.md/Reference/TypeScript+API/MarkdownRenderer/render, which should be used instead, it will also take care of syntax highlighting and MathJax rendering.

[document.getElementById('content').innerHTML = marked.parse(${JSON.stringify(content)}, { renderer: renderer });](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L123)
`innerHTML` is dangerous with untrusted user content, marked does not do any sanitization.

[const frameWindow = printFrame.contentWindow;](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L59-L138)
Also not a fan of putting everything in an iframe like this.

[containerEl.createEl('h2', { text: 'Print Plugin Settings' });](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L163)

[setName('Enable Print Option')](https://github.com/gitatmax/printsidian/blob/de738d8221b00ed79f62a773b043acb30c78dc3c/src/main.ts#L166-L173)
This should not be a setting, and it's not used anyway.

AnAngryRaven:
The delay variables I should be able to fix -- forgot I had them set to `var`.

The `isHovering` variable may prove to be slightly trickier though..the only reason it is written this way is because the title bar that the buttons (search, bookmark, etc) are in does not actually count towards either the `leftSplit` or the `leftRibbon`, meaning it collapses when attempting to hit any of these buttons.

I'll see if I can't come up with a different solution to avoid using `var`.

liamcain:
- [callback: () => this.printCurrentNote()](https://github.com/gitatmax/printsidian/blob/0364fcce4d9e6ef5c1740cad64540845b6e6c7af/src/main.ts#L21) you should use checkCallback instead and check to see if there is an active markdown file
- [const activeView = this.app.workspace.getActiveViewOfType(MarkdownView);](https://github.com/gitatmax/printsidian/blob/0364fcce4d9e6ef5c1740cad64540845b6e6c7af/src/main.ts#L36) you should use [`getActiveFile`](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L4146) instead. This will help if the user has focused a sidebar view.

PR #3974, opened 2024-07-30
Add plugin: Color Highlighter

heatherpiper:

## Repo URL

Link to my plugin: https://www.github.com/heatherpiper/color-highlighter

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -12822,5 +12822,12 @@
         "author": "logonoff",
         "description": "Adds Discord-like syntax for inline spoilers in reader mode.",
         "repo": "logonoff/obsidian-inline-spoilers"
+    },
+    {
+	"id": "color-highlighter",
+	"name": "Color Highlighter",
+	"author": "Heather Piper",
+	"description": "Highlight color codes in your notes.",
+	"repo": "heatherpiper/color-highlighter"
     }
 ]


heatherpiper:
/skip

5: This inline style is necessary for the settings UI layout. It's applied conditionally to a dynamically created element and is specific to the settings interface.

19: This inline style is necessary for proper rendering of Dataview inline query results. It's applied only to specific elements identified during document processing and can't be reliably targeted with a static CSS selector.

All other changes have been made and should be picked up by the bot shortly.

joethei:
Remove the `main.js` file from your `src` folder.

heatherpiper:
`main.js` has been removed.

liamcain:
- [await this.app.vault.modify(activeFile, newContent);](https://github.com/heatherpiper/color-highlighter/blob/17b3e527258de0fd07838eeebe83de99489be08c/main.ts#L196) don't attempt to parse and modify the frontmatter manually. There are race conditions and edge cases that are easy to run into. You should use [`processFrontmatter`](<https://github.com/obsidianmd/obsidian-api/blob/df9434bec6883bd32727f61b1ed1336f7663d0e4/obsidian.d.ts#L1450>) instead
- [refreshAllViews](https://github.com/heatherpiper/color-highlighter/blob/17b3e527258de0fd07838eeebe83de99489be08c/main.ts#L82) this is getting called a bit too aggressively. You really shouldn't need to iterate through every open leaf and rerender it. A better way to handle this is to use a MarkdownRenderChild and add it to current context <https://github.com/obsidianmd/obsidian-api/blob/df9434bec6883bd32727f61b1ed1336f7663d0e4/obsidian.d.ts#L2260>. You can then have your component register itself to metadataCache.on('changed') and see if it needs to be redrawn. Similarly with the editor extensions. You 
- [leaf.view.previewMode.rerender(true);](https://github.com/heatherpiper/color-highlighter/blob/17b3e527258de0fd07838eeebe83de99489be08c/main.ts#L95) You should not be calling this.
- [this.app.workspace.iterateAllLeaves](https://github.com/heatherpiper/color-highlighter/blob/17b3e527258de0fd07838eeebe83de99489be08c/main.ts#L69) you should not use rebuildView. Instead, you should call updateOptions, which will automatically reload all views in the workspace <https://github.com/obsidianmd/obsidian-api/blob/df9434bec6883bd32727f61b1ed1336f7663d0e4/obsidian.d.ts#L4675>
- [refreshAllViews](https://github.com/heatherpiper/color-highlighter/blob/17b3e527258de0fd07838eeebe83de99489be08c/main.ts#L78) why are you calling refresh views right after rebuilding the views? This shouldn't be necessary either if you're using updateOptions.

PR #3983, opened 2024-08-02
Add plugin: Promerge

leknes:

## Repo URL

Link to my plugin: https://github.com/Leknes/promerge

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -12892,5 +12892,12 @@
         "author": "Technerium",
         "description": "Graph of the Number of Files in the Vault.",
         "repo": "technerium/obsidian-vault-size-history"
+    },
+    {
+        "id": "promerge",
+        "name": "Promerge",
+        "author": "Leonhard Senkel",
+        "description": "Provides a unique way of merging files and directories into an intelligible folder where the latest version of a project can be observed.",
+        "repo": "Leknes/promerge"
     }
 ]


joethei:
[Copyright (c) [year] [fullname]](https://github.com/Leknes/promerge/blob/5efc63fe795de617d2704710acb970831398e471/LICENSE#L3)
Fill this.

Leknes:
Alright, I did both things.

clehene:
@leknes I found this randomly while reviewing pull requests (the name "promerge" sounded interesting) - so asking mostly out of curiosity. 

Your plugin seems to be enabling some sort of version control? 
If so two things come to mind:
- version control is super hard to get right through a tool (a few decades of version control management)
- there are great tools like git and github that do a fantastic job (git has been around for 20 years) 

Why create a tool / plugin for something that is hard to solve and already has a good solution rather than maybe document how soemthing like git / github can solve the problem?

leknes:
@clehene I initially made promerge, as I was trying to manage my projects more cleanly on obsidian. I wanted to be able to look into documents that are outdated, to get an impression of the whole development process, while also maintaining a neat organization.

But you are right, that git or github offer a much more sophisticated form of version control, which probably is the best approach for most cases. I think promerge still has a niche usecase though. For instance, when the user wants to be able to observe every version of a document without leaving obsidian, or does not want to use an external tool for version control, promerge might be a simple but adequate solution

clehene:
@leknes thank you, I think I tangentially understand the use-case. I use timestamped entries for all notes with a slightly related use-case of tracking changes and automating around this workflow. However I consider this more or less the main format and looking at getting aggregation of notes (or hubs) by composing these. 
So there's an explicit versisoning-like aspect derived from timestamped entries, but it's improper to call it versioning since it is data rather than metadata. 
At the same time I'm using the git plugin but mostly to protect from Obsidian corruption. 

TL;DR I believe there are legit funcitonal use-cases tangential to versioning and I guess yours is one of those (I guess also zettelkasten is in the same class)

Upon trying to read your README again - just blunt feedback - it's unclear how specific to your workflow this is and whether is soemthing that could be easily adopted by someone else. I'm not sure as I can't really _easily_ figure that out from the text example. I believe some more thorough examples could help as well as a more comprehensive description (e.g. blog post like) description of your workflow and how it relates to other etc. 

I'm giving you this feedback since I know there's a non-trivial effort to get a plugin in the hands of everyone else, and so you clearly see the value, so it's probably worth to make the extra effort to validate and ensure its success. :) 
Cheers!

clehene:
@joethei followign up with a note related to a comment I made on the forum to manage 1800+ plugins - both as a marketplace steward (you) as well as a user (me).

This is unrequested feedback and likely the wrong place to write it but since it's a good example I'll drop it here and maybe cross post it on forums :) 

I think BRAT https://github.com/TfTHacker/obsidian42-brat goes a long way and could be adopted natively. 

It would be nice to have a clear separation even within the marketplace between plugins that pass the test of time and users, and the more experimental ones (such as this as well as my remember-view-state plugin).

I believe a good analogy is that of brew which allows taps and there are few more like Kubernetes kubectl's Krew (https://krew.sigs.k8s.io/)  - both are CLI-based but discoverability and management are similar issues. 

So there's a few principles I'm conflating

- a memory garbage collector principle that leverages the idea that most objects will be short lived - so thriough analogy most plugins will be short lived (this is a very general principle). Objects that "survive" garbage colleciton are "promoted" to a "tenured" memory space. 
- release channels principles (think of beta, alpha/nightly channels) but more like a separation of "official" vs "experimental" plugins alowing users to browse both directly from obisdian. This could prioritize reviews too with more effort going inti the "official" ones and making it easier for everyone to get theirs in the "experimental" 

Combining these two ideas could be through different repositories providing a way for users to explicitely enable the either channel. This could open up the door for more enterprise features too (like private repositories, etc.). 
One-offs with BRAT as a native plugin.

leknes:
@clehene Yeah, I honestly don't think that this plugin will have like a very wide range of use-cases, but as I think, it at least has some theoretical usage scenarios, so that's fine. 

About the README, first of all, thanks for the feedback; I never really made a README for a published application so writing one is quite new to me. I will try to improve a bit on it, so that it is more understandable and not all to abstract.

PR #3988, opened 2024-08-02
Add plugin: Scripture Indexer

jdrbrn:

## Repo URL

Link to my plugin: https://github.com/jdrbrn/obsidian-scripture-indexer

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13480,5 +13480,12 @@
         "author": "imfenghuang",
         "description": "A free-to-use translatation service for foreign language markdown file.",
         "repo": "imfenghuang/obsidian-immersive-translate"
+  },
+  {
+	"id": "scripture-indexer",
+	"name": "Scripture Indexer",
+	"author": "jdrbrn",
+	"description": "Indexes references to scriptures in notes.",
+	"repo": "jdrbrn/obsidian-scripture-indexer"
     }
 ]


joethei:

["minAppVersion": "0.15.0",](https://github.com/jdrbrn/obsidian-scripture-indexer/blob/b623f231d41cea4b6f65be36e260de61d935fc96/manifest.json#L5)

[name: 'Index All Files',](https://github.com/jdrbrn/obsidian-scripture-indexer/blob/b623f231d41cea4b6f65be36e260de61d935fc96/main.ts#L66), [name: 'Index This File',](https://github.com/jdrbrn/obsidian-scripture-indexer/blob/b623f231d41cea4b6f65be36e260de61d935fc96/main.ts#L74), [name: 'Reset Index',](https://github.com/jdrbrn/obsidian-scripture-indexer/blob/b623f231d41cea4b6f65be36e260de61d935fc96/main.ts#L85)

[this.IndexFile(this.app.vault.getFileByPath(file.path)!);](https://github.com/jdrbrn/obsidian-scripture-indexer/blob/b623f231d41cea4b6f65be36e260de61d935fc96/main.ts#L98), [this.IndexFile(this.app.vault.getFileByPath(file.path)!);](https://github.com/jdrbrn/obsidian-scripture-indexer/blob/b623f231d41cea4b6f65be36e260de61d935fc96/main.ts#L115)
Don't assume that a file exists, always check.

jdrbrn:
Thanks for the code review! I believe the necessary changes have all been made. Please let me know if there's anything else.

As a follow up question, why should I check for files existing in the situations you linked? From my understanding of the documentation those events should only be triggered on file changes and return files that do exist. Is there a possibility that they return a file that doesn't exist (or is a folder as a TAbstractFile says it could represent)? Or is it just a best practices thing? Thanks again for your help.

liamcain:
- [Could be different than the previously loaded one if data was synced after plugin was loaded](https://github.com/jdrbrn/obsidian-scripture-indexer/blob/717ff91aebe8e54c2f6adce6279dd147fd54be42/main.ts#L148) If settings were loaded after the plugin, then you can use onExternalSettingsChange (<https://github.com/obsidianmd/obsidian-api/blob/df9434bec6883bd32727f61b1ed1336f7663d0e4/obsidian.d.ts#L3088>). I highly recommend not loading settings here. This means reading from the filesystem on every edit, which is way more computationally expensive than it needs to be.
- [this.saveSettings](https://github.com/jdrbrn/obsidian-scripture-indexer/blob/717ff91aebe8e54c2f6adce6279dd147fd54be42/main.ts#L160) you should also [debounce](https://github.com/obsidianmd/obsidian-api/blob/1b4f6e2e5753c8ac2d538bc5d55c38f228450cb2/obsidian.d.ts#L772) your saveSettings method.
- [await files.forEach(file => this.ScrapeFile(file))](https://github.com/jdrbrn/obsidian-scripture-indexer/blob/717ff91aebe8e54c2f6adce6279dd147fd54be42/main.ts#L140) this is not configured correctly. `ScrapeFile` is async, so to properly wait for it to get called on every file in the vault, you need to `await` each call to `ScrapeFile`. Something like would be better:

```ts
for (let file of files) {
  await this.ScrapeFile(file)
}
```

jdrbrn:
Thanks for the advice.

- Updated to use the onExternalSettingsChange function and removed the manual reloading of the settings file
- Debounced the saveSettings method and also most cases of WriteIndex as it's a similar idea, just writing to an actual note not the settings file
- Updated IndexAllFiles to properly scrape files async

PR #3991, opened 2024-08-03
Add plugin: Open Sidebar on Hover

AnAngryRaven:

## Repo URL

Link to my plugin: https://github.com/AnAngryRaven/obsidian-open-sidebar-on-hover

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13487,5 +13487,12 @@
 	"author": "jdrbrn",
 	"description": "Indexes references to scriptures in notes.",
 	"repo": "jdrbrn/obsidian-scripture-indexer"
+  },
+  {
+        "id": "open-sidebar-on-hover",
+        "name": "Open Sidebar on Hover",
+        "author": "AnAngryRaven",
+        "description": "Enables the ability to open the sidebar by hovering over it.",
+        "repo": "AnAngryRaven/obsidian-open-sidebar-on-hover"
     }
 ]


liamcain:
Hi @AnAngryRaven just wanted to check in and see if you saw the comments made by our automated review system. I took a peak and everything else looks okay, so just changing the `var`s to `let` should be enough and I'll merge this.

AnAngryRaven:
@liamcain Ah, apologies!! I did indeed, though I see why I had given myself a few days (then subsequently forgot.. 😞) to think things over.

The `delayTime` variable should be easily changed, and I'll be sure to do that. The trickier one is the `isHovering` variable, as I guess I'm essentially inducing a race condition. From what I remember, certain areas of the leftSplit (the title bar surrounding the files, search, and bookmark buttons) aren't actually considered part of the overall document, meaning things like `mouseleave` don't work as expected. So this was the solution I came up to get around that.

From my understanding of `let`, it wouldn't really be possible to use anything other than `var`, as the variable needs to be accessed throughout different scopes. Though I wouldn't be surprised if there's a different way to do it that I just haven't seen yet, so I'll still see what I can do 🙂‍↕️

It has just occurred to me that this probably isn't actually intended behaviour to begin with though, so perhaps a bug report would be more appropriate...

liamcain:
For `isHovering`, it looks like you can just change `var` to `let`. It should work exactly as expected.

AnAngryRaven:
Ahh, apologies!! Appears as though I misunderstood all of the explanations of how let works.. have now changed them accordingly

PR #3993, opened 2024-08-03
Add plugin: Daily Note Collector

LA:

## Repo URL

Link to my plugin: https://github.com/la/obsidian-daily-note-collector

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13158,6 +13158,13 @@
         "author": "Ram Rachum",
         "description": "Easily select text in backticks.",
         "repo": "cool-RR/backtick-text-selector"
+    },
+    {
+        "id": "daily-note-collector",
+        "name": "Daily Note Collector",
+        "author": "Adar Butel",
+        "description": "Adds links to new notes to your daily note.",
+        "repo": "LA/obsidian-daily-note-collector"
     }
 ]
 


joethei:
["minAppVersion": "0.15.0",](https://github.com/LA/obsidian-daily-note-collector/blob/5d7e92f1411e2ff25a5de82c702c57d62799898e/manifest.json#L5)

LA:
> ["minAppVersion": "0.15.0",](https://github.com/LA/obsidian-daily-note-collector/blob/5d7e92f1411e2ff25a5de82c702c57d62799898e/manifest.json#L5) You are using newer API's, set this to the latest public build number.

I only left what the [sample plugin default is](https://github.com/obsidianmd/obsidian-sample-plugin/blob/master/manifest.json#L5), but I have updated my [minAppVersion to 1.6.7](https://github.com/LA/obsidian-daily-note-collector/blob/master/manifest.json#L5) and created [a new release](https://github.com/LA/obsidian-daily-note-collector/releases/tag/1.0.2).

liamcain:
- [this.app.vault.on("create", this.onCreateFile.bind(this))](https://github.com/LA/obsidian-daily-note-collector/blob/54243029e4f6baacd51dbf7f19f532c29f729803/main.ts#L10) you should not register the 'create' vault event directly in `onload`. This registration should be wrapped inside a [`onLayoutReady`](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3626) callback. To see why, try restarting Obsidian and opening your daily note. "create" is fired for every note in your vault as part of its initialization process, so you will end up with a link to every note in your vault.
- [const date = new Date();](https://github.com/LA/obsidian-daily-note-collector/blob/54243029e4f6baacd51dbf7f19f532c29f729803/main.ts#L94-L99) The daily note path is configurable. You can't assume that the file path will be YYYY-MM-DD.md. You should use something like <https://github.com/liamcain/obsidian-daily-notes-interface> to get the path to the current daily note.
- [getLink(file: TFile) {](https://github.com/LA/obsidian-daily-note-collector/blob/54243029e4f6baacd51dbf7f19f532c29f729803/main.ts#L80) instead of creating this yourself, you should use generateMarkdownLink (<https://github.com/obsidianmd/obsidian-api/blob/df9434bec6883bd32727f61b1ed1336f7663d0e4/obsidian.d.ts#L1435> which will follow the user's link format settings.

LA:
Thank you @liamcain for your detailed feedback!

I have released a new version that I believe satisfies all of your concerns: https://github.com/LA/obsidian-daily-note-collector/releases/tag/1.0.3

PR #3997, opened 2024-08-05
Add plugin: Vitepress Publisher

tyrad:

## Repo URL

Link to my plugin:

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13053,6 +13053,13 @@
         "author": "James-Yu",
         "description": "Display file cards for external files.",
         "repo": "James-Yu/external-file-card"
+    },
+    {
+        "id": "vitepress-publisher",
+        "name": "Vitepress Publisher",
+        "author": "mistj",
+        "description": "Conveniently preview and publish Markdown files using vitepress.",
+        "repo": "tyrad/obsidian-vitepress"
     }
 ]
 


liamcain:
- [this.settingTab = new SettingTab(this.app, this)](https://github.com/tyrad/obsidian-vitepress/blob/35ed82fb88f4a43ea4550655aa2be39e4717e481/src/main.ts#L19) You shouldn't need a reference to this. Just create the setting tab inside onload
- [button.detach();](https://github.com/tyrad/obsidian-vitepress/blob/35ed82fb88f4a43ea4550655aa2be39e4717e481/src/main.ts#L98) why are you detaching the existing button? If the button already exists, you should just skip. Otherwise you'll be mutating the dom more than you need to.
- [setTimeout](https://github.com/tyrad/obsidian-vitepress/blob/35ed82fb88f4a43ea4550655aa2be39e4717e481/src/main.ts#L111) what's this setTimout for?
- [activeLeaf](https://github.com/tyrad/obsidian-vitepress/blob/35ed82fb88f4a43ea4550655aa2be39e4717e481/src/main.ts#L90) maybe check the view type of the leaf too
- [const activeLeaf](https://github.com/tyrad/obsidian-vitepress/blob/35ed82fb88f4a43ea4550655aa2be39e4717e481/src/main.ts#L90) this name is not accurate. it's not the active leaf.

PR #3998, opened 2024-08-05
Add plugin: Lazy Plugin Loader

alangrainger:

## Repo URL

Link to my plugin: https://github.com/alangrainger/obsidian-lazy-plugins

This plugin is similar to `obsidian-plugin-groups` in the store, however [that plugin is no longer under development](https://github.com/Mocca101/obsidian-plugin-groups/issues/41#issuecomment-1907752889), as confirmed by the plugin owner.

This plugin is also significantly more simple, and should be compatible with Obsidian long into the future.

It has [some good feedback on Reddit](https://www.reddit.com/r/ObsidianMD/comments/1ekea52/install_this_plugin_loader_to_make_your_obsidian/) already.

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13088,6 +13088,13 @@
 	  "author": "Henri Jamet",
 	  "description": "Explore your vault by iteratively adding or ignoring linked notes, ultimately generating a customizable canvas that visually represents the preserved notes and their connections.",
 	  "repo": "hjamet/Canvas-Explorer"
-	}
+    },
+	{
+        "id": "lazy-plugins",
+        "name": "Lazy Plugin Loader",
+        "author": "Alan Grainger",
+        "description": "Load plugins with a delay on startup, so that you can get your app startup down into the sub-second loading time.",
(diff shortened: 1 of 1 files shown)


liamcain:
I think it might be more reliable if you just disable all plugins (except this one) and manage the load state yourself. So call enablePlugin/disablePlugin instead of enableAndSave/disableAndSave. That way you can ensure that this plugin is loaded before all other plugins and you aren't fighting with the config. What do you think?

alangrainger:
That's a great idea. I had wondered if it could cause issues if there are some plugins which expect to be loaded the normal "Obsidian" way - but if the suggestion is coming from you then absolutely I will implement it that way 💯

alangrainger:
Done - this change is live in 1.0.11. Thanks for the suggestion :+1: 

https://github.com/alangrainger/obsidian-lazy-plugins/releases/tag/1.0.11

PR #4000, opened 2024-08-06
Update plugin author name: formatto-format

pixelbracket:


joethei:
Did you change your GitHub account, or did you take over maintenance from cosmostellar?
Either way, I need some clarification from the cosmostellar account.

PR #4023, opened 2024-08-10
Add plugin: Snippets Manager

ramandv:

## Repo URL

Link to my plugin:

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13109,6 +13109,13 @@
         "author": "ytisf",
         "description": "Enrich your notes with information from VirusTotal.",
         "repo": "ytisf/virustotal-enrich"
-     }
+    },
+    {
+        "id": "snippets-manager",
+        "name": "Snippets Manager",
+        "author": "Venkatraman Dhamodaran",
+        "description": "Snippets manager with full mobile support. You can manage code snippets, infos like passport number, email signature or anything.",
+        "repo": "ramandv/obsidian-snippets-manager"
+    }
 ]
 


liamcain:
- [read](https://github.com/ramandv/obsidian-snippets-manager/blob/e2ecb472798a63e7d9a2c52c056641026da92426/src/SnippetManagerPlugin.ts#L45) I think you're fine to use [cachedRead](https://github.com/obsidianmd/obsidian-api/blob/9bb30633586c6649474b2affdf5b87efdd86498f/obsidian.d.ts#L3373) instead. That way you don't need to read from the filesystem
- [md = content.split('\n');](https://github.com/ramandv/obsidian-snippets-manager/blob/e2ecb472798a63e7d9a2c52c056641026da92426/src/SnippetManagerPlugin.ts#L57) You can also leverage the metadata cache instead of parsing the markdown file yourself. That will be more performant and easier to deal with. [`getFileCache`](https://github.com/obsidianmd/obsidian-api/blob/bceb489fc25ceba5973119d6e57759d64850f90d/obsidian.d.ts#L2378). This will give you the start and end offset of each section. Then you can just get fileContents.slice(start.offset, end.offset)
- [addEventListener](https://github.com/ramandv/obsidian-snippets-manager/blob/e2ecb472798a63e7d9a2c52c056641026da92426/src/SnippetSuggestModal.ts#L12) If you want to add custom behavior for meta/ctrl keys, you can use Scope. 

```ts
this.scope.register(null, 'Enter', (evt: KeyboardEvent) => {
			if (evt.isComposing) {
				return;
			}
			this.chooser.useSelectedItem(evt);
			return false;
		});
```

You might need to include ts-ignore, since I think scope is still a private API but its safe to use.

- [callback](https://github.com/ramandv/obsidian-snippets-manager/blob/e2ecb472798a63e7d9a2c52c056641026da92426/src/SnippetManagerPlugin.ts#L29) this should use [editorCallback](https://github.com/obsidianmd/obsidian-api/blob/38dd22168d2925086371bfc59e36fd9121527a39/obsidian.d.ts#L614) instead. Then you should pass a reference to the editor into your modal. This will ensure that you are accessing the same editor that was focused before triggering the command.
- [const view = this.app.workspace.getActiveViewOfType(MarkdownView);](https://github.com/ramandv/obsidian-snippets-manager/blob/e2ecb472798a63e7d9a2c52c056641026da92426/src/SnippetSuggestModal.ts#L54) this won't work for Canvas files but if you have a reference to the editor passed in, you would use that here instead.

ramandv:
@liamcain All the requested changes have been done.

PR #4034, opened 2024-08-12
Add plugin: Battery Indicator

Opisek:

## Repo URL

Link to my plugin: https://github.com/Opisek/obsidian-battery-indicator

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13095,6 +13095,13 @@
         "author": "Alan Grainger",
         "description": "Load plugins with a delay on startup, so that you can get your app startup down into the sub-second loading time.",
         "repo": "alangrainger/obsidian-lazy-plugins"
-    }	
+    },
+	{
+        "id": "battery-indicator",
+        "name": "Battery Indicator",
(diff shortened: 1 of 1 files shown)


liamcain:
- [(async () => {](https://github.com/Opisek/obsidian-battery-indicator/blob/e0ff4034e4acea1ceb63b13c664466934b5e37e2/src/polling.ts#L15) why not just call this.poll()?
- [console.log(callback);](https://github.com/Opisek/obsidian-battery-indicator/blob/e0ff4034e4acea1ceb63b13c664466934b5e37e2/src/polling.ts#L48) remove this

Opisek:
Thanks for merging @liamcain 
Changed both points. Calling `this.poll()` was not possible with some previous logic that I since removed, so it was an artifact from old code. I'll keep it commented out in case I want to revisit the idea with dynamic polling intervals.

PR #4035, opened 2024-08-12
Add plugin: OnlyWorlds Builder

tmoosting:

## Repo URL
 
Link to my plugin: https://github.com/OnlyWorlds/obsidian-plugin

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
- [ x] My GitHub release contains all required files (as individual files, not just in the source.zip / source.tar.gz)
  - [x] `main.js`
  - [x] `manifest.json`


tmoosting:
Handled all required and optional issues

liamcain:
- [Handlebars.registerHelper('linkify',](https://github.com/OnlyWorlds/obsidian-plugin/blob/a27db9cb680120d7abed07ecde8e7a8879edce35/main.ts#L44) looks like this is duplicated belo

tmoosting:
Hi Liam,
Thank you for your attention & time. This is not an environment I am very familiar with so I appreciate your pointers.
I made changes accordingly, except for two cases:
1. I did not manage to use the Vault API for the second of those suggestions (5th bullet point, "same here" , for ImportWorldCommand). In this case, a file is accessed from the plugin files, not from the vault itself, so as I eventually assumed, I should not use the vault API for this. If I nonetheless should, please advise the relative path I should use with getAbstractFileByPath. 
2. Regarding the section pattern with getFileCache (8th bullet point): this piece of codes is meant to parse a specific JSON structure I have, from data that is directly inputted by the user, so I do not think I can use the cache for this. Again, please advise if I misunderstand.

liamcain:
Sorry for the long delay getting back to this. Your changes look good to me!

liamcain:
Plugin added here: https://github.com/obsidianmd/obsidian-releases/commit/727da498181fc9b92863b17430979316b79d0192

PR #4041, opened 2024-08-13
Add plugin: Verovio Music Renderer

kpaede:

## Repo URL

https://github.com/kpaede/Verovio-Music-Renderer/

Link to my plugin:
https://github.com/kpaede/Verovio-Music-Renderer/releases/tag/1.0.1

## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [X]  iOS _(if applicable)_ -> Should work but not 100% sure
  - [X] `main.js`
  - [X] `manifest.json`
      
      Info: The plugin accesses a soundfont file on Github for playback (just like similar midi playback plugins)

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13265,7 +13265,7 @@
         "repo": "Talal-A/obsidian-immich"
     },
     {
- 	 "id": "Verovio-Music-Renderer",
+ 	 "id": "verovio-music-renderer",
  	 "name": "Verovio Music Renderer",
  	 "author": "Kornelius Paede",
 	  "description": "Renders Musical Notation from various formats and plays it back.",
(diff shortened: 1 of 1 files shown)


joethei:
Your `styles.css` file is not in the repo.

[const response = await fetch(path);](https://github.com/kpaede/Verovio-Music-Renderer/blob/8facf9d02aedba57a8b457678d861fbca57b0aca/src/verovioProcessor.ts#L156)

[function createButton(iconSvg: string, onClick: () => void): HTMLButtonElement {](https://github.com/kpaede/Verovio-Music-Renderer/blob/8facf9d02aedba57a8b457678d861fbca57b0aca/src/verovioProcessor.ts#L300-L305)
This should be replaced by the `setIcon` functions, see <https://lucide.dev/icons/> for the available icon's

PR #4122, opened 2024-08-27
Add plugin: Simple Sync

xing393939:
Simple sync notes between local and cloud with Amazon S3, Dropbox, webdav.

## Repo URL
Link to my plugin: https://github.com/xing393939/simple-sync

## Release Checklist
  * [x]   Windows
  * [ ]   macOS
  * [ ]   Linux
  * [x]   Android _(if applicable)_
  * [ ]   iOS _(if applicable)_
* [x]  My GitHub release contains all required files (as individual files, not just in the source.zip / source.tar.gz)
  * [x]  `main.js`
  * [x]  `manifest.json`
  * [x]  `styles.css` _(optional)_
* [x]  I have read the developer policies at https://docs.obsidian.md/Developer+policies, and have assessed my plugins's adherence to these policies.
* [x]  I have read the tips in https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines and have self-reviewed my plugin to avoid these common pitfalls.
* [x]  I have given proper attribution to these other projects in my `README.md`.


xing393939 (commented):
Add a release

joethei:
What sets your plugin apart from the "Remotely save" and "Remotely sync" plugins?
Your changes should be contributed to one of these, instead of creating another fork.

xing393939:
The settings of those two plugins is too complex for most people. I want options for minimizing settings.

joethei:
I don't see any simplifications in the settings so far.
At the moment, it would be much better as a PR to the original plugin.
We can only accept forks of existing plugins if there is a significant difference visible.

PR #4125, opened 2024-08-27
Update Chat with bard plugin

Artel250:
The plugin was originally developed to enable users to use Google Bard in Obsidian. Google Bard has now been replaced with gemini and a completely different api, so the plugin has also been completely reworked to adapt to those changes.

This includes also a name change: from "Talk with Bard" to "Gemini AI Assistant"

Would it be possible to make the changes so that the plugin is updated in the obsidan community plugins, or should I create and submit a new plugin alltogether?

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -8512,11 +8512,11 @@
         "repo": "Tatz884/RescueTime-Obsidian"
     },
     {
-        "id": "chat-with-bard",
-        "name": "Chat with Bard",
+        "id": "gemini-ai-assistant",
+        "name": "Gemini AI Assistant",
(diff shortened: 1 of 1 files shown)


joethei:
Don't update the plugin id, any existing users won't be able to update the plugin.

Artel250:
Hi, I have reverted the id to the original one. I have also kept the original id in the manifest.json of the plugin.

joethei:
Don't include "Obsidian" in the description (also you made a mistake in the capitalisation here)

Artel250:
Hello, I have updated the description.

PR #4174, opened 2024-09-06
Add plugin: Pronounce Selected Text

GoBeromsu:

## Repo URL

Link to my plugin: https://github.com/GoBeromsu/Pronounce-Selected-Text

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13158,6 +13158,13 @@
         "author": "Ram Rachum",
         "description": "Easily select text in backticks.",
         "repo": "cool-RR/backtick-text-selector"
+    },
+    {
+        "id": "pronounce-selected-text",
+        "name": "Pronounce Selected Text",
+        "author": "Beomsu Koh",
+        "description": "A plugin to pronounce selected text",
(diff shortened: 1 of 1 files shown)


GoBeromsu:
I added the default hotkeys :) 
[Releases · GoBeromsu/Pronounce-Selected-Text](https://github.com/GoBeromsu/Pronounce-Selected-Text/releases)

joethei:
Your plugin currently only supports British English, while the README states that it supports many languages, please update the README.

Also, what sets your plugin appart from the existing Text to Speech plugin, other than the above?

[hotkeys: [{ modifiers: ["Mod", "Alt"], key: "p" }],](https://github.com/GoBeromsu/Pronounce-Selected-Text/blob/bbc0e2d7f57dc7b8a2914393850e0d2775c8568e/main.ts#L9)

GoBeromsu:
Thank you for your response. I agree with you, and I will come back after making some changes

PR #4293, opened 2024-09-30
Add plugin: Command Palette Enhancer

lakca:

## Repo URL

Link to my plugin: https://github.com/lakca/obsidian-command-palette-enhancer

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13466,5 +13466,12 @@
       "author": "ChenPengyuan",
       "description": "Show mermaid diagrams in a draggable and zoomable popup",
       "repo": "gitcpy/obsidian-mermaid-pop"
+  },
+  {
+      "id": "command-palette-enhancer",
+      "name": "Command Palette Enhancer",
+      "author": "lakca",
+      "description": "Enhance command palette. Pin recent commands, fuzzy search supports command id...",
+      "repo": "lakca/obsidian-command-palette-enhancer"
   }
 ]


lakca:
/skip, this value is from user settings.

joethei:
[# Obsidian Command Palette Enhancer](https://github.com/lakca/obsidian-command-palette-enhancer/blob/0f62e46b73c63ce4af3963fc18152231a87c7847/README.md#L1)

[historyIcon: '<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="lucide lucide-history"><path d="M3 12a9 9 0 1 0 9-9 9.75 9.75 0 0 0-6.74 2.74L3 8"/><path d="M3 3v5h5"/><path d="M12 7v5l4 2"/></svg>',](https://github.com/lakca/obsidian-command-palette-enhancer/blob/0f62e46b73c63ce4af3963fc18152231a87c7847/src/settings.ts#L29)
This should just be `history` by default, and you can add suggestion of all the icons provided in Obsidian by implementing a `AbstractInputSuggest` combined with the `getIconIds()` function.

[renderIcon() {](https://github.com/lakca/obsidian-command-palette-enhancer/blob/0f62e46b73c63ce4af3963fc18152231a87c7847/src/providers/history.ts#L130-L142)
And here you should use the `setIcon` function from the Obsidian API instead.

[export function addClassTo(el: HTMLElement, classes: string|string[]) {](https://github.com/lakca/obsidian-command-palette-enhancer/blob/0f62e46b73c63ce4af3963fc18152231a87c7847/src/utils.ts#L34-L38)
Use `el.addClass()` instead.

PR #4297, opened 2024-09-30
Add plugin: Metadata Auto Classifier

GoBeromsu:

## Repo URL

Link to my plugin: https://github.com/GoBeromsu/Metadata-Auto-Classifier

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13690,5 +13690,12 @@
       "author": "Patrick Chiang",
       "description": "Adds replace all functionality into search.",
       "repo": "patrickchiang/obsidian-replace-all"
+  },
+  {
+	"id": "metadata-auto-classifier",
+	"name": "Metadata Auto Classifier",
+	"author": "Beomsu Koh",
+	"description": "Automatically classifies and applies metadata to your notes.",
+    "repo": "GoBeromsu/Metadata-Auto-Classifier"
   }
 ]


joethei:
[hotkeys: [{ modifiers: ['Mod', 'Shift'], key: 'F' }],](https://github.com/GoBeromsu/Metadata-Auto-Classifier/blob/e1b33eac11fd2354e0d3c6522fdaab1ee33fe877/src/main.ts#L42)
In this specific case, this hotkey is colliding with the default hotkey to open the search tab.

[// Frontmatter tags](https://github.com/GoBeromsu/Metadata-Auto-Classifier/blob/e1b33eac11fd2354e0d3c6522fdaab1ee33fe877/src/frontmatterHandler.ts#L58-L71)
use the [getAllTags](https://docs.obsidian.md/Reference/TypeScript+API/getAllTags) function from the API.

[new Setting(containerEl).setName('Frontmatter settings').setHeading();](https://github.com/GoBeromsu/Metadata-Auto-Classifier/blob/e1b33eac11fd2354e0d3c6522fdaab1ee33fe877/src/settings/index.ts#L69)

PR #4335, opened 2024-10-05
Add plugin: random-retrieval

JeanJean-rxl:

## Repo URL

Link to my plugin:

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13728,8 +13728,8 @@
     "repo": "NilsLeo/obsidian-youtube-iframe-timestamps"
   },
   {
-    "id": "random-retrieval-plugin",
-    "name": "random-retrieval-plugin",
+    "id": "random-retrieval",
+    "name": "random-retrieval",
     "author": "Rachninomav",
     "description": "Random Note Retrieval based on LLM.",
     "repo": "JeanJean-rxl/random-retrieval-plugin"
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13731,7 +13731,7 @@
     "id": "random-retrieval-plugin",
     "name": "random-retrieval-plugin",
     "author": "Rachninomav",
-    "description": "Random Note Retrieval based on LLMs.",
+    "description": "Random Note Retrieval based on LLM.",
     "repo": "JeanJean-rxl/random-retrieval-plugin"
   }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13726,5 +13726,12 @@
     "description": "Allows you to embed YouTube videos with timestamps directly in your notes, enabling seamless referencing and note-taking without needing to open a separate browser window.",
     "author": "Nils Leo",
     "repo": "NilsLeo/obsidian-youtube-iframe-timestamps"
+  },
+  {
+    "id": "random-retrieval-plugin",
+    "name": "random-retrieval-plugin",
+    "author": "Rachninomav",
+    "description": "Random Note Retrieval based on LLMs.",
+    "repo": "JeanJean-rxl/random-retrieval-plugin"
   }
 ]


JeanJean-rxl:
Thank you for your check,

I have released a v1.0.4 that fixed the required issues above

joethei:
[import * as fs from 'fs';](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/main.ts#L18), [import { exec, ChildProcess } from 'child_process';](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/main.ts#L19)

[const absPath = app.vault.adapter.basePath;](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/main.ts#L24)
This will also fail on mobile, and don't use the global app instead, but the one provided by your plugin instance.

[PATH_TO_JSON: `${absPath}/.obsidian/plugins/random-retrieval-plugin/data.json`,](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/main.ts#L36-L37)
Don't hardcode the path to the plugin folder, users can change it.
Use `this.manifest.dir` instead.

[await fs.promises.writeFile(this.settings.PATH_TO_JSON, jsonString, 'utf8');](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/main.ts#L70)
This call serves no purpose, the `saveData` call already saves to that file.

[setEnableRibbonIcon = (value: boolean): void => {](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/main.ts#L103-L123), [.setName('Open in New Leaf')](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/set_tab.ts#L85)

[const axios = require('axios');](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/main.ts#L128)
Stick to one import style

[const response = await axios.get(`http://127.0.0.1:8000/search?query=${encodeURIComponent(query)}`);](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/main.ts#L132)

[const command = `osascript -e 'tell application "Terminal"](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/main.ts#L174-L178)
This will only work on Macs, which is not mentioned in the README, and there is also no validation that makes sure that the code is only being executed on a Mac.

[containerEl.createEl('h2', { text: 'Random Retrieval Settings' });](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/set_tab.ts#L17)

[.setName("LLM Retriever Language")](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/set_tab.ts#L20), [.setName("Open How Many Notes")](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/set_tab.ts#L33), [.setName("Conda Environment Name")](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/set_tab.ts#L46), [.setName('Open in New Leaf')](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/set_tab.ts#L85)

[new Setting(containerEl)](https://github.com/JeanJean-rxl/random-retrieval-plugin/blob/712176dd1c02ce68350b63910f9b224cdb065eea/src/set_tab.ts#L58-L82)
This should not be a setting, see above.

JeanJean-rxl:
Hi Johannes,

Thank you so much for your detailed reply and apologise for some mistakes that could have been avoided.

I have released a v1.0.7 according to your suggestions, please double check the logics below:
- this code is designed to be executed only on Mac so far
- data.json can still be correctly generated after the change on abs/relative path

As for other minor changes, I have checked that
- the axios.get() has been replaced with requestUrl
- ribbon toggle has been avoided
- top-level heading has been removed, sentence case has been made in UI
- redundant saving ops has been removed.

Danke schön!

PR #4336, opened 2024-10-05
Add plugin: Folder Tabulation

SpeedaRJ:

## Repo URL

Link to my plugin: https://github.com/SpeedaRJ/obsidian_folder_tabulation/

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -14159,6 +14159,13 @@
 	  "author": "kuzzh",
 	  "description": "Polish, summarize, translate, analyze, and interpret code using AI",
 	  "repo": "kuzzh/obsidian-ai-bot"
-}
+},
+{
+    "id": "folder-tabulation",
+    "name": "Folder Tabulation",
+    "author": "SpeedaRJ",
+    "description": "Enables navigation between files in a folder via hotkeys and commands.",
+    "repo": "SpeedaRJ/obsidian_folder_tabulation"
+  }
 
 ]


SpeedaRJ:
> ### Required
> [[1](https://github.com/SpeedaRJ/obsidian_folder_tabulation/blob/b5e960280f04d418c0de8b52e9f027076c69c586/src/helper.ts#L166)][[2](https://github.com/SpeedaRJ/obsidian_folder_tabulation/blob/b5e960280f04d418c0de8b52e9f027076c69c586/src/helper.ts#L167)]:You should avoid assigning styles via JavaScript or in HTML and instead move all these styles into CSS so that they are more easily adaptable by themes and snippets. /skip I can't really find a good way to implement this differently, some of the plugins I took as reference also handle color assignment in the same way. I understand that the purpouse of this requirement is to enable better theming, but this particular piece of code provides exactly that: Configuration of colors and only colors that can be set to anything from any theme or differentely.

joethei:
A screenshot / GIF in the README would be nice.

[nextFileColor: "#b3b3b3",](https://github.com/SpeedaRJ/obsidian_folder_tabulation/blob/9a116ec184570c9266bce8a279cb2d2106e7f581/main.ts#L13-L14)
This would be better if you would use a accent color from: https://docs.obsidian.md/Reference/CSS+variables/Foundations/Colors
Then it's automatically looking good with any theme, and users don't need to change any settings by default.
If they want it to look different, they can write a CSS snippet.

SpeedaRJ:
Hello @joethei, 
I'm sorry this took a while longer than I expected due to personal issues, I have updated the code-base and documentation with the requested changes.
If there's anything else need of changing please let me know.

joethei:
[# Obsidian File Tabulation Plugin](https://github.com/SpeedaRJ/obsidian_folder_tabulation/blob/0956c056cb16d415a7814a27d7b66b0ed6735d10/README.md?plain=1#L1)

SpeedaRJ:
The title of the README file has been updated.

PR #4368, opened 2024-10-10
Update: Move obsidian-findoc repo to studiowebux github organization

studiowebux:
Hello, 

I created the findoc plugin a while ago, 
I am cleaning up my github organizations and moving everything to my main account "studiowebux".

So this PR is only renaming the repo url.

Can I move my repo before your approval ? 

Thanks you !

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -6773,7 +6773,7 @@
         "name": "Financial Doc",
         "author": "Studio Webux",
         "description": "Financial documentation and tracking using CSV format and Chart.js.",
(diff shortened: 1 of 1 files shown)


joethei:
You need to move the repo first, that way I can validate that it's an authorized transfer and nobody will have issues installing the plugin during the move, as there is a fallback.

studiowebux:
Hello @joethei Thanks for you answer,

I just moved the repository to my main account/org

https://github.com/studiowebux/obsidian-findoc

let me know if you need actions from my side.

PR #4405, opened 2024-10-13
Add plugin: Session Notes

tabibyte:

## Repo URL

Link to my plugin: https://github.com/tabibyte/session-notes

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13789,5 +13789,12 @@
     "author": "Heesu Suh",
     "description": "AI chat with note context, smart writing assistance, and one-click edits for your vault.",
     "repo": "glowingjade/obsidian-smart-composer"
+  },
+  {
+    "id": "session-notes",
+    "name": "Session Notes",
+    "author": "tabibyte",
+    "description": "Create temporary one-off session notes.",
+    "repo": "tabibyte/session-notes"
 }
 ]


joethei:
["description": "A plugin to create temporary & session notes that will be deleted when session ends",](https://github.com/tabibyte/session-notes/blob/73c96f44106cd4a72c33444a6b333518e2774cbc/manifest.json#L6)
Avoid including sentences like "This is a plugin that does ..." in your description, it should be self-evident that this is a plugin for Obsidian.

[hotkeys: [{ modifiers: ['Ctrl', 'Shift'], key: ';' }],](https://github.com/tabibyte/session-notes/blob/73c96f44106cd4a72c33444a6b333518e2774cbc/main.ts#L11), [hotkeys: [{ modifiers: ['Ctrl', 'Shift'], key: '\'' }],](https://github.com/tabibyte/session-notes/blob/73c96f44106cd4a72c33444a6b333518e2774cbc/main.ts#L18)

tabibyte:
thank you for your time and comments @joethei. I corrected everything you've pointed out.

PR #4487, opened 2024-10-21
Update plugin: Pexels Banner

jparkerweb:
rename / rebrand `Pexels-Banner` to `Pixel-Banner`

# I am renaming my existing Community Plugin

old repo: https://github.com/jparkerweb/pexels-banner
rename repo: https://github.com/jparkerweb/pixel-banner

---

As I now support more than just the "Pexels" API it is fitting to update the Plugin name to be more generic.

Thank you :)

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13314,11 +13314,11 @@
         "repo": "Epiphany-Notes/epiphany-obsidian-plugin"
     },
     {
-        "id": "pexels-banner",
-        "name": "Pexels Banner",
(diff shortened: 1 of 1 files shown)


jparkerweb:
Renaming from `Pexels Banner` to `Pixel Banner` as I now support more than just the Pexels API

- id and name in manifest updated to match rename github repo / project

joethei:
The id should not be changed, if it is changed existing users won't be able to update, and the download count will be reset.

PR #4526, opened 2024-10-26
Add plugin: Export to HTML

kalvn:

## Repo URL

Link to my plugin: https://github.com/kalvn/obsidian-export-to-html

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -14635,5 +14635,12 @@
     "author": "wenlzhang",
     "description": "Maintain note links when splitting or reorganizing notes.",
     "repo": "wenlzhang/obsidian-link-maintainer"
+  },
+  {
+    "id": "export-to-html",
+    "name": "Export to HTML",
+    "author": "kalvn",
+    "description": "Export your Markdown notes as HTML, directly in the clipboard or as a file.",
+    "repo": "kalvn/obsidian-export-to-html"
   }
 ]


kalvn:
Hello there, it looks like there's a problem in the automated validation. The repo info is wrong:

```
Run actions/github-script@v6
Repo info: mProjectsCode/obsidian-lemons-search-plugin
Found issue: The newly added entry is not at the end, or you are submitting on someone else's behalf. The last plugin in the list is: `mProjectsCode/obsidian-lemons-search-plugin`. If you are submitting from a GitHub org, you need to be a public member of the org.
Error: Failed to validate plugin
```

joethei:
[const markdownData = await parse(view.data, vault);](https://github.com/kalvn/obsidian-export-to-html/blob/4eb8dc4fa2a7fc6103c3519c54774553b6750f6d/src/main.ts#L49), [return await marked.parse(markdownData);](https://github.com/kalvn/obsidian-export-to-html/blob/4eb8dc4fa2a7fc6103c3519c54774553b6750f6d/src/parser.ts#L95)
Use the `MarkdownRenderer.render` method to get the actual HTML in Obsidian, and not the version marked is producing.
It will also contain HTML produced by plugins.

[currentVault = vault;](https://github.com/kalvn/obsidian-export-to-html/blob/4eb8dc4fa2a7fc6103c3519c54774553b6750f6d/src/parser.ts#L90-L91)
This code serves no purpose, remove it.

kalvn:
Thank you for your feedback and analysis.

I removed `currentVault` in favor of `this.app.vault`.

But about Markdown parsing:
1. I don't get how to use `MarkdownRenderer.render` to get HTML out of Markdown in the context of the editorCallback.
2. It does not allow me to hook into the rendering process to transform images in base 64 for example, so it is not an equivalent.

Can you please advise?

joethei:
You would create a div, render onto that div, get the HTML content of said div, then remove the div.
Afterwards you will need to do some post processing, but you are already doing that with marked.

kalvn:
@joethei I just deployed v1.1.0 that drops the marked library and uses the Obsidian markdown renderer.

https://github.com/kalvn/obsidian-export-to-html/releases/tag/1.1.0

kalvn:
Not sure how to get back the **Ready for review** tag 👀

joethei:
[# Obsidian Export to HTML](https://github.com/kalvn/obsidian-export-to-html/blob/8cecbfe7a547fdc8e4a04553018e7076f4c6d0d2/README.md?plain=1#L1C1-L1C26)

kalvn:
@joethei Done, I just updated the header and the introduction in the README file, to make it clearer it is a community plugin.

PR #4531, opened 2024-10-27
Add plugin: Masking Type

Telehakke:

## Repo URL

Link to my plugin: https://github.com/Telehakke/masking-type

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -13964,6 +13964,13 @@
 	  "description": "Sorting your TODOs (checkboxes) in current note by completion status",
 	  "author": "Jiri Sifalda",
 	  "repo": "jsifalda/obsidian-todos-sort"
+	},
+    {
+        "id": "masking-type",
+        "name": "Masking Type",
+        "author": "Telehakke",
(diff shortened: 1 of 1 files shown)


joethei:
Why are you using comments to configure the masking options per note and not frontmatter values?
That would be easier to process.

[new Setting(containerEl)](https://github.com/Telehakke/masking-type/blob/a5ae2d4ae6e4b5f5668540f5694f0e38def42f68/src/views/maskingTypeSettingTab.ts#L32-L46)
Get the locale via the code below, and remove this setting.
```ts
import {moment} from "obsidian";
moment.locale();
```

Telehakke:
Thank you for your review, joethei.
The reason why I used the comments was because I didn't notice the existence of the frontmatter.
I was able to understand that the code is easier and user-friendly, so I changed it to use the frontmatter.

PR #4625, opened 2024-11-07
Add plugin: Note Annotations

jancbeck:

## Repo URL

Link to my plugin: https://github.com/jancbeck/obsidian-note-annotations/releases/tag/1.0.0

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -14908,5 +14908,12 @@
     "author": "Amin Sennour",
     "description": "Change the default filename used for new notes",
     "repo": "TheLoneWanderer4/obsidian-uuid-title"
+  },
+  {
+    "id": "omnidian",
+    "name": "Omnidian Active Reader",
+    "author": "Jan Beck",
+    "description": "Add multi-colored highlights and comments to your notes.",
+    "repo": "jancbeck/omnidian"
   }
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -14911,7 +14911,7 @@
   },
   {
     "id": "omnidian",
-    "name": "Omnidian Active Reader",
+    "name": "Omnidian",
     "author": "Jan Beck",
     "description": "Add multi-colored highlights and comments to your notes.",
     "repo": "jancbeck/omnidian"
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -14910,10 +14910,10 @@
     "repo": "TheLoneWanderer4/obsidian-uuid-title"
   },
   {
-    "id": "omnidian",
-    "name": "Omnidian",
+    "id": "note-annotations",
+    "name": "Note Annotations",
     "author": "Jan Beck",
     "description": "Add multi-colored highlights and comments to your notes.",
-    "repo": "jancbeck/omnidian"
+    "repo": "jancbeck/obsidian-note-annotations"
   }
 ]


jancbeck:
/skip 

> [1]:Obsidian's configuration directory isn't necessarily .obsidian, it can be configured by the user. You can access the configured value from Vault#configDir

this isn't used in the plugin code in any way and just a way for me to more easily run the plugin locally against a sandbox vault

> [1][2][3][4][5]:You should avoid assigning styles via JavaScript or in HTML and instead move all these styles into CSS so that they are more easily adaptable by themes and snippets

These are programmatic styles like positioning and element relative to another element. Although there is the new CSS anchor positioning API, it requires a polyfill at this time. 
The highlighted background colors are to be set by users in the settings or as values directly and not meant for theming via CSS. if the user has not set a color, a themeable default css variable is set in the Js

jancbeck:
@joethei i think the action was supposed to assign you but didn't work 
https://github.com/obsidianmd/obsidian-releases/actions/runs/11719552167/job/32643052368#step:3:1

joethei:
["id": "omnidian",](https://github.com/jancbeck/omnidian/blob/48b0f1517c601549ce3cf0801ef576938c026ecd/manifest.json#L2-L3)
Please change your plugin id and name to something that does not include parts of the word "Obsidian"

["description": "Better active reading experience in Obsidian",](https://github.com/jancbeck/omnidian/blob/48b0f1517c601549ce3cf0801ef576938c026ecd/manifest.json#L6)
Don't include "Obsidian" in the description.

[${](https://github.com/jancbeck/omnidian/blob/48b0f1517c601549ce3cf0801ef576938c026ecd/src/main.ts#L34-L36), [name: "Highlight Selection",](https://github.com/jancbeck/omnidian/blob/48b0f1517c601549ce3cf0801ef576938c026ecd/src/main.ts#L58), [`Highlighting Mode](https://github.com/jancbeck/omnidian/blob/48b0f1517c601549ce3cf0801ef576938c026ecd/src/main.ts#L143), [`Highlighting Mode ${](https://github.com/jancbeck/omnidian/blob/48b0f1517c601549ce3cf0801ef576938c026ecd/src/main.ts#L146-L147), [`Highlighting Mode: ${this.isHighlightingModeOn}`,](https://github.com/jancbeck/omnidian/blob/48b0f1517c601549ce3cf0801ef576938c026ecd/src/main.ts#L155)

[// This command will only show up in Command Palette when the check function returns true](https://github.com/jancbeck/omnidian/blob/48b0f1517c601549ce3cf0801ef576938c026ecd/src/main.ts#L66), [// This command will only show up in Command Palette when the check function returns true](https://github.com/jancbeck/omnidian/blob/48b0f1517c601549ce3cf0801ef576938c026ecd/src/main.ts#L78)
You don't return `false` anywhere in these two, so they should just be a `editorCallback`

[.is-live-preview::selection {](https://github.com/jancbeck/omnidian/blob/48b0f1517c601549ce3cf0801ef576938c026ecd/src/styles.css#L4-L6)

Have you tested that your plugin works with popup windows?

jancbeck:
thanks for your review @joethei 
I have made the requested changes

> Have you tested that your plugin works with popup windows?
if you mean the "open in new window" window then this doesn't work at the moment. If it's required for release, could you point me into a direction to how to make it work? At the moment it looks like my event handlers aren't triggered in the popup window.

jancbeck:
@joethei 

> Also seems like you missed one of the sentence case strings.

which one?

jancbeck:
@joethei 

> > Also seems like you missed one of the sentence case strings.
> 
> which one?

joethei:
hm, looks like you fixed it, but the bot did not detect that.

PR #4200, opened 2024-09-12
Add plugin: BetaX NAS Sync

skye-z:
# I am submitting a new Community Plugin

## Repo URL

Link to my plugin: https://github.com/skye-z/ons

## Release Checklist
- [x] I have tested the plugin on
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  Android _(if applicable)_
  - [ ]  iOS _(if applicable)_
- [x] My GitHub release contains all required files (as individual files, not just in the source.zip / source.tar.gz)
  - [x] `main.js`
  - [x] `manifest.json`
  - [ ] `styles.css` _(optional)_
- [x] GitHub release name matches the exact version number specified in my manifest.json (_**Note:** Use the exact version number, don't include a prefix `v`_)
- [x] The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.
- [x] My README.md describes the plugin's purpose and provides clear usage instructions.
- [x] I have read the developer policies at https://docs.obsidian.md/Developer+policies, and have assessed my plugins's adherence to these policies.
- [x] I have read the tips in https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines and have self-reviewed my plugin to avoid these common pitfalls.
- [x] I have added a license in the LICENSE file.
- [x] My project respects and is compatible with the original license of any code from other plugins that I'm using.
      I have given proper attribution to these other projects in my `README.md`.

PR #4639, opened 2024-11-09
Add plugin: BetaX NAS Sync

skye-z:

## Repo URL

Link to my plugin: https://github.com/skye-z/ons

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -14705,6 +14705,13 @@
     "author": "kbwo",
     "description": "Provide keyboard-driven hints for clickable elements.",
     "repo": "kbwo/obsidian-click-hint"
+  },
+  {
+    "id": "ons",
+    "name": "BetaX NAS Sync",
+    "author": "Skye",
+    "description": "Keep Vault synchronized with the NAS on your home intranet.",
+    "repo": "skye-z/ons"
   }
 ]
 


skye-z:
Due to prolonged lack of review, the original Pull #4200  was closed by the robot

The relevant modification records can be found in the old pull records

joethei:
Sorry for the delay in reviewing.

> Besides that, I would like to ask how to get the current language of Obsidian?
I'd like the plugin to be able to implement i18n, but I can't find an interface for it in the docs or the community.

The most common approach is this: <https://github.com/joethei/obsidian-rss/tree/master/src/l10n>
Once Obsidian 1.8 is publicly released, there will be a `getLanguage` function that will replace the `moment.locale()` call, the docs will include an example for this then.

[vault.on('create', (file) => {](https://github.com/skye-z/ons/blob/3e48e4f8ae53e3a44a455efe73499fbb0334cd2f/obsidian-plugin/main.ts#L69-L84)

[stat.vault.cachedRead(stat);](https://github.com/skye-z/ons/blob/3e48e4f8ae53e3a44a455efe73499fbb0334cd2f/obsidian-plugin/main.ts#L89), [file.vault.cachedRead(file)](https://github.com/skye-z/ons/blob/3e48e4f8ae53e3a44a455efe73499fbb0334cd2f/obsidian-plugin/src/peer-manager.ts#L498)
For syncing you will want to use the `read` function, not the cached variant.

[iceServers: [](https://github.com/skye-z/ons/blob/3e48e4f8ae53e3a44a455efe73499fbb0334cd2f/obsidian-plugin/src/peer-manager.ts#L50-L53)
These two should also use the value from the settings.

[this.channel.onclose = () => console.log('数据通道已关闭');](https://github.com/skye-z/ons/blob/3e48e4f8ae53e3a44a455efe73499fbb0334cd2f/obsidian-plugin/src/peer-manager.ts#L65)

skye-z:
Oh understandable
There will always be some tedious things to deal with at the end of the year 😀

1. First of all, thank you for providing the i18n example. I have already used it and the effect is very good 👍
2. Regarding the "event", I used this.registerEvent() as described in the document
3. Vault.catchRead() has all been replaced with vault.read()
I used to worry about performance issues when using vault.read() before, but after using it for a long time, I didn't find any performance bottlenecks, so I don't need catchRead()  now
4. The configuration items for iceServers have been actually loaded
5. I have annotated all console.log() files, and now only retain the output console.error() for error messages

I have been using the 'this. app. emulate Mobile (true)' during this period to test the plugin's performance on mobile devices. Currently, the latest version of the plugin can run normally on mobile devices. Do I need to wait for the completion of this pull to create an update pull or modify it directly?

PR #4655, opened 2024-11-11
Update community-plugins.json

kryptonianson:
added obsidian-markdown-tags

## Repo URL

Link to my plugin: https://github.com/binarynoir/obsidian-markdown-tags

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -1,5 +1,12 @@
 [
     {
+        "id": "obsidian-markdown-tags",
+        "name": "Markdown Tags",
+        "author": "John Smith III",
+        "description": "Enhance your documents with custom tags. Use predefined or custom labels, customizable colors, and arrow indicators to visually track tasks and statuses.",
+        "repo": "binarynoir/obsidian-markdown-tags"
+    },
+    {
         "id": "nldates-obsidian",
         "name": "Natural Language Dates",
         "author": "Argentina Ortega Sainz",


kryptonianson:
made required adjustments

joethei:
[.tags {](https://github.com/binarynoir/obsidian-markdown-tags/blob/c7f754b1361f887b7288948c672a56c4c22d7170/styles.css#L1)
This is a bit generic and could conflict with other plugins.

[const tagSyntaxRegex = /\[\[<?tag\|(?<label>[^\]|]+)(?:\|(?<bgcolor>[^\]|]*))?(?:\|(?<fgcolor>[^\]|]*))?\]\]/g;](https://github.com/binarynoir/obsidian-markdown-tags/blob/c7f754b1361f887b7288948c672a56c4c22d7170/main.ts#L14)
You should use a different syntax, the double brackets are already used in the core app, so this could lead to some conflicts.
Especially in this example from your README: `[[tag|label]]`, which is also a valid Wikilink.

kryptonianson:
> [.tags {](https://github.com/binarynoir/obsidian-markdown-tags/blob/c7f754b1361f887b7288948c672a56c4c22d7170/styles.css#L1) This is a bit generic and could conflict with other plugins.
> 
> [const tagSyntaxRegex = /[[<?tag|(?[^\]|]+)(?:|(?[^\]|]_))?(?:|(?[^\]|]_))?]]/g;](https://github.com/binarynoir/obsidian-markdown-tags/blob/c7f754b1361f887b7288948c672a56c4c22d7170/main.ts#L14) You should use a different syntax, the double brackets are already used in the core app, so this could lead to some conflicts. Especially in this example from your README: `[[tag|label]]`, which is also a valid Wikilink.

These were really good observations. I have changed the css styles to .bn-tags and .bn-arrow-tags. In addition I have changed the tag syntax to ((tag|label)) using double '((' instead which is actually more representative of the final render and makes more sense. I hope this is satisfactory. And thank you!

PR #4691, opened 2024-11-15
Add plugin: Link Preview

felipetappata:

## Repo URL

Link to my plugin:  [https://github.com/felipetappata/obsidian-link-preview](https://github.com/felipetappata/obsidian-link-preview)

## Release Checklist
  - [x] Windows
  - [x] macOS
  - [x] Linux
  - [x] Android _(if applicable)_
  - [x] iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -14208,5 +14208,12 @@
     "author": "dangehub",
     "description": "Blur anything you want to keep your privacy.",
     "repo": "dangehub/obsidian-blur-mode"
-  }
+  },
+  {
+    "id": "link-preview",
+    "name": "Link Preview",
+    "author": "Felipe Tappata",
+    "description": "Show a preview of external links on hover",
+    "repo": "felipetappata/obsidian-link-preview"
+}
 ]


joethei:
Some screenshots/GIFs would be nice in the README.

Is there any particular reason your plugin is desktop only?
[let hideTimeout: NodeJS.Timeout;](https://github.com/felipetappata/obsidian-link-preview/blob/e49b850318ae785d21b2470cb3ca4d5925a63aee/main.ts#L71)
This is the only thing that I can see that could break it on mobile, you should change this type to `number` and use the `activeWindow.setTimeout` / `activeWindow.clearTimeout` functions.

[this.registerDomEvent(document, 'mouseover', (evt: MouseEvent) => {](https://github.com/felipetappata/obsidian-link-preview/blob/e49b850318ae785d21b2470cb3ca4d5925a63aee/main.ts#L33)
You should also hook into the `window-open` event to add this listener there, then this plugin should also work with pop out windows.

[new Setting(containerEl)](https://github.com/felipetappata/obsidian-link-preview/blob/e49b850318ae785d21b2470cb3ca4d5925a63aee/main.ts#L181-L189)
This settings feels a bit pointless, as a user I would just disable the plugin itself.

felipetappata:
Thank you for the suggestions. I've implemented the changes in Release 0.1.2.

PR #4698, opened 2024-11-16
removing duplicate entry

ganesshkumar:
Fixing duplicate entry.

As you can see, the plugin `transfer-latex-from-gpt` has been listed twice in this file. 
- line 13849 - from where I am deleting the proposal
- line 13562


joethei:
thanks for spotting this.

PR #4701, opened 2024-11-16
Add plugin: Qbstamp

Telehakke:

## Repo URL

Link to my plugin: https://github.com/Telehakke/qbstamp

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -14062,6 +14062,13 @@
     "author": "Omri Levi",
     "description": "Updates numbered lists automatically to keep them in sequential order. Features live updates, smart pasting, and manual control options.",
     "repo": "OmriLeviGit/Automatic-Renumbering-Obsidian"
+  },
+  {
+    "id": "obstamp",
+    "name": "Obstamp",
+    "author": "Telehakke",
+    "description": "Register and paste any text.",
+    "repo": "Telehakke/obstamp"
   }
 ]
 
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -14064,11 +14064,11 @@
     "repo": "OmriLeviGit/Automatic-Renumbering-Obsidian"
   },
   {
-    "id": "obstamp",
-    "name": "Obstamp",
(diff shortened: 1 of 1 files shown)


joethei:
["id": "obstamp",](https://github.com/Telehakke/obstamp/blob/591c6858dd13ce08cf10e7739ff92a529022d7af/manifest.json#L2-L3)
Please rename your plugin, the name should not include "Obsidian" or shorthand version of "Obsidian"

[private handleMouseMove = (event: MouseEvent): void => {](https://github.com/Telehakke/obstamp/blob/591c6858dd13ce08cf10e7739ff92a529022d7af/src/main.ts#L39-L41), [this.app.workspace.containerEl.removeEventListener(](https://github.com/Telehakke/obstamp/blob/591c6858dd13ce08cf10e7739ff92a529022d7af/src/main.ts#L32-L36), [this.app.workspace.containerEl.addEventListener(](https://github.com/Telehakke/obstamp/blob/591c6858dd13ce08cf10e7739ff92a529022d7af/src/main.ts#L24-L28), [protected showFavoritesMenu = (editor: Editor): void => {](https://github.com/Telehakke/obstamp/blob/591c6858dd13ce08cf10e7739ff92a529022d7af/src/models/pluginAction.ts#L32-L57)
This is a very bad way to do this, for the `editor-menu` event you already have the existing menu, just append to that.
For the command you should show a suggest modal instead of opening a context menu in a potentially unexpected place, as the mouse does not have to be anywhere near the selected text if one triggers the command.
https://docs.obsidian.md/Plugins/User+interface/Modals#Select+from+list+of+suggestions>

Telehakke:
I edited the PR to change the id and name of the plugin.
Then I get a “fatal: couldn't find remote ref” error in the middle of running actions/checkout@v4.
I do not know how to solve this problem.

PR #4758, opened 2024-11-26
Update cover image location in theme metadata

abhimangs:
This pull request updates the cover image path in the community-css-themes.json file to reflect the new location of the image.

Changes made:

Updated the cover image path from **Assets/Obsidian-Cover.png** to **assets/cover-obsidian.png**.
This ensures the correct display of the theme preview in the community themes list.

### DIFF
--- a/community-css-themes.json
+++ b/community-css-themes.json
@@ -2178,7 +2178,7 @@
     "name": "Luminex",
     "author": "abhimangs",
     "repo": "abhimangs/Obsidian-Luminex",
-    "screenshot": "Assets/Obsidian-Cover.png",
+    "screenshot": "assets/cover-obsidian.png",
     "modes": ["dark", "light"]
   },
   {


abhimangs:
Hi Maintainers,

I have a quick question regarding theme repository and theme name changes:

If I change the name of my theme or the repository name for my theme, will the download statistics on the Obsidian Community Themes page remain intact, or will they reset and start from zero?
Thank you for your guidance!

joethei:
The download stats will reset if you change the name of your theme.

abhimangs:
Hi Maintainers,

I have another quick question:

Can a single contributor release more than one theme in the Obsidian Community Themes repository? Are there any specific guidelines or restrictions I should be aware of if I plan to release a second theme?

Thank you for your assistance!

abhimangs:
Hi Maintainers,

I understand that the cover image for themes must meet the size requirement of 512x288 pixels to pass the tests. However, I’ve noticed that many larger themes use higher-resolution images for better quality.

My question is:
If I initially upload a 512x288 image to meet the requirement and have my pull request merged, can I later replace the image with a larger, higher-quality version? Would this affect the community theme display or cause any issues with repository validations?

Thank you for your guidance!

joethei:
The size requirement for the cover is there to optimize the loading speed in the community themes view (the one with all the covers shown)
Some older themes might not use that resolution, as we did not set that limit from the start.
Please don't change the cover image for something with higher resolution, all that will do is cause your cover to load slower.

You can put the image with the higher resolution into the README, as that is only shown when your theme is selected from the list.

abhimangs:
Hi @joethei,

Thank you for your detailed response and for clarifying the importance of optimizing loading speeds for the community themes view.

I was wondering if you could kindly suggest a higher resolution than 512x288 that would still align with the community guidelines. My goal is to provide better visual quality for users with high-resolution screens without significantly impacting performance. Would 1024x576 or any other resolution you recommend be acceptable?

Thank you so much for your time and guidance!

joethei:
No, that is too big, it has to be fast to load of a slow mobile connection

abhimangs:
It seems the community guidelines prioritize quick loading times, especially for users on slower connections, which limits the resolution of theme preview images. Here's a potential approach you can take in response to joethei:

Hi @joethei,

Thank you for your feedback and clarification.

I understand the need to prioritize fast loading times for all users, especially those on slower mobile connections. I’ll ensure the preview image adheres to the recommended resolution of 512x288.

If there's room for slight flexibility in future guidelines, it would be great to explore resolutions that balance both visual quality and performance for users with high-resolution displays.

Thanks again for your time and insights!

PR #4767, opened 2024-11-28
Update community-plugins.json: header-navigation

talwrii:

## Repo URL

Link to my plugin:https://github.com/talwrii/obsidian-header-navigation

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
- [ x] My GitHub release contains all required files (as individual files, not just in the source.zip / source.tar.gz)
  - [x] `main.js`
  - [x] `manifest.json`
- [ x My README.md describes the plugin's purpose and provides clear usage instructions.
- [ x] I have read the developer policies at https://docs.obsidian.md/Developer+policies, and have assessed my plugins's adherence to these policies.


talwrii:
I've made those changes and created a new release 1.0.1 without and var's in the project own code.

talwrii:
Thanks for reviewing you. Probably a pointless comment, but you don't get to say what people name things so no, "Naming something "Obsidian xyz" is not *reserved* for you. You do get to decide what gets put in your (community) plugin list however, and I can see how within the editor this is apt to create confusion so I don't mind making the change (even though a whole bunch of plugins still do it [1]) Within *github* including Obsidian in the title seems to just be clear naming.

[1] Here is a short list of plugins that use Obsidian in their name (about 60% of repos you find when searching github for Obsidian)

https://github.com/blacksmithgu/obsidian-dataview
https://github.com/SilentVoid13/Templater
https://github.com/obsidianmd/obsidian-translations
https://github.com/obsidian-tasks-group/obsidian-tasks
https://github.com/liamcain/obsidian-calendar-plugin

Anyway. I've made all these changes:

>  Remove the styles.css file from your release, since it's empty it's just unnecessary bloat that Obsidian would download.

https://github.com/talwrii/obsidian-header-navigation/commit/ba65006f0f4e0f291b905c71b17e8bad168b5a0b

>  Naming something "Obsidian xyz" is reserved for first party products we create, please change this header.

https://github.com/talwrii/obsidian-header-navigation/commit/7764a338ad13fe615f07fbe191f740fbc9372a03

>  Do not use "Obsidian" as the author of this plugin, please make it your name or username.

Obviously, Just a left over that I forgot to change from the sample plugin. Updated this:

https://github.com/talwrii/obsidian-header-navigation/commit/5ef8fd64c9aca152eb73938d28d70267ad9fa63e

>  Don't prefix the command id with the plugin id, Obsidian will make sure that there are no conflicts with other plugins.

https://github.com/talwrii/obsidian-header-navigation/commit/fba13d322ba883ca14fc52f584f546768927306d

>  I would recommend changing this to your plugin's details. Obsidian doesn't read it but it may confuse others who are viewing your plugin's source code.

https://github.com/talwrii/obsidian-header-navigation/commit/451b881d190eadc4eea7f819800620473ec7c798

I've created a new release (without style.css) here. 

https://github.com/talwrii/obsidian-header-navigation/releases/tag/1.0.3

Thanks for your help. 

Tom

joethei:
> Thanks for reviewing you. Probably a pointless comment, but you don't get to say what people name things so no, "Naming something "Obsidian xyz" is not reserved for you. You do get to decide what gets put in your (community) plugin list however, and I can see how within the editor this is apt to create confusion so I don't mind making the change (even though a whole bunch of plugins still do it [1])

This is a branding thing, we have to protect our brand, or we risk getting the trademark getting taken away from us.
Having a trademark means we can decide who get's to use the name "Obsidian", and who does not. (only in a note taking context)
To provide a example, somebody tried to name their plugin "Obsidian AI", the name is in line with our existing product naming scheme, so I think you see why we don't want plugins to be named that, they had a whole website, Twitter account (before the name change) and everything.
Now, the thing with trademark law is, you have to protect your trademark everywhere strictly, so even if we are with "Obsidian Dataview plugin" being the README header, the law kinda forces us to stop people from doing that.
(having it in the GH repo name is fine, as it's not equivalent to a product name)
The reason why you are finding so many existing instances is that this a pretty new rule, and we haven't enforced this for existing plugins yet, but we are working on it, among other things we need to enforce on existing plugins (mostly UI related stuff)

Now, to get to the actual review, found nothing to mention here, will merge tomorrow most likely.

PR #5008, opened 2024-12-30
update description

jvsteiner:
# I need to correct the description

many apologies, but I had the wrong description when I submitted this.  This plugin has already been approved, but the description is wrong.  Please accept my pull request so the description will appear correct in the community plugins.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -14653,7 +14653,7 @@
   {
     "id": "send-note",
     "name": "Send Note",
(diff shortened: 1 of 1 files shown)


joethei:
Don't include "Obsidian" in the description, it should be obvious that this is for Obsidian since you can only see it when browsing in the app / on the website.

joethei:
and "URL" is capitalized

jvsteiner:
quite right - fixed

PR #5063, opened 2025-01-04
Add plugin: Periodic notes calendar

luiisca:

## Repo URL

Link to my plugin: https://github.com/luiisca/obsidian-periodic-notes-calendar

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -18142,5 +18142,12 @@
     "author": "David V. Kimball",
     "description": "Customize the Slides core plugin's navigation, styles, and behavior.",
     "repo": "davidvkimball/obsidian-custom-slides"
+  },
+  {
+    "id": "periodic-notes-calendar",
+    "name": "Periodic notes calendar",
+    "author": "Luis Cadillo",
+    "description": "Organize periodic notes from daily to yearly with a calendar interface. Navigate and summarize entries using stickers in the calendar view.",
+    "repo": "luiisca/obsidian-periodic-notes-calendar"
   }
 ]


luiisca:
still working on it

joethei:
[# 📅 Obsidian Periodic Notes Calendar Plugin](https://github.com/luiisca/obsidian-periodic-notes-calendar/blob/cb309d57f072d30dd58ecfd9865b7941ae5db60a/README.md#L1)

[export function isMobile() {](https://github.com/luiisca/obsidian-periodic-notes-calendar/blob/cb309d57f072d30dd58ecfd9865b7941ae5db60a/src/utils.ts#L78-L92)
Use the `Platform` API: https://docs.obsidian.md/Reference/TypeScript+API/Platform

[export function handleLocaleCommands() {](https://github.com/luiisca/obsidian-periodic-notes-calendar/blob/cb309d57f072d30dd58ecfd9865b7941ae5db60a/src/utils.ts#L57-L76), [import {](https://github.com/luiisca/obsidian-periodic-notes-calendar/blob/cb309d57f072d30dd58ecfd9865b7941ae5db60a/src/ui/modals/locales-picker.ts#L1-L53)
This command has nothing to do with the rest of the plugin functionality, remove it.

[export class CalendarView extends ItemView {](https://github.com/luiisca/obsidian-periodic-notes-calendar/blob/cb309d57f072d30dd58ecfd9865b7941ae5db60a/src/view.ts#L10-L36)
You should also implement the `getIcon` method here

[.setTitle("Create Note")](https://github.com/luiisca/obsidian-periodic-notes-calendar/blob/cb309d57f072d30dd58ecfd9865b7941ae5db60a/src/ui/popovers/file-menu.ts#L49), [<div class="setting-item-name">Date Format</div>](https://github.com/luiisca/obsidian-periodic-notes-calendar/blob/cb309d57f072d30dd58ecfd9865b7941ae5db60a/src/ui/components/Nldatepicker.svelte#L130)

[import { Moment } from 'moment';](https://github.com/luiisca/obsidian-periodic-notes-calendar/blob/cb309d57f072d30dd58ecfd9865b7941ae5db60a/src/ui/modals/nldate-picker.ts#L5), [import moment, { Moment } from 'moment';](https://github.com/luiisca/obsidian-periodic-notes-calendar/blob/cb309d57f072d30dd58ecfd9865b7941ae5db60a/src/ui/components/view/manager.ts#L7)
Import moment from the Obsidian package instead of shipping your own copy

[static async restartView({ active }: { active: boolean } = { active: true }) {](https://github.com/luiisca/obsidian-periodic-notes-calendar/blob/cb309d57f072d30dd58ecfd9865b7941ae5db60a/src/ui/components/view/manager.ts#L22-L55)
There's already default settings, hotkeys, etc for opening panes in split mode. There are setting for whether new tabs should reuse or not reuse, and those are all handled for you when you use getLeaf
Implementing your own custom settings means every plugin behaves differently with respect to leaf/pane management, which is confusing for the user when using multiple plugins. It also means you can't take into account things like "Open in new window", which is often forgotten about, and not easy to implement correctly because it only works on the desktop app.

luiisca:
Hi @joethei,

Thanks so much for the detailed review and the specific feedback on leaf/pane management!

I've refactored the view handling based on your input to better align with Obsidian's defaults. Key changes include:

- Removed the dedicated setting for view placement.
- Made preview splitting an opt-in setting, adding a warning in the description about how it might alter standard panel behavior compared to the default.
- Added context menu actions for users who want to explicitly trigger a split for the preview.

luiisca:
Hi @joethei,

Following up on this PR. I've addressed your feedback (details above).

Could you please take another look when you have a moment? To prevent the automated system from closing it due to inactivity, a re-review would be much appreciated.

Thanks!

luiisca:
Hi @joethei, just a quick reminder that this PR may be auto closed soon. All feedback has been addressed. Could you review when you have a moment?

luiisca:
Hi @joethei, thanks for the follow-up.

The second link points to an old commit. That code isn't present in the latest version, so no change is needed there.

The other two points have now been addressed.

Let me know if there's anything else.

PR #5096, opened 2025-01-08
Add plugin: md-to-slack

NateCashmore:
Addition of md-to-slack plugin which converts markdown messages to Slack Block Kit messages.


NateCashmore:
Wrong process followed so closing this in favour of later PR

PR #5097, opened 2025-01-08
Add plugin: Md to Slack

NateCashmore:

## Repo URL
(https://github.com/NateCashmore/md-to-slack-plugin)

Link to my plugin:

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [ ] `main.js`
  - [ ] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -15209,5 +15209,5 @@
     "author": "Nathan Cashmore",
     "description": "Converts markdown messages to Slack Block Kit messages",
     "repo": "NateCashmore/md-to-slack-plugin"
-  },	
+  }	
 ]
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -15202,5 +15202,12 @@
 	"description": "Import bookmarks from Readeck.",
 	"author": "Makebit",
     "repo": "makebit/obsidian-readeck-importer"
-  }
+  },
+  {
+    "id": "md-to-slack",
+    "name": "Md to Slack",
+    "author": "Nathan Cashmore",
+    "description": "Converts markdown messages to Slack Block Kit messages",
+    "repo": "NateCashmore/md-to-slack-plugin"
+  },	
 ]


joethei:
What sets your plugin apart from the other plugins that allow exporting to Slack?

[Md](https://github.com/NateCashmore/md-to-slack-plugin/blob/019905a32d6453e7707f900821aa0b9c357d8c10/manifest.json#L3), [markdown](https://github.com/NateCashmore/md-to-slack-plugin/blob/019905a32d6453e7707f900821aa0b9c357d8c10/manifest.json#L6), [Md](https://github.com/NateCashmore/md-to-slack-plugin/blob/019905a32d6453e7707f900821aa0b9c357d8c10/README.md#L1), [Md](https://github.com/NateCashmore/md-to-slack-plugin/blob/019905a32d6453e7707f900821aa0b9c357d8c10/src/main.ts#L17)
Please use the proper capitalization, which would be `MD` and `Markdown` respectively.

Remove the `styles.css` file from your release, since it's empty it's just unnecessary bloat that Obsidian would download.

[const slackifyMarkdown = require('slackify-markdown');](https://github.com/NateCashmore/md-to-slack-plugin/blob/019905a32d6453e7707f900821aa0b9c357d8c10/src/converter.ts#L1-L2)
Please stick to one import style.

NateCashmore:
"What sets your plugin apart from the other plugins that allow exporting to
Slack?"
I wrote the plugin as at the time the other two options which I had looked
into did not work when installed.  I assumed as they were both over 2 years
old and had no updates in that period that they had not been maintained.
As a result I wrote this.

Today I have managed to install them both and they both work.  This then
becomes a bit embarrassing as I also don't see the need for yet another
Markdown to Slack converter so this request can be closed.

I'll contact the other developers as the only differentiator now is:
* Conversion includes title of note
* Better formatting of > elements in line with presentation in obsidian
* Configurable endpoint for Slack Block Builder URL.

I've updated the code based on your suggestions and pushed but will close
the merge request based on the above.
Thanks for looking into this.

PR #5117, opened 2025-01-11
Add plugin: Better Paste

xyve7:

## Repo URL

Link to my plugin: https://github.com/xyve7/obsidian-better-paste

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -14782,6 +14782,13 @@
     "author": "zero",
     "description": "打造您的极致插件管理体验，让插件管理变得更加直观、高效，同时提升您的工作流程和个性化设置。",
     "repo": "0011000000110010/obsidian-manager"
+  },
+  {
+    "id": "better-paste",
+    "name": "Better Paste",
+    "author": "xyve",
+    "description": "Better renaming for pasted images in Obsidian.",
+    "repo": "xyve7/obsidian-better-paste"
   }
 ]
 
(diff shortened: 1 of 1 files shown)


joethei:
[import * as path from 'path';](https://github.com/xyve7/obsidian-better-paste/blob/3384a04ad19cf8c788eb9ecd0f3f6ef2d57b6630/main.ts#L2)

[this.app.vault.on('create', async (file) => {](https://github.com/xyve7/obsidian-better-paste/blob/3384a04ad19cf8c788eb9ecd0f3f6ef2d57b6630/main.ts#L27)

[// This event gets fired when obsidian is being loaded](https://github.com/xyve7/obsidian-better-paste/blob/3384a04ad19cf8c788eb9ecd0f3f6ef2d57b6630/main.ts#L28-L32)
The proper approach is to put the registration of the event inside of the `onLayoutReady` call.


[titleEl.setText('Renaming Pasted Image');](https://github.com/xyve7/obsidian-better-paste/blob/3384a04ad19cf8c788eb9ecd0f3f6ef2d57b6630/main.ts#L71)

xyve7:
Thank you for the tip about the registering the event within the `onLayoutReady` call. I made your requested changes.

xyve7:
Hello, what is the status on this PR?

PR #5159, opened 2025-01-16
Add plugin: Yesterday's note

trevortylerlee:

## Repo URL

Link to my plugin: https://github.com/trevortylerlee/yesterdays-note

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16196,6 +16196,13 @@
     "author": "Nathan Arthur",
     "description": "Use AI to generate and track research questions based on your notes.",
     "repo": "narthur/research-quest"
-	}
+},
+{
+    "id": "yesterdays-note",
+    "name": "Yesterday's note",
+    "author": "Trevor Tyler Lee",
+    "description": "Open yesterday's daily note.",
+    "repo": "trevortylerlee/yesterdays-note"
+}
 ]
 


trevortylerlee:
I've refactored the code to use the `instanceof` check to validate if the file/folder exists.

Zachatoo:
["fundingUrl": "https://trevortylerlee.com",](https://github.com/trevortylerlee/yesterdays-note/blob/6ba3bdb8498bee77fdcea5df0fa59bb42935233d/manifest.json#L9)

[return await this.app.vault.read(templateFile);](https://github.com/trevortylerlee/yesterdays-note/blob/6ba3bdb8498bee77fdcea5df0fa59bb42935233d/main.ts#L147)
[Prefer to use `Vault.cachedRead()` over `Vault.read()` when you aren't going to be modifying the file](https://docs.obsidian.md/Plugins/Vault#Read+files)

[const leaf = this.app.workspace.getUnpinnedLeaf();](https://github.com/trevortylerlee/yesterdays-note/blob/6ba3bdb8498bee77fdcea5df0fa59bb42935233d/main.ts#L173)
Prefer to not use deprecated APIs. Please use `getLeaf(false)` instead.

---

Is there an expected use case for the Daily Note settings and your plugin's settings to differ? For example, for the Daily Note's date format and your plugin's date format setting to be different? I would recommend either removing settings from your plugin that can be inherited from the Daily Notes plugin settings, or adding a description to each setting that defaults to the Daily Note plugin settings to make it clear that it will inherit the Daily Note plugin setting, and that the user does not need to fill out that setting.

trevortylerlee:
Thanks for the feedback, made the following changes:

- use Vault.cachedRead() instead of Vault.read()
- use getLeaf(false) instead of getUnpinnedLeaf()
- remove unnecessary settings and use default Daily note settings
  - remove custom date format
  - remove custom note location
  - remove custom template

joethei:
Please remove the "Obsidian" branding from the image in your README, that makes it look like this is an official product.

trevortylerlee:
Done.

trevortylerlee:
@joethei Changes have been made, can you take another look?

trevortylerlee:
@joethei can I get some help? Not sure why it's saying my latest release is missing the `main.ts` and `manifest.json` files.

Zachatoo:
@trevortylerlee The bot is correct, you are missing the main.js and manifest.json files from your latest release. Notice the difference between these two releases. If you use the GitHub action in your repo for releases then the step of adding those files to your release will be done for you.

image

PR #5185, opened 2025-01-19
Add plugin: Custom New Filename

godrix:

## Repo URL

Link to my plugin: https://github.com/godrix/obsidian-new-filename

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -14908,5 +14908,12 @@
     "author": "Amin Sennour",
     "description": "Change the default filename used for new notes",
     "repo": "TheLoneWanderer4/obsidian-uuid-title"
-  }
+  },
+{
+  "id": "custom-new-filename",
+  "name": "Custom New Filename",
+  "author": "Carlos Gabriel Godri",
+  "description": "Customize the default filename for new notes, with unique and date-based options.",
+  "repo": "godrix/obsidian-new-filename"
+}
 ]


Zachatoo:
[# Obsidian New FileName Plugin](https://github.com/godrix/obsidian-new-filename/blob/118fdc28ed6773b5b50720bcc00c42a25f8cf295/README.md#L1)

[file.basename = this.getLowestNonColidingFilename(filename);](https://github.com/godrix/obsidian-new-filename/blob/118fdc28ed6773b5b50720bcc00c42a25f8cf295/main.ts#L29)
This will behave weird in some scenarios. For example, if you create a new note with an explicit name (such as with the Quick Switcher or clicking on a link to a non-existing note), the note will appear to take the note name you gave it instead of the explicit name provided by the user. But when Obsidian is reloaded, it will have the explicit file name, not the file name your plugin tried to override it with.

I recommend making the following changes to avoid the above issues:

1. Use [FileManager.renameFile](https://docs.obsidian.md/Reference/TypeScript+API/FileManager/renameFile) to rename the file after creating it instead of updating the `basename`.
2. Only rename the file if it starts with "Untitled", so you don't rename explicitly named files.

---

I know you don't have it in your releases, but consider removing the styles.css file from your repo as well if you aren't using it.

PR #5247, opened 2025-01-27
Update community-plugins.json

jose-elias-alvarez:

## Repo URL

Link to my plugin: https://github.com/jose-elias-alvarez/obsidian-easy-fonts

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


jose-elias-alvarez:
/skip (modifying document styles via JS is necessary for the plugin's functionality)

joethei:
What sets your plugin apart from the "Custom Font Loader" plugin?

jose-elias-alvarez:
> What sets your plugin apart from the "Custom Font Loader" plugin?

I found out about Custom Font Loader about 50% of the way through making this plugin, and it does indeed achieve the same goal at a high level, but after reviewing that plugin's features and codebase, I kept working on mine for these reasons:

1. Much faster font loading mechanism: [FontFace API](https://developer.mozilla.org/en-US/docs/Web/API/FontFace) vs. creating and appending a `<style>` tag
2. Cleaner code / architecture, leading to other speed boosts + fewer bugs (for example, I don't think Custom Font Loader ever cleans up old base64 font files, and it doesn't always unload cleanly)
3. Ability to configure a different font for interface / text / monospace, mirroring Obsidian's own options
4. No restrictions on where fonts are located in the vault

Custom Font Loader could theoretically be patched to use the FontFace API / have a bunch of bugs fixed, but the second two changes could break existing user configs, so I thought it made sense to provide an alternative and start fresh.

PR #5283, opened 2025-02-01
Update community-plugins.json

OmriLeviGit:
# I am changing the plugin's name and repo link

## Repo URL
https://github.com/OmriLeviGit/Auto-List-Reordering-Obsidian

As the renumbering feature was implemented in 1.8.3, i have made some edits and removed the main feature from newer versions, leaving only the secondary feature for automatic checkbox sorting.
The old name does not fit the description so i have changed it.

I left the plugin ID as it is, and already made changes to the manifest.


OmriLeviGit:
Hey @joethei, tagging as it might have been missed,
a small rename and a description change <3

PR #5324, opened 2025-02-07
Add plugin: Shortcuts

Quorafind:

## Repo URL

Link to my plugin: https://github.com/Quorafind/Obsidian-Shortcuts

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -15048,5 +15048,12 @@
     "author": "Ben Rotholtz",
     "description": "Track your goals with a calendar view",
     "repo": "GizmoRay/obsidian-goal-tracker"  
+  },
+  {
+    "id": "shortcuts",
+    "name": "Shortcuts",
+    "author": "Johnny & Boninall",
+    "description": "A shortcuts plugin for power users.",
+    "repo": "quorafind/obsidian-shortcuts"  
   }
 ]


Quorafind:
/skip Style is used for absolute position. All these elements would be generated and destoryed at the same time. And would not update their position after first loading,

joethei:
Your README is a bit confusing, "Shortcuts" is another word for "Hotkeys" in a lot of cases, which adds to the confusion.
Even after reading the code, I am still confused what the plugin can do, because a lot of things are not described at all.
Like the `aria-label` feature for example.

["description": "A shortcuts plugin for power users.",](https://github.com/Quorafind/Obsidian-Shortcuts/blob/a47aeb599db1d3fe3487874d6972e1c0537c0fa4/manifest.json#L6)
This description is also very confusing, same reasons as above.
Also, please don't use the word "plugin" in the description.

[const keycode = require("keycode");](https://github.com/Quorafind/Obsidian-Shortcuts/blob/a47aeb599db1d3fe3487874d6972e1c0537c0fa4/src/shortcutsSettingTab.ts#L25-L26)
Please stick to one import style.

Quorafind:
All you mentioned issues are fixed.

Quorafind:
Already added to https://github.com/Quorafind/Obsidian-Shortcuts?tab=readme-ov-file#overview

Quorafind:
Hello, any updates?

PR #5414, opened 2025-02-17
Add plugin: File Renamer

carlosmunoz:
Submit file renamer plugin for addition to the community plugin list.

## Repo URL

Link to my plugin:

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


carlosmunoz:
Thank you for the review. I have addressed all comments except for one cited below.

/skip
> [[2](https://github.com/carlosmunoz/obsidian-file-renamer/blob/37b5fb63c4b011faec4ef5a64a00c45f0f5e9706/suggesters/FileSuggester.ts#L26)]:You should not cast this, instead use a `instanceof` check to make sure that it's actually a file/folder.

There is already an `instanceof` check in the `filter` call in the previous line. This would remove all non TFile instances, but the subsequent calls still need these instances cast.

joethei:
[import * as path from 'path';](https://github.com/carlosmunoz/obsidian-file-renamer/blob/b3b0ad24b55b6824691c30f0587e14390cb0a10f/main.ts#L2)

[const newFilePath = file.parent?.path + path.sep + newFileName;](https://github.com/carlosmunoz/obsidian-file-renamer/blob/b3b0ad24b55b6824691c30f0587e14390cb0a10f/main.ts#L30)
The above can be easily fixed by using a regular slash `/` here instead of `path.sep`, the Obsidian API already handles different path seperators for you.

[console.log(`File renamer - renaming file: ${file.path} to ${newFilePath}`);](https://github.com/carlosmunoz/obsidian-file-renamer/blob/b3b0ad24b55b6824691c30f0587e14390cb0a10f/main.ts#L32)

[// Credit for this goes to SilentVoid13's templater plugin suggesters:](https://github.com/carlosmunoz/obsidian-file-renamer/blob/b3b0ad24b55b6824691c30f0587e14390cb0a10f/suggesters/suggest.ts#L1-L202)

[containerEl.createEl("h2", { text: "Renaming Rules" });](https://github.com/carlosmunoz/obsidian-file-renamer/blob/b3b0ad24b55b6824691c30f0587e14390cb0a10f/settings/SettingsTab.ts#L21)
```ts
```

[containerEl.createEl("h2", { text: "Renaming Rules" });](https://github.com/carlosmunoz/obsidian-file-renamer/blob/b3b0ad24b55b6824691c30f0587e14390cb0a10f/settings/SettingsTab.ts#L21), [["File Name (Regex)", "New Name (Regex)", "Template to Apply", "Actions"].forEach((text) => {](https://github.com/carlosmunoz/obsidian-file-renamer/blob/b3b0ad24b55b6824691c30f0587e14390cb0a10f/settings/SettingsTab.ts#L28), [btn.setButtonText("Add Entry")](https://github.com/carlosmunoz/obsidian-file-renamer/blob/b3b0ad24b55b6824691c30f0587e14390cb0a10f/settings/SettingsTab.ts#L106)

carlosmunoz:
Thanks for the review @joethei. I am actually rethinking this plugin and might cancel the PR. I'll address your comments if I decide to move forward.

PR #5420, opened 2025-02-17
Add plugin: Invert Pasted Image

4yt4c4yd1n:

## Repo URL

Link to my plugin: https://github.com/4yt4c4yd1n/invert-pasted-image

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -15188,5 +15188,12 @@
     "author": "Nick de Bruin",
     "description": "Creates a searchable gallery and sidecar files for attachments such as images and videos. The sidecar files allow you to add notes and tags to your media files.",
     "repo": "Nick-de-Bruin/obsidian-media-companion"
+  },
+  {
+    "id": "invert-pasted-image",
+    "name": "Invert Pasted Image",
+    "author": "4yt4c4yd1n",
+    "description": "Plugin to automatically invert pasted images",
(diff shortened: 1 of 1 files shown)


Zachatoo:
[import { inspect } from 'util'](https://github.com/4yt4c4yd1n/invert-pasted-image/blob/3464a11983427b6ea8cc4465b6be98e41a4fb3a4/main.ts#L2)
This import is only available from a Node.js runtime, which means that this will throw errors on Mobile (that's a problem since you have `isDesktopOnly` marked as `false` in your manifest.json file. Please remove this import.

[console.log(evt.clipboardData)](https://github.com/4yt4c4yd1n/invert-pasted-image/blob/3464a11983427b6ea8cc4465b6be98e41a4fb3a4/main.ts#L24)

Is it intentional that using the "Toggle inverting pasted images" command does not persist the change after reloading Obsidian? It feels odd to run the command and see the setting change in the settings tab, but it isn't actually changed because you didn't call `saveSettings()`.

The plugin doesn't appear to invert images for me, I'm guessing it's because the latest release does not reflect the code. Can you please create a new release once changes are made to make this easier to test?

PR #5439, opened 2025-02-18
Add plugin: Rainbow-Colored Sidebar

Kovah:

## Repo URL

Link to my plugin: https://github.com/Kovah/obsidian-rainbow-colored-sidebar

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux (not available to me)
  - [ ]  Android _(if applicable)_ (not available to me)
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16854,6 +16854,13 @@
   "author": "Maclean Dunkin",
   "description": "Sync your Limitless AI lifelog entries",
   "repo": "Maclean-D/obsidian-limitless-lifelogs"
-}
+},
+{
+    "id": "rainbow-colored-sidebar",
+    "name": "Rainbow-Colored Sidebar",
+    "author": "Kevin Woblick",
+    "description": "Automatically color your sidebar like a rainbow. No configuration needed.",
+    "repo": "Kovah/obsidian-rainbow-colored-sidebar"  
+  }
 ]
 


Kovah:
/skip Settings tab was renamed. The usage of innerHTML is not a security risk in this case, as static HTML is applied with some text and a link. Styles need to be set via JS, otherwise the whole plugin wouldn't work, as the user selects different static settings which are then applied. In this case, CSS variables are set which are used in the plugin stylesheet.

joethei:
[`<h1 align="center">Obsidian Rainbow-Colored Sidebar Plugin</h1>`](https://github.com/Kovah/obsidian-rainbow-colored-sidebar/blob/c6a64f3e0940fda03d7d6883eda29f1189ec1f1e/README.md#L1), [Obsidian Rainbow Colored Sidebar](https://github.com/Kovah/obsidian-rainbow-colored-sidebar/blob/c6a64f3e0940fda03d7d6883eda29f1189ec1f1e/styles.scss#L2)

[.setName('Color Scheme')](https://github.com/Kovah/obsidian-rainbow-colored-sidebar/blob/c6a64f3e0940fda03d7d6883eda29f1189ec1f1e/main.ts#L63), [.setName('Increase Contrast')](https://github.com/Kovah/obsidian-rainbow-colored-sidebar/blob/c6a64f3e0940fda03d7d6883eda29f1189ec1f1e/main.ts#L89)

Kovah:
Point 2 has been changed, config dir is now excluded by this.app.vault.configDir.

> You should avoid assigning styles via JavaScript
Plugin won't work without that.

> Using innerHTML, outerHTML or similar API's is a security risk.
In this case, it's not, as stated above. A hard-coded string is placed using a 1-liner, instead of having to write a dozen lines of code.

Kovah:
Release fixed

PR #5440, opened 2025-02-18
Add plugin: Note Links Citations

stretchyboy:

## Repo URL
https://github.com/stretchyboy/obsidian-note-link-citation

Link to my plugin:

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -15307,5 +15307,12 @@
     "author": "Pavlo Kobyliatskyi",
     "description": "Memorizing words or phrases",
     "repo": "pavlokobyliatskyi/obsidian-memodack-plugin"
+  },
+  {
+    "id": "note-links-citations",
+    "name": "Note Links Citations Plugin",
+    "author": "Martyn Eggleton",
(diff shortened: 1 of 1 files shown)


Zachatoo:
Remove the `styles.css` file from your release, since it's empty it's just unnecessary bloat that Obsidian would download. Also remove it from your repo.

[if ("source" in oPage && oPage.source){](https://github.com/stretchyboy/obsidian-note-link-citation/blob/b722840e590bc3daf3673dc09ad8508830582fd8/main.ts#L38-L39)
This logic doesn't make sense. If a page has a source, then add the page's title, but the title may be undefined. Also not sure if it makes sense to only show the page number if the link has authors, and otherwise removing it.

The README should more clearly indicate how to use this plugin. I had to read through the code to see that I needed to add specific properties to notes. The README should specify which properties are available to use, what the data type of that property should be (list, text, etc), and some examples.

---

Consider removing unused code.

PR #5441, opened 2025-02-18
Add plugin: Tier List

MoxAlehin:

## Repo URL

Link to my plugin: https://github.com/MoxAlehin/TierList

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -15769,5 +15769,12 @@
     "author": "Belinda Mo, Athena Leong",
     "description": "Review, generate, and add flashcards for your notes using AI",
     "repo": "ai-learning-tools/obsidian-spaced-repetition-ai"
+  },
+  {
+    "id": "tier-list",
+    "name": "Tier List",
+    "author": "Mox Alehin",
+    "description": "Visual ranking and organizing content into customizable tier lists.",
+    "repo": "MoxAlehin/TierList"
   }
 ]


MoxAlehin:
/skip I removed some JS that assigned styles directly. However, the remaining assignments allow users to modify CSS variables, which adjust the appearance of my plugin through the plugin settings rather than themes and snippets.

joethei:
[new Setting(containerEl).setName('Default settings').setHeading();](https://github.com/MoxAlehin/TierList/blob/86c9401ae674c233c9fd20a1b0dbff90e72bec72/src/settings.ts#L101)

[let content = await app.vault.read(file);](https://github.com/MoxAlehin/TierList/blob/86c9401ae674c233c9fd20a1b0dbff90e72bec72/src/utils/file-utils.ts#L9-L24), [const fileContent = await app.vault.read(activeFile);](https://github.com/MoxAlehin/TierList/blob/86c9401ae674c233c9fd20a1b0dbff90e72bec72/src/utils/file-utils.ts#L31-L41), [const fileContent = await app.vault.read(activeFile);](https://github.com/MoxAlehin/TierList/blob/86c9401ae674c233c9fd20a1b0dbff90e72bec72/src/utils/file-utils.ts#L52-L61), [const content = await app.vault.read(activeFile);](https://github.com/MoxAlehin/TierList/blob/86c9401ae674c233c9fd20a1b0dbff90e72bec72/src/utils/file-utils.ts#L82-L89), [const fileContent = await app.vault.read(activeFile);](https://github.com/MoxAlehin/TierList/blob/86c9401ae674c233c9fd20a1b0dbff90e72bec72/src/utils/file-utils.ts#L96-L103)
Use the `process` API to make sure that your edits do not conflict with other plugins.
Take a look at: <https://youtu.be/RvtV9DKr3ag?si=IBYnWr73UntlbVLr&t=52> for the proper usage.

[// Credits go to Liam's Periodic Notes Plugin: https://github.com/liamcain/obsidian-periodic-notes](https://github.com/MoxAlehin/TierList/blob/86c9401ae674c233c9fd20a1b0dbff90e72bec72/src/suggesters/suggest.ts#L1-L184)

[await MarkdownRenderer.render(app, str, this.renderEl, '', this.plugin);](https://github.com/MoxAlehin/TierList/blob/86c9401ae674c233c9fd20a1b0dbff90e72bec72/src/modals/slot-modal.ts#L207)
Don't pass your plugin instance as the component here, that is too long lived and can lead to memory leaks.
The component should be pretty short lived, only for as long as necessary, i.e. only as long as the content is visible.

PR #5558, opened 2025-03-02
Add plugin: MP Preview

Yeban8090:

## Repo URL

Link to my plugin:https://github.com/Yeban8090/mp-preview

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


Yeban8090:
1. The main.js file has been deleted and ignored in .gitignore.
/skip 2. You should avoid assigning styles through JavaScript or HTML and instead move all these styles to CSS so that the theme and code snippets can adapt to them more easily. Skipping the check is necessary because my theme requires real-time preview of changes, which can only be set via JavaScript.
/skip 3. Using innerHTML, outerHTML, or similar APIs poses security risks. These are also required as they are needed for dynamic settings.
4. Modified.

Yeban8090:
Do I still need to make changes? Why does it prompt: "This branch cannot be rebased due to conflicts"?

PR #5607, opened 2025-03-06
Add plugin: Folders Graph

d7sd6u:

## Repo URL

Link to my plugin: https://github.com/d7sd6u/obsidian-folders-graph

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -15762,5 +15762,12 @@
     "author": "Gareth Ng",
     "description": "Generate blog posts from your notes using OpenAI",
     "repo": "garethng/obsidian-blog-generator"
+  },
+  {
+    "id": "folders-graph",
+    "name": "Folders Graph",
+    "description": "Add folders to graph views.",
+    "author": "d7sd6u",
+    "repo": "d7sd6u/obsidian-folders-graph"
   }
 ]


joethei:
[red: { light: 0xe93147, dark: 0xfb464c },](https://github.com/d7sd6u/obsidian-folders-graph/blob/ae9dd5b09ef2769075a1c0916e83ae8d5248d15d/src/main.ts#L55-L62)
These colors should not be hardcoded, but user changeable.
In the core app for example you can change the colors using some CSS variables.

[this.app.metadataCache.resolvedLinks =](https://github.com/d7sd6u/obsidian-folders-graph/blob/ae9dd5b09ef2769075a1c0916e83ae8d5248d15d/src/main.ts#L166-L181)
Please don't change the metadataCache data, this can lead to serious data loss if other plugins react to this.
You are also missing error handling, if there is a exception then the metadata cache will remain broken.

d7sd6u:
> [this.app.metadataCache.resolvedLinks =](https://github.com/d7sd6u/obsidian-folders-graph/blob/ae9dd5b09ef2769075a1c0916e83ae8d5248d15d/src/main.ts#L166-L181) Please don't change the metadataCache data, this can lead to serious data loss if other plugins react to this. You are also missing error handling, if there is a exception then the metadata cache will remain broken.

This is a synchronous code in the one-threaded language. There is no way other plugins can react to this (assuming they don't patch `render` function as well and in such a way that has mutating side-effects - that would be ridiculous) except for adding getter to the `resolvedLinks`. Again this is a bad practice and is a problem with that other plugin, _not_ with my patch.

I can avoid changing the `resolvedLinks`, but that would mean I would have to rewrite the `render` function from scratch. I can do that if you insist, but I find that to be meaningless work that is not worth it especially considering that I would have to constantly keep my re-engineered version up to date with new Obsidian versions.

Nonetheless, indeed I should have wrapped the `next.apply()` in try/catch, my bad. I'll fix that.

d7sd6u:
I have added the color pickers for the new nodes colors.

PR #5637, opened 2025-03-09
Add plugin: One Step Wiki Link

busyoGG:

## Repo URL

Link to my plugin:https://github.com/busyoGG/OneStepWikiLink

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`


joethei:
[const path = require("path");](https://github.com/busyoGG/OneStepWikiLink/blob/bc474b6727562e3bdd5a94bf28e93dabc7bd7f20/src/main.ts#L4)

[language: Language;](https://github.com/busyoGG/OneStepWikiLink/blob/bc474b6727562e3bdd5a94bf28e93dabc7bd7f20/src/main.ts#L15), [language: Language.CN,](https://github.com/busyoGG/OneStepWikiLink/blob/bc474b6727562e3bdd5a94bf28e93dabc7bd7f20/src/main.ts#L24), [new Setting(containerEl)](https://github.com/busyoGG/OneStepWikiLink/blob/bc474b6727562e3bdd5a94bf28e93dabc7bd7f20/src/setting.ts#L85-L99)
This should not be a setting, use `getLanguage` from the Obsidian API to get the language configured in the app.
Note that this requires the `minAppVersion` to be set to 1.8.0

[autoTimer: NodeJS.Timeout | undefined;](https://github.com/busyoGG/OneStepWikiLink/blob/bc474b6727562e3bdd5a94bf28e93dabc7bd7f20/src/main.ts#L54-L55)
Use `number` as the type here and use `window.setTimeout` / `window.clearTimeout` where this is being used.

[name: "Convert All Matching Words to Wiki Links",](https://github.com/busyoGG/OneStepWikiLink/blob/bc474b6727562e3bdd5a94bf28e93dabc7bd7f20/src/main.ts#L131)
[Use sentence case in UI](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Use+sentence+case+in+UI) and you should also add this to your i18n configuration.

busyoGG:
> [const path = require("path");](https://github.com/busyoGG/OneStepWikiLink/blob/bc474b6727562e3bdd5a94bf28e93dabc7bd7f20/src/main.ts#L4) This import is only available from a Node.js runtime, which means that this will throw errors on Mobile (that's a problem since you have `isDesktopOnly` marked as `false` in your manifest.json file.
> 
> [language: Language;](https://github.com/busyoGG/OneStepWikiLink/blob/bc474b6727562e3bdd5a94bf28e93dabc7bd7f20/src/main.ts#L15), [language: Language.CN,](https://github.com/busyoGG/OneStepWikiLink/blob/bc474b6727562e3bdd5a94bf28e93dabc7bd7f20/src/main.ts#L24), [new Setting(containerEl)](https://github.com/busyoGG/OneStepWikiLink/blob/bc474b6727562e3bdd5a94bf28e93dabc7bd7f20/src/setting.ts#L85-L99) This should not be a setting, use `getLanguage` from the Obsidian API to get the language configured in the app. Note that this requires the `minAppVersion` to be set to 1.8.0
> 
> [autoTimer: NodeJS.Timeout | undefined;](https://github.com/busyoGG/OneStepWikiLink/blob/bc474b6727562e3bdd5a94bf28e93dabc7bd7f20/src/main.ts#L54-L55) Use `number` as the type here and use `window.setTimeout` / `window.clearTimeout` where this is being used.
> 
> [name: "Convert All Matching Words to Wiki Links",](https://github.com/busyoGG/OneStepWikiLink/blob/bc474b6727562e3bdd5a94bf28e93dabc7bd7f20/src/main.ts#L131) [Use sentence case in UI](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Use+sentence+case+in+UI) and you should also add this to your i18n configuration.

I have removed path, replaced it with manually concatenated strings, and fixed the other mentioned issues.

I added a Chinese command while keeping both language versions of the command, as the English command is more convenient for typing.

PR #5638, opened 2025-03-09
Add plugin: File Compressor

networkmastered:

## Repo URL

Link to my plugin:https://github.com/networkmastered/obsidian-compress

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -15804,5 +15804,12 @@
     "author": "Marc André Ueberall",
     "description": "Weather generator for the HârnMaster: Roleplaying in the World of Kèthîra rule system.",
     "repo": "marcueberall/obsidian-harn-weather"
-  }
+  },
+  {
+    "id": "compress",
+    "name": "File Compressor",
+    "author": "networkmastered",
+    "description": "This plugin will add the .ctxt file type which will allow you to compress notes to a smaller size.",
+    "repo": "networkmastered/obsidian-compress"
+  },
 ]


networkmastered:
Thanks, they have now been replaced.

staticmatrix (approved):
I like the way you have now updated the repository to have multiple types of compression for different cases. I want to recommend to add blocks to the data so if something fails you can find the issue in a much smaller area. Maybe also make your own view class instead of using obsidians so you dont have to do that hacky solution. Ive tested it out though and it seems to work pretty well. Just also want to note that sometimes its not loading the text when i open obsidian on one of its files.

networkmastered:
> I like the way you have now updated the repository to have multiple types of compression for different cases. I want to recommend to add blocks to the data so if something fails you can find the issue in a much smaller area. Maybe also make your own view class instead of using obsidians so you dont have to do that hacky solution. Ive tested it out though and it seems to work pretty well. Just also want to note that sometimes its not loading the text when i open obsidian on one of its files.

oh wow, thanks for the info the blocking idea is good, ill definitely implement that with some checksums. Ill maybe work on the view at some point later, as of now its just sort of been as you say a hacky fix. Ill need to recreate obsidians editor and markdown whichll take a while. The loading sends a request to the plugin which causes it to then reload the leaf(editor). Once there is a custom view/leaf that should also fix  the issue. If you have any further problems you could maybe put them in the issues on the repo ;)

joethei:
[Copyright (C) 2020-2025 by Dynalist Inc.](https://github.com/networkmastered/obsidian-compress/blob/4d2bac86c077ee518356a0928e5ca9b88451de41/LICENSE#L1)
Update the license file.

["fundingUrl": "https://github.com/networkmastered",](https://github.com/networkmastered/obsidian-compress/blob/4d2bac86c077ee518356a0928e5ca9b88451de41/manifest.json#L9)

[obsidian](https://github.com/networkmastered/obsidian-compress/blob/4d2bac86c077ee518356a0928e5ca9b88451de41/README.md#L2)
Obsidian is always capitalized.

[statusBarItemEl2.title = "File Size"](https://github.com/networkmastered/obsidian-compress/blob/4d2bac86c077ee518356a0928e5ca9b88451de41/main.ts#L200)

[alert("In order to view CTXT files you will need to re-enable the plugin. If you are deleting the plugin, please convert your files back before.")](https://github.com/networkmastered/obsidian-compress/blob/4d2bac86c077ee518356a0928e5ca9b88451de41/main.ts#L238)
Please remove this message, this will trigger in a few cases where the message makes no sense.

[(function (global, factory) {](https://github.com/networkmastered/obsidian-compress/blob/4d2bac86c077ee518356a0928e5ca9b88451de41/util/pakoMinified.js#L1-L663), [//compressed json. (dictionaryBIG.json)](https://github.com/networkmastered/obsidian-compress/blob/4d2bac86c077ee518356a0928e5ca9b88451de41/util/networkmastercompression.js#L7-L14)
Please include the actual source code and not a minified / compiled version.

networkmastered:
> [alert("In order to view CTXT files you will need to re-enable the plugin. If you are deleting the plugin, please convert your files back before.")](https://github.com/networkmastered/obsidian-compress/blob/4d2bac86c077ee518356a0928e5ca9b88451de41/main.ts#L238)
Please remove this message, this will trigger in a few cases where the message makes no sense.

I have removed the alert. It was there as when the plugin is disabled it means that they user wont be able to edit any of the compressed files. Is there any other way that could be implemented or should it just stay removed?

joethei:
The warning in the README is enough.

PR #5649, opened 2025-03-10
Add plugin: White Noise

zhouhua:

## Repo URL

Link to my plugin: https://github.com/zhouhua/obsidian-white-noise

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16476,6 +16476,13 @@
     "author": "Aava",
     "description": "Rename your file to the current text selection.",
     "repo": "AavaGames/obsidian-rename-file-to-selection"
+  },
+  {
+    "id": "white-noise",
+    "name": "White Noise",
+    "author": "Zhou Hua",
+    "description": "Play white noise to help you focus on your work.",
+    "repo": "zhouhua/obsidian-white-noise"
   }
 ]
 


zhouhua:
/skip Due to the need for dynamically setting icon sizes, and since Obsidian's ⁠setIcon method doesn't provide the corresponding interface, it can only be done through element attributes.

joethei:
Some screenshots in the README would be great.

["description": "Play white noise in Obsidian to help you focus on your work",](https://github.com/zhouhua/obsidian-white-noise/blob/6d23be9d29098f1ce151d98362f41d9d4935bf5b/manifest.json#L6)
Don't include "Obsidian" in the description, this should be obvious given the context in which this is shown.

[.status-bar-item {](https://github.com/zhouhua/obsidian-white-noise/blob/6d23be9d29098f1ce151d98362f41d9d4935bf5b/src/main.css#L5)

[locale = (window.i18next.language || '').startsWith('zh') ? 'zh' : 'en';](https://github.com/zhouhua/obsidian-white-noise/blob/6d23be9d29098f1ce151d98362f41d9d4935bf5b/src/L.ts#L17)
Get the language from the `getLanguage()` function (requires `minAppVersion` of 1.8.0`

[config: 'White Noise Settings',](https://github.com/zhouhua/obsidian-white-noise/blob/6d23be9d29098f1ce151d98362f41d9d4935bf5b/src/i18n/en/index.ts#L22)

zhouhua:
@joethei I have fixed all the issues you mentioned, except for the last one. Because this text isn't used in the settings tab: 

image

joethei:
It should not be there either

zhouhua:
I don't think so. I believe it's best for this pop-up panel to have a title indicating its purpose. How about making a slight adjustment? I've removed the word "setting".

image

PR #5857, opened 2025-03-30
Add plugin: LLM docs

shane-lamb:

## Repo URL

Link to my plugin: https://github.com/shane-lamb/obsidian-llm-docs

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -17470,5 +17470,12 @@
     "author": "Zarstensen",
     "description": "Evaluate, solve and much more within LaTeX blocks using Sympy.",
     "repo": "zarstensen/obsidian-latex-math"
-  }
+  },
+{
+    "id": "llm-docs",
+    "name": "LLM docs",
+    "author": "Shane Lamb",
+    "description": "Simple LLM (AI chat) client using pure markdown documents",
+    "repo": "shane-lamb/obsidian-llm-docs"
+}
 ]


Zachatoo:
What sets your plugin apart from the other existing Chat-GPT/GPT-3 plugins?

[new Notice("You're on fire!")](https://github.com/shane-lamb/obsidian-llm-docs/blob/2e47abd77b620c2b92d43fb295157b6441be8cb3/src/main.ts#L118)
Can you please include more helpful information for the user to self-diagnose why this command wouldn't be working?

[await app.fileManager.processFrontMatter(file, (frontmatter) => {](https://github.com/shane-lamb/obsidian-llm-docs/blob/2e47abd77b620c2b92d43fb295157b6441be8cb3/src/llm-doc.ts#L27-L32)
Do not use `processFrontMatter` to only retrieve frontmatter from a file and not update the frontmatter. Please use [`MetadataCache.getFileCache`](https://docs.obsidian.md/Reference/TypeScript+API/MetadataCache/getFileCache) or [`parseYaml`](https://docs.obsidian.md/Reference/TypeScript+API/parseYaml) instead.

[const response = await fetch(`${settings.baseUrl}/v1/models`, {](https://github.com/shane-lamb/obsidian-llm-docs/blob/2e47abd77b620c2b92d43fb295157b6441be8cb3/src/open-ai.ts#L26)
Use the `requestUrl` function from the Obsidian API instead, it will handle some things like CORS automatically. Note that `requestUrl` does not support streaming.

["node-fetch": "^3.3.2"](https://github.com/shane-lamb/obsidian-llm-docs/blob/2e47abd77b620c2b92d43fb295157b6441be8cb3/package.json#L32)
Use [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) built in to Electron instead of adding a package to support streaming.

[const existingText = await this.app.vault.read(this.file)](https://github.com/shane-lamb/obsidian-llm-docs/blob/2e47abd77b620c2b92d43fb295157b6441be8cb3/src/llm-doc.ts#L59-L62)
[Prefer to use `Vault.process()` instead of `Vault.modify()` to modify a file in the background](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Prefer+%60Vault.process%60+instead+of+%60Vault.modify%60+to+modify+a+file+in+the+background)

[const linkText = this.app.metadataCache.fileToLinktext(file, '')](https://github.com/shane-lamb/obsidian-llm-docs/blob/2e47abd77b620c2b92d43fb295157b6441be8cb3/src/main.ts#L154)
Please provide the `sourcePath`, should be `file.path` in this case.

[new Setting(containerEl)](https://github.com/shane-lamb/obsidian-llm-docs/blob/2e47abd77b620c2b92d43fb295157b6441be8cb3/src/settings-tab/index.ts#L28-L39)
Prefer to use [AbstractInputSuggest](https://docs.obsidian.md/Reference/TypeScript+API/AbstractInputSuggest) when letting a user select a file/folder to add type-ahead support.

---

I highly recommend using the Editor interface when possible. Modifying the file as a string and using `vault.modify` might seem easier, but it has a number of downsides:
- Information like the user's cursor position, selections, and any folded content is lost.
- It's less efficient. Using the editor allows you to make short positional edits very quickly.

shane-lamb:
Thanks @Zachatoo for your useful feedback.

"What sets your plugin apart from existing plugins?"

The basic idea of LLM docs was to use plain markdown files as the chat interface (and chat history) for LLM chats. Most other plugins for chatting with LLMs use a separate chat window which is not built using the markdown editor.

I've had a look around and the only other plugin with very similar functionality seems to be [ChatGPT MD](https://github.com/bramses/chatgpt-md) - admittedly I didn't notice the existence of this plugin at the time I started development of LLM docs, and if I had, I may not have started development at all.

Having said that, here are some advantages that I could identify over the ChatGPT MD plugin:
- I've integrated a codemirror plugin which styles headings + shows a loading indicator + displays a button to submit the chat to the chat to the LLM, to make usage more intuitive - while ChatGPT doesn't seem to have any dynamic elements in the markdown document
- I've supported link-following to other documents for inclusion in chat context, and supported inline images as context

I'm currently working on tool use, to enable chatting with code repositories, for example.

Overall though ChatGPT MD does seem to be a pretty solid plugin, so I'm worried that it may not be worth the effort to address all of your points, if my submission will ultimately be rejected on the basis that it is too similar to an existing plugin. How much differentiation is needed in order to be accepted? Is it tolerable to have multiple similar plugins in the official list?

Thanks!

Zachatoo:
Your differences are substantial enough that your plugin will be accepted. I would recommend ensuring your README is clear in how it differs from other similar plugins to make it easier for those looking for AI plugins to choose which plugin they want to use.

joethei:
[const file: TFile = view.file!](https://github.com/shane-lamb/obsidian-llm-docs/blob/c10189b485f2e24be5312cd377f17e4ed21d8b84/src/main.ts#L39), [const file: TFile = view.file!](https://github.com/shane-lamb/obsidian-llm-docs/blob/c10189b485f2e24be5312cd377f17e4ed21d8b84/src/main.ts#L48)
`view.file` could be undefined, please don't just assume that it's there just because there is an editor.

shane-lamb:
Thanks @joethei, I've made the requested change.

PR #5858, opened 2025-03-30
Add plugin: About Blank

Ai-Jani:

## Repo URL

Link to my plugin: https://github.com/Ai-Jani/about-blank

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16777,6 +16777,13 @@
     "author": "Natalie Sumbo",
     "description": "Automatically downloads and embeds images from URLs when pasting",
     "repo": "sky150/obsidian-image-embedder"
-  }
+  },
+  {
+    "id": "about-blank",
+    "name": "About Blank",
+    "author": "Ai-Jani",
+    "description": "Customize the empty file (New tab) itself a little bit by adding \"Commands\" or \"Open files\". And edit these: Grouping, Set icon, Ask before execution, Register as a new command.",
+    "repo": "Ai-Jani/about-blank"
+}
 ]
 


Ai-Jani:
/skip

All the indicated places use `document.documentElement.style.setProperty` to change CSS variable values (or remove changes with `document.documentElement.style.removeProperty`). These CSS variables are plugin-specific and not existing CSS variables.

## Current UX

- In `styles.css`, plugin-specific CSS variables are defined in `:root` and used as values for properties of specific elements as follows:

```css
:root {
  --about-blank-empty-state-display: none;
  --about-blank-icon-text-gap: 10px;
}

.workspace-split.mod-root .empty-state-action:not(.about-blank-visible),
.workspace-split.mod-root .empty-state-title:not(.about-blank-visible) {
  display: var(--about-blank-empty-state-display);
}

.about-blank-button-container:has(.svg-icon) .about-blank-button {
  padding-left: var(--about-blank-icon-text-gap);
}
```

- This plugin rewrites only the values of these CSS variables using `document.documentElement.style.setProperty` based on settings to change the appearance.
- Since the code only changes the values of plugin-specific CSS variables, users can freely change the appearance through themes or snippets.

```css
.workspace-split.mod-root .empty-state-action:not(.about-blank-visible),
.workspace-split.mod-root .empty-state-title:not(.about-blank-visible) {
  display: block;
}

.about-blank-button-container:has(.svg-icon) .about-blank-button {
  padding-left: 12px;
}
```

## Alternatives

1. Toggling CSS classes
    - It is possible to express this using only `styles.css` without using `document.documentElement.style.setProperty` by switching classes applied to target elements according to settings.
    - Reasons for not adopting this approach:
        - This plugin uses `this.app.workspace.on("layout-change", ...)` to display additional actions when the "New tab" (empty file) view is displayed.
        - If CSS classes are changed when this view is newly generated and displayed, the following will happen each time the view is displayed:
            1. The view without the class applied will be displayed momentarily
            2. Immediately after, the class is applied and the appearance changes
        - This would result in very poor UX. To avoid this, the plugin currently rewrites CSS variables in advance.
2. Using the "Style Settings" plugin
    - Using "Style Settings" would eliminate the need for `document.documentElement.style.setProperty` in this plugin.
    - Reasons for not adopting this approach:
        - Preference to avoid increasing dependencies if possible.
        - Although the logic is slightly different, it's doing something similar to what this plugin is currently doing, so I've determined it's not worth adopting at the cost of increasing dependencies.
3. User-side CSS snippet approach
    - A method where users create snippets if they want to change the settings for the target areas. The method would be displayed in the README and settings UI.
    - If my argument is rejected and `document.documentElement.style.setProperty` cannot be used, this approach will be adopted.
    - However, this depends on users' knowledge of CSS, etc., so I would prefer to avoid it if possible.

joethei:
[.setName("Add actions to New tabs")](https://github.com/Ai-Jani/about-blank/blob/5e9cca0ac99dd5bde4328cce5176c49afa76173b/src/settings/settingTab.ts#L162)

Ai-Jani:
Thanks for your time.
The phrase was used to distinguish between " New tab", which refers to an empty file view, and the more general meaning of "new tab". However, I now use a clearer name and follow the "sentence case" rule.
And now I have released [1.1.0](https://github.com/Ai-Jani/about-blank/releases/tag/1.1.0).

PR #5890, opened 2025-04-02
Add plugin: Tick Tones

DontBlameMe99:

## Repo URL

Link to my plugin: https://github.com/DontBlameMe99/Tick-Tones

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16812,6 +16812,13 @@
     "description": "Reveal the current file in the file explorer while collapsing all other tree items.",
     "author": "d7sd6u",
     "repo": "d7sd6u/obsidian-reveal-folded"
+  },
+  {
+  "id": "tick-tones",
+  "name": "Tick Tones",
+  "author": "DontBlameMe",
+  "description": "Plays a tone when you tick a checkbox",
+  "repo": "DontBlameMe99/Tick-Tones"
   }
 ]
 
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16814,6 +16814,13 @@
     "repo": "d7sd6u/obsidian-reveal-folded"
   },
   {
+  "id": "data-fetcher",
+  "name": "Data Fetcher",
+  "author": "qf3l3k",
+  "description": "Fetch data from multiple sources (REST APIs, RPC, gRPC, GraphQL) and insert results into notes",
+  "repo": "qf3l3k/obsidian-api-fetcher"
+ },
+ {
   "id": "tick-tones",
   "name": "Tick Tones",
   "author": "DontBlameMe",


Zachatoo:
[if (!fs.existsSync(assetsPath)) {](https://github.com/DontBlameMe99/Tick-Tones/blob/bcfa75808f92988d82077f21aecc02b3b96a9b3d/soundLoader.ts#L25), [fs.readdirSync(assetsPath).forEach((file: string) => {](https://github.com/DontBlameMe99/Tick-Tones/blob/bcfa75808f92988d82077f21aecc02b3b96a9b3d/soundLoader.ts#L30), [const base64Data = `data:audio/wav;base64,${fs.readFileSync(filePath).toString("base64")}`;](https://github.com/DontBlameMe99/Tick-Tones/blob/bcfa75808f92988d82077f21aecc02b3b96a9b3d/soundLoader.ts#L34)
[Prefer to use `Vault.adapter` APIs instead of `fs` APIs](https://docs.obsidian.md/Reference/TypeScript+API/DataAdapter) to enable your plugin to work on mobile.

[return `${adapter.getBasePath()}/${this.app.vault.configDir}/plugins/${this.manifest.id}`;](https://github.com/DontBlameMe99/Tick-Tones/blob/bcfa75808f92988d82077f21aecc02b3b96a9b3d/soundLoader.ts#L53)
Prefer to use [`PluginManifest.dir`](https://docs.obsidian.md/Reference/TypeScript+API/PluginManifest/dir).

[if (adapter instanceof FileSystemAdapter) {](https://github.com/DontBlameMe99/Tick-Tones/blob/bcfa75808f92988d82077f21aecc02b3b96a9b3d/soundLoader.ts#L52)
Prefer not to use `FileSystemAdapter` APIs directly, as they only work on Desktop. If you use `Vault.adapter` APIs instead of `fs` APIs then you don't need to get the base path.

[const assetsPath = path.join(this.pluginPath, "assets");](https://github.com/DontBlameMe99/Tick-Tones/blob/bcfa75808f92988d82077f21aecc02b3b96a9b3d/soundLoader.ts#L21), [const filePath = path.join(assetsPath, file);](https://github.com/DontBlameMe99/Tick-Tones/blob/bcfa75808f92988d82077f21aecc02b3b96a9b3d/soundLoader.ts#L33)
Concatenate using `/` as a separator instead of using `path.join`, only `/` will work when using Obsidian APIs.

[.setName("Checkbox Tick Sound")](https://github.com/DontBlameMe99/Tick-Tones/blob/bcfa75808f92988d82077f21aecc02b3b96a9b3d/settings.ts#L33), [.setName("Test Selected Sound")](https://github.com/DontBlameMe99/Tick-Tones/blob/bcfa75808f92988d82077f21aecc02b3b96a9b3d/settings.ts#L46), [button.setButtonText("Play Sound");](https://github.com/DontBlameMe99/Tick-Tones/blob/bcfa75808f92988d82077f21aecc02b3b96a9b3d/settings.ts#L49)

---

To make the manual review process and future plugin maintenance more efficient, we recommend that you split your code into multiple files and move all source code files into a `src` directory.

I would not recommend committing your assets folder to GitHub, it gives a false impression that your plugin comes bundled with sounds but it does not, since [only the main.js, manifest.json, and styles.css (optional) files will be downloaded by Obsidian from your release](https://docs.obsidian.md/Plugins/Releasing/Submit+your+plugin#Step+2+Create+a+release).

Please update your README to be more clear on how to use this plugin. I don't expect most users to understand what the "plugin's assets folder" means in your plugin's settings tab. I think it would be wise to include step by step instructions for how to add a new sound to your README.

Consider removing all uses of `fs` and `path` so you can update your plugin manifest's `isDesktopOnly` to `false`.

PR #5891, opened 2025-04-02
Add plugin: Model Viewer

janispritzkau:

## Repo URL

Link to my plugin: https://github.com/janispritzkau/obsidian-model-viewer

## Release Checklist
  - [ ]  Windows
  - [x]  macOS (tested on Obsidian 1.7.6, November 12, 2024, set as minAppVersion)
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16828,11 +16828,11 @@
   "repo": "DontBlameMe99/Tick-Tones"
   },
   {
-  "id": "model-viewer",
-  "name": "Model Viewer",
-  "author": "Janis Pritzkau",
-  "description": "View and embed interactive 3D models directly in your vault, powered by Google's <model-viewer> component. Supports the glTF and GLB file formats.",
-  "repo": "janispritzkau/obsidian-model-viewer"
+    "id": "model-viewer",
+    "name": "Model Viewer",
+    "author": "Janis Pritzkau",
+    "description": "View and embed interactive 3D models directly in your vault, powered by Google's <model-viewer> component. Supports the glTF and GLB file formats.",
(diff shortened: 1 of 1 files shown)


janispritzkau:
I think the screenshot in the README might cause issues. I will look into it.

janispritzkau:
Do I need to include any screenshots used in the README in the GitHub release?

On another note, the included `@google/model-viewer` package uses lazy-loaded dependencies from a Google CDN when the model needs them, by default. This would leave me with two options: bundle it into the plugin, or leave it as is (maybe with a warning to the user). I tried bundling them (in the branch `bundle-external-deps`), but I could only get it to work by adding them directly to my repository. I would imagine this would cause problems with the review process.

janispritzkau:
@joethei I was able to find a solution for bundling the resources by inlining them as base64 and writing the assets to the plugin directory by using the `vault.adapter` API. Do you think this is acceptable?

[Here is the codegen script for that.](https://github.com/janispritzkau/obsidian-model-viewer/blob/bundle-external-deps/scripts/bundle.ts)

I tested it on MacOS and Android. No idea if this works on iOS too.

This blows up the main.js considerably, from 6 MB to 18 MB. I could gzip it (using DecompressionStream or other methods if available) if size is really a problem.

I don't think there is another way to do this without modifying Model Viewer or Three.js itself, to change the way resources are loaded.

janispritzkau:
I think I might be able to get rid of some files. Also, some of them already contain base64, so I'm essentially encoding stuff twice.

janispritzkau:
Thanks for merging! Excited to see where this project is going

PR #5909, opened 2025-04-04
Add plugin: LawList: Custom List Styles

willem-schlieter:

## Repo URL

Link to my plugin: https://github.com/willem-schlieter/lawlist

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16889,6 +16889,13 @@
     "author": "Luigi Cerone",
     "description": "Keep track of how much water you drunk",
     "repo": "LuigiCerone/obsidian-water-tracker"
-  }
+  },
+  {
+  "id": "lawlist",
+  "name": "LawList: Custom List Styles",
+  "author": "Willem Schlieter",
+  "description": "Adds support for custom list styles in Obsidian (Edit Mode and Read Mode).",
+  "repo": "willem-schlieter/lawlist"
+}
 ]
 
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16894,7 +16894,7 @@
   "id": "lawlist",
   "name": "LawList: Custom List Styles",
   "author": "Willem Schlieter",
-  "description": "Adds support for custom list styles in Obsidian (Edit Mode and Read Mode).",
+  "description": "Adds support for freely configurable, custom list styles (Edit Mode and Read Mode).",
   "repo": "willem-schlieter/lawlist"
 }
 ]


Zachatoo:
[new Setting(containerEl).setName("LawList: Custom List Styles").setHeading();](https://github.com/willem-schlieter/lawlist/blob/2a2d27d525fb6b721cbbae41cb0f3460883850a6/main.ts#L141)

[this.registerEditorExtension(lawListCMViewPlugin);](https://github.com/willem-schlieter/lawlist/blob/2a2d27d525fb6b721cbbae41cb0f3460883850a6/main.ts#L29)
I notice you're doing some workarounds to communicate with your editor extension. Here is a simpler way to accomplish that. In this case, I'm passing the plugin instance.
```js
this.registerEditorExtension(ViewPlugin.define((view) => new MyViewPlugin(view, this)));
```

Do not create stylesheets in your plugin without cleaning them up when your plugin is unloaded.

Please include in your README that your plugin does not support lists above a certain amount of list items. In my testing I couldn't go above 15 list items.

[color: #568;](https://github.com/willem-schlieter/lawlist/blob/2a2d27d525fb6b721cbbae41cb0f3460883850a6/styles.css#L2)
Why specifically this color? If you're intending to match the list item color in reading mode, then prefer to use Obsidian's CSS variable for [`list-marker-color`](https://docs.obsidian.md/Reference/CSS+variables/Editor/List) to match the user's theme.

---

To make the manual review process and future plugin maintenance more efficient, we recommend that you move all source code files into a `src` directory.

Have you considered using [@counter-style](https://developer.mozilla.org/en-US/docs/Web/CSS/@counter-style) or [counters](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_counter_styles/Using_CSS_counters) instead of styling individual list items? It should make it much easier to manage counters, you could have a single stylesheet and only modify it when the user changes their settings.

willem-schlieter:
Hi, thank you so much for your suggestions! I just completed them all, merged the branch that I worked in during the last days into master and updated the release files.

Thanks especially for the idea with `@counter-styles`! It makes things so much easier. I also managed to completely remove the limit to 15 items. Both have now resulted in a major reorganisation of my code...

So what are the next steps?

PR #5950, opened 2025-04-07
Add plugin: Inline Callouts

gapmiss:

## Repo URL

Link to my plugin: https://github.com/gapmiss/inline-callouts

## Release Checklist
  - [ ]  Windows
  - [X]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16924,6 +16924,13 @@
   "author": "natri0",
   "description": "Adds diagrams that show how structures are laid out in memory / network.",
   "repo": "natri0/obsidian-bytefield"
+},
+{
+  "id": "inline-callouts",
+  "name": "Inline Callouts",
+  "author": "@gapmiss",
+  "description": "Add inline callouts/badges/icons to notes.",
+  "repo": "gapmiss/inline-callouts"
 }
 ]
 


joethei:
Why did you create a new plugin instead of updating the old one?

[Copyright (c) 2023 gapmiss](https://github.com/gapmiss/inline-callouts/blob/0fd72c2ccb3c7a9d91698a73f4e4f3d583a57327/LICENSE#L3)
Wrong year.

gapmiss:
> Why did you create a new plugin instead of updating the old one?

I changed to a different, simpler syntax and I didn't want to break user's already established badges. I understand I could have probably made it backwards compatible, but I assume users would rather just switch to a new-and-improved plugin.

Out of curiosity, how would you approach something like that (upgrading and making it backwards compatible)? Seems like a difficult task to accomplish and also maintain.

> [Copyright (c) 2023 gapmiss](https://github.com/gapmiss/inline-callouts/blob/0fd72c2ccb3c7a9d91698a73f4e4f3d583a57327/LICENSE#L3) Wrong year.

Fixed

PR #6038, opened 2025-04-14
Add plugin: JIRA links shortener

rplatonovs:

## Repo URL

Link to my plugin: https://github.com/rplatonovs/obsidian-jira-links-shortener

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -17030,12 +17030,19 @@
     "description": "Format files with Prettier using built-in settings for configuration.",
     "repo": "dylanarmstrong/obsidian-prettier-plugin"
   },
-    {
-        "id": "filename-styler",
-        "name": "FileName Styler",
-        "author": "Marc Feininger",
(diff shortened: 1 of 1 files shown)


joethei:
[const jiraRegex = /^http.*\/browse\/([A-Z]+-\d+)/g;](https://github.com/rplatonovs/obsidian-jira-links-shortener/blob/97912e0a4c067aabe207a88e3a7aacd756c7e804/main.ts#L12)
This regex is a bit to generic, and could trigger with other URL's
It would be better if you let users configure their JIRA instance URL.

rplatonovs:
@joethei fair enough, I added the configuration you suggested. I also identified and fixed the case were query string can be lost from the URL. A few unit tests added for the core function that does the transformation of a pasted text.

PR #6047, opened 2025-04-15
Add plugin: External Links Icon

moziar:

## Repo URL
Link to my plugin: https://github.com/moziar/obsidian-external-links-icon

## Release Checklist
  - [ ] Windows
  - [x] macOS
  - [ ] Linux
  - [ ] Android (if applicable)
  - [x] iOS (if applicable)
  - [x] `main.js`
  - [x] `manifest.json`
- [x] GitHub release name matches the exact version number specified in my manifest.json (Note: Use the exact version number, don't include a prefix `v`)

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16511,6 +16511,13 @@
     "author": "qf3l3k",
     "description": "Fetch data from multiple sources (REST APIs, RPC, gRPC, GraphQL) and insert results into notes",
     "repo": "qf3l3k/obsidian-api-fetcher"
-  }
+  },
+{
+    "id": "external-links-icon",
+    "name": "External Links Icon",
+    "author": "Moz",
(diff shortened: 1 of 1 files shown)


moziar:
"Link Favicons" is a powerful plugin, while "External Links Icon" is simpler and easier to use.
- Speed is the primary feature. Since all the icon files are locally stored and optimized, you hardly notice the loading process, even if there are over a thousand links in a single note. 
- I focus on out-of-the-box usability, so you can enjoy link icons right after installing the plugin without any configuration, which is particularly convenient in _URL Scheme_ scenarios.
- However, at the same time, local storage and lack of customization also mean that every time you need to adapt to a new software, you need to update it, which is also the reason for using a plugin (relying on the plugin market for updates) rather than using CSS Snippets.

The inappropriate title has been modified.

PR #6050, opened 2025-04-15
Add plugin: Log Keeper

JimJamBimBam:

## Repo URL

Link to my plugin: https://github.com/JimJamBimBam/obsidian-log-keeper/tree/1.3.0

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -17477,5 +17477,12 @@
     "author": "Shane Lamb",
     "description": "Simple LLM (AI chat) client using pure markdown documents",
     "repo": "shane-lamb/obsidian-llm-docs"
+  },
+  {
+    "id": "log-keeper",
+    "name": "Log Keeper",
+    "author": "James Sonneveld",
+    "description": "Generate time stamps automatically as changes are made to a note.",
+    "repo": "JimJamBimBam/obsidian-log-keeper"
   }
 ]


Zachatoo:
[:disabled {](https://github.com/JimJamBimBam/obsidian-log-keeper/blob/23f45e0d7765ed404d535d755d1fd96f1b6ff1d4/styles.css#L1-L5)

[console.log(\`Activating: ${this.manifest.name}\n](https://github.com/JimJamBimBam/obsidian-log-keeper/blob/23f45e0d7765ed404d535d755d1fd96f1b6ff1d4/src/main.ts#L19-L23)

[Toggle between tracking modifications made to notes per day or by the interval set by Update Interval.](https://github.com/JimJamBimBam/obsidian-log-keeper/blob/23f45e0d7765ed404d535d755d1fd96f1b6ff1d4/src/settings.ts#L57), [Setting this to true will also disable Update Interval.](https://github.com/JimJamBimBam/obsidian-log-keeper/blob/23f45e0d7765ed404d535d755d1fd96f1b6ff1d4/src/settings.ts#L59)

---

Consider updating your classes in your styles.css file to be more specific to your plugin, other plugins may have the same classes and your styles will conflict.

Your esbuild.config.mjs file looks incorrect, your main.ts file is in a src folder, but your esbuild config is pointing to a main.ts file in the root of your repo.

Consider extending [`FuzzySuggestModal`](https://docs.obsidian.md/Reference/TypeScript+API/FuzzySuggestModal) instead of `SuggestModal` and calling `super.renderSuggestion` in your `renderSuggestion` method to get highlighting of matching keywords built in, instead of implementing it yourself.

PR #6115, opened 2025-04-21
Add plugin: Exchange Rate

KarasuShin:

## Repo URL

Link to my plugin: https://github.com/KarasuShin/obsidian-exchangerate

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16511,6 +16511,13 @@
     "author": "qf3l3k",
     "description": "Fetch data from multiple sources (REST APIs, RPC, gRPC, GraphQL) and insert results into notes",
     "repo": "qf3l3k/obsidian-api-fetcher"
-  }
+  },
+{
+    "id": "exchange-rate",
+    "name": "Exchange Rate",
+    "author": "Karasu",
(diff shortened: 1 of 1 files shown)


Zachatoo:
["author": "Karasu <karasu.shing@gmail.com>",](https://github.com/KarasuShin/obsidian-exchangerate/blob/88bad743510d21dbb14e74ff75c33fed26e4e283/manifest.json#L7)
This field is displayed in the sidebar and in the description for the plugin details page. Are you sure you want to have your email listed so prominently?
image

Please remove the styles.css file from your repo if you aren't using it.

joethei:
The current functionality can easily be replicated inside of a DV/Templater script without this plugin, i'd recommend adding some UI features, like a codeblock that one can insert into a note to see the current exchange rate/converted currency values.

It would be better if you would use the bundled moment library instead of dayjs, as it means your bundle size is smaller and the plugin will be faster to load.

PR #6264, opened 2025-05-04
Add plugin: Ace Code Editor

RavenHogWarts:

## Repo URL

Link to my plugin: [https://github.com/RavenHogWarts/obsidian-ace-code-editor](https://github.com/RavenHogWarts/obsidian-ace-code-editor)

## Release Checklist

- [x]  Windows
- [ ]  macOS
- [ ]  Linux
- [ ]  Android *(if applicable)*
- [ ]  iOS *(if applicable)*
- [x] `main.js`
- [x] `manifest.json`
- [x] `styles.css` *(optional)*
- [x] GitHub release name matches the exact version number specified in my manifest.json (***Note:** Use the exact version number, don't include a prefix `v`*)
- [x] I have read the developer policies at [https://docs.obsidian.md/Developer+policies](https://docs.obsidian.md/Developer+policies), and have assessed my plugins's adherence to these policies.
- [x] I have read the tips in [https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines) and have self-reviewed my plugin to avoid these common pitfalls.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16826,6 +16826,13 @@
     "author": "Adam Fletcher",
     "description": "Adds toggleable colored highlights to sentences based on their length so you can easily see the rhythm of your writing.",
     "repo": "adamfletcher/obsidian-sentence-rhythm"
-}
+},
+    {
+        "id": "ace-code-editor",
+        "name": "Ace Code Editor",
+        "author": "RavenHogWarts",
+        "description": "An enhanced code editor using Ace editor, providing syntax highlighting, code folding, and other advanced editing features.",
+        "repo": "RavenHogWarts/obsidian-ace-code-editor"
+    }
 ]
 


Zachatoo:
[if (this.settings.snippetsManager.location === "Ribbon") {](https://github.com/RavenHogWarts/obsidian-ace-code-editor/blob/0dd21cd7ad94198a1f33125f09ba1b8aac8d6338/src/main.ts#L103-L111)
In your case, you should always add it to the ribbon, and a user can reorder or remove it if they wish in Obsidian's ribbon settings.

[.rht-toolkit-modal {](https://github.com/RavenHogWarts/obsidian-ace-code-editor/blob/0dd21cd7ad94198a1f33125f09ba1b8aac8d6338/style/BaseModal.css#L72-L84)
This style doesn't seem relevant to your plugin.

["minAppVersion": "1.6.7",](https://github.com/RavenHogWarts/obsidian-ace-code-editor/blob/0dd21cd7ad94198a1f33125f09ba1b8aac8d6338/manifest.json#L5)

[const existingFile = app.vault.getAbstractFileByPath(fullPath);](https://github.com/RavenHogWarts/obsidian-ace-code-editor/blob/0dd21cd7ad94198a1f33125f09ba1b8aac8d6338/src/components/Modal/CreateCodeFileModal.tsx#L80-L83)
When using [`Vault.getAbstractFileByPath`](https://docs.obsidian.md/Reference/TypeScript+API/Vault/getAbstractFileByPath), the result will be either a `TFile`, `TFolder`, or `null`. Your current code does not account for all of these scenarios. If you are only expecting a `TFile`, prefer to use [`Vault.getFileByPath`](https://docs.obsidian.md/Reference/TypeScript+API/Vault/getFileByPath). If you are only expecting a `TFolder`, prefer to use [`Vault.getFolderByPath`](https://docs.obsidian.md/Reference/TypeScript+API/Vault/getFolderByPath).

[const normalizedFolderPath = folderPath.endsWith("/")](https://github.com/RavenHogWarts/obsidian-ace-code-editor/blob/0dd21cd7ad94198a1f33125f09ba1b8aac8d6338/src/components/Modal/CreateCodeFileModal.tsx#L50-L57)
[Use `normalizePath` to clean up user defined paths](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Use+%60normalizePath()%60+to+clean+up+user-defined+paths)

[const content = await app.vault.read(file);](https://github.com/RavenHogWarts/obsidian-ace-code-editor/blob/0dd21cd7ad94198a1f33125f09ba1b8aac8d6338/src/core/utils/CodeBlock.ts#L140-L167)
[Prefer to use `Vault.process()` instead of `Vault.modify()` to modify a file in the background](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Prefer+%60Vault.process%60+instead+of+%60Vault.modify%60+to+modify+a+file+in+the+background)
You could also consider using the Editor APIs instead since you're always editing the active file. You can get the Editor instance by passing it from the `this.app.workspace.on("editor-menu")` event, or from from `activeView.editor` (though it's possible it could be a different editor at that point).

[\<h2>Edit Code Block<\/h2>](https://github.com/RavenHogWarts/obsidian-ace-code-editor/blob/0dd21cd7ad94198a1f33125f09ba1b8aac8d6338/src/components/Modal/EditCodeBlockModal.tsx#L56)

[<button](https://github.com/RavenHogWarts/obsidian-ace-code-editor/blob/0dd21cd7ad94198a1f33125f09ba1b8aac8d6338/src/components/Settings/SettingsTab.tsx#L44-L49)
Please do not provide a button to reload your plugin. All plugin settings should apply automatically, without having to reload your plugin.

Your plugin downloads external resources from a cdn when loading an Ace Editor. Please consider bundling those resources into your plugin so your plugin can work offline. If that is not possible, please disclose in your README that your plugin requires an internet connection to download resources for your editor to function.

RavenHogWarts:
@Zachatoo Thank you for your review. I have addressed all the points you raised and made the necessary changes. Would you please review the updates?

joethei:
[const content = await app.vault.process(file, (data) => {](https://github.com/RavenHogWarts/obsidian-ace-code-editor/blob/492b16a81f7145c92107e2e411d804ac7c1d7867/src/core/utils/CodeBlock.ts#L140-L171)
The data processing needs to be wrapped inside of the `process` callback.
Your current approach can still lead to conflicts with other plugins editing the file, which the `process` function is meant to avoid.

["https://cdn.jsdelivr.net/npm/ace-builds@1.40.1/src-min-noconflict/";](https://github.com/RavenHogWarts/obsidian-ace-code-editor/blob/492b16a81f7145c92107e2e411d804ac7c1d7867/src/core/services/AceLanguages.ts#L455)
This needs to be bundled into the plugin.

PR #6294, opened 2025-05-06
Add Atlassian Jira Auto-Linker plugin

Kovah:

## Repo URL

Link to my plugin: https://github.com/Kovah/obsidian-jira-auto-linker

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux (not available)
  - [ ]  Android _(if applicable)_ (not available)
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -16889,6 +16889,13 @@
     "author": "Luigi Cerone",
     "description": "Keep track of how much water you drunk",
     "repo": "LuigiCerone/obsidian-water-tracker"
+  },
+  {
+    "id": "atlassian-jira-auto-linker",
+    "name": "Atlassian Jira Auto-Linker",
+    "author": "Kevin Woblick",
+    "description": "Automatically create links to Jira from issue IDs like APP-1426.",
+    "repo": "Kovah/obsidian-jira-auto-linker"  
   }
 ]
 


Kovah:
/skip Not applicable as the code only adds static text to the settings page.

joethei:
The first call to `innerHTML` is not related to the static text on the settings page, please replace it with a safer alternative.

Kovah:
What would be a suitable, safer way to add HTML to the rendered view after the Markdown processing? 

The Editor Extensions docs specifically say 
> If you want to change how to convert Markdown to HTML in the Reading view, consider building a [Markdown post processor](https://docs.obsidian.md/Plugins/Editor/Markdown+post+processing).

Those docs do only give a text example but doesn't say anything about HTML. The linked docs about HTML elements are also not helpful, as the `createEl()` doesn't allow specifying any HTML, but only plain text. As far as I can see, there is no other way to directly replace text with HTML within nodes without using innerHTML.

Kovah:
I am still waiting for feedback on my question.

Kovah:
I am still waiting for feedback on my question.

xamgore:
As I see it, you can't do `.innerHTML = user-provided-input`. Instead, you call `el.createEl('a', { ... } })` and `el.createTextNode(..)` to construct a DOM node, as well as `el.appendText()` and `el.appendChild`. That is not possible at the present code, as you accumulate an HTML string (`modifiedText`).

```ts
	private replaceWithLinks(text: string): string {
		return this.settings.registrations.reduce((modifiedText, registration) => {
			const regex = new RegExp(`\\b(${registration.projectKey}-\\d+)\\b`, 'gi');
			return modifiedText.replace(
				regex,
				(match) => 
					`<a href="${registration.baseUrl}/browse/${match}" target="_blank" rel="noopener noreferrer">
								${match}
					</a>`
			);
		}, text);
	}
```

Instead, you should build a single `RegExp`, a union of all project keys like `\\b(KEY1|KEY2|KEY3)-\\d+)\\b`. And escape them before passing to `new RegExp(...)`. Escaping can be done via https://github.com/tc39/proposal-regex-escaping or https://www.npmjs.com/package/regexp.escape. With this regular expression you navigate the text node. I believe ChatGPT or Claude Opus is able to one-shot this.

Kovah:
/skip `innerHTML` is just used for one static link in the foot, which would require 4 times the code if done via the requested method. I don't see any issue with that.

PR #6301, opened 2025-05-07
Add plugin: deepseek-ai-assistant

mali-i:

## Repo URL

Link to my plugin: https://github.com/mali-i/obsidian-Deepseek-ai-assistant-plugin

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -17899,8 +17899,8 @@
     "repo": "goderyu/obsidian-blueprint-renderer"
   },
   {
-    "id": "Deepseek-ai-assistant",
-    "name": "Deepseek-ai-assistant",
+    "id": "deepseek-ai-assistant",
+    "name": "deepseek-ai-assistant",
     "author": "mali-i",
     "description": "Help you to study with ai-prompting",
     "repo": "mali-i/obsidian-Deepseek-ai-assistant-plugin"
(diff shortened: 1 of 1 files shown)


Zachatoo:
Can you double check that you've made all the requested changes? I found at least one change that wasn't made for using sentence case in UI, I stopped checking once I found one.

mali-i:
Sorry, I’ll double-check the other issues. 

Just to avoid self-doubt and overthinking, I’d like to confirm—when you mentioned the 'sentence case' issue, were you suggesting that the word 'key' in 'API key' should also be capitalized (i.e., 'API Key')?

Zachatoo:
The opposite. It should be `API key`, not `API Key`. Only the first letter of the first word in a sentence should be capitalized, except for acronyms, they can be all caps.

mali-i:
I've double-checked the changes. Please review the latest code in my plugin repo. Thanks!

mali-i:
oh. OK, thanks.

Zachatoo:
Can you please make a new release with your latest changes? Should be good to go after that.

mali-i:
Oh right, I forgot. The latest release version has been updated! Thanks.

Zachatoo:
["minAppVersion": "1.1.0",](https://github.com/mali-i/deepseek-ai-assistant/blob/8d138aca21178d00554e34ee35aed6ac872d682c/manifest.json#L5)
Keep in mind that this is the minimum version of Obsidian that your plugin can run on, it is not related to your plugin versions. I'm calling that out since you updated it at the same time you updated your version, you only need to update this when you use newer APIs that weren't available on older versions of Obsidian.

PR #6311, opened 2025-05-07
Add plugin: Better Mind Map

linem-davton:

## Repo URL

Link to my plugin: https://github.com/linem-davton/obsidian-better-mindmap

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -17883,5 +17883,12 @@
     "author": "Rerurate_514",
     "description": "Select the Arknights image and place it in the banner property.",
     "repo": "Rerurate514/insert_ArknightsURL_banners"
+  },
+  {
+    "id": "better-mindmap",
+    "name": "Better Mind Map",
+    "author": "Utkarsh Raj",
+    "description": "Visualize notes as interactive mind maps",
+    "repo": "linem-davton/obsidian-better-mindmap"
   }
 ]


linem-davton:
/skip

The issues flagged ([1][2]) refers to the `handleRanksepChange` (and similar) function(s) where JavaScript is used to update a CSS Custom Property, for example: `document.documentElement.style.setProperty("--mindmap-ranksep-multiplier", String(newValue));`

This approach is intentionally used to allow user-configurable settings (like layout spacing) to dynamically influence the plugin's appearance. The CSS Custom Property (`--mindmap-ranksep-multiplier`) is defined and then consumed by the plugin's stylesheet.

This method doesn't hardcode specific styles directly onto elements. Instead, it updates a variable that the CSS itself uses, making it adaptable. Themes or CSS snippets can still override or utilize this CSS variable if they wish, or even set its initial/default value. This is a standard pattern for bridging user-controlled dynamic values from JavaScript into a CSS-driven styling system. The actual styling rules that consume this variable are located in the CSS files.

Therefore, I believe this usage is appropriate for providing dynamic, user-controlled theming and layout adjustments, and aligns with best practices for making such values accessible to CSS.

Zachatoo:
[name: "Open Mind‑Map",](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/main.tsx#L11), [name: "Reset Mind Map View",](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/main.tsx#L22), ["Mind-Map"](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/view/ObsidianMindView.tsx#L31), ["Mind Map"](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/view/ObsidianMindView.tsx#L153), ["Horizontal Layout" : "Vertical Layout"](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/ui/canvas/MindCanvas.tsx#L217), ["Expand All" : "Collapse All"](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/ui/canvas/MindCanvas.tsx#L219), [Rank Sep](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/ui/canvas/MindCanvas.tsx#L291), [`Rank Separation Multiplier (Default: ${DEFAULT_RANKSEP_MULTIPLIER})`](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/ui/canvas/MindCanvas.tsx#L310), [Node Sep](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/ui/canvas/MindCanvas.tsx#L316), [`Node Separation Multiplier (Default: ${DEFAULT_NODESEP_MULTIPLIER})`](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/ui/canvas/MindCanvas.tsx#L336)

[console.log("Mindmap view closed");](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/view/ObsidianMindView.tsx#L109)

[.react-flow__edge-path {](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/style/extra.css#L18-L79)
Please increase the specificity of these selectors by adding another class that is specific to your plugin (for example, `.mindmap-container`)

[<ReactMarkdown](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/ui/canvas/toReactFlow.tsx#L71-L90)
Obsidian provides a built in mechanism to render Markdown using [`MarkdownRenderer.render`](https://docs.obsidian.md/Reference/TypeScript+API/MarkdownRenderer/render), that will also handle clicking links, rendering math, etc. Please use this instead of using a different library unless you have a compelling use case to use something else.

[const headingMatch = line.match(/^(#{1,6})\s+(.\*)$/);](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/parser/parseOutline.ts#L23), [const bulletMatch = line.match(/^(\s*)[-*+]\s+(.\*)$/);](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/parser/parseOutline.ts#L37)
This is going to return false positives, for example, a line in a codeblock that starts with a `#` or `-`. Please use [`MetadataCache.getFileCache`](https://docs.obsidian.md/Reference/TypeScript+API/MetadataCache/getFileCache) instead. This method will also provide you with positional information about the headings and list items.
```ts
this.app.metadataCache.getFileCache(file)?.headings;
this.app.metadataCache.getFileCache(file)?.listItems;
```

[const sourcePath = this.currentLeaf.view.getState()?.file ?? "";](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/view/ObsidianMindView.tsx#L176)
You have not implemented the `setState` method so `getState?.file` will always return null. I suspect you won't need this method after making the other changes I've requested, but if this is still needed, you'll need to implement the [`getState`](https://docs.obsidian.md/Reference/TypeScript+API/View/getState) and [`setState`](https://docs.obsidian.md/Reference/TypeScript+API/View/setState) methods.

---

[style: {](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/ui/canvas/MindCanvas.tsx#L187-L191), [style: {}, // clear the override](https://github.com/linem-davton/obsidian-better-mindmap/blob/23dad6e6cb0c32575969fd821705ac6f8475387c/src/ui/canvas/MindCanvas.tsx#L204)
If possible, use classes instead of styles.

Zachatoo:
[MarkdownRenderer.render(app, markdown, container, sourcePath, {});](https://github.com/linem-davton/obsidian-better-mindmap/blob/c02729355b460ffd9c3ce80dac71d9e7d7941ddf/src/ui/canvas/toReactFlow.tsx#L29)
You need to pass a class that extends [`Component`](https://docs.obsidian.md/Reference/TypeScript+API/Component) to avoid memory leaks. For your case, you can create a new `Component` and call `unload()` on it when your `ObsidianMarkdownRenderer` React component is unmounted.

```tsx
useEffect(() => {
    const component = new Component();
    // Existing code...
    // Make sure to pass component to render function
    return () => {
        component.unload();
    }
}, [/* ... */]);
```

---

I expect this will be the final feedback I have before merging.

linem-davton:
@Zachatoo 

Since the Validation job failed, I have updated the top comment to reflect what the validation needs. Please re-trigger the validation workflow.

linem-davton:
/skip

The issues flagged ([1][2]) refer to the handleRanksepChange (and similar) function(s) where JavaScript is used to update a CSS Custom Property, for example: document.documentElement.style.setProperty("--mindmap-ranksep-multiplier", String(newValue));

This approach is intentionally used to allow user-configurable settings (like layout spacing) to dynamically influence the plugin's appearance. The CSS Custom Property (--mindmap-ranksep-multiplier) is defined and then consumed by the plugin's stylesheet.

This method doesn't hardcode specific styles directly onto elements. Instead, it updates a variable that the CSS itself uses, making it adaptable. Themes or CSS snippets can still override or utilize this CSS variable if they wish, or even set its initial/default value. This is a standard pattern for bridging user-controlled dynamic values from JavaScript into a CSS-driven styling system. The actual styling rules that consume this variable are located in the CSS files.

Skipping `innerHTML` check because it is used here only to clear the DOM element before it is safely repopulated by Obsidian's MarkdownRenderer API. No external or user-generated HTML is being injected,

joethei:
["description": "Mind‑map plugin to preview notes",](https://github.com/linem-davton/obsidian-better-mindmap/blob/d22a3a77e9ad38868615be4c5c41d3c1f9560ee9/manifest.json#L7)
Avoid including sentences like "xyz plugin that does ..." in your description, it should be a given since this is only shown when browsing plugins through the app or on our website.

linem-davton:
Fixed new description: [Visualize notes as interactive mind maps](https://github.com/linem-davton/obsidian-better-mindmap/blob/a92827a997c1f67109e42cb19cbc8ddca52ea29f/manifest.json#L7)

PR #6327, opened 2025-05-09
Add plugin: Tree Diagram

limpido:

## Repo URL

Link to my plugin: https://github.com/limpido/obsidian-tree-diagram

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -17512,5 +17512,12 @@
     "author": "Nitish Khurana",
     "description": "Moves all the files linked to a open md file to a folder and updates the link in md file.",
     "repo": "nitishkhurana/obsidian-move-files-plugin"
+  },
+  {
+    "id": "tree-diagram",
+    "name": "Tree Diagram",
+    "author": "limpido",
+    "description": "Convert directory structure formatted in tab-indented text to a tree diagram.",
+    "repo": "limpido/obsidian-tree-diagram"
   }
 ]


Zachatoo:
[# Obsidian Tree Diagram](https://github.com/limpido/obsidian-tree-diagram/blob/c78177a5e460491b85b8b92ebd8949c19134b1e2/README.md#L1)

[console.log("tree load");](https://github.com/limpido/obsidian-tree-diagram/blob/c78177a5e460491b85b8b92ebd8949c19134b1e2/src/main.ts#L8), [console.log("tree unload");](https://github.com/limpido/obsidian-tree-diagram/blob/c78177a5e460491b85b8b92ebd8949c19134b1e2/src/main.ts#L27)

[this.component = new Component();](https://github.com/limpido/obsidian-tree-diagram/blob/c78177a5e460491b85b8b92ebd8949c19134b1e2/src/main.ts#L9)
Creating a component like this is going to lead to memory leaks, since you aren't managing the component's lifecycle. You should create a new class that extends the [`MarkdownRenderChild`](https://docs.obsidian.md/Reference/TypeScript+API/MarkdownRenderChild) class and call `ctx.addChild(...)` to have Obsidian manage your component lifecycle for you. If it's helpful, you can refer to [this example](https://github.com/Zachatoo/obsidian-rerender-markdown-codeblock-processor-example).

PR #6717, opened 2025-06-12
Update community-plugins.json

fredoliveira:

## Repo URL

Link to my plugin: https://github.com/fredoliveira/obsidian-typechecker/

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -17358,5 +17358,12 @@
     "author": "ZigHolding",
     "description": "Sync notes or plugins between vaults.",
     "repo": "zigholding/obsidian-notesync-plugin"
+  },
+  {
+    "id": "property-typechecker",
+    "name": "Property Type Checker",
+    "author": "fredoliveira",
+    "description": "Property type checker makes sure your frontmatter properties are of the correct type. It adds a sidebar view that lists issues in your open files as well as the rest of the vault.",
+    "repo": "fredoliveira/obsidian-typechecker"
   }
 ]


Zachatoo:
["version": "0.9.3",](https://github.com/fredoliveira/obsidian-typechecker/blob/7962543ac1529902e775bad90111b2bbef9c2e14/manifest.json#L4)
There is not a matching release with this tag. Obsidian looks at this to determine which release to download. Please ensure there is always a release with a matching tag so your plugin can be downloaded when it's added to community plugins.

[# Obsidian Property Type Checker](https://github.com/fredoliveira/obsidian-typechecker/blob/7962543ac1529902e775bad90111b2bbef9c2e14/README.md#L1)

[.setting-item-description {](https://github.com/fredoliveira/obsidian-typechecker/blob/7962543ac1529902e775bad90111b2bbef9c2e14/styles.css#L118-L123)

[this.app.workspace.on("active-leaf-change", async () => {](https://github.com/fredoliveira/obsidian-typechecker/blob/7962543ac1529902e775bad90111b2bbef9c2e14/src/main.ts#L51), [this.app.workspace.on("active-leaf-change", async () => {](https://github.com/fredoliveira/obsidian-typechecker/blob/7962543ac1529902e775bad90111b2bbef9c2e14/src/main.ts#L60)
[`on('file-open')`](https://docs.obsidian.md/Reference/TypeScript+API/Workspace/on('file-open')) would be the better event to listen to.

[if (this.settings.enableAutoCheck) {](https://github.com/fredoliveira/obsidian-typechecker/blob/7962543ac1529902e775bad90111b2bbef9c2e14/src/main.ts#L49-L63)
Why have this twice? If the setting is true, then your `updateViewCurrentFile` is going to run twice on every active leaf change.
Please only register this at most once.

[containerEl.createEl("h2", { text: "Type Checker Settings" });](https://github.com/fredoliveira/obsidian-typechecker/blob/7962543ac1529902e775bad90111b2bbef9c2e14/src/TypeCheckerSettingsTab.ts#L17)

[containerEl.createEl("h3", {](https://github.com/fredoliveira/obsidian-typechecker/blob/7962543ac1529902e775bad90111b2bbef9c2e14/src/TypeCheckerSettingsTab.ts#L31-L34)
```ts
```

---

[private updateTimeout: NodeJS.Timeout | null = null;](https://github.com/fredoliveira/obsidian-typechecker/blob/7962543ac1529902e775bad90111b2bbef9c2e14/src/main.ts#L20)
Use a regular number here for the type, and use `window.setTimeout` / `window.clearTimeout` where this being used.

[value.every((v) => typeof v === "string" && v.startsWith("#"))](https://github.com/fredoliveira/obsidian-typechecker/blob/7962543ac1529902e775bad90111b2bbef9c2e14/src/main.ts#L249)
Nit: tags in frontmatter don't have to start with a `#`, it's optional.

PR #6818, opened 2025-06-20
Add plugin: Vault Stats

blueheron786:

## Repo URL

Link to my plugin: https://github.com/blueheron786/obsidian-stats-plugin/

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`


blueheron786:
@Zachatoo @joethei I've been trying to get this merged in for a month (one of two plugins), any pointers on where I'm going wrong?

Zachatoo:
["author": "Obsidian",](https://github.com/blueheron786/obsidian-stats-plugin/blob/d805d47c59b71886f17a231d7f49d6704f29b7bd/manifest.json#L7)
Please update this to your name or username.

[const content = await this.app.vault.read(file);](https://github.com/blueheron786/obsidian-stats-plugin/blob/d805d47c59b71886f17a231d7f49d6704f29b7bd/main.ts#L20)
[Prefer to use `Vault.cachedRead` over `Vault.read` when you only want to read a file and not modify it](https://docs.obsidian.md/Plugins/Vault#Read+files)

---

You have to loop over every file in the vault 3 times in your `showStats` method, consider only looping once for better performance.

Zachatoo (approved):

joethei:
https://github.com/blueheron786/obsidian-stats-plugin/blob/f13125d597cb27e06fa8690676830205e1677b67/manifest.json#L5
There is no 1.18.0 of Obsidian.
The app checks the minAppVersion, and if that is equal or lower than the currently installed version it uses the `version`, if it's not then it checks `versions.json`, which has the same problem in your case.

blueheron786:
Thank you for the prompt feedback. Seems like something we can add to the automated checks. I'll target 1.8.3 instead.

joethei:
You don't need to change anything on this end, you just need to modify the manifest.json and versions.json in your repo.

blueheron786:
OK, sounds good, thanks for all the info. I updated both files and I still see the problem after restarting Obsidian, so maybe it's an issue with caching somewhere. I'll wait and see what happens in a couple of hours.

blueheron786:
I deleted versions.json since I only have one version, and updated the manifest.json version to 1.18.10, which is what I have installed. Same error. 😔

PR #6842, opened 2025-06-22
Add plugin: Pretty Properties

anareaty:

## Repo URL

Link to my plugin: https://github.com/anareaty/pretty-properties

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -17533,5 +17533,12 @@
     "author": "Christ Degni",
     "description": "Autocomplete text to type more efficiently.",
     "repo": "c-degni/text-autocomplete"
+  }, 
+  {
+  "id": "pretty-properties",
+  "name": "Pretty Properties",
+  "author": "Anareaty",
+  "description": "Makes note properties look more fun: adds side image, banners, list property colors and allows to hide specific properties.",
+  "repo": "anareaty/pretty-properties"
   }
 ]


joethei:
[containerEl.createEl("h3", {text: i18n.t("BANNERS")})](https://github.com/anareaty/pretty-properties/blob/8a3e673d7a5ed405936e41af2f311cda7191257a/src/settings.ts#L73), [containerEl.createEl("h3", {text: i18n.t("COVERS")})](https://github.com/anareaty/pretty-properties/blob/8a3e673d7a5ed405936e41af2f311cda7191257a/src/settings.ts#L166), [ontainerEl.createEl("h3", {text: i18n.t("TASKS")})](https://github.com/anareaty/pretty-properties/blob/8a3e673d7a5ed405936e41af2f311cda7191257a/src/settings.ts#L301), [containerEl.createEl("h3", {text: i18n.t("OTHER")})](https://github.com/anareaty/pretty-properties/blob/8a3e673d7a5ed405936e41af2f311cda7191257a/src/settings.ts#L386)
```ts
```

joethei:
In the styles.css:
```css
:root {
	--my-custom-color: red;
}

.custom-class-name {
	color: var(--my-custom-color);
}
```

Changing the variable:
```ts
document.body.setCssProps({'--my-custom-color': 'blue'});
```

anareaty:
The problem is, not only variables themselves, but also the classes they are assigned to are dynamicaly changing depending on the user settings, so this approach wouldn't work. If I want to set user-selected colors to user-selected tags, I will have to either create an editable stylesheet and update it when user changes settings, or constantly search for every tag from the user's list to directly assign color to it. Is the second approach more preferable?

anareaty:
I made a part of the requested changes, but I am unsure what to do with the styles that should be applied to the individual properties. Should I directly apply them to elements instead?

PR #6921, opened 2025-06-29
Add plugin: Better Trash

Fertion:

## Repo URL

Link to my plugin: https://github.com/Fertion/Obsidian_Better_trash

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -17575,5 +17575,12 @@
     "author": "Albert O'Shea",
     "description": "Changes click behavior for unsupported files to work similar to Windows File Explorer: single-click selects, double-click opens in default app.",
     "repo": "Alb-O/obsidian-click-unsupported-files"
+  },
+  {
+    "id": "better_trash",
+    "name": "Better Trash",
+    "author": "Fertion",
+    "description": "Advanced trash with file viewing, automatic cleanup, and original path restoration.",
+    "repo": "Fertion/Obsidian_Better_trash"
   }
 ]


Zachatoo:
[# Obsidian Better Trash](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/README.md#L1)

[loadStyles() {](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L899-L980), [const styleEl = document.getElementById('better-trash-styles');](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L300-L303)
If you includes a styles.css file in your release, Obsidian will manage loading and unloading that css file for you. Please use a styles.css file instead of loading your own stylesheet using JavaScript.

[if (!this.originalDelete) {](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L380-L396), [if (this.originalDelete) {](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L289-L294), [if (this.originalDelete) {](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L1015-L1020)
Overriding methods in this way is unsafe in a plugin ecosystem. Please consider if there is a way to do what you want without overriding methods.
In your case, you could consider listening to the [`Vault.on('delete')`](https://docs.obsidian.md/Reference/TypeScript+API/Vault/on('delete')) event.
If you must override this method, then please use a tool such as [monkey-around](https://github.com/pjeby/monkey-around) to safely patch the method in a way that patches can be uninstalled in arbitrary order.

[this.checkIntervalId = window.setInterval(](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L319-L322)

[callback: () => this.restoreFileCommand(),](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L244)
This should be a [conditional command](https://docs.obsidian.md/Plugins/User+interface/Commands#Conditional+commands), with the condition being an active file that is in the trash folder.

[if (!await vault.adapter.exists(originalDir)) {](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L562), [while (await vault.adapter.exists(newPath)) {](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L574), [while (await vault.adapter.exists(newPath)) {](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L649)
[Prefer the Vault API over the Adapter API](https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Prefer+the+Vault+API+over+the+Adapter+API)

[this.trashInfoComponents.set(leaf, component);](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L438-L439)
The lifecycle of these components are not being properly managed and is resulting in a memory leak. They should be unloaded when the related MarkdownView is no longer visible. Right now, if you open a MarkdownView that should load your component and then close the tab, your component doesn't get unloaded until your plugin is unloaded.
You can register the components to the MarkdownView using [`view.addChild()`](https://docs.obsidian.md/Reference/TypeScript+API/Component/addChild), which will have the MarkdownView manage it's lifecycle.
```ts
leaf.view.addChild(component);
```
Note that you will also want to ensure that your components are unloaded when your plugin is unloaded. One way you could do this is by emitting your own event to the workspace when your plugin is unloaded (using [`Workspace.trigger()`](https://docs.obsidian.md/Reference/TypeScript+API/Events/trigger)) and have your component listen to that event and unload itself. You don't have to do this option but this option will not result in a memory leak which is what we're looking for.

[this.app.workspace.on('layout-ready', () => {](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L271)
Use [`Worspace.onLayoutReady()`](https://docs.obsidian.md/Reference/TypeScript+API/Workspace/onLayoutReady) instead.

[setTimeout(() => {](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L280-L285)
You shouldn't need this, `Workspace.onLayoutReady()` should account for this.

[const base = file.name.substring(0, file.name.length - ext.length - 1);](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L651)
Use [`file.basename`](https://docs.obsidian.md/Reference/TypeScript+API/TFile/basename) instead.

[const lang = this.settings.language || 'en';](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L1025), [new Setting(containerEl)](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L1067-L1081)
This should not be a setting, use the `getLanguage()` function from the API to get the language configured in the Obsidian settings (requires `minAppVersion: "1.8.0"`)

[new Setting(containerEl)](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L1062-L1064)

[new Setting(containerEl)](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L1084-L1111)
Use [AbstractInputSuggest](https://docs.obsidian.md/Reference/TypeScript+API/AbstractInputSuggest) when letting a user select a file/folder to add type-ahead support.

---

Optional feedback:

To make the manual review process and future plugin maintenance more efficient, we recommend that you split your code into multiple files and move all source code files into a `src` directory. You can use a bundling tool such as `esbuild` to bundle your code into a single `main.js` file for releases. See [obsidian-sample-plugin](https://github.com/obsidianmd/obsidian-sample-plugin) for an example.

[this.app.workspace.setActiveLeaf(this.leaf);](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L191-L192), [this.app.workspace.setActiveLeaf(this.leaf);](https://github.com/Fertion/Obsidian_Better_trash/blob/2aeafdab913e8abbfc1ea0f4a9448c4ccb78adfa/main.js#L204-L205)
Why do you need to set the active leaf and wait for the event queue in order to restore or delete a file?

PR #7054, opened 2025-07-08
Add plugin: Paste as file link

mbedded:

## Repo URL

Link to my plugin: https://github.com/mbedded/obsidian-paste-file-link

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

---
Remarks:

The guidelines says, that `this.app.vault.getFiles().find(..)` should not be used. The idea behind my plugin is, that I can paste a file name (note) so a file-link is created. I've tested the method `this.app.vault.getFileByPath()` but when I'm looking for "my note" it does not find "folder/my note" for example.

In my workflow I use F2 to select the filename, press CTRL+C to copy it, switch to another note and paste the name. The selection with F2 does not contain the full path of the file. That's why I use `getFiles().find(..)`instead. If there is a way for improvement, I'm happy to update my plugin.

I'm not sure how big the impact is. In my personal vault (around 1000 files) I had no issues but this may depend on the system (e.g. dual-core instead of quad-core).

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -18427,7 +18427,7 @@
     "id": "paste-as-file-link",
     "name": "Paste as file link",
     "author": "Matthias Büge",
-    "description": "Allows pasting clipboard content as file links to existing notes.",
+    "description": "Paste clipboard content as file links into existing notes, when a file with this name is existing.",
     "repo": "mbedded/obsidian-paste-file-link"
   }
 ]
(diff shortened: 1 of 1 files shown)


Zachatoo:
Great!

One last thing. Can you bump the version in the manifest.json at the root of your repo and create a new release with your latest changes, so I can validate that you're familiar with how to make updates to your plugin that Obsidian will pick up?
["version": "1.0.0",](https://github.com/mbedded/obsidian-paste-file-link/blob/c7393b289d280bda9f80b730ed58c00cae8e9b81/manifest.json#L4)

Zachatoo (approved):

PR #7062, opened 2025-07-10
Add plugin: Yandex Tracker Issue

CubieProg:

## Repo URL

Link to my plugin: https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -18415,5 +18415,12 @@
     "author": "weshell",
     "description": "Render xiangqi(Chinese chess) positions diagrams in a note.",
     "repo": "west-shell/obsidian-xiangqi"
+  },
+  {
+    "id": "yandex-tracker-issue",
+    "name": "Yandex Tracker Issue",
+    "author": "Pavel Sokolov",
+    "description": "Display Yandex Tracker issues in your notes",
+    "repo": "CubieProg/Obsidian-Yandex-Tracker-Issue"
   }
 ]


Zachatoo:
[manifest.json](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/.gitignore#L27)
You should be tracking changes to this file in git. Obsidian looks at the manifest.json file at the root of your repo on your default branch to determine which release to download. If this isn't being updated then users will never get updates to your plugin.

[--cell-size: 50px;](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/styles.css#L2), [.board {](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/styles.css#L5), [.row {](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/styles.css#L9), [.square {](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/styles.css#L14), [.white {](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/styles.css#L24), [.black {](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/styles.css#L28), [.red-hihglight {](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/styles.css#L32), [.blue-hihglight {](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/styles.css#L36), [.green-hihglight {](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/styles.css#L40), [.processor-gantt-margin-bottom {](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/styles.css#L45), [.gantt-wrapper {](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/styles.css#L49), and others
Please update your classes and custom variables to be more specific to your plugin to avoid conflicts with other plugins.

[style: `margin-left: ${offset}%;`](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/src/MarkdownParser/Gantt/NavDateLine.ts#L7)

[this.markdownParser.unregisterProcessors()](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/main.tsx#L27)
This isn't necessary. Since you're using `this.plugin.registerMarkdownCodeBlockProcessor` to register the processors, Obsidian will unload them for you when your plugin is unloaded.

[const auth_header = createEl('h1', { text: "Авторизация." })](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/src/Settings/SettingsTab.ts#L78)
```ts
```

["OAuth Token"](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/src/Settings/SettingsTab.ts#L83), ['Укажите ваш OAuth Token'](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/src/Settings/SettingsTab.ts#L84), ['Укажите Org ID вашей организации'](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/src/Settings/SettingsTab.ts#L92)

---

[attr: { text: "test caption" },](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/src/MarkdownParser/MarkdownParser.tsx#L147)
This looks like test code, do you need it?

You have a lot of dependencies listed in your package.json that you don't appear to be using, can you please uninstall unused dependencies?

[private rerender(full: boolean = false) {](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/main.tsx#L30-L34)
This is a bit extreme for what you're trying to do, which is re-render your Markdown codeblock processors.
Please consider having your processors be classes that extend [`MarkdownRenderChild`](https://docs.obsidian.md/Reference/TypeScript+API/MarkdownRenderChild), calling [`ctx.addChild`](https://docs.obsidian.md/Reference/TypeScript+API/MarkdownPostProcessorContext/addChild) to have the MarkdownCodeblockProcessor manage loading and unloading renderers as it sees fit, and [registering](https://docs.obsidian.md/Reference/TypeScript+API/Component/register) your `yandex-tracker-issue:rerender` event to your processor classes and re-render themselves when you receive an event.
[Here](https://github.com/Zachatoo/obsidian-rerender-markdown-codeblock-processor-example) is an example if it's helpful.

CubieProg:
Hello! I`ve made all changes. But, I have a question about this:

> [style: margin-left: ${offset}%;](https://github.com/CubieProg/Obsidian-Yandex-Tracker-Issue/blob/ebfada741c2e6f307bd74af62f1f6cf2cbcd1ab0/src/MarkdownParser/Gantt/NavDateLine.ts#L7)

Yes, I know that managing styles in scripts is a bad practice but I need to assign this margins dynamically to each singular div.

All of this divs must be margined by a different (unique) value.

I`ve commented this code but may be you can give me an advice to manage this issue?

Thanks in advance.

Zachatoo:
> Yes, I know that managing styles in scripts is a bad practice but I need to assign this margins dynamically to each singular div.
> 
> All of this divs must be margined by a different (unique) value.

Use [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascading_variables/Using_CSS_custom_properties) to set highly dynamic values.
   ```css
   .my-class {
        width: var(--my-width, 10px); /* default to 10px, modified if --my-width is set */
   }
   ```
   ```js
   myHTMLElement.style.setProperty("--my-width", this.settings.myWidth);
   ```

Zachatoo (approved):
I appreciate your decision to only re-render specific processors based on which setting was updated. Nice work!

PR #7099, opened 2025-07-14
feat: add commentator plugin

Fevol:

## Repo URL

Link to my plugin: https://github.com/Fevol/obsidian-criticmarkup

## Release Checklist
  - [X]  Windows
  - [ ]  macOS
  - [X]  Linux
  - [ ]  iOS _(if applicable)_
  - [X] `main.js`
  - [X] `manifest.json`

## Additional notes
At the time of writing, this plugin is not yet fully ready for review. Many more things need to be added, improved and finished, but these should be finished by the time this plugin reaches the top of the queue.

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -17890,5 +17890,12 @@
     "author": "Utkarsh Raj",
     "description": "Visualize notes as interactive mind maps",
     "repo": "linem-davton/obsidian-better-mindmap"
+  },
+  {
+    "id": "commentator",
+    "name": "Commentator",
+    "author": "Fevol",
+    "description": "Suggest edits, add comments, and annotate your notes using CriticMarkup syntax.",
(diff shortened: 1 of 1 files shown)


Fevol:
/skip Explanation below.

### Inline JS styles
All direct element style assignments used in the plugin are necessary. They apply on an individual element basis, or are instances where it would not make sense to be transferred to CSS variables. In any case, the user is never supposed to override these styles.

### Casting `TFile`
No actual `TAbstractFile` is being cast in the plugin, the highlighted code fragments are used to mock a `TFile` for instantiating a temporary view, as it is not possible to create a dummy `TFile` via existing API.
(I did fix one instance of `TFile` casting where it was not necessary, but the argument was already a TFile.)
 
### Casting to `any`
The `any` casts are used to avoid having to specifically type these objects. (Two instances of `any` usage have been removed)

PR #7128, opened 2025-07-17
Add Better Page Preview

desaixie:

## Repo URL

Link to my plugin: [https://github.com/desaixie/better-page-preview](https://github.com/desaixie/better-page-preview)

## Release Checklist
  - [ ]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -17904,5 +17904,12 @@
     "author": "algernon",
     "description": "Help you to study with ai-prompting",
     "repo": "mali-i/deepseek-ai-assistant"
+  },
+  {
+    "id": "better-page-preview",
+    "name": "Better Page Preview",
+    "author": "Desai",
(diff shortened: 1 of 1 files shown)


joethei:
[modifier: Platform.isMacOS ? "Meta" : "Control",](https://github.com/desaixie/better-page-preview/blob/9a583fadf7ff82be65be61cdee7fd2a5c239cf15/main.ts#L11)
This should use "Mod" by default, and only during rendering / event handling should it differentiate between Mac and Windows/Linux.

[window.addEventListener("keydown", this.onKeyDown, true);](https://github.com/desaixie/better-page-preview/blob/9a583fadf7ff82be65be61cdee7fd2a5c239cf15/main.ts#L50-L53)
This does not support pop-out windows, see here how to update your plugin to support pop out windows:
https://obsidian.md/blog/how-to-update-plugins-to-support-pop-out-windows/

PR #7133, opened 2025-07-18
Rename Plugin: Virtual Footer → Virtual Content

Signynt:
# I am changing the name of an existing Community Plugin

Since I initially released my plugin it's scope has increased, and the original name is no longer accurate.
I will update the Repo url and release a new version of my plugin with the updated name once this PR has been merged.

## Repo URL
Link to my plugin: https://github.com/Signynt/virtual-footer

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -17059,11 +17059,11 @@
     "repo": "jaspersurmont/obsidian-comments"
   },
   {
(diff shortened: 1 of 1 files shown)


joethei:
The plugin ID should not be changed.
It will make it impossible for existing users to update.

PR #7180, opened 2025-07-21
Add plugin: Table Checkbox Renderer

dannns:
Add plugin: Table Checkbox Renderer

## Repo URL

Link to my plugin: https://github.com/dannns/obsidian-table-checkbox-renderer

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -17911,5 +17911,12 @@
     "author": "Tiago Jacinto",
     "description": "Add inheritance-like behavior to notes.",
     "repo": "TiagoJacinto/obsidian-object-oriented-thinking"
+  },
+  {
+    "id": "table-checkbox-renderer",
+    "name": "Table Checkbox Renderer",
+    "author": "Daniel Aguerrevere",
+    "description": "Interactive checkboxes for Markdown tables. Toggle checkboxes in Reading Mode to instantly update the Markdown file. Supports multiple checkboxes per cell and any table layout.",
+    "repo": "dannns/obsidian-table-checkbox-renderer"
   }
 ]


dannns:
Updated authorUrl to point to my Github profile

joethei:
Remove the `main.js` file from the repo and add it to the `.gitignore`, release assets should not be mixed with source assets.

[# Obsidian Table Checkbox Renderer](https://github.com/dannns/obsidian-table-checkbox-renderer/blob/e44b4d31cc9254dd99535c00202aa5cd767fa37c/README.md#L1)

[const content = await plugin.app.vault.read(file);](https://github.com/dannns/obsidian-table-checkbox-renderer/blob/e44b4d31cc9254dd99535c00202aa5cd767fa37c/src/dom-helpers.ts#L32-L42)
It would be safer to use the `process` function from the API, as that will make sure that the change happens atomically.

dannns:
> Remove the `main.js` file from the repo and add it to the `.gitignore`, release assets should not be mixed with source assets.
> 
> [# Obsidian Table Checkbox Renderer](https://github.com/dannns/obsidian-table-checkbox-renderer/blob/e44b4d31cc9254dd99535c00202aa5cd767fa37c/README.md#L1) Naming something "Obsidian xyz" is reserved for first party products we create, please change this header.
> 
> [const content = await plugin.app.vault.read(file);](https://github.com/dannns/obsidian-table-checkbox-renderer/blob/e44b4d31cc9254dd99535c00202aa5cd767fa37c/src/dom-helpers.ts#L32-L42) It would be safer to use the `process` function from the API, as that will make sure that the change happens atomically.

Thank you. Addressed review comments in:
https://github.com/dannns/obsidian-table-checkbox-renderer/releases/tag/1.0.0

PR #7240, opened 2025-07-27
Add plugin: Web viewer Bookmarks

stefandanzl:

## Repo URL

Link to my plugin: https://github.com/stefandanzl/webviewer-bookmarks

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -18555,5 +18555,12 @@
     "author": "Max Margenot",
     "description": "Implements AI indexing and search to surface related thoughts to your current work. Look into the mirror.",
     "repo": "mmargenot/tezcat"
+  },
+  {
+    "id": "granola-sync",
+    "name": "Granola Sync",
+    "author": "Tom Elliot",
+    "description": "Sync Granola notes to your vault.",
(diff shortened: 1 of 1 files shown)


stefandanzl:
I just tried creating a web viewer bookmark with the regular bookmarks feature - I didn't think of using that before

Opening links can globally be set to either open in built-in web viewer or with the default web browser (my prefered way for regular use). For special cases (when wanting to use bookmarks for certain routine work tasks in obsidian) I would like to define bookmarks and open them in web viewer, which can't be defined in bookmarks at the moment.

Another aspect is that bookmarks can only be set as commands or ribbons with additional setup through other plugins like "Commander" which is also kind of tedious, if instead you could do all of those settings in one setup per bookmark in this plugin as can be seen in this screenshot example:

{50F0FE29-9EA8-4A79-9136-66571B2A6D00}

Additional features: open in new tab, lucide icon

mnaoumov:
Merged

PR #7292, opened 2025-08-01
Update community-plugins.json

AlphaHasher:
Just updating the existing [plugin's](https://github.com/AlphaHasher/obsidian-remove-hyperlinks) name and description


joethei:
Don't change the plugin id, all existing users of your plugin won't be able to update.

PR #7312, opened 2025-08-03
Add plugin: Easily Habit Tracker

yodaliu0414:
Add a new plugin "Easily Habit Tracker"

## Repo URL

Link to my plugin:

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/.github/workflows/validate-plugin-entry.yml
+++ b/.github/workflows/validate-plugin-entry.yml
@@ -54,7 +54,7 @@ jobs:
                 'My GitHub release contains all required files (as individual files, not just in the source.zip / source.tar.gz)', 'GitHub release name matches the exact version number specified in my manifest.json',
                 'The `id` in my `manifest.json` matches the `id` in the `community-plugins.json` file.',
                 'My README.md describes the plugin\'s purpose and provides clear usage instructions.',
-                'I have read the developer policies at https://docs.obsidian.md/Developer+policies, and have assessed my plugin\'s adherence to these policies.',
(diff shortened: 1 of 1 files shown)


yodaliu0414:
Hello! My apologies for the multiple pull requests.

I accidentally closed the original PR (#7246) and was unable to reopen it. I also mistakenly created #7309 while trying to fix the issue.

This new pull request is the correct one with all the required changes. I'm very sorry for any confusion this may have caused, and thank you for your time and consideration.

yodaliu0414:
Hello, thank you for the automated review! I've pushed updates to address the feedback and would like to provide some clarification on the remaining points.

Optional: any Casts
I have fixed most of the any casts. The remaining ones in periodicNotesUtils.ts and settings.ts are necessary to access the settings of the external "Periodic Notes" plugin. Since my plugin doesn't control the types for that external dependency, casting to any is the required approach for interoperability in this case.

Required: Inline Styles
/skip

The remaining inline styles flagged by the bot are all dynamic values set using style.setProperty. This is intentional, as I am using CSS Custom Properties (CSS variables) to control dynamic layouts (like the number of columns in a grid) from JavaScript.

The actual styling rules that use these variables (e.g., grid-template-columns: repeat(var(--habits-per-row), 1fr);) are located in the styles.css file. This approach follows the best practice of separating styling from logic while still allowing for the dynamic layouts required by the plugin's features.

Thank you for your time and for considering these points!

PR #7445, opened 2025-08-15
Update community-plugins.json

namikaze-40p:

## Repo URL

Link to my plugin: https://github.com/namikaze-40p/obsidian-cliplet

## Release Checklist
  - [x]  Windows
  - [x]  macOS
  - [ ]  Linux
  - [x]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -18205,5 +18205,12 @@
     "author": "ikmolbo",
     "description": "Transform handwritten documents and scanned images into editable text with Handwriting OCR's AI-powered handwriting to text conversion.",
     "repo": "ikmolbo/handwriting-ocr-obsidian-plugin"
+  },
+  {
+    "id": "cliplet",
+    "name": "Cliplet",
+    "author": "namikaze-40p",
+    "description": "Combines a dedicated clipboard with a snippet feature, allowing you to efficiently manage frequently used text and snippets and retrieve them instantly when needed.",
+    "repo": "namikaze-40p/obsidian-cliplet"
   }
 ]


namikaze-40p:
@Zachatoo 
Thank you very much for your thorough review!
I’ve addressed the points you mentioned in [v0.2.1](https://github.com/namikaze-40p/obsidian-cliplet/releases/tag/0.2.1).
Could you please take another look when you have a moment?

---
> Naming something "Obsidian xyz" is reserved for first party products we create, please change this header.

-> Fixed in [this commit](https://github.com/namikaze-40p/obsidian-cliplet/commit/7dce64d95eeb6446b12185507bbe6fb264d42d51).

---
> Please do not manage your own stylesheet, you should only be referencing styles in your styles.css file. If you need highly dynamic styling, please use CSS custom properties instead.

-> Fixed in [this commit](https://github.com/namikaze-40p/obsidian-cliplet/commit/5b049def6ef568c0c042883ec34068f377ac7266).

---
> This should be an editorCheckCallback, with the condition being a selection in the editor.

-> Fixed in [this commit](https://github.com/namikaze-40p/obsidian-cliplet/commit/214b88d0530b1199b0422e9634490495fd254491).

---
> Use [Scope.register](https://docs.obsidian.md/Reference/TypeScript+API/Scope/register) to add keymap handlers to this scope.

-> Fixed in [this commit](https://github.com/namikaze-40p/obsidian-cliplet/commit/180a45879fd07136c0647c21416a55a9becf5127).

---
> [this.close = () => {};](https://github.com/namikaze-40p/obsidian-cliplet/blob/aea369d41474739139c0d3fd41aac816fc0de4ac/src/ui/cliplet-search-modal.ts#L212)
Why are you overriding this method?
> 
> Can you explain what the purpose your mutation observers are? There is likely a better way to do what you're trying to do.

-> Fixed in [this commit](https://github.com/namikaze-40p/obsidian-cliplet/commit/180a45879fd07136c0647c21416a55a9becf5127).

For now, I’ve removed the override of the close method.

I use mutation observers to detect when the highlighted suggestion changed so I could update a live preview. Since there’s no public “selection changed” hook on FuzzySuggestModal and arrow keys are handled internally, we used DOM mutations as a workaround.

Zachatoo (approved):

PR #7449, opened 2025-08-15
Add plugin: Sync-safe files names

j-maas:

## Repo URL

Link to my plugin: https://github.com/j-maas/sync-safe-file-names

## Release Checklist
  - [ ]  Windows
  - [ ]  macOS
  - [x]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -18205,5 +18205,12 @@
     "author": "ikmolbo",
     "description": "Transform handwritten documents and scanned images into editable text with Handwriting OCR's AI-powered handwriting to text conversion.",
     "repo": "ikmolbo/handwriting-ocr-obsidian-plugin"
+  },
+  {
+    "id": "sync-safe-file-names",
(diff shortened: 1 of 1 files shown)


j-maas:
@Zachatoo Thanks for the feedback!

1. I've added a description to the README with the differences to [Safe Filename Linter](https://github.com/sneakyfoxes/obsidian-safe-filename-linter). (That plugin does not run automatically for new files.)
2. I've added a description to the alias setting. (If the file name is already safe, no alias will be created.)
3. I've removed the styles.css. (The [release script](https://docs.obsidian.md/Plugins/Releasing/Release+your+plugin+with+GitHub+Actions) did [fail](https://github.com/j-maas/sync-safe-file-names/actions/runs/16999467623/job/48198022598) when I removed it, which is why I left it in previously.)

PR #7467, opened 2025-08-17
Add plugin: ZenBoard

Shadabdullah:

## Repo URL

Link to my plugin: https://github.com/Shadabdullah/obsidian-zenboard

## Release Checklist
  - [x] Windows
  - [x] macOS
  - [x] Linux
  - [ ] Android *(if applicable)*
  - [ ] iOS *(if applicable)*
  - [x] `main.js`
  - [x] `manifest.json`
  - [ x] `styles.css` *(optional)*

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -18212,5 +18212,12 @@
     "author": "Jeffry",
     "description": "Collect related notes based on links/backlinks to provide focused context for external AI chatbots. Explore note relationships visually and export the bundle.",
     "repo": "shuxueshuxue/PackUp4AI"
+  },
+  {
+  "id": "zenboard",
+	"name": "ZenBoard",
+  "author": "Shad Abdullah",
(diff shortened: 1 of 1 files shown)


Zachatoo:
The rule of thumb is to never use inline styles, because it makes it difficult for themes/snippets to style.

> A) Use CSS custom properties: set --habit-color on the element and
reference it in CSS
> B) Generate dynamic CSS classes for each habit

Either of these options are good.

Shadabdullah:
Thanks a lot for the clarification, @Zachatoo! 

I appreciate the detailed guidance. I’ll refactor the inline styles to use CSS custom properties (--habit-color) so that themes and snippets can easily override them.

Also noted on the casing and the author URL — I’ll update the UI labels to sentence case and point the authorUrl to a page that properly represents me.

Thanks again for reviewing and helping me align everything with the Obsidian guidelines!

PR #7550, opened 2025-08-23
Add plugin: Cyclearn

noahboos:

## Repo URL

Link to my plugin: https://github.com/NoahBoos/obsidian-flashcards

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`

### DIFF
--- a/community-plugins.json
+++ b/community-plugins.json
@@ -18228,10 +18228,10 @@
     "repo": "bao-tg/kindle-vocab"
   },
   {
-    "id": "flashcards",
-    "name": "Flashcards",
+    "id": "cyclearn",
+    "name": "Cyclearn",
     "author": "Noah Boos AKA Rift",
     "description": "The all-in-one solution tailored for flashcards user. With no external apps required, everything's made for and used by your vault.",
-    "repo": "NoahBoos/obsidian-flashcards"
+    "repo": "NoahBoos/obsidian-cyclearn"
   }
 ]


NoahBoos:
Changed.

claremacrae:
Hi @NoahBoos,

> [[1](https://github.com/NoahBoos/obsidian-cyclearn/blob/d1954ee8606a83f476b1f49f12bee4bf1db9d4c6/src/database/Database.ts#L8)]:Obsidian's configuration directory isn't necessarily `.obsidian`, it can be configured by the user. You can access the configured value from `Vault#configDir`

I see the [code on main](https://github.com/NoahBoos/obsidian-cyclearn/blob/393a3685d0592fe232cb7fbb4b998f3abda0587a/src/database/Database.ts#L8) is now this:

```ts
    const PATH_DATABASE = `${this.app.vault.adapter.basePath}/${this.app.vault.configDir}/plugins/obsidian-${cyclearn.manifest.id}/database.json`;
```

What's the reason for the `obsidian-` prefix?

I believe the plugin will be installed in the `manifest.id` folder - so I would have thought any files would go in that same folder.

Are you separately creating a second directory for your plugin data, or will that `obsidian-${cyclearn.manifest.id}/` not be found?

I presume you've tested the change - I'm just trying to understand what I don't understand.

claremacrae:
Yes, I found it...

The Obsidian [eslint-plugin tool](https://github.com/obsidianmd/eslint-plugin/tree/1053f94fee7703abbc2d876fbcdaff9621bfcd17) does warn agains using `obsidian` and `plugin` in `id`:

https://github.com/obsidianmd/eslint-plugin/blob/1053f94fee7703abbc2d876fbcdaff9621bfcd17/lib/rules/validateManifest.ts#L240-L243

This is the definition of FORBIDDEN_WORDS, used by that function:

https://github.com/obsidianmd/eslint-plugin/blob/1053f94fee7703abbc2d876fbcdaff9621bfcd17/lib/rules/validateManifest.ts#L24

mnaoumov:
The best way to get access to the plugin dir is via https://docs.obsidian.md/Reference/TypeScript+API/PluginManifest/dir

claremacrae:
@mnaoumov: nice, but why the ? - why might there be no value?

NoahBoos:
Hey everyone, 

I think I understood what @claremacrae tried to warn me about. 

Just to be sure, I must remove "obsidian-" from the path where I create the database, in order to avoid conflict in folder when Obsidian creates plug-in related files (I think this can happen as soon as the user wants to download the plug-in... As far as I understood.), right ?

I will do this as soon as I get back home, in a couple of hours. I wasn't available lately. 

Thank you very much for your help anyway. 

- Noah

claremacrae:
Hi @NoahBoos,

### Suggestion 1 - fix the behaviour

Yes, the smallest change you can make is to do this, which I was suggesting, and should give the correct behaviour:

```ts
    const PATH_DATABASE = `${this.app.vault.adapter.basePath}/${this.app.vault.configDir}/plugins/${cyclearn.manifest.id}/database.json`;
```

### Suggestion 2 - and simplify the code further

@mnaoumov has suggested a much cleaner way, using this value:

https://docs.obsidian.md/Reference/TypeScript+API/PluginManifest/dir

You could write `cyclearn.manifest.dir` to the console, and you will probably find it most of the work for you - and you could simplify the code a lot more.

For my plugin, `manifest.dir` is:

```
.obsidian/plugins/obsidian-tasks-plugin
```

You need to test it, but you will probably end up with something like this:

```ts
    const PATH_DATABASE = `${this.app.vault.adapter.basePath}/${cyclearn.manifest.dir}/database.json`;
```

NoahBoos:
Hello again,

What @mnaoumov suggested seems to work.

Thank you all for the help, now... Gotta wait for the review ! :>

PR #7716, opened 2025-09-05
Add obsidian-webdav-sync

LunarWatcher:

## Repo URL

Link to my plugin: https://github.com/LunarWatcher/obsidian-webdav-sync

## Release Checklist
  - [x]  Windows
  - [ ]  ~~macOS~~ Not supported
  - [x]  Linux
  - [x]  Android
  - [ ]  ~~iOS~~ Not supported
  - [x] `main.js`
  - [x] `manifest.json`

## Some comments ahead of time

* The adapter API is forced; the vault API refuses to index dotfiles, which is a dealbreaker. I'm operating with full vault sync, which requires those files. There may be individual calls to the adapter API that could be replaced with the vault API, but I do not trust it due to the lack of reliability around dotfiles, so I have not attempted to do so. This is not a plugin that does :sparkles: stuff :sparkles: without user intervention, so the performance tradeoff is acceptable in my opinion
* 2 instances of `innerHTML` that don't take user input are used to create icon buttons with text. It's nasty, but perfectly safe

And although not used in the plugin, the global `window.app` is used in the Python-based integration tests, because it's convenient. I mention this purely as another vote for "don't remove it entirely" - integration testing is already hard enough


LunarWatcher:
/skip already skipped, just doing a second one in case the review bot re-assignment was a mistake, and it got stuck in a weird state after I figured I'd resolve some conflicts. Guess that'll have to wait

PR #7996, opened 2025-09-30
Add plugin: DarkSlide

Th0rGarden:

## Repo URL

Link to my plugin: https://github.com/Th0rGarden/obsidian-darkslide

## Release Checklist
  - [x]  Windows
  - [ ]  macOS
  - [ ]  Linux
  - [ ]  iOS _(if applicable)_
  - [x] `main.js`
  - [x] `manifest.json`
- [x] I have read the developer policies at https://docs.obsidian.md/Developer+policies, and have assessed my plugin's adherence to these policies.


Th0rGarden:
Is this one of those situations you guys are going to tell us that propping up Obsidian's ecosystem with free plugins is something we should be grateful for because we're making these plugins for free? Or is that just something an out of touch moderator seems to think is an appropriate thing to say to the developers who actually make the only thing that keep obsidian from being the same as any other markdown editor? What is going on in the obsidian community where we are being spoken to that way? A very sad state of affairs, to be honest. 

I would reign those moderators in, if I were you. Do you pay these sad, out-of-touch souls? Otherwise I really am questioning the leadership on this team in a very serious manner.

I should add that I'm not personally bothered by this as much as it worries me for what is happening to Obsidian in the last couple years. The founders have essentially entirely disappeared from the public eye.

